PR #3, opened 2017-12-18
[WFCORE-3435] Expose capability/requirements to deployments

jmesnil:
JIRA: https://issues.jboss.org/browse/WFCORE-3435


jmesnil:
@bstansberry @dmlloyd Could you please review?

I'm not familiar with that code so I'd like to get your review before proposing this change to Swarm dev team.

core/WFCORE-3435_capability_requirements_deployments.adoc
```diff
@@ -0,0 +1,184 @@
...
++---------------------------------------+
+----
+
+== Proposed Design
+
+We would need to introduce a new API so that an user could activate services that depends on capabilities.
+This API is provided by WildFly Core (in its wildfly-controller artifact).
+
+The user must implement the `org.jboss.as.controller.CapabilityActivator` interface.
+This interface provides a `org.jboss.as.controller.CapabilityActivatorContext` interface that extends `org.jboss.msc.service.ServiceActivatorContext` and add a method to access the `org.jboss.as.controller.capability.CapabilityServiceSupport`.
```
dmlloyd: Why do we need all these new types? Could we not simply add the `hasCapability`/`getCapability` methods to `ServiceActivatorContext` directly?

core/WFCORE-3435_capability_requirements_deployments.adoc
```diff
@@ -0,0 +1,184 @@
...
+| getServiceRegistry(): ServiceRegistry |      +---------------------------------------------------------+
++---------------------------------------+          ^
+                                                  / \
+                                                   |
+                                               +---------------------------------------------------------------+
+                                               |                 << WildFly Core >>                            |
+                                               |             CapabilityServiceSupport                          |
+                                               +---------------------------------------------------------------+
+                                               | hasCapability(name: String): boolean                          |
+                                               | getCapabilityServiceName(capabilityName: String): ServiceName |
```
dmlloyd: What if a capability is not backed by a service?

core/WFCORE-3435_capability_requirements_deployments.adoc
jmesnil: `CapabilityServiceSupport` is an existing class that DUP can already uses.

I don't like the idea to introduce in MSC API the notion of "capabilities" that are only relevant to WildFly Core.
If you think it is ok to add to the `ServiceActivatorContext` all the methods from `CapabilityServiceSupport` that will ease my work but it does not feel conceptually clean.

For reference, all the methods are:
```
boolean hasCapability(String capabilityName);
<T> T getCapabilityRuntimeAPI(String capabilityName, Class<T> apiType) throws NoSuchCapabilityException;
<T> T getCapabilityRuntimeAPI(String capabilityBaseName, String dynamicPart, Class<T> apiType) throws NoSuchCapabilityException;
ServiceName getCapabilityServiceName(String capabilityName);
ServiceName getCapabilityServiceName(String capabilityBaseName, String ... dynamicParts);
```

core/WFCORE-3435_capability_requirements_deployments.adoc
jmesnil: This feature is to install an user service that depends on another service through the use of capabilities.
If the capability is not backed by a service, I'm not sure what the user service would be depending on.

jmesnil:
@bstansberry this proposal does not allow to *install* a capability from an user deployment.

Is that in the scope too?

core/WFCORE-3435_capability_requirements_deployments.adoc
dmlloyd: OK, I forgot that that `ServiceActivatorContext` was in MSC.

core/WFCORE-3435_capability_requirements_deployments.adoc
dmlloyd: If the activated services are using the same interface which includes `getCapabilityRuntimeAPI` then I guess that works.  In this case the dependency is simply on the WildFly container existing, which is (AFAIK) always true when services are running.

jmesnil:
@bobmcwhirter, Could you review this proposal (esp. wrt the changes you will have to make in Swarm) and tell if that's ok for you?

core/WFCORE-3435_capability_requirements_deployments.adoc
```diff
@@ -0,0 +1,184 @@
...
+| getServiceRegistry(): ServiceRegistry |
++---------------------------------------+
+----
+
+== Proposed Design
+
+We would need to introduce a new API so that an user could activate services that depends on capabilities.
+This API is provided by WildFly Core (in its wildfly-controller artifact).
+
+The user must implement the `org.jboss.as.controller.CapabilityActivator` interface.
```
bstansberry: Should this be in controller? I don't think that module's ever going to not be jboss.api=private as it has too many internals that aren't readily separable.

Other deployment framework classes are in the server module. I think the "deployment framework" classes should be broken out from server, as that module's a mix of internals and subsystem-dev usable stuff, and we should separate that. (I believe that's something fairly straightforward to do.)

This is something I think we need to discuss separately and report back here as it's likely a quite wide ranging conversation.

bobmcwhirter:
From the WildFly Swarm POV, this does look to solve our problems.

core/WFCORE-3435_capability_requirements_deployments.adoc
dmlloyd: One possibility might be to add this API into MSC with an SPI that WildFly could implement.  Ugly, perhaps, but it'd solve the problem.

PR #4, opened 2017-12-18
Create a generic design documentation template.

jamezp:


design-doc-template.adoc
```diff
@@ -0,0 +1,39 @@
+= Title
+:author:            Your Name
+:email:             your.email@redhat.com
+:toc:               left
+:icons:             font
+:keywords:          comma,separated,tags
```
dmlloyd: I guess the bullet list is not correct here?

design-doc-template.adoc
```diff
@@ -0,0 +1,39 @@
...
+:idprefix:
+:idseparator:       -
+:issue-base-url:    https://issues.jboss.org
```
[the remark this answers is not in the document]
dmlloyd: In Jeff's example, which I copied, the JIRAs were listed up here in the metadata.  I guess that is not the right thing to do?

design-doc-template.adoc
jamezp: I don't really know to be honest. I just know when I tried to generate the document with `asciidoctor` it had errors.

design-doc-template.adoc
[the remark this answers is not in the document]
jamezp: This could really be deleted. I just thought *maybe* it would be useful to reference the issue like `{issue-base-url}/WCORE-XXX`. It might not really be worth it though.

I did base this mostly off https://developer.jboss.org/wiki/DesignNotesForGracefulShutdownUponReceiptOfASIGTERM.

design-doc-template.adoc
```diff
@@ -0,0 +1,39 @@
+= Title
+:author:            Your Name
+:email:             your.email@redhat.com
+:toc:               left
+:icons:             font
+:keywords:          comma,separated,tags
+:idprefix:
```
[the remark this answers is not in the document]
jamezp: Yes. By default it uses the header name like `_issue_metadata`. With the two `id` settings it would make it `issue-metadata`.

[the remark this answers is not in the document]
dmlloyd: That's handy.  I've been doing my `id` by hand.

rhusar:
What's holding this up? I would suggest merging this and if people want to make further changes to the template they can open further PRs.

jamezp:
@rhusar It doesn't look like any of the PR's have been merged yet. I'm not sure what the plan is for them.

rhusar:
I assume that has to do with the process, but a template is helpful to have.

design-doc-template.adoc
```diff
@@ -0,0 +1,39 @@
...
+
+== Overview
+
+== Issue Metadata
+
+=== Issue:
+
+* {issue-base-url}/WFCORE[WFCORE-XXXX]
+
+=== Related Issues:
```
rhusar: The trailing colon should be removed.

design-doc-template.adoc
```diff
@@ -0,0 +1,39 @@
...
+
+=== Affected Projects or Components:
+
+=== Other Interested Projects:
+
+== Requirements
+
+=== Hard Requirements
+
+=== Nice-to-Have Requirements
```
rhusar: Isn't this a bit paradoxical?

design-doc-template.adoc
rhusar: This is wrong to begin with, the correct url prefix is `https://issues.jboss.org/browse/`

Though, this really goes against the ideas behind asciidoc which is to be human readable even unrendered. Let alone the fact that it now generates wrong links..

design-doc-template.adoc
bstansberry: Perhaps the wording is odd, but oh well. This outlines things that are being considered for the feature as the analysis is being done but which if not practical (technically or in terms of necessary resources) when it comes time to do the development, their absence should not mean the feature has not met its objectives.

These should be resolved one way or the other as soon as possible so people don't waste time (e.g. writing tests) on things that won't happen. But in the initial analysis phase it's useful to note everything being considered while being clear about the must-haves.

design-doc-template.adoc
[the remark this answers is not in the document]
bstansberry: +1 to removing this.

design-doc-template.adoc
rhusar: I agree its useful to jot down ideas being considered which are not hard requirements.

But makes me wonder, what is the difference (i.e. where is the line) between these so-called "nice to have requirements" and "non-requirements" section below? Does this mean "probably won't be implemented" and "obviously out of scope", respectively?

bstansberry: More like "Might not be implemented" and "Definitely out of scope".

The distinction between "might not" and "probably won't" isn't a big deal though. The text for the items can say something about probability.

Sometimes a Non-Requirement is because you just know you aren't going to do it (e.g. no time or not worth the effort), so state that up front and save someone asking.  Other times you're just being thorough and stating what might seem obvious to you but surprisingly often isn't obvious to others. 30 secs of writing a non-requirement can save you 10 minutes of writing a response to a question. It's also a get-out-of-jail-free card if someone later comes along and says the missing thing is a Bug.

rhusar:
Should we include a small section on testing? I think outlining at least an approach to testing would be useful. A full test plan might be later populated by QE.

For example, latest feature "support for configuring scattered caches" could have a note of testing in terms of "Run the standard suite of failover, stress and scalability tests with configured scattered cache for web sessions". This should quickly give an idea for QE what's the scope of the work.

jamezp:
I just added a section for "Test Plan". I think, for now at least, that should be enough.

PR #7, opened 2018-01-05
EAP7-832 Add option for all cookies over a secure channel to be marke…

stuartwdouglas:
…d secure

### DIFF
--- a/undertow/EAP7-832_Secure_Cookie.adoc
+++ b/undertow/EAP7-832_Secure_Cookie.adoc
@@ -1,0 +1,53 @@
+= EAP7-832 Secure Cookie Handling
+:author:            Stuart Douglas
+:email:             sdouglas@redhat.com
+:toc:               left
+:icons:             font
+:keywords:          comma,separated,tags
+:idprefix:
+:idseparator:       -
+:issue-base-url:    https://issues.jboss.org
+
+== Overview
+
+The goal of this feature is to add an option to mark all cookies that are set on a secure connection with the
+secure attribute. This means that a cookie that is set over a secure connection will never be sent over an unsecure
+connection.
+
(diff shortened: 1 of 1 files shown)


undertow/EAP7-832_Secure_Cookie.adoc
```diff
@@ -0,0 +1,51 @@
...
+:icons:             font
+:keywords:          comma,separated,tags
+:idprefix:
+:idseparator:       -
+:issue-base-url:    https://issues.jboss.org
+
+== Overview
+
+The goal of this feature is to add an option to mark all cookies that are set on a secure connection with the
+secure attribute. This means that a cookie that is set over a secure connection will never be sent over an unsecure
```
jstourac: typo - `set` -> `sent`

undertow/EAP7-832_Secure_Cookie.adoc
```diff
@@ -0,0 +1,51 @@
...
+=== Related Issues:
+
+* {issue-base-url}/UNDERTOW-1182[UNDERTOW-1182]
+
+=== Dev Contacts:
+
+* mailto:{email}[{author}]
+
+=== QE Contacts:
+
```
jstourac: feel free to put me in

undertow/EAP7-832_Secure_Cookie.adoc
stuartwdouglas: Both set and sent work here, but set is less ambiguous imho as it makes it clear it is referring to set-cookie.

jstourac: Oh, of course, you are right. In this case `set` is better. Keep it that way, then :)

PR #10, opened 2018-01-24
WFLY-9563 Integrate KUBE_PING discovery protocol

rhusar:
https://issues.jboss.org/browse/WFLY-9563
https://issues.jboss.org/browse/EAP7-861

PR #12, opened 2018-02-05
WFLY-9664 Add support for configuring scattered caches

pferraro:
https://issues.jboss.org/browse/EAP7-867
https://issues.jboss.org/browse/WFLY-9664

### DIFF
--- a/clustering/infinispan/WFLY-9664_Scattered_Cache.adoc
+++ b/clustering/infinispan/WFLY-9664_Scattered_Cache.adoc
@@ -1,0 +1,67 @@
+= Scattered Cache
+:author:            Paul Ferraro
+:email:             paul.ferraro@redhat.com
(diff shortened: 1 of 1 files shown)


rhusar:
I think we should vaguely mirror the WildFly repository directory structure, so here I suggest we use `clustering/infinispan/`. E.g. here https://github.com/wildfly/wildfly-proposals/pull/10 I am creating the jgroups dir `clustering/jgroups/`

rhusar:
Also, since the process is open, I suggest we name documents by their upstream and publicly visible issues (not internal ones).

kabir:
+1 - we have not been strict for this for WF12 due to the time pressure, but want to do this for things now coming for WF13

mvinkler:
QE feedback: looks OK.

PR #13, opened 2018-02-08
EAP7-839 CDI 2.0 support.

manovotn:
Note that this is sent retrospectively as per request on https://issues.jboss.org/browse/EAP7-839.
E.g. the job is mostly done/merged into wfly as we speak.

### DIFF
--- a/weld/EAP7-839_CDI_2_0_support.adoc
+++ b/weld/EAP7-839_CDI_2_0_support.adoc
@@ -1,0 +1,53 @@
+= EAP7-839 CDI 2.0 Support
+:author:            Matej Novotny
+:email:             manovotn@redhat.com
+:toc:               left
+:icons:             font
+:keywords:          comma,separated,tags
+:idprefix:
+:idseparator:       -
+:issue-base-url:    https://issues.jboss.org
+
+== Overview
+
+This issue has two parts.
+Firstly it brings Weld 3.x/CDI 2.0 into Wildfly.
+Secondly, it introduces support for switch between CDI 1.2 (EE7) and CDI 2.0 (EE 8) for the sake of certification.
+
+== Issue Metadata
+
+=== Issue:
+
+* {issue-base-url}/EAP7-839[EAP7-839]
+
+=== Related Issues:
+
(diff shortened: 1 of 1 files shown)


weld/EAP7-839_CDI_2_0_support.adoc
```diff
@@ -0,0 +1,52 @@
...
+* {issue-base-url}/WFLY-9269[WFLY-9269]
+* {issue-base-url}/WFLY-9768[WFLY-9768]
+
+=== Dev Contacts:
+
+* mailto:{email}[{author}]
+
+=== QE Contacts:
+
+* mailto:{email}[{author}]
```
mkouba: This should be you @manovotn, right?

weld/EAP7-839_CDI_2_0_support.adoc
```diff
@@ -0,0 +1,52 @@
...
+
+=== Related Issues:
+
+* {issue-base-url}/WFLY-6465[WFLY-6465]
+* {issue-base-url}/WFLY-9269[WFLY-9269]
+* {issue-base-url}/WFLY-9768[WFLY-9768]
+
+=== Dev Contacts:
+
+* mailto:{email}[{author}]
```
mkouba: This should be you and me, right?

weld/EAP7-839_CDI_2_0_support.adoc
manovotn: Unless you want to join in as well? ;-)
Nice catch, will update

weld/EAP7-839_CDI_2_0_support.adoc
manovotn: Yes, will correct

weld/EAP7-839_CDI_2_0_support.adoc
manovotn: This is actually correct, those are fields from the beginning of the document and it will render my name and email (sorry it was probably too late when I wrote last comment) - it will look like [this](https://github.com/manovotn/wildfly-proposals/blob/EAP7-839/weld/EAP7-839_CDI_2_0_support.adoc)

PR #28, opened 2018-02-16
[WFCORE-3205] Design document/requirements for fixing embedded logging

jamezp:
https://issues.jboss.org/browse/WFCORE-3205

### DIFF
--- a/embedded/WFCORE-3205-embedded-logging.adoc
+++ b/embedded/WFCORE-3205-embedded-logging.adoc
@@ -1,0 +1,74 @@
+= Embedded Logging Changes
+:author:            James R. Perkins
+:email:             jperkins@redhat.com
+:toc:               left
+:icons:             font
+:keywords:          embedded,logging
+:idprefix:
+:idseparator:       -
+
+== Overview
+
+The embedded server should not require the JBoss Log Manager to be used. A user should be able to use their
+own log manager/logging configuration. The server itself uses JBoss Logging which works with several log
+managers. The embedded server should honor the log manager being used.
+
+== Issue Metadata
+
+=== Issue
+
+* https://issues.jboss.org/browse/WFCORE-3205[WFCORE-3205]
+
+=== Dev Contacts
+
+* mailto:{email}[{author}]
+
+=== QE Contacts
+
+* mailto:nziakova@redhat.com[Nikoleta Ziakova]
+
+=== Affected Projects or Components
+
+* Any consumer of the WildFly Embedded Server
+* Offline CLI
+* The new provisioning tool
+* Arquillian Embedded
+
+== Requirements
+
+=== Hard Requirements
+
+* Needs to work in:
+** Offline CLI
+** In the provisioning tool
+** Any environment created by a user; an executable main method, custom user configuration, maven/gradle plugin, etc.
+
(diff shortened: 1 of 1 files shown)


bstansberry:
There should be a specific hard requirement that the documented console log handling features of the offline CLI continue to work as documented (with my original wildfly.org blog post counting as documentation.) I'm basically talking about the --stdout param to embed-[server|host-controller] here. If anything else relevant is specified in the docs lets talk about it.

Modifying the CLI code to allow that to happen with a revised embedded server is fine, but that should be specifically part of the scope of this work.

jamezp:
That was my intent when I said it should work with offline CLI, but I do think it makes sense to explicitly state the `--stdout` parameter. 

One question though is should this only be a requirement if running CLI from the `$JBOSS_HOME/bin/jboss-cli.sh` script? I ask because CLI does offer an API which would be similar to an embedded server. By that I mean it could be invoked from any context like a maven plugin or just a simple standalone application.

bstansberry:
"+* Do we want to write the `logging.properties` when configuring embedded?"

We need to be able to use the provisioning tool / embedded to build a server that matches what we would build using the current build system. That's a hard requirement.

It's also a requirement that our build once we switch to provisioning is completely driven by the provisioning tool. This requirement exists because we want end users to be able to use the tool to produce the same thing that we ship as a zip. So, there can't be any post-provisioning tool magic in our build to make logging.properties match if provisioning doesn't make it match.

If we need to revisit the entire notion of logging.properties matching standalone.xml on the very first boot, that's a valid discussion.

jamezp:
Okay. I think then the `logging.properties` issue is likely out of scope for this specific topic then. That will be more of a logging subsystem/log manager change.

bstansberry:
"* If we don't use JBoss Log Manager like we have been, will it be confusing to users as by default JUL would likely be used if they haven't explicitly defined a log manager?"

This response may be somewhat a tangent.

The embedding app should be able to control the logging of the embedded server. I don't think that means the embedding app is *required* to just accept whatever logging it's user has configured for it and pass those settings directly through to the embedded server. The embedding app should be able to set up an appropriate log context that the server will respect.  Specifically in the case of the offline CLI I'm pretty sure we very much don't want to pass the CLI logging settings through.

I think the provisioning tool would be much the same.

I'm not sure about Arquillian embedded or other general uses of embedded.

bstansberry:
"One question though is should this only be a requirement if running CLI from the $JBOSS_HOME/bin/jboss-cli.sh script? I ask because CLI does offer an API which would be similar to an embedded server. By that I mean it could be invoked from any context like a maven plugin or just a simple standalone application."

I haven't thought hard about it but I think the --stdout thing is pretty simple (either discard server console logging or pass it to the CLI's output) and I don't see why that should vary depending how the CLI is used.

jamezp:
I've updated this document, moved it and removed the questions. Hopefully that's not an issue with some of the previous comments looking out of context.

nziakova:
There's hard requirement to work in provisioning tool, but I'm not sure if it can be tested now. What is the current status with it, please?

jamezp:
@nziakova I think @aloubyansky has done some testing with the changes. AIUI you'd have to use a snapshot of core with Galleon to test it.

PR #31, opened 2018-03-02
[WFLY-5603] Get data source information

gaol:
https://issues.jboss.org/browse/WFLY-5603
https://issues.jboss.org/browse/EAP7-940

### DIFF
--- a/jca/WFLY-5603_DataSource_Class_Information.adoc
+++ b/jca/WFLY-5603_DataSource_Class_Information.adoc
@@ -1,0 +1,80 @@
+= WFLY-5603 - Get Datasource Class Information
+:author:            Lin Gao
+:email:             lgao@redhat.com
+:toc:               left
+:icons:             font
+:keywords:          datasource,metadata
+:idprefix:
+:idseparator:       -
+:issue-base-url:    https://issues.jboss.org/browse
+
+== Overview
+
+The JDBC `datasource-class` is the recommended way to set up the DataSource connections in the WildFly `datasources` subsystem,
+but it requires user to know what `connection-properties` are defined in the `datasource-class`.
+
+This feature will facilitate it to show basic connection properties information somehow.
+
+== Issue Metadata
+
+=== Issue:
+
+* {issue-base-url}/WFLY-5603[WFLY-5603]
+
+=== Related Issues:
+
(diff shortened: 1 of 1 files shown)


jca/WFLY-5603_DataSource_Class_Information.adoc
```diff
@@ -0,0 +1,79 @@
...
+    }
+}
+--
+
+* The output is a list type of DMR contains maximum 2 elements with DMR type as Object.
+** One element is for the `datasource-class` if any, the other element is for the `xa-datasource-class` if any.
+** The key is the Java FQCN of the `(xa-)datasource-class`.
+** The value is another Object DMR which contains connection-property name and Java FQCN mappings.
+** The connection-property names are sorted alphabetically.
+* There is no model version update because only a runtime metrics is introduced.
```
kabir: @gaol I think you need to bump it as metrics are part of the management model. Take the situation where a user runs both WF12 and 13 and they have the same subsystem version. The user writes his script against WF13 and is able to access this metric. Then they run it against 12 which should be the same, and it blows up. Transformers are not needed though

gaol: Makes sense, I will update it.

simkam:
LGTM
I don't know which logic is under `:get-installed-driver` and `:installed-drivers-list` operations and high level command `jdbc-driver-info` but they should display properties discovered by this new feature

gaol:
Updated with the `:get-installed-driver` and `:installed-drivers-list` operations.

simkam:
@claudio4j @hpehl fyi, might be interesting for web console

claudio4j:
Good to known this proposal,once it is implemented in subsystem we might use it in web console. Thanks.

bstansberry:
Please let Luis Barreiro know about this proposal, as it's relevant to Agroal too.

gaol:
@barreiro Would you please take a look at this too? Thank you :smile:

barreiro:
This looks good. The current implementation on the Agroal subsystem has a slightly different model for driver, a 1:1 relation to a connection provider class (DataSource or XADatasource). Besides that, this is something I consider a nice feature and something that I can implement on Agroal as well.

PR #32, opened 2018-03-07
Proposal for CLI elytron auth security commands

jfdenise:


cli/EAP7-943_Security_Auth_Commands.adoc
```diff
@@ -0,0 +1,241 @@
+= Title
```
marekkopecky: Jira id and RFE name should probably be here ...

cli/EAP7-943_Security_Auth_Commands.adoc
```diff
@@ -0,0 +1,241 @@
...
+* Take benefit of existing Out Of The Box elytron configuration.
+* Drive users to select the right set of options thanks to CLI completion that 
+controls visibility of options.
+* Help users discover the underlying elytron model thanks to the echo-dmr command 
+to print out the set of sent model operations.
+* Allow user to visualise the management or application authentication configuration.
+
+=== Hard Requirements
+
+* For SASL supported mechanisms are EXTERNAL, DIGEST-MD5. PLAIN, JBOSS-LOCAL-USER, SCRAM-*, DIGEST-SHA-*
```
marekkopecky: Code `*, DIGEST-SHA-*` is evaluated as bold font and `, DIGEST-SHA-`

marekkopecky: See https://github.com/jfdenise/wildfly-proposals/blob/master/cli/EAP7-943_Security_Auth_Commands.adoc

cli/EAP7-943_Security_Auth_Commands.adoc
```diff
@@ -0,0 +1,241 @@
...
+                                    [--new-auth-factory-name=<a string>]
+                                    [--new-realm-name=<a string>]
+                                    [--new-security-domain-name=<a string>]
+                                    [--no-reload] [--super-user]_
+ 
+ 
+The mininal command to associate the out of the box elytron SASL factory to the HTTP interface +
+_security enable-sasl-management_
+ 
+The mininal command to enable EXTERNAL:+
```
marekkopecky: A space should probably be before '+' char, like on line 84 ( https://github.com/jfdenise/wildfly-proposals/blob/master/cli/EAP7-943_Security_Auth_Commands.adoc )

cli/EAP7-943_Security_Auth_Commands.adoc
```diff
@@ -0,0 +1,241 @@
...
+ 
+_security disable-http-auth-http-server --security-domain=<a string>
+                                           [--mechanism=<a string>]
+                                           [--no-reload]_
+=== Nice-to-Have Requirements
+
+=== Non-Requirements
+
+No support for mechanisms not listed in Hard Requirements. +
+These actions are not available in domain mode.
```
marekkopecky: Analysis could contains the reason for this. This should be also in documentation.

AFAIK (thanks to @mchoma's explanation), CLI can't support domain, because Elytron doesn't support domain. And Elytron can't support domain because Domain doesn't provide mechanisms for keystore distribution to domain servers ...

marekkopecky:
I checked CLI's point of view. Security's point of view of this analysis needs to be checked by @mchoma

mchoma:
We have had long discussion about the topic in https://developer.jboss.org/wiki/AuthenticationCommandsForCLI and this version is what we had agreed on there.

However now another note came to my mind while reading:
* There is also possibility to configure multiple HTTP authentication mechanisms in authentication factory. So it is generally anallogy to SASL mechanisms way. So to be consistent we should either remove reorder operation for SASL mechanisms or add it to HTTP mechanisms.

cli/EAP7-943_Security_Auth_Commands.adoc
```diff
@@ -0,0 +1,241 @@
...
+
+* https://issues.jboss.org/browse/WFCORE-3672
+
+=== Related Issues
+
+* https://issues.jboss.org/browse/EAP7-943
+
+=== Dev Contacts
+
+* mailto:{email}[{author}]
```
rhusar: Fill in your info here btw

marekkopecky: Which info do you prefer? Btw, mail and author is filled by properties, see the output: https://github.com/jfdenise/wildfly-proposals/blob/master/cli/EAP7-943_Security_Auth_Commands.adoc

rhusar: Ah, my bad! Cool then

jfdenise: @mchoma , for SASL, the order is important, the first mechanism is sent to the client.
For http all mechanisms are sent, so the order is not meaningful.

mchoma: @jfdenise you are right, I haven't realized it.

cli/EAP7-943_Security_Auth_Commands.adoc
bstansberry: It's unclear to me what effects these commands have on the target process, beyond the obvious one of changing the management model and its persisted xml config file. For example, is the target process actually creating the properties files if the command params specify those, or is this just a case of manipulating config to reference files that already exist?

I think the analysis should clarify this regardless, but I raise it here because it relates to the domain mode question @marekkopecky raised.

The xxx-management commands are not applicable to a domain overall, as there are no relevant domain-wide resources. They could apply to a HostController though. An HC is analogous to a standalone server. There's no question of distributing files to multiple hosts in the domain, as the command targets just one.

The xxx-http-server commands could work fine in a domain, so long as none of the commands are meant to change things outside the config model (e.g. create properties files, which would require non-existent distribution support.)

There are other things that would have to be handled for domain mode support, e.g. the user specifying what profile the xxx-http-server commands relate to, and what host the xxx-management commands relate to. 

I also think it's fine to say domain mode isn't covered just because it's out of scope for this version of the Feature but *may* be considered in the future. 

Anyway, I think some comment about why domain mode isn't covered should be added.

mchoma: That is good point. 

It seems to me these commands are meant to rely on existing artifacts. Otherwise they are missing many parameters. In origin document there was explicitly stated <cite>Adding users to properties files or file-system based is out of the scope of the following commands.<cite>

I also believe host for the xxx-management could be find out automatically and user does not have to provide this information.

jfdenise:
@marekkopecky , thank-you, I updated the docs with the issues you identified.

mchoma:
I have noticed Copy&Paste Error. In sentence "Associate an elytron HTTP factory to the security domain. In case a SASL factory already exists, ..."  should apparently be HTTP instead of SASL

jfdenise:
Thank-you, fixed.

cli/EAP7-943_Security_Auth_Commands.adoc
```diff
@@ -0,0 +1,241 @@
...
+* Drive users to select the right set of options thanks to CLI completion that 
+controls visibility of options.
+* Help users discover the underlying elytron model thanks to the echo-dmr command 
+to print out the set of sent model operations.
+* Allow user to visualise the management or application authentication configuration.
+
+=== Hard Requirements
+
+* For SASL supported mechanisms are EXTERNAL, DIGEST-MD5. PLAIN, JBOSS-LOCAL-USER, SCRAM-\*, DIGEST-SHA-*
+* For HTTP supported mechanisms are CLIENT-CERT, BASIC, DIGEST and FORM
```
mchoma: There is some confusion regarding CLIENT-CERT (servlet api name) vs. CLIENT_CERT (HttpServletRequest name). Elytron use primarly CLIENT_CERT, although understand CLIENT-CERT name. I think in this case we should talk rather about CLIENT_CERT authentication.

[1] https://issues.jboss.org/browse/JBEAP-7534

jfdenise: Thank-you, that is a typo. In help and implementation we do use CLIENT_CERT. I fixed proposal.

cli/EAP7-943_Security_Auth_Commands.adoc
```diff
@@ -192,26 +200,29 @@ _security disable-http-auth-management [--mechanism=<a string>]
...
+_security enable-http-auth-http-server --security-domain=<name> (
+                                          [--file-system-realm-name=<realm name>]
+                                          [--user-role-decoder=<role-decoder name>] |
+                                          [--user-properties-file=<file path>]
+                                          [--group-properties-file=<file path>]
+                                          [--relative-to=<system property>] |
+                                          [--properties-realm-name=<realm name>] | 
+                                          [--key-store-realm-name=<a string>] |
+                                          [--key-store-name=<key-store name>] )
+                                          [--mechanism=<mechanism name>]
```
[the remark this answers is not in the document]
jfdenise: Yes, no way to specify roles. To be honest, I didn't think at it. Do you think tha tis a common use-case?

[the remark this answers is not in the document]
mchoma: I would say it is pretty common. Then other question is what to expose, probably just --group-properties-file will be enough.

But maybe we just hit the border - "High level commands can do less then native elytron commands".

In practice CLIENT_CERT (auth) + LDAP store (authz) is pretty common. For puspose of "hello world" scenarios CLIENT_CERT + property file will be enough.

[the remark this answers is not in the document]
mchoma: I have make it work without roles, but it still make me thinking that it is much less useful than possibility to specify roles as well.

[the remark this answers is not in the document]
jfdenise: I looked at role + EXTERNAL SASL mechanism (I guess same for HTTP CLIENT_CERT) and it seems that we would need to:
1) allow to set an x500-principal-decoder if we want to map from a certificate to a user name
2) allow to provide the role-decoder (can be groups or roles properties file, could be a constant set of roles).
3) modify the SSL interactive support, it seems that the alias of the certificate inside the truststore needs to be exactly the same as what the X500 decoder decodes. Today we generate the alias, no way for a user to set it.

Would you agree with this? I would need to experiment with it to see the exact impact.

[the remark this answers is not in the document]
mchoma: Yes that will be the same.
1. +1
2. +1
3. +1 having alias option support seems useful in any case.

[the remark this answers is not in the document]
jfdenise: Having the EXTERNAL mechanism + roles derived from properties file group brings a complexity, a principal-decoder must be set on the security-domain to map from the DN to an entry in the mgmt-groups.properties file. Unfortunately the DN can't be used in the file directly. I am not confortable in making changes to the security domain. This can impact all realms present in it. There doesn't seem to have ways to set the principal-decoder on the referenced realm inside the security domain. Perhaps @fjuma can comment on this. I would prefer not to deal with x500-principal-decoder, this implies the trust-store alias to be aligned with the decoder output.

Management groups are legacy, a constant role mapper could be more adequate in our case. We could propose to re-use an existing constant-role-mapper or provide a list of roles something like: --roles and --role-mapper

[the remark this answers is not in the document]
fjuma: Instead of x500-principal-decoder, another option might be to specify a principal-transformer on the referenced realm.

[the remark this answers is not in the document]
jfdenise: @fjuma , thank-you, yes that works. I noticed that the user identity is not impacted by a transformer (the DN is kept), seems only used by the authentication layer. With a decoder, the identity is replaced by what the decoder returns. That is visible using whoami op.
One source of possible complexity is the fact that in both cases (decoder or transformer),  the returned "identity" must be a valid alias in the trust-store. Why do we have this limitation? For information I am using an aggregate-realm containing a key-store-realm and a properties file realm.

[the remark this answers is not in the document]
fjuma: One thing to note is that the principal-decoder resource also exposes the principal-transformer capability. This means that a principal-decoder can also be referenced as a principal-transformer on the mechanism-configuration and server-ssl-context - more details on this can be found in https://issues.jboss.org/browse/JBEAP-11084.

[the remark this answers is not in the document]
jfdenise: @fjuma , so it seems that the way to go is to use a decoder as a realm transformer. In this case the user identity is what the decoder decodes. This seems what I would expect as a user when setting some kind of decoder.
I looked at https://issues.jboss.org/browse/JBEAP-11084 but don't see alias related information. I can see an HHTP/2 issue, but the 1 to 1 mapping between alias and decoder output is already taken into account as a constraint.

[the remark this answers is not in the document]
mchoma: To map between client principal to key-store alias I think you need post-realm-principal-transformer as described here http://darranl.blogspot.cz/2017/07/wildfly-elytron-principal-transformers.html
```
    The Principal created by the pre-realm-principal-transformers is: -
        The Principal used to map the SecurityRealm
        The Principal that will be associated with the resulting SecurityIdentity.
    The Principal created after the final principal transformers is: -
        The Principal that will be passed to the SecurityRealm to obtain the RealmIdentity.
```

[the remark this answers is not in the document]
jfdenise: I guess that the problem comes from the aggregate-realm. I am setting the principal-transformer on it so the authentication realm is impacted by it and requires a valid alias in the trust-store. I should be able to set a principal-decoder for the authorization realm only. @fjuma does it make sense?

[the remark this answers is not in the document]
jfdenise: FYI I have logged https://issues.jboss.org/browse/WFCORE-3767 that would separate alias and principal in this case. For now anyway we are relying on a constant set of roles.

PR #34, opened 2018-03-13
WFLY-9417 Add ability to control deployment start behaviour

stuartwdouglas:


kabir:
@jstourac please take a look

undertow/WFLY-9417_deployment_start_behaviour.adoc
```diff
@@ -0,0 +1,74 @@
...
+by an incomplete deployment.
+
+== Test Plan
+
+This is complex to test, as our existing test infrastructure is not set up to perform
+tests while the server is starting up. Such a test would also likely be fragile as
+it would be totally dependent on the amount of time the server takes to start.
+
+It is anticipated that automated testing cannot reasonably be done for this feature,
+and as such manual verification will be performed only.
```
jstourac: we have some basic tests for graceful shutdown in our internal testsuite, I think these could be extended to cover this using SCI

jstourac: This section might be updated as there have been created some simple tests in our internal TS. Not sure whether we should mention internal testsuites here or whether we should keep such information just in tests-plans...

kabir:
@stuartwdouglas please update and then we should be able to merge this and the main PR

PR #37, opened 2018-03-20
WFLY-9724 Allow unescaped characters in URL

stuartwdouglas:


undertow/WFLY-9724_Allow_unescaped_characters_in_URL.adoc
```diff
@@ -0,0 +1,46 @@
...
+:toc:               left
+:icons:             font
+:keywords:          comma,separated,tags
+:idprefix:
+:idseparator:       -
+:issue-base-url:    https://issues.jboss.org
+
+== Overview
+
+This feature is to add an additional an addition option into the Undertow subsystem configuration to allow unescaped
```
jstourac: `an additional an addition` typo

undertow/WFLY-9724_Allow_unescaped_characters_in_URL.adoc
```diff
@@ -0,0 +1,46 @@
...
+
+=== Related Issues:
+
+* {issue-base-url}/EAP7-965[EAP7-965]
+
+=== Dev Contacts:
+
+* mailto:{email}[{author}]
+
+=== QE Contacts:
```
jstourac: feel free to put me here

undertow/WFLY-9724_Allow_unescaped_characters_in_URL.adoc
```diff
@@ -0,0 +1,46 @@
...
+
+This will be implemented as an additional listener option under the Undertow subsystem called
+`allow-unescaped-characters-in-url`. This will be false by default for security reasons, however if it is set to true
+then this will let Undertow process unescaped non ASCII characters in the URL.
+
+=== Test Plan
+
+A new test will be created that adds a new listener with this option enabled. A request with UTF-8 in the URL
+will be sent to both the new listener and the current listener, it is expected that the new listener will reject it
+and the existing listener will process it.
```
jstourac: hm, I think the test behavior should be vice versa, that means: new listener (`allow-unescaped-characters-in-url` = true) will process that request whether existing listener (`allow-unescaped-characters-in-url` = false) will reject it.

undertow/WFLY-9724_Allow_unescaped_characters_in_URL.adoc
stuartwdouglas: fixed

undertow/WFLY-9724_Allow_unescaped_characters_in_URL.adoc
stuartwdouglas: Done

undertow/WFLY-9724_Allow_unescaped_characters_in_URL.adoc
stuartwdouglas: Fixed, this test is now part of the pull request

jstourac:
Thank you for the update, Stuart. I think this is good to go.

undertow/WFLY-9724_Allow_unescaped_characters_in_URL.adoc
```diff
@@ -0,0 +1,49 @@
...
+
+=== Affected Projects or Components:
+
+* Wildfly
+
+== Requirements
+
+This will be implemented as an additional listener option under the Undertow subsystem called
+`allow-unescaped-characters-in-url`. This will be false by default for security reasons, however if it is set to true
+then this will let Undertow process unescaped non ASCII characters in the URL.
```
rsvoboda: This can be really nice source of CVEs :) 

What is the main motivation to support unescaped characters in URL ?

stuartwdouglas: Because people want it (e.g. it appears that some older versions of IE don't encode properly, and potentially other clients as well).

The sort of CVE that results from this is in general mostly theoretical, to actually exploit it you need a frontend that is both responsible for security and is also interpreting the characters in a different manner to the backend.

rsvoboda:
I know IE 8 had some troubles and many pages ended up with `<meta http-equiv="X-UA-Compatible" content="IE=7" />`

But that's history, currently only IE 11 and Microsoft Edge are to be considered.
IE 10 and lower are unsupported - https://www.whatismybrowser.com/guides/the-latest-version/internet-explorer

Client's which do not support url encoding - no strong opinion on this, but I lean towards 'it's not our concern, clients should support url encoding' position.

I read https://issues.jboss.org/browse/UNDERTOW-1185 complain, did you notice other requests on forums ?
I'm concerned from product perspective, in general I do not like many configuration options as it potentially generates huge config matrices for end user to end up with / for us to support.

stuartwdouglas:
I think Korean users still need to support older versions of IE (and obviously rely on non-ascii characters), the issue that generated this was https://issues.jboss.org/browse/JBEAP-13472 (quite a long conversation in the comments with the details).

rsvoboda:
JBEAP-13472 is pretty huge, support involved too.

You made it to the final round and have my ack ;)

PR #38, opened 2018-03-22
[WFCORE-3522] Part 1: Validate deployment content

dmlloyd:


yelhouti:
It would be great to fix this.
In my case when registering a module manually in standalone.xml my code works. but when testing it with arquillian it doesn't, specifically for this reason.

deployment/WFCORE-3522_Deployment content validation.adoc
```diff
@@ -0,0 +1,67 @@
...
+* All deployment-related subsystems
+
+=== Other Interested Projects
+
+== Requirements
+
+=== Hard Requirements
+
+* The deployer must provide a way for subsystems to register packages and resources that are indicative of invalid deployment content.
+This likely entails introducing a multi-valued capability which is consumed by the deployment system.
```
mjurc: Is the integration of subsystems with the new capability a concern of this feature? Or is this feature just about the implementation of capability itself, and subsystems using it will be a later step?

dmlloyd: That was my initial plan.  I don't really know if it's realistic to expect 100% coverage at first (maybe it is) but my thought was that if we don't put it in subsystems right away, the feature would not be useful.

That said, I think that any number of subsystems could be verified using just one test where one resource is generated into each forbidden package and verified in a loop, or something like that.  Then it's just a question of ensuring that we found all the packages we want to forbid, which I think would have to be a manual verification.

deployment/WFCORE-3522_Deployment content validation.adoc
```diff
@@ -0,0 +1,72 @@
...
+** The deployer must correlate this information in such a way that if the subsystem is disabled for a given deployment via ``exclude-subsystem`` in ``jboss-deployment-structure.xml``, the deployment can proceed without error.
+** There may be APIs that must always be forbidden, even when the ``exclude-subsystem`` flag is set.
+In these cases, the subsystem must have a way to indicate this situation.
+* The deployer must ignore invalid content.
+* The deployer must produce clear, user-friendly warnings about invalid content, ideally with a suggested remedy.
+
+=== Nice-to-Have Requirements
+
+* Implement an optional deployment mode where users could specify that they wish to reject deployments with invalid content.
+This could become the default setting in some future version.
```
mjurc: Is there any reasoning against default fast fail on deployment? I think it could be more user friendly (for starters).

deployment/WFCORE-3522_Deployment content validation.adoc
```diff
@@ -0,0 +1,72 @@
...
+== Requirements
+
+=== Hard Requirements
+
+* The deployer must provide a way for subsystems to register packages and resources that are indicative of invalid deployment content.
+This likely entails introducing a multi-valued capability which is consumed by the deployment system.
+** The deployer must correlate this information in such a way that if the subsystem is disabled for a given deployment via ``exclude-subsystem`` in ``jboss-deployment-structure.xml``, the deployment can proceed without error.
+** There may be APIs that must always be forbidden, even when the ``exclude-subsystem`` flag is set.
+In these cases, the subsystem must have a way to indicate this situation.
+* The deployer must ignore invalid content.
```
mjurc: What about case when deployment contains enriched API or overriding content with additional dependencies that the 'valid' deployment content depends on?

deployment/WFCORE-3522_Deployment content validation.adoc
dmlloyd: It would be a compatibility breakage to do so, unfortunately.

dmlloyd: One possibility would be to use an attribute to configure whether invalid content should be ignored or rejected.  The default shipped configuration could be our preferred behavior which is reject invalid content, but the transformers for older model versions could use a default value of "ignored" for compatibility.  @bstansberry are we allowed to do things like that?

deployment/WFCORE-3522_Deployment content validation.adoc
dmlloyd: Since packages cannot be split, it is not allowed for deployments to supplement container-provided APIs with additional classes.  Additional packages which are not part of container-provided APIs are allowed.

Container classes cannot link to deployment classes, but they can load them (via ServiceLoader for example).  Some APIs (like JAXP or JSONP) support loading different per-deployment implementations.  Some (like JSF) support complete replacement of the API, but as stated in the requirements, users must exclude the subsystem to do so.  Some container provided APIs cannot be implemented or replaced by deployments, like JTA.

I've been working on [a spreadsheet](https://docs.google.com/spreadsheets/d/1sH0L8XLulQrMzXd8-ZWnxbej9xsD6W-NKzf9BJ7xB-c/edit#gid=0) which lists all of our container-available APIs and their level of replacement/implementation support.  It is not yet complete but it starting to give a picture.

mjurc: Thanks for the summary!

With regards to subsystems that do not allow to be disabled/overriden from deployments, is it planned to ensure that all the subsystems will allow deployments to disable/override them in future iterations of this feature? I kind of agree with your previous assessment that this feature would be way more useful and usable if all of the subsystems supported it.

dmlloyd: @mjurc I don't yet know if it's possible for every subsystem to support disabling/excluding itself from deployments.  Every subsystem that can possibly support doing so should be made to allow doing so.

deployment/WFCORE-3522_Deployment content validation.adoc
bstansberry: Answering a very old question I just saw. :)

No that would not be valid. Behavior should be consistent across servers using a profile, so if the standard config specifies something a server can't support that should result in a failure when the server joins the domain.

Our standard profiles need not work with legacy servers. Any time we introduce a new subsystem they don't. Users who want to run older version servers should adapt our profiles to suit their needs or introduce new ones tailored to those older servers. The EAP docs emphasize the latter approach.

It is certainly possible though to have the default value for a new attribute be different from what our standard configs specify, with the default resulting in previous behavior. That can be helpful for migration as existing configs still work as they did before.

PR #42, opened 2018-03-27
WFLY-8449 Return Context Data to EJB Clients

stuartwdouglas:


ejb/WFLY-8449_Return_context_data_to_client.adoc
```diff
@@ -0,0 +1,81 @@
...
+
+=== Hard Requirements
+
+An `EJBClientInterceptor` will be able to request data that it is interested in by calling
+`org.jboss.ejb.client.EJBClientInvocationContext#addReturnedContextDataKey(String key)`. This tells the server that the
+client interceptor would like any data that is stored under the provided key in the context data map to be returned
+to the client.
+
+Internally this will create/add to a `Set` in the clients context data stored under
+`org.jboss.ejb.client.invocation.returned-context-data`, which will then be sent to the server as part of the invocations
```
dmlloyd: This key is pretty long.  Our keys start with just ``jboss.`` and the ``ejb.client.invocation`` is all inferrable (becuase it's an EJB invocation).  Maybe ``jboss.returned-keys``?

ejb/WFLY-8449_Return_context_data_to_client.adoc
```diff
@@ -0,0 +1,81 @@
...
+
+=== Nice-to-Have Requirements
+
+=== Non-Requirements
+
+It is not a requirement that all data in the context data map be returned to the client. This approach would
+be problematic for several reasons:
+
+ *  Most invocations simply do not want or need this data, so for the majority of clients this would increase network
+ and CPU usage for no reason
```
dmlloyd: This will cause a disparity with local remote invocation.  Maybe we should consider then dropping context data on local remote invocations as well unless the key is in the set?

ejb/WFLY-8449_Return_context_data_to_client.adoc
```diff
@@ -0,0 +1,81 @@
...
+ * There is no requirement that data that is placed in the context data map be Serializable. The only way to actually
+ determine this is to try and serialize it. As such any working implementation of this would need to catch and ignore
+ exceptions which happen during the Serialization process, or break backwards compatibility by causing invocations that
+ previously would have worked to fail.
```
dmlloyd: The definition of compatibility is that code which meets the specification (whatever that may be) continues to work, and that the specification is backwards-compatible.  It is possible to write code which works yet can reasonably break because it does not follow specification.  If there is a case where the specification is wrong, then we would retroactively update the specification and fix any interim problems as bugs.

Therefore we're not breaking compatibility because previously working invocations begin to fail.  If anything it would be because the specification did not evolve compatibly.  In this case we broke compatibility when we stopped sending context data back to the client in JBoss AS 7.  So, only selectively sending back context data is already breaking compatibility with our previous specification (such as it is).

That said, evolving the specification to not require serializable data in every key _is_ a backwards-compatible change (as long as we accept the incompatible change of restricting what context data is sent back).  But it is not going to be possible to just catch exceptions during serialization and then carry on like nothing has happened.  I think that it must be specified that keys listed in the returned-data set _must_ be serializable, otherwise invocation may fail.  Other keys may or may not be serializable (we don't care, as they will be dropped).

ejb/WFLY-8449_Return_context_data_to_client.adoc
dmlloyd: Thinking more about this, I think we _must_ make this change.  All the problems which are described with regards to serialization also apply with cloning; it's just less likely that a class might be missing in this case.

ejb/WFLY-8449_Return_context_data_to_client.adoc
```diff
@@ -0,0 +1,81 @@
...
+ * More seriously the client may not always be able to deserialize this context data, as it may not have the relevant
+ class definitions on the class path. This will break deserialization and cause applications that were previously working
+ to fail, breaking our backwards compatibility guarantees.
```
dmlloyd: This isn't a (non-)requirement per se, but I think it's safe to require that any key that the user is explicitly listing as being sent back to the client _must_ be present on the client, in the same way that return values must be.  Again we really can't do much to recover if an exception is thrown.  At least not without a great deal of work.

ejb/WFLY-8449_Return_context_data_to_client.adoc
```diff
@@ -0,0 +1,81 @@
...
+=== Other Interested Projects
+
+== Requirements
+
+=== Hard Requirements
+
+An `EJBClientInterceptor` will be able to request data that it is interested in by calling
+`org.jboss.ejb.client.EJBClientInvocationContext#addReturnedContextDataKey(String key)`. This tells the server that the
+client interceptor would like any data that is stored under the provided key in the context data map to be returned
+to the client.
```
dmlloyd: We might want to consider something more generic; if it is reasonable for one server-side interceptor to use context data to pass information between interceptors (but not the client), then surely it's reasonable for the client side to do the same (without passing to the server)?  Wouldn't we want the set of keys to be marked as generically transportable, instead of sending all keys from client to server but only some from server to client?

ejb/WFLY-8449_Return_context_data_to_client.adoc
stuartwdouglas: I am pretty sure this data is not cloned on local invocations, at least it did not appear to be when I was looking at the code, I will double check tomorrow.

ejb/WFLY-8449_Return_context_data_to_client.adoc
stuartwdouglas: makes sense

ejb/WFLY-8449_Return_context_data_to_client.adoc
stuartwdouglas: The EJB specification does not impose any serialisability requirement on context data. It is also not in any way implied it is available to remote clients. The EJB/interceptors spec uses the context data as a way to pass data between interceptors. 

Adding an additional serialisability requirement to something that is governed by the EE specs is a break in compatibility.

It is ok for us to have a selective mechanism for them to be returned, but we can't just start serialising them all.

ejb/WFLY-8449_Return_context_data_to_client.adoc
stuartwdouglas: Yes, this is describing a consequence that would happen if we just sent back everything. If a user specifies a key they want, it is up to them to make sure they can handle it

ejb/WFLY-8449_Return_context_data_to_client.adoc
stuartwdouglas: One option that Brad suggested would be for the client to provide a regex that can match key names

ejb/WFLY-8449_Return_context_data_to_client.adoc
dmlloyd: When I say "the specification" I mean "the way JBoss promises this stuff will work", which is generally a superset of the Java EE (and other) specifications.

ejb/WFLY-8449_Return_context_data_to_client.adoc
dmlloyd: Using a regex makes sense.  I'm just saying that we should treat client-to-server the same way we're treating server-to-client.

dmlloyd: Having just one regex key would probably be substantially more efficient than a set in most cases, at any rate.

ejb/WFLY-8449_Return_context_data_to_client.adoc
dmlloyd: Additional constraints, sure.  The more you promise, the larger the set of allowed behaviors.  The more you constrain, the smaller the set.  We offer features and behaviors beyond Java EE, for example, and we cannot constrain Java EE-specified behavior to a subset without breaking compliance.

stuartwdouglas: Yes, but this particular bit of the document is talking about problems that arise if you just try and serialise everything. If you attempt that then you add an additional requirement that everything that is in this map is serialisable and available on the client class path, neither of which are required by the EE spec. The serialisability one is easy enough to hack around on the server side, but AFAIK marshalling cannot recover from a failure on the client side due to a class not being available. 

In addition to this the majority (99%+) of deployments will have no client interceptors that can make use of this data anyway, so the data would simply be discarded on the client side.

The reason why this has been explicitly singled out is because so far all discussion on various JIRAs has been about making the whole thing available to every invocation. This section is about the problems that arise with this approach. 

For the majority of customers that have no interest in this feature just sending all the data back means the following:
- Extra CPU usage for serialization
- Extra network utilisation for the extra data that will be discarded
- The possibility that existing applications will stop working because a server side class is not present on the client, even though the client has no interest in that class, and the server is simply using standard EE spec functionality. 

That is why this is explicitly outlined in this section. This is a feature that will not see broad use, if you are not using it then it should have zero impact.

ejb/WFLY-8449_Return_context_data_to_client.adoc
mjurc: +1 for regex.

ejb/WFLY-8449_Return_context_data_to_client.adoc
mjurc: The issue with backwards compatibility also has other dimension - in between minor releases of EAP, we promise the users their applications will keep on working without outsides change in behaviour inside single major release. Breaking changes should not happen. If somebody starts using their new feature in the existing application wrongly and it breaks, however, that's a different story.

So, to be concise, applications that worked in EAP 7.x have to work in 7.2 out of the box without change in behaviour.

dmlloyd: OK. Even with the regex solution, we'd technically we'd be breaking compatibility with any existing app which happened to use our exact key name (unlikely as that may be); it might be a good idea to document (somewhere) that undefined attachment keys starting with "jboss" or "org.jboss" are reserved and are not part of any compatibility guarantees.

ejb/WFLY-8449_Return_context_data_to_client.adoc
dmlloyd: As per client/server discussion below, maybe ``jboss.send-keys`` or something woudl be better.

ejb/WFLY-8449_Return_context_data_to_client.adoc
mjurc: Agreed.

stuartwdouglas:
So the issue with regex is that all the known use cases so far basically involve and server and client side interceptor communicating by putting data in a specific key. I really don't want to add regex support because I really don't see a valid use case for a client interceptor to start trying to examine an undefined number of keys.

If a client really does have a use case for this then it is very easy to implement by simply installing a server side interceptor that does the inspection, and sends the relevant data to the client under a known key.

Regex is an idea that sounds good on the surface, but IMHO anything that is sent over the wire should be carefully considered, rather than just vacuumed up as a regex glob.

ejb/WFLY-8449_Return_context_data_to_client.adoc
stuartwdouglas: I don't really understand what you mean here

ejb/WFLY-8449_Return_context_data_to_client.adoc
stuartwdouglas: So the issue with regex is that all the known use cases so far basically involve and server and client side interceptor communicating by putting data in a specific key. I really don't want to add regex support because I really don't see a valid use case for a client interceptor to start trying to examine an undefined number of keys.

If a client really does have a use case for this then it is very easy to implement by simply installing a server side interceptor that does the inspection, and sends the relevant data to the client under a known key.

Regex is an idea that sounds good on the surface, but IMHO anything that is sent over the wire should be carefully considered, rather than just vacuumed up as a regex glob.

ejb/WFLY-8449_Return_context_data_to_client.adoc
dmlloyd: Sending keys from client to server should work the same way as sending keys from server to client.  When sending from client to server, we're not "returning" the key, we're simply "sending" it.

The latter question is, do we always include the special key itself in the set of keys transmitted between client and server, or do we exclude it?

ejb/WFLY-8449_Return_context_data_to_client.adoc
dmlloyd: If we just send one key, we'll eventually get a request to send multiple.  If we send a set, we'll eventually get a complaint that the set is too big.  I think a regex is our best shot so far at not having to revisit this issue in the future, and it has the advantage of being a simple string, which is pretty hard to mess up.

stuartwdouglas: The customer controls both the server and client interceptors that send and receive the data. The server interceptor can send what is needed, the client interceptor can receive it. If they really actually want to just sent everything they can just have the interceptor put a copy the context data map into the context data.

bmaxwell: The current behavior with client interceptors is: anything they put into the contextData is sent to the server. I have not really seen any users using more than 1 client interceptor.  The regex, I was mainly thinking if a user made heavy use of this, they may just want everything returned so a *.  I have not really seen any users sending contextData through multiple interceptors, the use cases are typically client <-> server.  And if I were writing an app, I would call contextData.remove() if I did not want it passed along.

stuartwdouglas:
I have updated this (and the corresponding WF PR) to make the key name specification a regex.

ejb/WFLY-8449_Return_context_data_to_client.adoc
```diff
@@ -0,0 +1,81 @@
...
+map to be returned to the client.
+
+Internally this will create/add to a `Set` in the clients context data stored under
+`org.jboss.ejb.client.invocation.returned-context-data`, which will then be sent to the server as part of the invocations
+context data, so no protocol change is required.
+
+If the server receives this set of keys then after the invocation has completed it will iterate the set of keys,
+and if a matching regular expression is found in the context data of the invocation then it will be copied into the
+context data map that is returned to the remote client. Any EJBClientInterceptor will be able to examine this data
+in the invocations context data map.
```
dmlloyd: So there's a set _and_ a regex on both sides?

dmlloyd: If it's going to be a set then regex is probably not better than going back to a set of strings.  Thinking more about it, if we send (for example) a HashSet of keys to include, if any of those keys overlaps with the set of keys already in the request, then they'll be deduplicated.  In the case of "here's a set of strings that I'm sending you and I want back again", the overhead is only an array of backreferences (most likely no more than two bytes each).

So I guess I've flipped back to being in favor of a set of plain strings.

ejb/WFLY-8449_Return_context_data_to_client.adoc
```diff
@@ -0,0 +1,81 @@
...
+
+=== Related Issues
+
+* https://issues.jboss.org/browse/EAP7-976[EAP7-976]
+
+=== Dev Contacts
+
+* mailto:{email}[{author}]
+
+=== QE Contacts
```
mjurc: Feel free to put me here

PR #43, opened 2018-04-03
[WFCORE-3596] Introduce named permission sets in the Elytron subsystem configuration

fjuma:
https://issues.jboss.org/browse/WFCORE-3596

### DIFF
--- a/elytron/WFCORE-3596-named-permission-sets.adoc
+++ b/elytron/WFCORE-3596-named-permission-sets.adoc
@@ -1,0 +1,112 @@
+= WFCORE-3596 Add support for named permission sets to the Elytron subsystem configuration
+:author:            Farah Juma
+:email:             fjuma@redhat.com
+:toc:               left
+:icons:             font
+:keywords:          comma,separated,tags
+:idprefix:
+:idseparator:       -
+:issue-base-url:    https://issues.jboss.org/browse
+
+== Overview
+
+Currently, permission mappings and constant permission mappers in the Elytron subsystem configuration
+can only reference individual permissions. This task is to introduce named permission sets and allow
+these to be referenced from permission mappings and constant permission mappers. Referencing individual
+permissions from permission mappings and constant permission mappers will still be allowed but will
+deprecated to discourage further use.
+
+== Issue Metadata
+
+=== Issue
+
+* https://issues.jboss.org/browse/WFCORE-3596[WFCORE-3596]
+
+=== Related Issues
+
+* https://issues.jboss.org/browse/EAP7-978[EAP7-978]
+
+=== Dev Contacts
+
+* mailto:{email}[{author}]
+
+=== QE Contacts
+
+* TBD
+
+=== Affected Projects or Components
+
+* WildFly, Security
+
+=== Other Interested Projects
+
+== Requirements
+
+=== Hard Requirements
+
+* It should be possible to configure named permission sets in the Elytron subsystem configuration. Each permission
+  set should contain zero or more permissions. It should be possible to reference permission sets from permission
+  mappings and from constant permission mappers. As an example, the default configuration would be as shown below.
+  Notice that the default permissions will be in a default permission set, "default-permissions", that is referenced by
+  the "default-permission-mapper". For the provisioning tool, the default permission set will be the single place to add
+  permissions as subsystems are being added and it will be the single place to remove permissions as subsystems are being
+  removed. Without named permission sets, the provisioning tool would not know which permission mapping a subsystem
+  permission should be added to or removed from. Note that the provisioning tool will only be touching the "default-permissions".
+  It won't attempt to touch any user-specified permission sets.
+
+[source,xml]
+----
+<mappers>
+...
+    <simple-permission-mapper name="default-permission-mapper" mapping-mode="first">
+        <permission-mapping>
+            <principal name="anonymous"/>
+            <permission-set name="default-permissions"/>
+        </permission-mapping>
+        <permission-mapping match-all="true">
+            <permission-set name="login-permission" />
+            <permission-set name="default-permissions"/>
+        </permission-mapping>
+    </simple-permission-mapper>
+...
+</mappers>
+
+<permission-sets>
+    <permission-set name="login-permission">
+        <permission class-name="org.wildfly.security.auth.permission.LoginPermission"/>
+    </permission-set>
+    <permission-set name="default-permissions">
+        <permission class-name="org.wildfly.extension.batch.jberet.deployment.BatchPermission" module="org.wildfly.extension.batch.jberet" target-name="*"/>
+        <permission class-name="org.wildfly.transaction.client.RemoteTransactionPermission" module="org.wildfly.transaction.client"/>
+        <permission class-name="org.jboss.ejb.client.RemoteEJBPermission" module="org.jboss.ejb-client"/>
+    </permission-set>
+</permission-sets>
+----
+
+* A permission set should be a private capability.
+
+* When transforming the model for legacy slaves, the permission sets should be in-lined, i.e., a permission set that's
+  referenced from a permission mapping or from a constant permission mapper would just be replaced with the permissions
+  that make up the permission set. For example, the "default-permissions" permission set referenced from the permission
+  mappings in the above example would be transformed to:
+
+[source,xml]
+----
+<permission class-name="org.wildfly.extension.batch.jberet.deployment.BatchPermission" module="org.wildfly.extension.batch.jberet" target-name="*"/>
+<permission class-name="org.wildfly.transaction.client.RemoteTransactionPermission" module="org.wildfly.transaction.client"/>
+<permission class-name="org.jboss.ejb.client.RemoteEJBPermission" module="org.jboss.ejb-client"/>
+----
+
+* Referencing individual permissions from permission mappings and constant permission mappers should still be allowed but
+  should deprecated to discourage further use.
+
+* Since we are not removing the permission element from permission mappings or constant permission mappers, there is no
+  need to migrate XML configuration for prior versions to the new version.
+
+=== Nice-to-Have Requirements
+
+=== Non-Requirements
+
+== Test Plan
+
+Subsystem parsing and transformer tests will be added.


mchoma:
IMHO, with suggested solution only benefit for provisioning tool is it knows in initial default configuration it should look into default-permissions. So don't need to look into multiple places, but just one place. But still must somehow identified items matching to affected subsystem (most probably based on module attribute)

When users starts to edit permissions-sets provisioning tool may be in same situation as now. It will have to iterate multiple possible places to look for change.

As we understood from discussion [1], problem is permission items are missing identifiers. Which is not resolved by moving permissions to permission-set because they still have not identity in permission-set.

What am I missing here?

[1] http://lists.jboss.org/pipermail/wildfly-dev/2018-March/006410.html

aloubyansky:
@mchoma you are right in a sense that this change doesn't address the main issue which is the ability to fine-tune this part of the configuration by a user (and/or merge changes during patch/update application or integration of other products into the installation) using the provisioning mechanism. By "this part of the configuration" I mean the list of the permission-sets in the permission mapper. The outcome of the discussion was that we don't need to support this kind of flexibility in this case. The list has to be treated as an atomic value. In other words, the mechanism (unlike the administrator) should not view it as a list of individual items but a single value and present it to the administrator as such.

mchoma:
Yes, it is as I understood this sentence

"So for now I am going to handle these kind of config
structures (i.e. lists of complex types) as atomic pieces (although
potentially big). They won't be mergeable during updates but one will be
replacing completely the other with the one coming in last winning."

But I thought you are talking about existing permissions LIST from model here and there is no need to create another LIST for this purpose.

elytron/WFCORE-3596-named-permission-sets.adoc
```diff
@@ -0,0 +1,91 @@
...
+    </permission-set>
+    <permission-set name="default-permissions">
+        <permission class-name="org.wildfly.extension.batch.jberet.deployment.BatchPermission" module="org.wildfly.extension.batch.jberet" target-name="*"/>
+        <permission class-name="org.wildfly.transaction.client.RemoteTransactionPermission" module="org.wildfly.transaction.client"/>
+        <permission class-name="org.jboss.ejb.client.RemoteEJBPermission" module="org.jboss.ejb-client"/>
+    </permission-set>
+</permission-sets>
+----
+
+* When transforming the model for legacy slaves, the permission sets should be in-lined.
```
mchoma: Could you elaborate more on this? Does that mean <permission-set> will be resolved to list of <permission> ?

elytron/WFCORE-3596-named-permission-sets.adoc
```diff
@@ -0,0 +1,91 @@
...
+        <permission-mapping match-all="true">
+            <permission-set name="login-permission" />
+            <permission-set name="default-permissions"/>
+        </permission-mapping>
+    </simple-permission-mapper>
+...
+</mappers>
+
+<permission-sets>
+    <permission-set name="login-permission">
```
mchoma: Will be permission set a capability?

elytron/WFCORE-3596-named-permission-sets.adoc
mchoma: Is there also automatic transformation to new version of model. I mean when I take WF12 standalone.xml and will put it in WF13 and restart the server I will end up with permission set representation?

In other words when I use WF12 standalone.xml in WF13, will I be able to manipulate permission-sets in permission-mapper ?

aloubyansky:
Let's clarify whether this change makes sense for Elytron other than do something for the provisioning. I assume it does. My approval was not so much about fixing the provisioning issue but more in a sense that we'll be OK with the change. If there was no other purpose for the change than I personally don't see it as necessary simply because it doesn't really avoid the complex atomic config structure.
Then whether it somewhat simplifies or brings in more complexity to the Elytron config will depend on how the permission-set is represented in the domain management model.

aloubyansky:
Sorry, I forgot about the obvious and important thing in the proposal (besides the structural changes) which is the single location where the permissions get collected - the default permission set. Which is the key of it.

elytron/WFCORE-3596-named-permission-sets.adoc
fjuma: It will be a private capability.

elytron/WFCORE-3596-named-permission-sets.adoc
fjuma: When transforming the model, the permission-set will be replaced with the permissions that make up the permission-set. 

XML configuration for prior versions will be converted to the new version, i.e., each list of permissions will be converted to a permission set.

fjuma:
I've updated the hard requirements in the analysis document based on the feedback.

In addition to providing a single collection point for permissions which is needed for provisioning, named permission sets simplify permission configuration by no longer requiring the same permission to be configured in multiple permission mappings and constant permission mappers.

mchoma:
@aloubyanskyAnd if mechanism is able to perform required change in one place, why is it problem to change in more places?

Amount of discussion around this topic shows something is weird here. But I can't identify what it is. Maybe it is only fact this is workaround for bigger problem. Then question is if this could be workarounded on provisioning mechanism level. @Brian I think your view will be also valueable here.

But as far as it will not affect users (no management API compatibility issues) and this will only change way of persisting permissions I am good with this change.

mchoma:
Ok I had a chat with alexey and I was missing 2 information.
- provisioning tool will touch only default-permissions. If I configure my-permission-set, provisioning tool will not touch that.
- with this change we are workarounding we can't name LIST in model (either permission-mappings, or permissions)

Could we get this information somehow to proposal, as this helped me to understand what new this proposal brings compared to current state.

elytron/WFCORE-3596-named-permission-sets.adoc
```diff
@@ -0,0 +1,108 @@
...
+  mappings in the above example would be transformed to:
+
+[source,xml]
+----
+<permission class-name="org.wildfly.extension.batch.jberet.deployment.BatchPermission" module="org.wildfly.extension.batch.jberet" target-name="*"/>
+<permission class-name="org.wildfly.transaction.client.RemoteTransactionPermission" module="org.wildfly.transaction.client"/>
+<permission class-name="org.jboss.ejb.client.RemoteEJBPermission" module="org.jboss.ejb-client"/>
+----
+
+* XML configuration for prior versions should be converted to the new version, i.e., each list of permissions should be
```
mchoma: Is this meant manually? If so, this is compatibility issue. I think user should be able to copy standalone.xml to newer version of server without issue.

fjuma: This will be done automatically.

elytron/WFCORE-3596-named-permission-sets.adoc
```diff
@@ -0,0 +1,108 @@
...
+  mappings in the above example would be transformed to:
+
+[source,xml]
+----
+<permission class-name="org.wildfly.extension.batch.jberet.deployment.BatchPermission" module="org.wildfly.extension.batch.jberet" target-name="*"/>
+<permission class-name="org.wildfly.transaction.client.RemoteTransactionPermission" module="org.wildfly.transaction.client"/>
+<permission class-name="org.jboss.ejb.client.RemoteEJBPermission" module="org.jboss.ejb-client"/>
+----
+
+* XML configuration for prior versions should be automatically converted to the new version, i.e., each list of permissions should be
```
mchoma: Could you elaborate on automatic conversion strategy? This does not seem to me as easy task.

mchoma:
There is probably alternative way to this proposal; Introduce required name attribute to permission-mappings. And named default permision-mappings with well-known names. 

Such alternative would be probably more light-weight compared to this proposal.

However it will suffer with same problem when migrating legacy config to newer version.

elytron/WFCORE-3596-named-permission-sets.adoc
fjuma: Updated the document with more information on this.

elytron/WFCORE-3596-named-permission-sets.adoc
```diff
@@ -0,0 +1,110 @@
...
+  converted to a permission set. In particular, each list of permissions in a permission mapping or constant permission mapper
+  should be automatically replaced with a reference to a newly created permission set that is made up of the list of permissions.
+  The name of the permission set could be a combination of the mapper name and the class-name of each permission in the list.
```
mchoma: But that won't help to provisioning tool which will manipulate only set named "default-permissions"

fjuma: @darranl Do you have any thoughts on this strategy for converting XML configuration for prior versions to the new version with named permission sets?

fjuma: The default permission set consists of the subsystem permissions only. Collecting all the permissions from all the mappings wouldn't be right as this would also include ```org.wildfly.security.auth.permission.LoginPermission``` (and also any other user-specified permissions from other mappings), which aren't meant to be part of the default permission set.

aloubyansky: That's good to know, thanks. Slightly OT, @fjuma you mentioned a class name as an element of a permission set name (which i assume has to be unique). Would it be safe to assume that the class name would work as a permission id?

fjuma: My initial thinking was that class name would work as a permission id. But, technically, one could have permissions with the same class name but different modules. So the combination of the class name and the module is likely better as a permission id.

fjuma:
@darranl +1. I've updated the analysis document to reflect this. Thanks.

darranl:
Yes my understanding is the starting point needs to be a provisionable config from the outset.  

This is no different to if an administrator defined a number of their own permission sets, some of which may use these permissions - I would not be expecting provisioning to be modifying those.

mchoma:
And why this is not sufficient solution? "Introduce required name attribute to permission-mappings". And name default permision-mappings with well-known names. 

In such case provisioning tool will know where to look.

<simple-permission-mapper name="default-permission-mapper" mapping-mode="first">
    <permission-mapping name="anonymous">
        <principal name="anonymous"/>
        <permission class-name="org.wildfly.extension.batch.jberet.deployment.BatchPermission" module="org.wildfly.extension.batch.jberet" target-name="*"/>
        <permission class-name="org.wildfly.transaction.client.RemoteTransactionPermission" module="org.wildfly.transaction.client"/>
        <permission class-name="org.jboss.ejb.client.RemoteEJBPermission" module="org.jboss.ejb-client"/>
    </permission-mapping>
    <permission-mapping name="all" match-all="true">
        <permission class-name="org.wildfly.security.auth.permission.LoginPermission"/>
        <permission class-name="org.wildfly.extension.batch.jberet.deployment.BatchPermission" module="org.wildfly.extension.batch.jberet" target-name="*"/>
        <permission class-name="org.wildfly.transaction.client.RemoteTransactionPermission" module="org.wildfly.transaction.client"/>
        <permission class-name="org.jboss.ejb.client.RemoteEJBPermission" module="org.jboss.ejb-client"/>
    </permission-mapping>
</simple-permission-mapper>

darranl:
All you have added there is an arbitrary name, it does nothing to make it safe for tooling to work out which ones it can modify.  

Administrators are then forced to add an arbitrary name which has not meaning other than "tooling needs this".

Moving to a named permission set has something we can describe to an administrator and they can choose to use it with an understanding of what it means.

In our documentation we will be able to have something along the lines of, the permission-set 'default' contains the full set of permissions that an identity would require to access any services on the server.  Once the permission set with the name 'default' has a describable meaning the administrator can then use it as they see fit in their own permission mappings - as it has meaning the administrator can also choose not to use it at all and maintain complete control of their permission mapping regardless of any provisioning actions.

The real problem we have today with provisioning is there is no clear dividing line between an administrator wanting to use provisioned permission mappings and the point they want to manage them themselves.

I would actally take it one step further, we have an open Jira issue to add some permission checks when connecting to management services i.e. management and JMX.  I would suggest at some point we end up with two permission sets 'application-all', and 'management-all' i.e. all the permissions needed to access management, and all the permissions needed to access application services.

At a later point I would like to consider if we can make these two specific permission sets internal and automatically populate them, but that gets more complex as we would have some permission sets dynamically registered and some configured whilst wanting to avoid conflicts in the naming.

mchoma:
Thank you for describing bigger picture. It is much clearer now.

PR #44, opened 2018-04-05
WFLY-10172 Java EE Security API

stuartwdouglas:


olukas:
Since this RFE is just about integration WildFly with reference implementation (JASPIC+Elytron is not part of this RFE as is also mentioned in this proposal) all requirements/non-requirements make sense. I think requirements are ready to go. Thanks Stuart.

PR #45, opened 2018-04-05
WFLY-6803 Add multi-server support to mod_cluster

rhusar:
https://issues.jboss.org/browse/WFLY-6803
https://issues.jboss.org/browse/EAP7-977

### DIFF
--- a/mod_cluster/WFLY-6803_Add_multi-server_support_to_mod_cluster.adoc
+++ b/mod_cluster/WFLY-6803_Add_multi-server_support_to_mod_cluster.adoc
@@ -1,4 +1,4 @@
-= Title
+= WFLY-6803 Add multi-server support to mod_cluster
 :author:            Radoslav Husar
 :email:             rhusar@redhat.com
 :toc:               left
@@ -9,11 +9,20 @@
 
 == Overview
 
-Undertow subsystem supports defining and running multiple server instances (each consisting of listeners, hosts, servlet container, etc).
-At the time of writing, mod_cluster subsystem registers only the default Undertow server returned by `org.wildfly.extension.undertow.UndertowService#getDefaultServer`
-to register with the proxy servers.
-Different Undertow server instances might be registered with different groups of reverse proxy servers.
-To support this, multiple mod_cluster configurations/profiles will be supported each associated with a specific Undertow server.
+Undertow and Undertow subsystem supports configuring and running multiple server instances (each consisting of set of listeners,
+set of hosts, servlet container, etc).
+Before implementing this proposed change, mod_cluster subsystem works only the default Undertow server returned by
+`org.wildfly.extension.undertow.UndertowService#getDefaultServer` to register with the proxy servers.
+Moreover, this means that the configured listener (currently called 'connector' in the model which is scheduled to be address
+as well with or after this change, see https://issues.jboss.org/browse/WFLY-10489[WFLY-10489]) must be one from the default
+servers configured listeners.
+The goal of this issue is so that Undertow other server can be registered with reverse proxy servers as well.
+To support this, multiple mod_cluster configurations (i.e. "profiles) will be supported,
+each associated with a specific Undertow server.
+Since the listener always identifies exactly one Undertow server, the listener configuration will drive which Undertow server
+this configuration pertains to, moreover, one profile can thus register only one server.
+However, multiple proxy configurations can reference a single server (e.g. registering an Undertow server with different group
+or reverse proxy servers with different configurations).
 
 == Issue Metadata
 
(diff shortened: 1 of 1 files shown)


mod_cluster/WFLY-6803_Add_multi-server_support_to_mod_cluster.adoc
```diff
@@ -0,0 +1,68 @@
...
+
+=== Affected Projects or Components
+
+* WildFly/EAP
+* mod_cluster
+* Undertow
+
+=== Other Interested Projects
+
+* EWS
```
jfclere: use JWS

rhusar: Fixed, thanks.

rhusar:
@honza-kasik is the QE contact for this feature.

mod_cluster/WFLY-6803_Add_multi-server_support_to_mod_cluster.adoc
```diff
@@ -0,0 +1,66 @@
...
+:keywords:          clustering,undertow,mod_cluster
+:idprefix:
+:idseparator:       -
+
+== Overview
+
+Undertow subsystem supports defining and running multiple server instances (each consisting of listeners, hosts, servlet container, etc).
+At the time of writing, mod_cluster subsystem registers only the default server returned by `org.wildfly.extension.undertow.UndertowService#getDefaultServer`
+to register with the proxy servers.
+Different servers might be registered with different groups of proxy servers.
```
honza-kasik: This means that multiple independent proxies may defined on one EAP instance running on different Undertow `server`s?

mod_cluster/WFLY-6803_Add_multi-server_support_to_mod_cluster.adoc
```diff
@@ -0,0 +1,66 @@
...
+:toc:               left
+:icons:             font
+:keywords:          clustering,undertow,mod_cluster
+:idprefix:
+:idseparator:       -
+
+== Overview
+
+Undertow subsystem supports defining and running multiple server instances (each consisting of listeners, hosts, servlet container, etc).
+At the time of writing, mod_cluster subsystem registers only the default server returned by `org.wildfly.extension.undertow.UndertowService#getDefaultServer`
```
honza-kasik: Does this includes any changes in `mod_cluster` subsystem?

honza-kasik:
@rhusar What is the default behavior, when multiple proxies (profiles – `/subsystem=modcluster/proxy=*`) are defined, each with different Undertow server? Should they all register to reverse proxy or only the one with default server when multicast discovery is enabled?

rhusar:
@honza-kasik It does not make sense to talk about a "default" here. There is no default. Each mod_cluster configuration functions on its own. What scenarios this change is bringing to the table:
* a configuration can use connector/listener associated with different than the default server (previously the default server was hard-coded)
* multiple configurations can be used register each undertow server 
* single application server node can be registered with multiple groups of proxies

honza-kasik:
@rhusar Sorry to bother you again, but I stumbled upon this when implementing tests and I don't know the expected behavior behind this so I don't know how to write the tests properly. This is why I am asking these questions.

> a configuration can use connector/listener associated with different than the default server (previously the default server was hard-coded)

This is the part I fully understand. It is nice not to have things hard-coded!

> multiple configurations can be used register each undertow server

So I can define multiple listeners on one Undertow server and then assign each of these listeners to configuration registered on `/subsystem=modcluster/proxy=*`? If so, this means, that the balancer (reverse proxy) should "register" this configuration as a separate node so I can for example, set different sets of excluded contexts for each? Which of these configurations should response on balancer advertisement? One of them? All?

> single application server node can be registered with multiple groups of proxies (e.g. register with groups of proxies in backup DNS-load balanced data center)

To be honest, I am little lost here. In this context, proxy is balancer or proxy as a configuration in mod_cluster subsystem?

rhusar:
> So I can define multiple listeners on one Undertow server and then assign each of these listeners to configuration registered on /subsystem=modcluster/proxy=*? If so, this means, that the balancer (reverse proxy) should "register" this configuration as a separate node so I can for example, set different sets of excluded contexts for each? Which of these configurations should response on balancer advertisement? One of them? All?

You theoretically could register same Undertow server with same group of proxies and then use `ProxyPass / balancer://othercluster/` to deal which balancer to use when.  So that would correspond to something like:

    /subsystem=modcluster/proxy=other:add(connector=default,advertise-socket=modcluster-other,balancer=othercluster)

Though typically would would have configured a second server and create configuration for it, so in this example there is other-server with ajp-other (offset port by 1):

```
/socket-binding-group=standard-sockets/socket-binding=ajp-other:add(port=8010)
/subsystem=undertow/server=other-server:add
/subsystem=undertow/server=other-server/ajp-listener=ajp-other:add(socket-binding=ajp-other)
/subsystem=undertow/server=other-server/host=other-host:add(default-web-module=other.war
/subsystem=undertow/server=other-server/host=other-host/location=other:add(handler=welcome-content
/subsystem=undertow/server=other-server/host=other-host:write-attribute(name=alias,value=[localhost]))
/subsystem=modcluster/proxy=other:add(connector=ajp-other,balancer=other-balancer,advertise-socket=modcluster)
```

> To be honest, I am little lost here. In this context, proxy is balancer or proxy as a configuration in mod_cluster subsystem?

So this would correspond to e.g. having 2 reverse proxy groups sending advertise on 224.0.1.105 and 224.0.1.106, you would add a configuration to listen for the 2nd groups advertisements, like:

    /socket-binding-group=standard-sockets/socket-binding=modcluster-other:add(multicast-address=224.0.1.106,multicast-port=23364)
    /subsystem=modcluster/proxy=other:add(connector=default,advertise-socket=modcluster-other)

honza-kasik:
@rhusar Thanks! I understand the scenario where I want to register one Wildfly instance with two reverse proxy groups. But I still have no idea how following one should behave:

There is one reverse proxy (A) and (for the purposes of this example) only one Wildfly instance (B) where I am using this shiny new feature. The A is using some multicast address and port to advertise its destination. I configure two `/subsystem=modcluster/proxy=*` resources on B, both with same multicast configuration as A. Now, will be both of these `/subsystem=modcluster/proxy=*` resources recognized by A as a separate nodes?

rhusar:
Added community docs section.

rhusar:
@honza-kasik So, maybe not a practical use case, but still doable. So (B) would be configured with 2 undertow servers, with each with host `subdomain1.example.com` and `subdomain2.example.com`, respectively. Both can share the same advertising config for simplicity of the example. Balancer names need to be distinct. Then, accessing the proxy (A), for domains `subdomain1.example.com` and `subdomain2.example.com` should service registered contexts.

The practical use case would be examples having a backend and frontend loadbalancers.

mod_cluster/WFLY-6803_Add_multi-server_support_to_mod_cluster.adoc
```diff
@@ -0,0 +1,94 @@
+= Title
```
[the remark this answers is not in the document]
rhusar: @jamezp PR for yours: https://github.com/wildfly/wildfly-proposals/pull/107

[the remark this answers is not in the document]
jamezp: Thanks for catching that :)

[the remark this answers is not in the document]
rhusar: Thank you!

PR #46, opened 2018-04-05
[WFLY-10178] Multi-release JAR support

dmlloyd:


rsvoboda:
@dosoudil @psakar please take a look at this and dd comments from productization perspective

psakar:
@dmlloyd @rsvoboda until actual implementation it is hard to judge from productization perspective. General rule should apply - it is good to provide info to productization about **proposed** changes to build environment **before those changes are merged** so we can provide feedback about possible issues on product side

rsvoboda:
`additionalClasspathElement` and `outputDirectory` tricks seem quite fragile, especially when I imagine overlay for Java 11 and further.
Looking at https://github.com/jboss-modules/jboss-modules/blob/1.x/pom.xml#L141

+1 to have MR support in JBoss parent POM

dmlloyd:
https://github.com/jboss/jboss-parent-pom/pull/53

PR #47, opened 2018-04-09
[WFCORE-2951 & WFCORE-2952] Add analysis documentation for JSON and X…

jamezp:
…ML formatters in the logging subsystem.

Analysis document for https://issues.jboss.org/browse/WFCORE-2951 and https://issues.jboss.org/browse/WFCORE-2952.

### DIFF
--- a/logging/WFCORE-2951_WFCORE-2952.adoc
+++ b/logging/WFCORE-2951_WFCORE-2952.adoc
@@ -1,0 +1,73 @@
+= JSON and XML Formatters
+:author:            James R. Perkins
+:email:             jperkins@redhat.com
+:toc:               left
+:icons:             font
+:keywords:          logging,formatter
+:idprefix:
+:idseparator:       -
+
+== Overview
+Pushing log messages to a centralized log collector/server seems to be a more common practice, especially with
+microservices. This can currently be done with a syslog-handler. However new collectors like fluentd or logstash accept
+JSON formatted logs. Having the ability format log messages in JSON will be a requirement in order to support newer log
+collectors.
+
+This will also be used in OpenShift as it currently uses an alpha version of the
+`org.jboss.logmanager:jboss-logmanager-ext` which is not meant to be consumed. Adding first class support for JSON
+formatting will allow OpenShift to not have to install a custom module and use standard configurations.
+
+Note that while XML may not be a requirement it may be useful to have and is essentially free since it's not much
+different than the JSON formatter implementation.
+
+== Issue Metadata
+
+=== Issue
+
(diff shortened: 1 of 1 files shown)


logging/WFCORE-2951_WFCORE-2952.adoc
```diff
@@ -0,0 +1,73 @@
+= Title
```
dmlloyd: Forgot the title :)

logging/WFCORE-2951_WFCORE-2952.adoc
```diff
@@ -0,0 +1,73 @@
...
+:author:            James R. Perkins
+:email:             jperkins@redhat.com
+:toc:               left
+:icons:             font
+:keywords:          comma,separated,tags
```
dmlloyd: and tags I guess

logging/WFCORE-2951_WFCORE-2952.adoc
```diff
@@ -0,0 +1,73 @@
...
+
+=== Nice-to-Have Requirements
+
+* Having the ability to override they keys/element names used when formatting the log record.
+
+=== Non-Requirements
+
+* The ability to write valid JSON or XML documents. What this means is if a JSON or XML formatter is used to write to a
+file, that file will not be well formed data. A separate handler would need to be created to ensure a full document is
+well formatted.
```
dmlloyd: Is this just because the JSON/XML comes out as a series of elements (as in, there is no wrapping root element)?

It would be pretty simple to add a header/footer which added the requisite enclosing bits.  Maybe this is more of a nice-to-have than a must-not-have.

jamezp: Yes because the formatter itself cannot wrap the elements. That would be up to a handler so using a header/footer _may_ work. The footer can be the challenge though since there is no guarantee it would be written.

dmlloyd: Ah of course.  Too bad the formatter doesn't have the opportunity to add headers and footers; it's really a natural fit.

Perhaps this would be a good thing to add to the ```ExtFormatter``` API at some future point.

jmesnil:
@jamezp fiy I have merged the JSON formatter feature in https://github.com/wildfly/wildfly-core/pull/3200. 

I'll merge this proposal once the XML formatter for WFCORE-2952 is also done and merged in core.

jamezp:
Okay thanks! I'll try to get that in today or tomorrow. It should be a pretty easy one.

jmesnil:
all dev PRs have been merged in wildfly-core

PR #48, opened 2018-04-10
EAP7-979 WFCORE-3568 Allow CLI output scrolling

MMarus:
WFCORE: https://issues.jboss.org/browse/WFCORE-3568
EAP7: https://issues.jboss.org/browse/EAP7-979

### DIFF
--- a/cli/EAP7-979_Allow_CLI_output_scrolling.adoc
+++ b/cli/EAP7-979_Allow_CLI_output_scrolling.adoc
@@ -1,0 +1,106 @@
+= EAP7-979 WFCORE-3568 Allow CLI output scrolling
+:author:            Marek Marusic
+:email:             mmarusic@redhat.com
+:toc:               left
+:icons:             font
+:keywords:          cli,output,scrolling
+:idprefix:
+:idseparator:       -
+
+== Overview
+
+Since https://issues.jboss.org/browse/WFCORE-426[WFCORE-426] and https://issues.jboss.org/browse/EAP7-735[EAP7-735]
+mouse scrolling is forbidden to large CLI out and user has to press Down or Enter to navigate.
+After pressing Down or Enter to display more output, user can not go up by pressing Up.
+Therefore, it would be convenient to allow CLI output scrolling by mouse and by Up/Down and PgUp/PgDown keys.
+
+== Issue Metadata
+
+=== Issue
+
+* https://issues.jboss.org/browse/EAP7-979[EAP7-979]
+
+=== Related Issues
+
+* https://issues.jboss.org/browse/WFCORE-3568[WFCORE-3568]
+
+=== Dev Contacts
+
+* mailto:{email}[{author}]
+* mailto:jdenise@redhat.com[Jean-François Denise]
+
+=== QE Contacts
+
+* mailto:eduda@redhat.com[Erich Duda]
+
+=== Affected Projects or Components
+
+* CLI
+
+=== Other Interested Projects
+
+== Requirements
+
+=== Hard Requirements
+* Fast scrolling using PgUp / PgDn. (ATM there is possibility to use space for fast scroll down)
+* Scrolling by mouse
+* Ability to scroll up
+* Since as mentioned in https://developer.jboss.org/wiki/CLI-PagingLongCLIOutput#comment-16541
+  the "less" behaviour would result in buffer alternation and therefore after
+  exiting the reading of output the output wouldn't be visible any more in the main buffer of the CLI.
+  This is not convenient since users may want to see at least part of the output
+  in the main buffer.
+  Therefore I would propose to print a part of the output from start to the last scrolled position
+  into the main buffer of the CLI
+  when exiting the more/less reading.
+
+
+==== List of keys and actions:
+Composed keys are unstable during the testing
+(this is tracked in https://issues.jboss.org/browse/WFCORE-3545[WFCORE-3545])
+Therefore I propose to use for every action a noncomposed key as well.
+This will enable us to create a stable tests.
+
+* PGDOWN / SPACE : scroll one page down
+* PGUP / BACKSLASH : scroll one page up
+* UP / SEMI_COLON: scroll one line up
+* DOWN / ENTER / CTRL_M : scroll one line down
+* Mouse scroll up : scroll up
+* Mouse scroll down : scroll down
+* ESC / 'q' / CTRL_C : quit
+* HOME / 'g' : go to start of the document
+* END / 'G' : go to end of the document
+
+=== Nice-to-Have Requirements
+
+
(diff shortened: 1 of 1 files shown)


cli/EAP7-979_Allow_CLI_output_scrolling.adoc
```diff
@@ -0,0 +1,99 @@
...
+* https://issues.jboss.org/browse/WFCORE-3568[WFCORE-3568]
+
+=== Dev Contacts
+
+* mailto:{email}[{author}]
+* mailto:jdenise@redhat.com[Jean-François Denise]
+
+=== QE Contacts
+
+* mailto:mkopecky@redhat.com[Marek Kopecký]
```
dudaerich: please update to eduda@redhat.com[Erich Duda]

cli/EAP7-979_Allow_CLI_output_scrolling.adoc
```diff
@@ -0,0 +1,99 @@
...
+
+* PGDOWN / SPACE : scroll one page down
+* PGUP / SLASH : scroll one page up
+* UP / SEMI_COLON: scroll one line up
+* DOWN / ENTER / CTRL_M : scroll one line down
+* Mouse scroll up : scroll up
+* Mouse scroll down : scroll down
+* ESC / 'q' / CTRL_C : quit
+
+=== Nice-to-Have Requirements
```
dudaerich: It would be great if the feature supported searching of patterns. Similar as it is done by less command.

cli/EAP7-979_Allow_CLI_output_scrolling.adoc
```diff
@@ -0,0 +1,99 @@
...
+.. Send SEMI_COLON to the console
+.. Check if the line is the same as the saved one from step 1.
+4. Move output one page down
+.. Send SPACE to the console
+.. Check if the output was moved down by one page (terminal height)
+5. Move output one page up
+.. Send SLASH to the console
+.. Check if the before last line is the same as the saved one from step 1. (last line will be the %More string)
+6. Send 'q' to the console
+7. Repeat step 0.
```
dudaerich: Should not it be here "Repeat step 1"?

cli/EAP7-979_Allow_CLI_output_scrolling.adoc
```diff
@@ -0,0 +1,99 @@
...
+=== Affected Projects or Components
+
+* CLI
+
+=== Other Interested Projects
+
+== Requirements
+
+=== Hard Requirements
+* Fast scrolling using PgUp / PgDn. (ATM there is possibility to use space for fast scroll down)
```
dudaerich: What about HOME and END for fast scrolling at start and end of the output.

cli/EAP7-979_Allow_CLI_output_scrolling.adoc
```diff
@@ -0,0 +1,99 @@
...
+This will enable us to create a stable tests.
+
+* PGDOWN / SPACE : scroll one page down
+* PGUP / SLASH : scroll one page up
+* UP / SEMI_COLON: scroll one line up
+* DOWN / ENTER / CTRL_M : scroll one line down
+* Mouse scroll up : scroll up
+* Mouse scroll down : scroll down
+* ESC / 'q' / CTRL_C : quit
+
```
mnovak1: Home : scroll at the start
End : scroll at the end

wdyt?

cli/EAP7-979_Allow_CLI_output_scrolling.adoc
```diff
@@ -0,0 +1,117 @@
...
+* DOWN / ENTER / CTRL_M : scroll one line down
+* Mouse scroll up : scroll up
+* Mouse scroll down : scroll down
+* ESC / 'q' / CTRL_C : quit
+* HOME : go to start of the document
+* END : go to end of the document
+
+=== Nice-to-Have Requirements
+
+Support for searching of patterns. Similar as it is done by less command.
```
MMarus: I have done an analysis for the search functionality as requested in the review.

However, the original issue [2] did not mention a possibility to search patterns and I think it would require a bigger changes.
Therefore, it could be better to create a separate "less" command for the search part and track that in separate Feature Request / Issue. The "less" implemented as a command would give user the possibility to choose whether he wants to use the auto "more" functionality or whether he wants to use the "less" command.
The "less" command could be used for operations like ":read-resource()" through a pipe e.g. ":read-resource() | less".

@jfdenise @dudaerich  what do you think about this ?

cli/EAP7-979_Allow_CLI_output_scrolling.adoc
```diff
@@ -0,0 +1,117 @@
...
+
+==== List of keys and actions for the search:
```
MMarus: I have done an analysis for the search functionality as requested in the review.

However, the original issue [2] did not mention a possibility to search patterns and I think it would require a bigger changes.
Therefore, it could be better to create a separate "less" command for the search part and track that in separate Feature Request / Issue. The "less" implemented as a command would give user the possibility to choose whether he wants to use the auto "more" functionality or whether he wants to use the "less" command.
The "less" command could be used for operations like ":read-resource()" through a pipe e.g. ":read-resource() | less".

@jfdenise @dudaerich what do you think about this ?

jfdenise: @MMarus, I fully agree that this should be covered by a separate issue. This is outside the scope of this one.
BTW, the aesh-extensions come with a less command, that could be a good start to see how we could use it. I suspect that we would have to fix/evolve it in order to have it run inside CLI.

dudaerich: I also agree that this deserves new RFE. I just tried to bring it here :)

However I don't agree with ":read-resource() | less". I think the _less_ behavior is stronger than _more_. It covers all functionality as _more_. IMO we don't need to support both. If we would do, the _less_ behavior was hidden and many users might miss it. @MMarus @jfdenise wdyt?

MMarus: @dudaerich I think you are suggesting to use the less behaviour as the default one. That could be a very good option as well. We could also print a part of the output from start to the last scrolled position into the main buffer of the CLI after exiting the less reading. This way the user could see a part of needed output even after exiting the less.

The less command from aesh-extensions looks very nice and it would be a very good start point for the implementation.

@jfdenise what do you think about this approach ?

MMarus: Thank you very much for the reviews.
 I have created a new Feature Request [1] and moved the search analysis there.

[1] https://issues.jboss.org/browse/WFCORE-3743

PR #49, opened 2018-04-17
[WFLY-10254][WFCORE-3752] Multi-Value Services API

ropalka:


msc/WFLY-10254_WFCORE-3752_MSC_Upgrade.adoc
```diff
@@ -0,0 +1,122 @@
...
+* https://issues.jboss.org/browse/WFLY-10254[WFLY-10254]
+
+=== Related Issues
+
+=== Dev Contacts
+
+* mailto:david.lloyd@redhat.com[David Lloyd]
+* mailto:ropalka@redhat.com[Richard Opalka]
+
+=== QE Contacts
```
simkam: Feel free to put me here

msc/WFLY-10254_WFCORE-3752_MSC_Upgrade.adoc
```diff
@@ -0,0 +1,122 @@
...
+ Supplier<Integer> port = serverBuilder.requires(HTTP_PORT);
+ serverBuilder.setInstance(new HttpServer(host, port));
+ serverBuilder.install();
+
+== Issue Metadata
+
+=== Issue
+
+* https://issues.jboss.org/browse/WFCORE-3752[WFCORE-3752]
+* https://issues.jboss.org/browse/WFLY-10254[WFLY-10254]
```
simkam: I believe we should have also EAP7 issue according to process.

msc/WFLY-10254_WFCORE-3752_MSC_Upgrade.adoc
ropalka: Done

msc/WFLY-10254_WFCORE-3752_MSC_Upgrade.adoc
ropalka: @dmlloyd WDYT?

bstansberry:
What is a multi-value service? Reading this I don't get it. The example code looks like a different way of wiring up injections, which is nice but from it I don't get what the term "multi-value service" means.

ropalka:
We plan to do an EAP team meeting with presentation on that topic with @dmlloyd soon.
Before MSC 1.4 every service could provide just single value to its dependents.
With implementation of MSC 203 we bring capability that every service can provide
more than just one value to its dependents.

For example in MSC 1.3 API users could lookup service controller from registry
then access its service via ServiceController.getService() method and finally access
the single value that service provides via Service.getValue() method.

Since MSC 1.4 both ServiceController.getService() and Service.getValue() methods become deprecated. The only way to retrieve values from service providing it is to properly set up
service dependencies ideally via new API.

ropalka:
Here is one example of what providing multiple values means @bstansberry :grinning: 
https://github.com/jboss-msc/jboss-msc/blob/master/src/test/java/org/jboss/msc/multi_value_services/MultiValueServicesTestCase.java#L62

ropalka:
I updated example in this document to see both scenarios in action @bstansberry :
 * service providing two values to its dependents (HTTP Config Service) - this was not possible before
 * service requiring two values from its dependencies (HTTP Server Service) - this was possible before
Both examples are leveraging new API.

bstansberry:
@ropalka Thanks; that's much clearer.

jmesnil:
@bstansberry @dmlloyd @simkam now that @ropalka has updated the proposal, could you approve this PR if you agree with the feature?

rsvoboda:
> They just reflect an engineering task to be tackled when appropriate.

Any estimation when this could happen? We already have a lot of deprecations in the code, some cleanup and API usage update would be really nice.

ropalka:
@jmesnil this PR have been approved by both @bstansberry & @simkam

bstansberry:
@rsvoboda This is JMO but I don't see updating a lot of code to use a different service builder API and get rid of deprecation warns as being something that's going to get much priority.  If code is being touched for some other reason, sure.

rsvoboda:
I would like to avoid situation when new code is using old API because current code uses it and people just reuse what's there.

Reviewer often does not open code from PR in IDE to see if it's deprecated or not
Maybe we could have some check in PRs to see if the new code is using deprecated methods/classes.
Naive way is to count number of warnings when using `-Dmaven.compiler.showWarnings=true`

ropalka:
@jmesnil @bstansberry who should merge this PR when it is approved?

PR #53, opened 2018-04-25
[WFLY-9818] XTS integration for Narayana providing WS-AT Integration with .NET

ochaloup:
https://issues.jboss.org/browse/WFLY-9818

### DIFF
--- a/transactions/WFLY-9818-ws-at-async-integration-with-.net.adoc
+++ b/transactions/WFLY-9818-ws-at-async-integration-with-.net.adoc
@@ -1,0 +1,119 @@
+= WS-AT Integration with .NET
+:author:            Ondrej Chaloupka
+:email:             ochaloup@redhat.com
+:toc:               left
+:icons:             font
+:keywords:          comma,separated,tags
+:idprefix:
+:idseparator:       -
+
+== Overview
+
+This feature brings support of interoperability of Windows MSDTC  WS transaction implementation with WildFly Narayana.
+By our knowledge the Narayana implementation follows the WS-AT/WS-Addressing specification
+while the MSDTC interpretation of the spec is extra strict which cause issue in the inter-operable communication.
+The feature provides an extension for Narayana WS-AT opening extra endpoints which can be contacted from .NET
+in the way the MSDTC interpret the specification and thus fixing the interoperability issue.
+
+=== Background
+
+When the WS-AT (XTS subsystem) is initiated in the Windows side, the Windows client creates a context with the Windows coordinator service and later it calls to the WildFly WS,
+the process fails at Registration operation.
+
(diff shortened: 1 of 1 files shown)


fmarchioni:
My understanding is that the behavior enforced by Windows MSDTC is coherent with the [specification](http://docs.oasis-open.org/ws-tx/wstx-wsat-1.2-spec.html) which says that 

> Non-terminal notification messages: MUST include a [source endpoint] property whose [address] property is not set to ‘http://www.w3.org/2005/08/addressing/anonymous’ or 'http://www.w3.org/2005/08/addressing/none’.

Assumed that this message does not indicate the end of a coordinator/participant relationship (non-terminal message) the behavior enforced by Windows MSDTC seems correct.
Comments?

ochaloup:
I will quote @adinn here, I hope he will be ok with it. The response comes from long mailing thread where discussion was held.

The reply response in WS-Addressing is specified in [1] and it says:
>   -> If the reply is a normal message, select the EPR from the related
> message's [reply endpoint] message addressing property.
>   -> If the [reply endpoint] message addressing property is not present
> the processor MUST fault. This could only occur when using an alternate
> representation of message addressing properties.

WSA spec was always unclear and
was very badly shoe-horned in with the related specs without ever really
fitting properly. So, getting the current implementation right took some
fiddling. However, that said, let's look at the text

> "If the reply is a normal message, select the EPR from the related
> message's [reply endpoint] message addressing property.

Note:
The [reply endpoint] message addressing property will always be present
when using the XML Infoset representation since, in the absence of a
wsa:ReplyTo element, the value of the [reply endpoint] message
addressing property defaults to an EPR with an [address] property of
"http://www.w3.org/2005/08/addressing/anonymous" - see section 3.2 XML
Infoset Representation of Message Addressing Properties.

If the [reply endpoint] message addressing property is not present the
processor MUST fault. This could only occur when using an alternate
representation of message addressing properties."

[1] https://www.w3.org/TR/2006/REC-ws-addr-core-20060509/#formreplymsg

The underlying transport implementation (CXF) does indeed ensure that
"anonymous" is passed as the ReplyTo value. However, I believe our
implementation actually passes "none" because this is meant to be a
one-way exchange. So, the element is /present/ (note that the spec
speaks about it being present, not about what value it has). A ReplyTo
of "anonymous" or "none" does not justify the FAULT.


@fmarchioni : We all agreed that this is interpretation point of view where we think Microsoft took wrong in our point of view.
I re-sent you the mailing discussion (several parts).
As I'm not deeply involved in the spec and implementation, please, after you read whole mail discussion(!!!), feel free to check probably with @rmartinc more doubts.

ochaloup:
@kabir do you think this proposal could be merged? And, probably after the WFLY13 will be released, then the PR for the https://issues.jboss.org/browse/EAP7-911 could be processed?

PR #54, opened 2018-04-27
[WFCORE-1427] Provide ability to suspend and resume the set of servers managed by a specific host controller

yersan:
Jira issues:
https://issues.jboss.org/browse/EAP7-601

https://issues.jboss.org/browse/WFCORE-3724
https://issues.jboss.org/browse/WFCORE-1427

### DIFF
--- a/core/WFCORE-1427_Suspend_Resume_servers_by_host_controller.adoc
+++ b/core/WFCORE-1427_Suspend_Resume_servers_by_host_controller.adoc
@@ -1,0 +1,128 @@
+= Convenient declaration of server suspend timeouts
+:author:            Brian Stansberry
+:email:             brian.stansberry@redhat.com
+:toc:               left
+:icons:             font
+:keywords:          comma,separated,tags
+:idprefix:
+:idseparator:       -
+:issue-base-url:    https://issues.jboss.org/browse
+
+== Overview
+
+In WildFly 10 we introduced the server suspend / graceful shutdown feature. This feature allows the server to complete the active requests normally, without accepting any new requests. A timeout value specifies how long the suspend or shutdown operation waits for active request to complete.
+
+The concept of graceful shutdown includes not only the 'shutdown' operation but also all the remaining operations that trigger the server to stop handling new requests. Those operations include stop, reload and restart and its variants stop-servers, reload-servers, restart-servers.
+
+The stop operation is adapted to use the "timeout" param and hence allowing a graceful stop. Technically, what the stop operation does is to suspend the affected servers before stopping them.
+
+An analysis of all the lifecycle management operations susceptible to use "timeout" parameter concluded that the use of the timeout parameter in all of them would increase the code complexity and will bring to the user confusion with corner cases.
+
+The graceful shutdown for those operations is available to the users via suspend operation. For example, if the user wants to reload all the servers in a domain, the following operations can be executed:
+
+[source]
+--
+[domain@localhost:9990 /] :suspend-servers(suspend-timeout=X)
+[domain@localhost:9990 /] :reload-servers
+--
+
+However, the user is unable to apply this feature conveniently to all the appropriate set of servers, specifically, the users cannot suspend or shutdown gracefully all the servers managed by a single host in a simple manner.
+
+For example, if they want to suspend the servers managed by a specific host, they have to identify which servers are being managed by that host and execute a suspend operation on each server.
+
+The objective of this RFE is to ensure there is a way to suspend and shutdown all the servers managed by a specific host.
+
+== Issue Metadata
+
+=== Issue
+
+* {issue-base-url}/EAP7-601[EAP7-601]
+* {issue-base-url}/WFCORE-1427[WFCORE-1427]
+
+=== Related Issues
+
+* {issue-base-url}/WFCORE-1632[WFCORE-1632]
+* {issue-base-url}/WFCORE-1917[WFCORE-1917]
+* {issue-base-url}/WFCORE-3804[WFCORE-3804]
+
+=== Dev Contacts
+
+* mailto:{email}[{author}]
+* mailto:yborgess@redhat.com[Yeray Borges]
+* mailto:jdenise@redhat.com[Jean-Francois Denise]
+
+=== QE Contacts
+
+* mailto:jstourac@redhat.com[Jan Stourac]
+
+=== Affected Projects or Components
+* WildFly Core kernel
+* WildFly Core CLI
+* HAL
+
+=== Other Interested Projects
+* N/A
+
+== Requirements
+
+=== Hard Requirements
+
+* Users should be able to suspend the servers managed by a specific host using a new host level operation called 'suspend-servers'.
+* Users should be able to resume the servers managed by a specific host using a new host level operation called 'resume-servers'.
+* Host level suspend operation should behave like any other suspend operations available for another set of servers, for example, the suspend-servers operation available at server-group level.
+* The host level operation 'suspend-servers' will have a timeout parameter named 'suspend-timeout' that controls how long the suspend operation should wait for in-flight requests to complete before returning.
+* The timeout is per server, not an aggregate overall timeout for operations that affect multiple servers.
+* Host controllers will send the suspend operation to the managed servers in parallel.
+* There are already operations which use a timeout parameter for similar purposes. The name of this parameter leads to confusion because it can be interpreted as an operation timeout. In order to have a consistency and to fix this confusion, this 'timeout' parameter will be deprecated in all relevant operations. A 'suspend-timeout' parameter will be added to be used instead of the deprecated one.
+
+=== Nice-to-Have Requirements
+* HAL support. This should be a hard requirement for a subsequent release but is a nice to have initially.
+* The use of 'suspend-timeout' param of the high level shutdown operation and adding a 'suspend-timeout' param to the low level 'shutdown' host operation in domain mode to get the host level graceful shutdown not oly via the suspend operation, but also using the shutdown operation.
+
+
+=== Non-Requirements
+* For host level 'suspend-servers' operation affecting multiple servers, treating the 'suspend-timeout' as being some overall timeout. The 'suspend-timeout' applies to each server.
+
+=== Design notes
+
+These are just examples of how the new operations looks like:
+
+- If the user wants to reload a host restarting its servers gracefully:
+
+[source]
+--
+[domain@localhost:9990 /] /host=master:suspend-servers(suspend-timeout=20)
+[domain@localhost:9990 /] /host=master:reload(restart-servers=true)
+--
+
+
+- If the user wants to shutdown gracefully a host
+
+[source]
+--
+[domain@localhost:9990 /] /host=master:shutdown(suspend-timeout=20)
+--
+
+The 'timeout' parameter will be deprecated in the following operations and 'suspend-timeout' parameter will be added to replace its function:
+
+[source]
+--
+[domain@localhost:9990 /] :stop-servers(timeout=X)
+[domain@localhost:9990 /] /host=*/server=*:stop(timeout=X)
+[domain@localhost:9990 /] /server-group=*:stop-servers(timeout=X)
+[domain@localhost:9990 /] :suspend-servers(timeout=X)
+[domain@localhost:9990 /] /host=*/server=*:suspend(timeout=X)
+[domain@localhost:9990 /] /server-group=*:suspend-servers(timeout=X)
+
+[standalone@localhost:9990  /] :suspend(timeout=X)
+[standalone@localhost:9990  /] :shutdown(timeout=X)
+[standalone@localhost:9990  /] shutdown --timeout
+--
+
+Notice the following: +
+{issue-base-url}/WFCORE-1917[WFCORE-1917] _stop_ and _suspend_ operations were deprecated under +/host=*/server-config=*+ resource. We are not going to modify the 'timeout' attribute on those operations. +
+{issue-base-url}/WFCORE-3804[WFCORE-3804] _stop_ and _suspend_ operations will be available under +/host=*/server=*+ resource once this issue is merged.
+
+== Test Plan
+
+Tests will be added to verify that a web application gets suspended and resumed when the operation is executed in a host.


core/WFCORE-1427_ Convenient_declaration_of_server_suspend_timeout.adoc
```diff
@@ -0,0 +1,235 @@
...
+
+- kill/destroy operations always should be allowed and will be executed immediately.
+- suspend / resume operations will be ignored.
+- stop operations will replace the previous operation.
+- restart operations will replace the previous operation.
+- reload operation at host level will reload the host and, if it was requested, it will restart the servers replacing the previous operation.
+- reload operation at server level will be only executed if the server status is STARTED.
+- start operation will be only executed only if the server status is STOPPED.
+- If the same operation is executed more than once, the new executions will be ignored.
+
```
yersan: Honestly, I do not know if those rules are what we want. I need some feedback here, an alternative could be just to ignore from any additional operation a server which is waiting for the suspend-timeout, but looks like weird if for example, I execute a stop operation and the servers which are waiting for a suspend in the background are finally restarted due to a previous operation. It sounds reasonable to me to stop them instead. I wait for any feedback.

core/WFCORE-1427_ Convenient_declaration_of_server_suspend_timeout.adoc
```diff
@@ -0,0 +1,235 @@
+= Title
```
jstourac: btw - it looks like there is an unintentional space in the name of this document - `WFCORE-1427_ Convenient...`

core/WFCORE-1427_ Convenient_declaration_of_server_suspend_timeout.adoc
yersan: Indeed, we were already going in this way, your example can be tested with our current server version, `stop` operation with a `timeout` and `reload-servers` at server group are already available. The result is `success` but the operation is partly effective since only the servers already started at the moment of `reload-servers` execution are reloaded. But I see your point and understand the problems if we even add operations which are ignored (`suspend` and `resume`).

>I am a little bit worried, that this approach brings unnecessary complexity to the code and maybe confusion to the customers in case some corner-case will occur?

Yes, lately I'm getting the same feeling so I do agree. The initial intention of adding a `suspend-timeout` to the operations which could require a server stop was to facilitate the server administration tasks using a single command instead of multiple commands to implement a graceful shutdown. In general, we understand that the timeouts should be short and almost for all operations (keep out the reload operation at host level restarting the servers), the alternative can be done with two commands, which can be executed together in a batch:

For example, a graceful restart of all servers in the domain:
```
[domain@localhost:9990 /] batch
[domain@localhost:9990 / #] :suspend-servers(timeout=60)
[domain@localhost:9990 / #] :restart-servers(blocking=false)
[domain@localhost:9990 / #] run-batch
```
If a user sees this sequence, it is comprehensible that even using `blocking=false` in the `restart-servers` operation, the whole batch operation has to wait until getting the servers suspended before restarting them. If we reduce both operations to just a single one:
```
[domain@localhost:9990 / #] :restart-servers(suspend-timeout=60, blocking=false)
```
As a user, I would not expect to wait for restarting because `blocking` attribute is false. 
This is the beginning of the problems because we will have more operations executing tasks in the background and we have to define the behaviour between all of them. Maybe we could realize that the complexity added is bigger than the advantages. We could, of course, allow this single operation to wait until getting the servers suspended before returns the control to the CLI session, but as I said, it looks like contra-intuitive the uses of `blocking=false` attribute.

At the moment, at least for the operation at host level, there is not two commands alternative.
That means, if you want to reload your host to apply the changes only to the servers managed by that host, you cannot suspend only the servers of that host before restarting them with the reload operation. Writing this I am thinking when this reloading restarting the servers must be done, but currently, I cannot find a use case. From the point of view of a user, can you give one?

The user usually applies changes selecting a profile, which affects to the server groups using that profile, which affects to the servers of each server group,. At server group level there are `suspend-servers` and `restart-servers` operations.

What seems to make sense is the `shutdown` operation. Shutting down a specific host affects only to the servers managed by that host, so shutting down the host gracefully makes sense.

In standalone mode, reload operation can be combined in two suspend+reload.

In general, what do you think?

core/WFCORE-1427_ Convenient_declaration_of_server_suspend_timeout.adoc
```diff
@@ -0,0 +1,235 @@
...
+
+=== Dev Contacts
+
+* mailto:{email}[{author}]
+* mailto:yborgess@redhat.com[Yeray Borges]
+* mailto:jdenise@redhat.com[Jean-Francois Denise]
+
+=== QE Contacts
+
+* mailto:{jstourac@redhat.com}[{Jan Stourac}]
```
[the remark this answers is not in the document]
yersan: @jstourac thanks, fixed

core/WFCORE-1427_ Convenient_declaration_of_server_suspend_timeout.adoc
```diff
@@ -0,0 +1,235 @@
...
+
+=== Other Interested Projects
+* N/A
+
+== Requirements
+
+=== Hard Requirements
+* All low level management operations that result in a server stopping handling of requests should accept a 'suspend-timeout' parameter that controls how long the server should wait for in-flight requests to complete before proceeding with the stop.
+** Meaning of the suspend-timeout should be the same as the 'timeout' param currently available for the 'shutdown' operation. A value of less than 0 means wait indefinitely, a value of 0 means don't wait, a value larger than zero means the server will wait up to this many seconds for all active requests to finish. Default value is 0. Value is in seconds.
+** Operations like 'shutdown' that already have a 'timeout' param are excepted from this requirement; no need to add a semantically equivalent 'suspend-timeout'.
```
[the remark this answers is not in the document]
yersan: The idea of using `suspend-timeout` instead of `timeout` was because the attribute being introduced was not really a "shutdown servers timeout". It was a timeout for suspending the servers prior a subsequent action, which could be a restart. There was already a timeout in the standalone shutdown operation and the idea was to avoid confusion with it.

core/WFCORE-1427_ Convenient_declaration_of_server_suspend_timeout.adoc
```diff
@@ -0,0 +1,235 @@
...
+* Relevant operations include 'reload' for a server or HC and 'restart-servers' on the domain root resource and on the server group resource. An audit should be performed to check for others as well.
+* All high level CLI commands that result in a server stopping handling of requests should also accept a 'suspend-timeout' parameter.
+* For a Host Controller 'reload' operation or high level CLI command, the suspend-timeout applies to the servers, not to the HC process. HC suspend is not supported. Any 'suspend-timeout' is irrelevant if the 'restart-servers' param is set to 'false' since in that case no server will stop normal handling of requests.
+* The 'suspend-timeout' is a per server, not an aggregate overall timeout for operations that affect multiple servers.
+* The 'suspend-timeout' is only for a server suspending before stopping normal request handling, not for the time take for any other aspect of the operation, e.g. the time to restart/reload once the server is suspended.
```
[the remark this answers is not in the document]
yersan: 'the time taken' Sorry, misspelled, fixed

core/WFCORE-1427_ Convenient_declaration_of_server_suspend_timeout.adoc
```diff
@@ -0,0 +1,235 @@
...
+** Operations like 'shutdown' that already have a 'timeout' param are excepted from this requirement; no need to add a semantically equivalent 'suspend-timeout'.
+* Relevant operations include 'reload' for a server or HC and 'restart-servers' on the domain root resource and on the server group resource. An audit should be performed to check for others as well.
+* All high level CLI commands that result in a server stopping handling of requests should also accept a 'suspend-timeout' parameter.
+* For a Host Controller 'reload' operation or high level CLI command, the suspend-timeout applies to the servers, not to the HC process. HC suspend is not supported. Any 'suspend-timeout' is irrelevant if the 'restart-servers' param is set to 'false' since in that case no server will stop normal handling of requests.
+* The 'suspend-timeout' is a per server, not an aggregate overall timeout for operations that affect multiple servers.
```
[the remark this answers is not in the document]
yersan: It means if you use a `suspend-timeout=30` and you have two servers, `server-one`, `server-two`, the timeout of 30 seconds is applied to each server, it is not a timeout for the overall operation. The overall operation could take more time to end.

Keep in mind that parallelization of multi-server suspend was a nice-to-have requirement, not all operations are now being executed in parallel, for example, `restart-servers` is done server by server.

core/WFCORE-1427_ Convenient_declaration_of_server_suspend_timeout.adoc
```diff
@@ -0,0 +1,235 @@
...
+[standalone@localhost:9990 /] reload
+[standalone@localhost:9990 /] :reload
+--
+
+=== Domain mode
+
+In practice, graceful shutdowns should be short, and it should be documented as the best practices. If it is expected to have a long period, reloading a Domain Controller could end up in a long period with an unmanageable domain blocking any additional incoming operations or new CLI sessions.
+To avoid it, a new attribute named _blocking_ and defaulted to _true_, will be added to the reload operation in domain mode.
+
+With this attribute, the users can decide whether they want to wait until the servers are stopped (default behaviour) or they want to continue using the CLI to execute additional operations.
```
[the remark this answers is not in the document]
yersan: Yes, by default the idea is always to do a blocking operation when the host controller is reloaded, that means, any additional operations will not possible. The section `Effects of subsequent operations in Domain mode` is irrelevant in this case, but relevant if the user uses `blocking=false`

core/WFCORE-1427_ Convenient_declaration_of_server_suspend_timeout.adoc
yersan: Fixed, thanks

core/WFCORE-1427_ Convenient_declaration_of_server_suspend_timeout.adoc
```diff
@@ -0,0 +1,235 @@
...
+* For an operation affecting multiple servers, treating the 'suspend-timeout' as being some overall timeout. The timeout applies to each server.
+* Cancellation of the operations using the :resume operation. For example, cancellation of a restart meanwhile the servers are waiting for the suspend-timeout.
+
+== Design Notes
+
+The following list shows the operations that will use the suspend-timeout attribute:
+
+[source]
+--
+[domain@localhost:9990 /] /host=*:reload
```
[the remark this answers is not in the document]
yersan: Yes, you're right, looks like this operation is also a candidate to include a suspend timeout. Keep in mind this operation is not using yet an attribute called timeout, but the high-level `shutdown` operation includes one, although it is not used in domain mode, see:

```
[domain@localhost:9990 /] help shutdown
```

core/WFCORE-1427_ Convenient_declaration_of_server_suspend_timeout.adoc
yersan: Let me sum up the current status, @jstourac it would be great if you can give me your opinion so we can move this enhancement forward. 

In general, we agreed on the uses of non-blocking operations like restart/reload in the background could confuse the users and those operations bring in more complexity than advantages. Right now users can use two operations to reload/restart the servers in a domain. If they want to stop the servers, they can use a single operation, because stop already has a timeout attribute.

For reload/restart servers gracefully, users can execute a suspend using a timeout. Those operations are available for all servers in a domain, server groups or single servers, but it is not available at host level.

There are operations at host level which require that the servers managed by the host be restarted. In general changes in `/host=master/core-service=management` are susceptible to required a server reload. 

So the missing part here is a convenient way to allow servers in a host be restarted gracefully. In order to allow it being consistent with the other operations, we can add a `suspend-servers` at host level. Users could suspend the servers first when they are going to reload a HC and restart their servers. This operation will allow as well a HC graceful shutdown. It could be done consistently suspending the servers before executing a shutdown.

@jstourac What do you think if we use this approach?

core/WFCORE-1427_ Convenient_declaration_of_server_suspend_timeout.adoc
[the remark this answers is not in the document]
jstourac: Thanks, just wanted to be sure.

core/WFCORE-1427_ Convenient_declaration_of_server_suspend_timeout.adoc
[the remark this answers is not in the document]
jstourac: Thanks for explanation. I did not realize that suspend operation may not occur in paralel for all the servers... maybe we could mention that here as an example?

core/WFCORE-1427_ Convenient_declaration_of_server_suspend_timeout.adoc
jstourac: @yersan, my original intention was about removing the `blocking` attribute from the operations, although I did not realize that such attribute is already present there and thus we cannot remove it actually. I agree with you that causing operation to block is also confusing...

I like your proposal as it brings same approach of gracefully suspend/reload/restart server operations for all variants - `all servers in a domain`, `server-groups`, `single-servers` and `host-level`. No unneccessary confusion introduced.

Although, with such change, this is quite a diferent RFE now. Such big change in RFE requirements probably needs to be communicated with eg. @bstansberry (as original reporter) or Bilge. If this RFE is build on some customer case, maybe there are also other consequences.

Another question is - what about the `shutdown` operation? As you described in a comment behind currently the timeout attribute is ignored in case of a domain mode. Is there any plan to unify this or is this a final solution?

core/WFCORE-1427_ Convenient_declaration_of_server_suspend_timeout.adoc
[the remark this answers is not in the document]
bstansberry: My instinct without a ton of thought is that for a 'shutdown' op or command the 'timeout' param should be deprecated and a 'suspend-timeout' added.  For a 'suspend' op the name of the op provides context about what the timeout is. For 'shutdown', 'reload' etc the meaning is not clear at all and to met the intuitive meaning is that it applies somehow to the entire operation not just the suspend part.

core/WFCORE-1427_ Convenient_declaration_of_server_suspend_timeout.adoc
bstansberry: I'm aware of all this and am fine with it. :) Having to execute two ops isn't as simple on the surface as one, but if to get it down to one we end up with a lot of complex behaviors users have to reason about, and probably get wrong, then really one op isn't as simple as it appears. 

I originated this RFE via WFCORE quite a while ago as it seemed like there were noticeable holes in how we handled graceful shutdown. I filed the EAP7 because significant RFEs should go through the kanban process. I'm not aware of end user demand.

core/WFCORE-1427_ Convenient_declaration_of_server_suspend_timeout.adoc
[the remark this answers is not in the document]
jstourac: Generally, I agree and I think it is a good approach. But to be honest - I would prefer to have same name of the attribute also for `suspend` op/cmd. Reason is when one uses different commands, these two attribute names might be easily exchangable and potentionally confusing (one command uses `timeout`, others `suspend-timeout`). So maybe the best option is to use just `suspend-timeout` for all the operations, wdyt?

core/WFCORE-1427_ Convenient_declaration_of_server_suspend_timeout.adoc
jstourac: I have no problem to modify and exploit this RFE unless there are reasons to have this tracked for some possible future reference? Don't know, probably it is okay to just reuse it.

core/WFCORE-1427_ Convenient_declaration_of_server_suspend_timeout.adoc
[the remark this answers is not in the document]
jstourac: Yeah, that is exactly how I understood this feature initialy. :) After reading that sentence in analysis, I got confused why it is mentioned there that way. Maybe I was simply just overthinking. Thanks for clarification Yeray!

core/WFCORE-1427_ Convenient_declaration_of_server_suspend_timeout.adoc
bstansberry: +1 to using this RFE.  Once this doc is in good enough state we can run it by Bilge (put a comment on the EAP7 JIRA and move it to PM Input state) and then once we have that feedback if needed adjust the title / description of the JIRAs better reflect what will be done.

yersan: @jstourac I've just updated the analysis doc with the new approach, I will wait for your review, thanks!

core/WFCORE-1427_ Convenient_declaration_of_server_suspend_timeout.adoc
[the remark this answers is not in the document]
yersan: I haven't taken into account this change yet in the new analysis doc version, but to me, it makes sense the deprecation of 'timeout' and the uses of 'suspend-timeout' instead for all of these operations including the suspend op. It makes a consistency between all of them.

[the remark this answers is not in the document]
yersan: @jstourac I reviewed this part, now the analysis includes a note to deprecate the timeout parameter and to use suspend-timeout instead. Please, take a look at the current document version and let us know your concerns, thanks.

core/WFCORE-1427_Convenient_declaration_of_server_suspend_timeout.adoc
```diff
@@ -0,0 +1,129 @@
...
+[domain@localhost:9990 /] /server-group=*:suspend-servers(timeout=X)
+
+[standalone@localhost:9990  /] :suspend(timeout=X)
+[standalone@localhost:9990  /] :shutdown(timeout=X)
+[standalone@localhost:9990  /] shutdown --timeout
+--
+
+Notice the following: +
+{issue-base-url}/WFCORE-1917[WFCORE-1917] _stop_ and _suspend_ operations were deprecated under +/host=*/server-config=*+ resource. We are not going to mofify the 'timeout' attribute on those operations. +
+{issue-base-url}/WFCORE-3804[WFCORE-3804] _stop_ and _suspend_ operations will be available under +/host=*/server=*+ resource once this issue is merged.
```
jstourac: Well, which operations then? Here is stated that `stop` and `suspend` but in the jira title and description, there is mentioned `reload` and `suspend` operations instead. I think that those operations mentioned in jira are correct.

core/WFCORE-1427_Convenient_declaration_of_server_suspend_timeout.adoc
```diff
@@ -0,0 +1,129 @@
...
+[domain@localhost:9990 /] /host=master:suspend-servers(timeout=20)
+[domain@localhost:9990 /] /host=master:reload(restart-servers=true)
+--
+
+
+- If the user wants to shutdown gracefully a host
+
+[source]
+--
+[domain@localhost:9990 /] /host=master:suspend-servers(timeout=20)
```
jstourac: `timeout` -> `suspend-timeout`

core/WFCORE-1427_Convenient_declaration_of_server_suspend_timeout.adoc
jstourac: Hm, now I realized that effectively this should be true for all three operations, right?

core/WFCORE-1427_ Convenient_declaration_of_server_suspend_timeout.adoc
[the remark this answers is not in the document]
yersan: Resolving this conversation, the above comments are now obsolete because we finally are not going to add a new attribute to the `shutdown` operation, instead we will be able to use  first a `/host=*:suspend-servers(suspend-timeout=*)` to suspend the servers and later, execute the shutdown.

core/WFCORE-1427_Convenient_declaration_of_server_suspend_timeout.adoc
yersan: Already fixed.

core/WFCORE-1427_ Convenient_declaration_of_server_suspend_timeout.adoc
yersan: Bilge agreed in reuse this proposal and jira issues. I've just renamed relevant commits and jiras.
I resolve this conversation and start the issue development.

core/WFCORE-1427_Suspend_Resume_servers_by_host_controller.adoc
```diff
@@ -0,0 +1,129 @@
...
+* The timeout is per server, not an aggregate overall timeout for operations that affect multiple servers.
+* Host controllers will send the suspend operation to the managed servers in parallel.
+* There are already operations which use a timeout parameter for similar purposes. The name of this parameter leads to confusion because it can be interpreted as an operation timeout. In order to have a consistency and to fix this confusion, this 'timeout' parameter will be deprecated in all relevant operations. A 'suspend-timeout' parameter will be added to be used instead of the deprecated one.
+
+=== Nice-to-Have Requirements
+* HAL support. This should be a hard requirement for a subsequent release but is a nice to have initially.
+* The use of 'timeout' param of the high level shutdown operation and adding a 'timeout' param to the low level 'shutdown' host operation in domain mode to get the host level graceful shutdown not oly via the suspend operation, but also using the shutdown operation.
```
jstourac: As mentioned in previous paragraph, the current name of the attribute 'timeout' is deprecated and will be changed to 'suspend-timeout'. Shall not we rather refer to the new attribute name here - 'suspend-timeout'?

core/WFCORE-1427_Suspend_Resume_servers_by_host_controller.adoc
```diff
@@ -0,0 +1,129 @@
...
+
+
+=== Non-Requirements
+* For host level 'suspend-servers' operation affecting multiple servers, treating the 'timeout' as being some overall timeout. The suspend-timeout applies to each server.
```
jstourac: again, it's probably clearer to use 'suspend-timeout' instead of just 'timeout' here too, just to be consistent

core/WFCORE-1427_Suspend_Resume_servers_by_host_controller.adoc
```diff
@@ -0,0 +1,129 @@
...
+[domain@localhost:9990 /] /host=master:reload(restart-servers=true)
+--
+
+
+- If the user wants to shutdown gracefully a host
+
+[source]
+--
+[domain@localhost:9990 /] /host=master:suspend-servers(suspend-timeout=20)
+[domain@localhost:9990 /] /host=master:shutdown()
```
jstourac: Actually, in this case, the shutdown operation for host offers me to define `suspend-timeout` attribute. So, this can be simplified to just
```
[domain@localhost:9990 /] /host=master:shutdown(suspend-timeout=20)
```
or did I miss anything?

yersan: You're right, the issue with this sentence is we had this possibility as a nice-to-have requirement. Since finally it was implemented, I'm going to remove it.

yersan: Changed

jstourac: Thanks, looks good now. Sorry, I didn't realize that this was in a `nice-to-have` requirements. Thus original variant was probably also okay. Anyway, thank you for the update!

jmesnil:
https://github.com/wildfly/wildfly-core/pull/3632 has been merged

PR #55, opened 2018-04-27
[WFCORE-3769][WFLY-10305] Use Galleon for the build

bstansberry:
https://issues.jboss.org/browse/WFCORE-3769
https://issues.jboss.org/browse/WFLY-10305

Proposal for converting the WildFly (Core) build to use Galleon in WildFly 13.

This is the first step in the shift to Galleon. It's not really a feature per se, as it is an internal detail of how we build. But it's useful to have the details specified.

I put some stuff in the "Test Plan" section which is beyond my remit I suppose. My intent there is to share some concepts.

I authored this initial effort on behalf of @aloubyansky but this is really his proposal; i.e. he's the main person to address any inputs.

### DIFF
--- a/build/WFCORE-3769_WFLY-10305_galleon_build.adoc
+++ b/build/WFCORE-3769_WFLY-10305_galleon_build.adoc
@@ -9,11 +9,15 @@
 
 == Overview
 
-The Galleon tool (https://github.com/wildfly/galleon) is a powerful new tool for doing platform 
-independent provisioning and updating of a software installation. The WildFly project intends 
-to use Galleon for its provisioning needs. This initial proposal is for a very limited use of
-Galleon -- using it produce the 'thin' and 'fat' distribution zips that are part of a
-WildFly Core and WildFly release.
+The Galleon tool is a powerful new tool for doing platform independent provisioning 
+and updating of a software installation. The overall Galleon project consists of 
+core tooling (https://github.com/wildfly/galleon) that isn't specific to provisioning 
+any particular kind of software, and then plugin s(https://github.com/wildfly/galleon-plugins) 
+that allow the tool to deal with particular types of software, e.g. WildFly.
+
+The WildFly project intends to use Galleon for its provisioning needs. This initial proposal 
+is for a very limited use of Galleon -- using it produce the 'thin' and 'fat' distribution 
+zips that are part of a WildFly Core and WildFly release.
 
 This initial work is not really a feature, it is an internal detail of the builds.
 
@@ -34,10 +38,13 @@ This initial work is not really a feature, it is an internal detail of the build
 
 === QE Contacts
 
+ * mailto:msimka@redhat.com
+
 === Affected Projects or Components
 
  * WildFly
  * WildFly Core
+ * EAP Installer
 
 === Other Interested Projects
 
--- a/build/WFCORE-3769_WFLY-10305_galleon_build.adoc
+++ b/build/WFCORE-3769_WFLY-10305_galleon_build.adoc
@@ -1,0 +1,166 @@
+= [WFCORE-3769][WFLY-10305] Use Galleon for the WildFly and WildFly Core build
+:author:            Brian Stansberry
+:email:             brian.stansberry@redhat.com
+:toc:               left
+:icons:             font
+:keywords:          
+:idprefix:
+:idseparator:       -
+
+== Overview
+
+The Galleon tool (https://github.com/wildfly/galleon) is a powerful new tool for doing platform 
+independent provisioning and updating of a software installation. The WildFly project intends 
+to use Galleon for its provisioning needs. This initial proposal is for a very limited use of
+Galleon -- using it produce the 'thin' and 'fat' distribution zips that are part of a
+WildFly Core and WildFly release.
+
+This initial work is not really a feature, it is an internal detail of the builds.
+
+== Issue Metadata
+
+=== Issue
+
+* https://issues.jboss.org/browse/WFCORE-3769
+* https://issues.jboss.org/browse/WFLY-10305
+
+=== Related Issues
+
+* 
+
+=== Dev Contacts
+
+* mailto:alexey.loubyansky@redhat.com
+
+=== QE Contacts
+
+=== Affected Projects or Components
+
+ * WildFly
+ * WildFly Core
+
+=== Other Interested Projects
+
+ * WildFly Swarm
+ * Teiid
+ * Keycloak
+ * Infinispan
+
+== Requirements
+
+=== Hard Requirements
+
+ * Produce a 'core' Galleon feature pack as part of the default WildFly Core maven build.
+ * Use that 'core' feature pack to produce a 'thin' core distribution as part of
+   the default WildFly Core maven build. 
+ ** This distribution should be produced in the 'build' maven module and should
+    have the org.wildfly.core:wildly-core-build maven groupId and artifactId. This
+    is the traditional location and maven GA for our 'thin' core distribution.
+ * Use that 'core' feature pack to produce a 'fat' core distribution as part of
+   the default WildFly Core maven build. 
+ ** This distribution should be produced in the 'dist' maven module and should
+    have the org.wildfly.core:wildly-core-dist maven groupId and artifactId. This
+    is the traditional location and maven GA for our 'thin' core distribution.
+ * Produce a 'servlet' Galleon feature pack as part of the default WildFly maven build.
+ ** This feature pack will depend on the 'core' feature pack.
+ * Use that 'servlet' feature pack to produce a 'thin' servlet distribution as part of
+   the default WildFly maven build. 
+ ** This distribution should be produced in the 'servlet-build' maven module and should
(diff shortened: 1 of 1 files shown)


build/WFCORE-3769_WFLY-10305_galleon_build.adoc
```diff
@@ -0,0 +1,166 @@
+= Title
```
jamezp: I'm starting to wonder if the "Title" should just be blank in the template as it seems to not get changed a lot. I know I've forgotten several times :). Anyway this could probably use a title.

build/WFCORE-3769_WFLY-10305_galleon_build.adoc
```diff
@@ -0,0 +1,166 @@
...
+:author:            Brian Stansberry
+:email:             brian.stansberry@redhat.com
+:toc:               left
+:icons:             font
+:keywords:          comma,separated,tags
```
jamezp: Maybe this should be deleted too from the template as well as here.

build/WFCORE-3769_WFLY-10305_galleon_build.adoc
```diff
@@ -0,0 +1,166 @@
...
+
+=== Nice-to-Have Requirements
+
+None 
+
+=== Non-Requirements
+
+ * Support for any use of the Galleon tooling outside of the maven build steps described in this document. 
+ * Long term support for the initial content details of the Galleon feature packs. The initial version
+   of these feature packs should be regarded as 'Tech Preview', i.e. subject to change in the next WildFly
```
rsvoboda: Is 'Tech Preview' language used on community side or it's just product related note?

Will be Galleon feature packs (GFP) uploaded to the public maven repository? As it's just internal build tool for WF13 scope I guess no uploads of GFP.

build/WFCORE-3769_WFLY-10305_galleon_build.adoc
```diff
@@ -0,0 +1,166 @@
...
+   of these feature packs should be regarded as 'Tech Preview', i.e. subject to change in the next WildFly 
+   release if we discover the need to correct things after this initial use. The intent though is to produce 
+   feature packs external parties can look at to learn about how WildFly+Galleon will work and to
+   start planning how their own software can be based on our feature packs. 
+ * Publication of legacy 'thin' and 'fat' distributions to public maven repositories. These are
+   only meant for testing use to verify the Galleon output against the current mechanism output;
+   they are not meant for other consumption. Publishing these is actually an anti-requirement.
+ * Absolutely identical output content of a thin or fat distribution produced using the legacy tooling and 
+   one produced using Galleon. The two are meant to be very, very similar but minor differences that do
+   not affect functionality are acceptable. The intent though is to minimize these.
```
rsvoboda: `differences that do not affect functionality are acceptable` - what fits to this category ?

Some real examples would be nice. The tool has PR to be merged to WildFly Core, PR for WildFLy should come soon. We should be aware of existing minor differences.

build/WFCORE-3769_WFLY-10305_galleon_build.adoc
```diff
@@ -0,0 +1,166 @@
...
+
+=== Related Issues
+
+* 
+
+=== Dev Contacts
+
+* mailto:alexey.loubyansky@redhat.com
+
+=== QE Contacts
```
simkam: Feel free to add me here.

simkam:
and @thauser fyi, this might affect also installer

build/WFCORE-3769_WFLY-10305_galleon_build.adoc
```diff
@@ -0,0 +1,166 @@
...
+
+ * WildFly Swarm
+ * Teiid
+ * Keycloak
+ * Infinispan
+
+== Requirements
+
+=== Hard Requirements
+
```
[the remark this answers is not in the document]
aloubyansky: This is done.

build/WFCORE-3769_WFLY-10305_galleon_build.adoc
```diff
@@ -0,0 +1,166 @@
...
+   using the legacy org.wildfly.build:wildfly-server-provisioning-maven-plugin. These will
+   use the 'legacy servlet feature pack' and 'legacy full feature pack' discussed previously.
+ ** This maven profile should not run by default.
+ * These legacy distributions should be produced in such a way that their output, except 
+   for their maven GAVs, would be acceptable for release as the primary build output if 
+   this shift to Galleon had not occurred. In other words, these distributions should 
+   provide a valid base for comparison to the output of the Galleon build.
+ ** The only purpose of these legacy distributions is to facilitate such comparisons.
+ * The content of the Galleon-produced distributions should functionally match the output
+   of the legacy distributions, with only minimal minor cosmetic differences.
```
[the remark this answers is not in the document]
aloubyansky: Unfortunately I won't guarantee identical content. We used to have comments in our config files, for example. New configs are generated using embedded controller. There is no way atm to generate comments, for example.
Another example, the undertow subsystem config element allows to configure a few attributes that if not explicitly configured assume default values. These values are not present in the current config but are initialized to their default values in the galleon generated config. In general, it is not required to initialize all the parameters but this specific case does require it because these attributes are used to resolve capabilities that are ultimately used to validate the config before the installation. We can remove those attributes but once we do that we can't guarantee the configuration produced by galleon is going to work. In all our configs this is the only place where I had to explicitly use the default values.
Other possible cosmetic diff, the order of some attributes or elements may be different. I am trying to preserve the order but in a few places that could still end being slightly different.

build/WFCORE-3769_WFLY-10305_galleon_build.adoc
aloubyansky: While they are going to be mainly used internally (by the build), possible advantages of making them public with the 'tech preview' notice is that the community and other teams could easily try using them, get some experience and give feedback on how it worked or not for them.

build/WFCORE-3769_WFLY-10305_galleon_build.adoc
aloubyansky: Like the ones I mentioned before. There is for example a new .galleon dir which records the provisioned state. This directory could be removed actually assuming nobody will try reprovisioning the installation using galleon.

build/WFCORE-3769_WFLY-10305_galleon_build.adoc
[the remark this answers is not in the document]
bstansberry: Comments in our standard config files are bugs, as they result in irrelevant diffs when the file gets repersisted (since we can't store comments + their xml location in the model and thus can't persist them.) So not having them in galleon is not a flaw, it's a reason to clean up existing bugs.

The comments in our current standard configs are in the domain.xml and host[...].xml variants. We long ago cleaned them out of the standalone variants.

https://issues.jboss.org/browse/WFCORE-3771
https://issues.jboss.org/browse/WFLY-10311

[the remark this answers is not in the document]
bstansberry: I don't believe xml *attribute* ordering in the standard config file should be different due to galleon. If there is a difference, that indicates a bug in the legacy configs or a bug in the xml persistence (i.e. not related to galleon).  We've long had a rule that our standard configs should follow xsd order, as should our xml persisters.  If a config doc created by our xml persistence code (which is what the galleon produced docs are) is in a different *attribute* order from the legacy configs, then that means either the xml persister was wrong or the legacy config was wrong. Most every release we have an exercise to find and correct those kinds of problems. So for WF 13 we should use the diff of the galleon build vs legacy build to help find and clean up those.  In core I've found and sent patches for two such issues:

https://issues.jboss.org/browse/WFCORE-3772
https://issues.jboss.org/browse/WFCORE-3773

XML element order is a different matter. XML element order typically reflects the order in which independent management resources are added, and that can vary between the legacy configs and what galleon produces. That's for valid reasons.  The legacy configs were manually hand coded as a single entire config with element order set to whatever the human author manually wrote. A galleon config is created from groups of features, with the content of those groups organized around convenient reuse and convenient exclusion of less commonly used features. Composing a config from those distinct groups of features can result in a different order of elements. 

The legacy config generator also seems to have a bug in the order in which it creates the extension elements. (This is one area of the legacy configs that is not under manual control.) Galleon does not have this bug.

build/WFCORE-3769_WFLY-10305_galleon_build.adoc
```diff
@@ -0,0 +1,166 @@
...
+:email:             brian.stansberry@redhat.com
+:toc:               left
+:icons:             font
+:keywords:          
+:idprefix:
+:idseparator:       -
+
+== Overview
+
+The Galleon tool (https://github.com/wildfly/galleon) is a powerful new tool for doing platform
```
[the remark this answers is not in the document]
bstansberry: I added a bit on this.

build/WFCORE-3769_WFLY-10305_galleon_build.adoc
bstansberry: Done; thanks.

PR #57, opened 2018-05-01
Add Implementation Plan section

kabir:


design-doc-template.adoc
```diff
@@ -37,3 +37,11 @@
 === Non-Requirements
 
 == Test Plan
+
+== Implementation Plan
```
jamezp: Minor detail, but should this be above the test plan? It's not a huge deal though.

bstansberry: My 2 cents: It's a less significant section than a test plan and often (aka usually) won't even be filled out, so it shouldn't be put in a higher priority section.

jamezp: That makes sense. I'll retract my comment :)

dmlloyd: I agree it should come before the test plan, not because of priority but because the test plan will likely reference the implementation.  For example "Implement xyz in module A" could be in the implementation plan, whereas "Create unit tests for implementation of xyz in module A" would refer to it.  It might be confusing to refer to something that comes later in the document.

kabir: @dmlloyd @jamezp @bstansberry How about I move it up and comment out the header as well? Then people who want to use it can uncomment and fill it in

bstansberry: Sounds good.

PR #62, opened 2018-05-11
WFLY-10328 Agroal subsystem

barreiro:
EAP feature https://issues.jboss.org/browse/EAP7-983

### DIFF
--- a/agroal/WFLY-10328_EAP7-983_Agroal-subsytem.adoc
+++ b/agroal/WFLY-10328_EAP7-983_Agroal-subsytem.adoc
@@ -1,0 +1,80 @@
+= WFLY-10328 - Agroal subsystem
+:author:            Luis Barreiro
+:email:             lbarreiro@redhat.com
+:toc:               left
+:icons:             font
+:idprefix:
+:idseparator:       -
+:issue-base-url:    https://issues.jboss.org/browse
+
+== Overview
+
+This issue is about providing an alternative to the current datasources subsystem based on IronJacamar. The new one is based on the Agroal connection pool library.
+Agroal is a lightweight and very fast connection pool implementation that performs exceptionally well in situations with high contention on the pool. For most of Wildfly's use cases Agroal can be a replacement for IronJacamar.
+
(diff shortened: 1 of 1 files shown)


barreiro:
@dmlloyd yes it was. Security integration has been on the requirements from day one.

simkam:
@istraka @fmarchioni I suppose this will need tests also in transaction test suites, can you review this too? 
@barreiro have you considered also pool-less setup for scenarios where other pool is used instead of driver? f.ex Oracle UCP

barreiro:
@simkam yes, we had discussions around pool-less and is something we may add in the near future (probably as a third datasource type besides XA and non-XA)

jamezp:
We should document whether or not something like the `*-ds.xml` will be supported this subsystem.

jmesnil:
@jamezp by `*-ds.xml` you mean the IronJacamar's deployment files? If that's the case, I don't think the new subsystem would have to support it (as a comparison, the `messaging-activemq` subsystem does not support the `*-jms.xml` from the legacy `messaging` subsystem).

Besides, the new subsystem supports the standard Java EE7 datasource definition annotations and XML file as defined Java EE7 spec (chapter EE.5.18.3)

johnathonlee:
@jmesnil `*-ds.xml` predated IronJacamar and is not an IJ deployment file.  For EAP it was considered an unmanaged connection: https://access.redhat.com/solutions/85163

kabir:
@barreiro Can you please add in the new Community Documentation section I just mentioned on wildfly-dev?

barreiro:
Added the Community Documentation section

darranl:
Has there been any consideration to running this in a HostController process? Getting access to databases from the HostController process is still an issue for some.  I don't think this needs to be in the initial work but could be a good follow up.

barreiro:
@darranl no, that use case hasn't taken into consideration.

darranl:
@barreiro Ok, could be a good one to discuss under it's own RFE, let me know if you want to chat some time about when it would be used.

PR #65, opened 2018-05-15
Upgrade path to JPA 2.2

Sanne:


jpa/WFLY-6464_JPA_22_HibernateUpgrade.adoc
```diff
@@ -0,0 +1,86 @@
+= <INSERT TITLE HERE>
```
jamezp: The title should be updated

Sanne:
@jamezp done, sorry.

jpa/WFLY-6464_JPA_22_HibernateUpgrade.adoc
```diff
@@ -0,0 +1,86 @@
...
+
+== Overview
+
+Support JPA 2.2 by including Hibernate ORM 5.3.x and matching dependencies.
+
+The new libraries will be added in addition to the existing version of Hibernate ORM so that EE7 certification can happen via the JPA 2.1 based implementation.
+
+The new Hibernate ORM version will need to be activated explicitly via the `ee8.preview.mode=true` flag.
+
+An Hibernate ORM upgrade implies and Hibernate Search upgrade, which in turn implies an Apache Lucene upgrade.
```
kabir: Nitpick:
s/and Hibernate Search/an Hibernate Search

simkam:
Reading Compatibility analysis of Hibernate ORM (from 5.1 to 5.3) 

> Failures Mitigated using Compatibility Transformer
@Sanne can you please elaborate in document on what compatibility transformer is and what it does. 

> Deprecated in 5.1; Removed in 5.3
This might be fine from Hibernate PoV but it is not in EAP world. As far I know it is not allowed to remove things in EAP minor version.

Sanne:
@simkam regarding the Compatibility Transformer: good point I'll find some references and update the document.

Regarding any method which was removed: these were discussed with @n1hility and @dmlloyd and they reviewed it too specifically from a perspective of "EAP world".

Is there any specific method which concerns you? Probably best to call out any need for further details in the document.

Sanne:
@rsvoboda all details are in the documents, I just mean we've not been following Hibernate deprecation rules only and we're aware of the deprecation rules of EAP as we've been following David's and Jason's guidance and they've helped review those reports and docs personally.

So while I will defer to them, my personal understanding is that those methods which have removed are fine as they have always been deprecated in the life of EAP7 and therefore have completed their "deprecation cycle".

That's just my understanding, but perhaps best if you talk with @dmlloyd directly?

simkam:
My understanding is that deprecated in EAP7 means that it can only be removed in EAP8. I'd like to include also @bmaxwell for his opinions.

dmlloyd:
You can remove things in an EAP minor version, as long as they were deprecated in the previous major version.

simkam:
in that case, EAP 6.4/Hibernate 4.2.x, for example [o.h.ejb.HibernatePersistence](https://github.com/hibernate/hibernate-orm/blob/4.2/hibernate-entitymanager/src/main/java/org/hibernate/ejb/HibernatePersistence.java), only one method was deprecated and whole class is removed in 5.3. I didn't check other classes from analysis, but there might be more with the same case

dmlloyd:
Another thing to consider is, was this class considered API or SPI?  We allow SPI to be removed in a minor version.

Sanne:
I pushed an update to:

- clarify the requirements, especially around expected backward compatibility
- reference David's _Compatibility Transformer_

Sanne:
@simkam regarding o.h.ejb.HibernatePersistence: 

I think you indeed identified something that could have been done better in the past, but that cathegory of issues doesn't look like a blocking problem to allow people to (optionally) try Hibernate ORM 5.3.

Firstly like David suggested it's arguable if this is an API at all; the `PersistenceProvider` implementation is not something that user code will invoke directly nor import in their own code.
It is commonly just _named_ in a `persistence.xml` configuration file to override which provider one wants to use.

This is a very old one and was phased out eventually by logging a warning if the old name was being used; AFAIR the class has also been deprecated for some time, probably not yet in 4.2 (which was released in 2013!).

More importantly, I wonder why you're comparing 5.3 with 4.2. Even if I agree that could have been handled better, that ship has sailed as this class is already missing in WildFly today.

We can, of course, re-introduce such things. That's indeed the purpose of those documents and I hope your team will help us identify things we might have missed and improve it all.

This specific proposal though should focus on the differences between what is already in WildFly and what we want it to have to eventually get the EE8 certification. I changed the title of the document, as well as the "nice to have" requirements: hopefully that will make the intent clearer.

Thanks!

Sanne:
pushed an update: adding related reference to

 -  https://issues.jboss.org/browse/WFLY-10393 Upgrade Infinispan to 9.2.3.Final

scottmarlow:
IMO, it is too risky to not include ORM 5.3 in WF13.

Whether this change goes into WF13 or not, in WF14, we remove ORM 5.1.14.Final and only use ORM 5.3.0.Final (or 5.3.something)  If this changes goes into WF13, we will get community testing, if not, we will have less testing for WF14.

PR #67, opened 2018-05-23
WFCORE-3873 SNI support

stuartwdouglas:


stuartwdouglas:
@dmlloyd @darranl can you please review?

security/WFCORE-3873_SNI_Support.adoc
```diff
@@ -0,0 +1,76 @@
...
+=== Other Interested Projects
+
+== Requirements
+
+=== Hard Requirements
+
+This will involve adding a new `server-sni-ssl-context` resource to the Elytron subsystem. This resource will have
+a non-optional attribute named `default-ssl-context` that references an existing server-ssl-context.
+
+In addition this resource can have any number of sni-match subresources. These subresources map names (which may be
```
jstourac: `sni-match` may be put into between `` so it is clearly visible in the text

security/WFCORE-3873_SNI_Support.adoc
```diff
@@ -0,0 +1,76 @@
...
+wildcard names) to `server-ssl-context` resources, and will be used by the SNI support to determine the correct context
+to choose. Wildcard names use `*` as a wildcard, and can only be used to match a single level of subdomain in much the
+same way as with wildcard certificates.
```
jstourac: Out of curiosity - is something like `*.*.example.com` or just `*.*.com` possible? I guess we want to support only one wildcard in the `sni-match`, correct?

kabir:
@stuartwdouglas Can you please add in the new Community Documentation section I just mentioned on wildfly-dev?

security/WFCORE-3873_SNI_Support.adoc
jstourac: @stuartwdouglas I've just realized this question was not answered. I still suppose the right behavior is not to allow this.

jstourac:
I am missing an answer for the review but generally this is okay to go.

security/WFCORE-3873_SNI_Support.adoc
stuartwdouglas: Yea, we only want to allow a single wildcard

security/WFCORE-3873_SNI_Support.adoc
```diff
@@ -0,0 +1,76 @@
...
+* Undertow
+* Elytron
+
+=== Other Interested Projects
+
+== Requirements
+
+=== Hard Requirements
+
+This will involve adding a new `server-sni-ssl-context` resource to the Elytron subsystem. This resource will have
```
mchoma: How is `server-sni-ssl-context` implemented? Does it "inherit" from `server-ssl-context`? Will any update of `server-ssl-context` (e.g. TLSv1.3) automatically propagate to `server-sni-ssl-context` or do we need pay attention to get to both now?

darranl: That comment is likely more appropriate on the TLS 1.3 RFE and any resulting analysis, this is quite an old RFE now so we wouldn't retrospectively start worrying about TLS 1.3 at this stage which is a new RFE.

mchoma: My note is related to implementaiton of this feature. So I htink it belongs here. TLS 1.3 is just example. I mean in general if there is any change on `server-ssl-context`, e.g. new attribute. Will it be automatically available in `server-sni-ssl-context`. I dont see any PR with implementation 
anywhere. 

Or `server-sni-ssl-context` is copy of `server-ssl-context` with additional attributes?

darranl: The pull request has now been merged to WildFly Core containing the subsystem changes to configure an SNI aware SSLContext.

mchoma: I looked into implementation and I have realized my concern was wrong. `server-sni-ssl-context` has only 2 attributes: DEFAULT_SSL_CONTEXT, HOST_CONTEXT_MAP. So if anything changed in `server-ssl-context` it will be in `server-sni-ssl-context` as well.

PR #72, opened 2018-05-30
EAP7-984 Allow to search in the CLI multi-page output

jfdenise:


cli/EAP7-984_Search_Output.adoc
```diff
@@ -0,0 +1,70 @@
...
+* mailto:eduda@redhat.com[Erich Duda]
+
+=== Affected Projects or Components:
+
+* WildFly Core CLI
+
+
+== Requirements
+
+=== Hard requirements
```
dudaerich: Can you please elaborate what type the pattern is? Is it simple string or regular expression?

dudaerich: Hmm, as Linux/Git user, I am used to regular expressions when I search text in long outputs (e.g. `git log`). Is there some particular reason why you chose simple text?

jfdenise: The reason is to keep the feature surface small, really address the minimal search requirements. If that is a too big limitation, that would make the feature unusable, then this could be revisited.

dudaerich: I see. IMO it is not big limitation and it can be added later. Thank you.

cli/EAP7-984_Search_Output.adoc
```diff
@@ -0,0 +1,70 @@
...
+== Issue Metadata
+
+=== Issue:
+
+* {issue-base-url}EAP7-984[EAP7-984]
+
+=== Related Issues:
+
+* {issue-base-url}WFCORE-3743
+* {issue-base-url}WFCORE-3568 (dependency on this one).
```
dudaerich: How is this feature related to [WFCORE-3568 Allow CLI output scrolling](https://issues.jboss.org/browse/WFCORE-3568)?

In the design notes it is written that the feature will search only in boundaries of the current paged content. This is OK for "more" behavior. However the WFCORE-3568 will implement the "less" behavior and it would be more natural if this feature will search also outside boundaries of the current paged content. Will be this enhancement implemented in scope of WFCORE-3568?

jfdenise: We will stick to the more behavior. The CLI has a custom paging that is not similar to less. The scrolling of WFCORE-3568 stays in the boundary of the page, so we keep the CLI approach there too.

dudaerich: The first use case which came to my mind when I heard about this feature was: I will execute read-resource-description on messaging subsystem and then I will type `/discovery-group` and the output will be moved to this attribute. However what I actually will have to do is:

```
/discovery-group
PAGE_DOWN
/discovery-group
PAGE_DOWN
/discovery-group
PAGE_DOWN
...
```

Am I right?

jfdenise: I think that it will behave the way you want.  You type /XXX then the display will be moved to the first occurence of XXX if any. Then typing 'n' will move you to the next occurence. 'N' to go back to the previous one if any. PAGE_DOWN will be unaffected, if you type PAGE_DOWN you will go one page down. If some match are found during navigation they are highlighted. Would that work?

dudaerich: What if I type /XXX but the XXX is not present on display but it is present in the last line of the output (which is not displayed)? Will be the display moved to the last line of output?

jfdenise: It will moved to the last page in order to visualize the highlighted match.

dudaerich: Then it is ok. Thank you :)

jfdenise: yes, I can understand your point with less you go to the actual line, in CLI you will go to the last (of first) page that contains content.
I realize that one limitation willl be resizing. Resizing when paging content is not currently well supported. We should explicitly say that it is a limitation

jfdenise: We should at least support resizing terminal height. When paging it makes full sense for users to make to terminal window longer to display more content.

dudaerich: I bit misunderstood you what you think by more behavior. I think it is ok that the display won't be moved to actual line if the text will be highlighted.

Can you please put your concerns about resizing into the document to not lost in the comments?

jfdenise: I added terminal re-sizing as a soft requirement.

kabir:
@jfdenise Can you please add in the new Community Documentation section I just mentioned on wildfly-dev?

jfdenise:
@kabir , done. Sorry for that.

PR #81, opened 2018-06-12
Add section for community documentation

kabir:


kabir:
-1 on modifying existing and merged documents
A halfhearted +1 on trying to get people to update their existing PRs

rsvoboda:
> -1 on modifying existing and merged documents

Why? You can mass-edit them and add empty section like:
```
== Community Documentation
```

The structure should be kept the same even for older docs.

You have transformers in WF to keep old configs functional, this is almost the same.

kabir:
Transformers are for making sure that new stuff does not make its way to the old versions which is the opposite IMO. If you want to open a PR adding this, but all means go ahead, but the main driver for this is to not have to keep asking again and again whether a feature needs documentation and/or where it is

PR #84, opened 2018-06-14
Analysis for EAP7-1021, CLI key navigation

jfdenise:


cli/EAP7-1021_Command_Navigation.adoc
```diff
@@ -0,0 +1,84 @@
...
+
+* Ctrl+A or HOME: Linux, Windows, Solaris, Mac OSX
+
+=== Go to the end of the line
+
+* Ctrl+E or END: Linux, Windows, Solaris, Mac OSX
+
+=== Go left (back) one character
+
+* Ctrl+B: Linux, Windows, Solaris, Mac OSX
```
dudaerich: I know that this is obvious but to be fully complete there should be "or Arrow left". The same applies for Go right (forward) one character.

jfdenise: Good catch, I updated for both left/right arrow.

jfdenise:
@rsvoboda , tested on HP-UX and updated analysis.

jfdenise:
@dudaerich , I updated the analysis to state that HOME/END key are not supported for Mac OSX, these keys go to beginning/end of document (terminal in our case).

dudaerich:
@jfdenise , did you push the changes? I don't see any mention that HOME/END key are not supported for Mac OSX.

jfdenise:
@dudaerich , pushed.

dudaerich:
@jfdenise I've reviewed the change and it's ok for me. Thanks.

kabir:
@jfdenise Can you please add in the new Community Documentation section I just mentioned on wildfly-dev?

jfdenise:
@kabir , done. Sorry for that.

PR #88, opened 2018-06-18
[WFCORE-3932]: Adding support for standalone to standalone model synchronization / model difference computation

ehsavoie:
Jira: https://issues.jboss.org/browse/WFCORE-3932
Current code status: https://github.com/ehsavoie/wildfly-core/tree/WFCORE-3932

### DIFF
--- a/controller/model-diff.adoc
+++ b/controller/model-diff.adoc
@@ -1,0 +1,81 @@
+Export manual configuration changes as a Galleon Feature Pack
+:author:            Emmanuel Hugonnet
+:email:             ehugonne@redhat.com
+:toc:               left
+:icons:             font
+:idprefix:
+:idseparator:       -
+
+== Overview
+A Galleon provisioned server may be reconfigured or modified manually by the user via jboss-cli or direct editing of configuration.
+Also files may be added or removed like deployments or modules for instance;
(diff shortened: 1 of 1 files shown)


aloubyansky:
From Galleon and provisioning perspective and roadmap the described functionality is a highly critical one that we need, ASAP. Thanks.

kabir:
@ehsavoie Can you please add in the new Community Documentation section I just mentioned on wildfly-dev?

LittleJohnII:
@kabir can we get an EAP7 planning issue for this?
@ehsavoie I have a question about the requirements in the proposal. In hard requirements you write that this should target the complete model, including deployments. In non-requirements, there's written that files (including deployments) are not targeted. IIUC, this means that the result of the diff will contain the information about what was deployed, but not the actual deployment files?

ehsavoie:
@LittleJohnII yes you are correct, Galleon provides already the filesystem changes so those files are already available. In a could environment I guess the deployment files would me mounted shared filesystem or whatever thus only the information about which files we are going to use is pertinent.

aloubyansky:
Reabsed (had to fix the build) commit is https://github.com/aloubyansky/wildfly-core/commit/426251ab7c0e80ae135440085b86e7c16cbff095

PR #89, opened 2018-06-19
[WFLY-10518] Connections to remote Artemis broker

jmesnil:
JIRA: https://issues.jboss.org/browse/WFLY-10518

### DIFF
--- a/messaging/WFLY-10518_artemis_no_local_broker.adoc
+++ b/messaging/WFLY-10518_artemis_no_local_broker.adoc
@@ -1,0 +1,168 @@
+= [WFLY-10518] Connections to remote Artemis broker must not require local Artemis instance
+:author:            Jeff Mesnil
+:email:             jmesnil@redhat.com
+:toc:               left
+:icons:             font
+:keywords:          messaging,jms,openshift
+:idprefix:
+:idseparator:       -
+
+== Overview
+
+Connections to remote Artemis-based broker (including A-MQ 7) are handled in
+WildFly by the
+https://wildscribe.github.io/WildFly/13.0/subsystem/messaging-activemq/server/pooled-connection-factory/index.html[pooled-connection-factory]
+resource that is in the `messaging-activemq`.
+
+This `pooled-connection-factory` resource can be used to connect to the local
+ Artemis server that "owns" it or to a remote Artemis broker.
+
+However, this `pooled-connection-factory` resource always require the presence
+ of a local Artemis server even when it connects only to a remote Artemis broker.
+This use case will become more prevalent as WildFly is deployed on OpenShift to
+connect to a remote A-MQ 7 broker.
+
+The presence of a local Artemis server has impact on the memory and CPU
+ footprint of WildFly image on OpenShift.
+This RFE will improve WildFly so that a connection to a remote Artemis-based
+ broker will no longer require the presence of a local instance of Artemis.
+
+== Issue Metadata
+
+=== Issue
+
+* https://issues.jboss.org/browse/WFLY-10518[WFLY-10518]
+
+=== Related Issues
+
+* https://issues.jboss.org/browse/EAP7-1014[EAP7-1014]
+* https://issues.jboss.org/browse/EAP7-1006[EAP7-1006] is a RFE related to Java EE7 JMS resource
+definitions that will benefit from this RFE.
+
+=== Dev Contacts
+
+* mailto:{email}[{author}]
+
+=== QE Contacts
+
+* mailto:mnovak@redhat.com[Miroslav Novak]
+
+=== Affected Projects or Components
+
+* This is a internal restructuring of the `messaging-activemq` subsystem that requires no changes
+from Artemis or other messaging components.
+
+=== Other Interested Projects
+
+== Requirements
+
+* It must be possible to configure WildFly to connect to a remote Artemis-based broker (using the Artemis resource adapter)
+  without having to run a local Artemis broker in WildFly.
+* Use Capabilities/Requirements to ensure that Artemis client or server resources are properly configured.
+* Previous and current subsystem schemas and models must remain compatible (using aliases to the new resources).
+
+=== Nice-to-Have Requirements
+
+* Refactor the Artemis modules to have a proper split between Artemis client and server code.
+* The Artemis resource adapter should only need to depend on Artemis client module.
+
+== Proposed Design
+
+Move resources up from `/subsystem=messaging/server` to `/subsystem=messaging/`:
+
+  * `connector`, `http-connector`, `remote-connector`, `in-vm-connector`
+  * `discovery-group`
+  * `pooled-connection-factory`
+  * `connection-factory`
+
+Note that a `server` can still be able to reference some of these resources
+(e.g. `discovery-group`).
+
+Other resources that remains in `/subsystem=messaging/server`:
+
+  * `acceptor`, `http-acceptor`, `remote-acceptor`, `in-vm-acceptor` (they corresponds to Artemis broker configuration)
+  * `broadcast-group` (this kind of resource is only used by Artemis brokers to broadcast cluster information)
+
+=== JMS Destination resources
+
+With the proposed design, a `pooled-connection-factory` resource is able to connect to a
+remote Artemis broker.
+However, to provide a fully functional JMS application, the user also needs to inject (or
+  lookup in JNDI) JMS Destinations object that represents the queue or topic managed by the remote
+  broker.
+
+There are `jms-queue` and `jms-topic` resources under the `server` resources but they require
+broker code to create and configure the queue on the local Artemis broker.
+
+We need another type of resources that represents JMS Destination object to store in JNDI.
+
+These new resources will be under `/subsystem=messaging-activemq`:
+
+  * `client-jms-queue`
+  * `client-jms-topic`
+
+These resources have a single attribute:
+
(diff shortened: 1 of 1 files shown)


messaging/WFLY-10518_artemis_no_local_broker.adoc
```diff
@@ -0,0 +1,166 @@
...
+
+These new resources will be under `/subsystem=messaging-activemq`:
+
+  * `client-jms-queue`
+  * `client-jms-topic`
+
+These resource have only 2 attributes:
+
+  * `name` - the name of the destination that *MUST* be identical to the name of the destinations configured
+    in the remote Artemis broker
```
bstansberry: How is a 'name' attribute distinct from the value of the last element in the resource address?

messaging/WFLY-10518_artemis_no_local_broker.adoc
```diff
@@ -0,0 +1,166 @@
...
+* mailto:{email}[{author}]
+
+=== QE Contacts
+
+* mailto:mnovak@redhat.com[Miroslav Novak]
+
+=== Affected Projects or Components
+
+* This is a internal restructuring of the `messaging-activemq` subsystem that requires no changes
+from Artemis or other messaging components.
```
bstansberry: It will result in changes to our OpenShift images though.

messaging/WFLY-10518_artemis_no_local_broker.adoc
jmesnil: sorry, the description is not correct. 
What I meant is that the `name` of the resource is meaningful and must be identical to the name of the destination in the remote Artemis broker.

messaging/WFLY-10518_artemis_no_local_broker.adoc
```diff
@@ -0,0 +1,168 @@
...
+  * `client-jms-topic`
+
+These resources have a single attribute:
+
+  * `jndi-name` - the JNDI binding to store the Destination object
+
+The name of these resources are meaningful: it *MUST* be identical to the name of the destinations configured
+    in the remote Artemis broker
+
+These resources have *no* management operations (e.g. to flush messages, etc.). They are just local
```
bstansberry: s/*no* management operations/*no* runtime management operations/g

I'm picking nits. :)

messaging/WFLY-10518_artemis_no_local_broker.adoc
rsvoboda: @jmesnil @bstansberry will be there some migration path for
a) current OpenShift deployments
b) classical .zip - e.g. some equivalent for migrate operation
?

messaging/WFLY-10518_artemis_no_local_broker.adoc
```diff
@@ -0,0 +1,168 @@
...
+https://wildscribe.github.io/WildFly/13.0/subsystem/messaging-activemq/server/pooled-connection-factory/index.html[pooled-connection-factory]
+resource that is in the `messaging-activemq`.
+
+This `pooled-connection-factory` resource can be used to connect to the local
+ Artemis server that "owns" it or to a remote Artemis broker.
+
+However, this `pooled-connection-factory` resource always require the presence
+ of a local Artemis server even when it connects only to a remote Artemis broker.
+This use case will become more prevalent as WildFly is deployed on OpenShift to
+connect to a remote A-MQ 7 broker.
```
rsvoboda: EAP classical .zip supports integration with A-MQ 7.
Is this RFE going to be the new way of EAP classical .zip integration with A-MQ 7?

If this is going to be the way for EAP classical .zip integration with A-MQ7 aka 3rd party messaging, we need new RFE for this. 

Currently EAP classical .zip integration with A-MQ7 is done via `/subsystem=messaging/server` and if we change the supported way we are basically dropping feature ...

messaging/WFLY-10518_artemis_no_local_broker.adoc
```diff
@@ -0,0 +1,168 @@
...
+* mailto:mnovak@redhat.com[Miroslav Novak]
+
+=== Affected Projects or Components
+
+* This is a internal restructuring of the `messaging-activemq` subsystem that requires no changes
+from Artemis or other messaging components.
+
+=== Other Interested Projects
+
+== Requirements
```
rsvoboda: This feature must be transparent for end deployment, only server configuration changes are acceptable.

Will be this transparent standalone client too ? Like it will be able to do all the stuff it was able to do with embedded Artemis broker.

messaging/WFLY-10518_artemis_no_local_broker.adoc
jmesnil: I don't follow your point.

This RFE is not dropping any feature. It adds the feature to be able to connect to a remote A-MQ7 *without running a local broker*.
Currently the integration with A-MQ7 is done with `/subsystem=messaging/server/pooled-connection-factory`. With this RFE, it can be done instead with  `/subsystem=messaging/pooled-connection-factory` (no local server required).

messaging/WFLY-10518_artemis_no_local_broker.adoc
jmesnil: For deployments and standalone clients that uses JMS by looking up resources from JNDI, that will be transparent.

messaging/WFLY-10518_artemis_no_local_broker.adoc
rsvoboda: My point is that we would have 2 ways now for classical .zip

For 3rd party messaging scope we cover `/subsystem=messaging/server` way.
For new way in 3rd party messaging scope we would require new RFE, this is one is OpenShift related and I would like to keep it that way.

Supporting 2 ways on classical .zip has some price, especially on HW load/execution time ++ impact to CP stream.

messaging/WFLY-10518_artemis_no_local_broker.adoc
rsvoboda: What about native api  and direct queue addressing ?

messaging/WFLY-10518_artemis_no_local_broker.adoc
bstansberry: AIUI and IMHO this would be the recommended way on bare metal as well. I don't see anything specific to OpenShift here beyond an offhand reference to using fewer resources there.

Running an embedded broker simply to connect to a remote broker is a waste of resources regardless of the operational environment. And running unnecessary software means greater vulnerability to bugs and security issues.

Someone is going to have to test this for it to be used in the images. If it's tested on OS but we tell bare metal customers it's not supported because <???> that's a strange message.

messaging/WFLY-10518_artemis_no_local_broker.adoc
bstansberry: For bare metal or any cloud situation where the user is bringing their own standalone.xml, I don't see a need for a migrate operation. We've used migrate operations when we have removed something and the user must make a config change to have a working server. That's not the case here. The old way still works fine.

The migration tool has gone beyond that a bit in trying to adapt the user's old config to whatever "best practices" are in the new release, but I doubt that's feasible here. I don't know how the tool would know that the embedded broker in the old config isn't needed.

For these kinds of situations it sounds to me like a docs issue. Perhaps  I'll feel differently when we get a bunch of migration code contributors who are up for long term maintenance too. ;)

As for OpenShift deployments moving from EAP 7.1 to 7.2 where they are relying on our tooling to generate their config document we'll need to work out a migration path so people who need the embedded broker don't lose the ability to have it. All these things are controlled by env vars, so perhaps something like this:

USE_OLD_WAY -- default is true in 7.2, no value results in a WARN to stdout saying default will change to false in 7.3 and recommending setting....
USE_NEW_WAY -- default is false in 7.2, true in 7.3

Of course those aren't the real env var names.

It's possible in our 7.3 images we'd not generate an embedded broker config at all any more. Our basic direction (and not just us engineers) is that an embedded broker is a poor fit on the cloud. The reasons for that go well beyond resource utilization. Our docs already state that users shouldn't use an embedded broker.

@luck3y fyi.

messaging/WFLY-10518_artemis_no_local_broker.adoc
rsvoboda: I agree that new was should be more efficient, but existing customers witch are using standalone AMQ won't be probably migrating to this if they are not pushed.

>AIUI and IMHO this would be the recommended way on bare metal as well.

It must be explicitly stated in description of the RFE so it gets documented.


If this is going to be the default way, we can refocus our testing and do limited testing of the old way to limit risks.

messaging/WFLY-10518_artemis_no_local_broker.adoc
rsvoboda: @tremes this feature will affect OpenShift

messaging/WFLY-10518_artemis_no_local_broker.adoc
jmesnil: Clients using ActiveMQ Core protocol are connecting directly to the server. this feature does not change anything about that.

messaging/WFLY-10518_artemis_no_local_broker.adoc
bstansberry: @tremes Note though that we (aka @luck3y) are likely not going to develop any image scripting changes to configure standalone-openshift.xml this way before the WF 14 feature freeze. If you want to experiment with this in an image before then let's chat.

PR #90, opened 2018-06-21
[WFLY-10614] Referencing WildFly Elytron SecurityDomains as an alternative to the current http-authentication-factory reference.

darranl:


undertow/WFLY-10614_Elytron_SecurityDomain_Referencing.adoc
```diff
@@ -0,0 +1,98 @@
...
+
+=== Other Interested Projects
+
+== Requirements
+
+=== Hard Requirements
+
+The `application-security-domain` resource in the Undertow subsystem will support a new `security-domain` attribute as an alternative to the existing `http-authentication-factory` attribute - these two attributes will be defined as 'alternatives' to ensure mutual-exclusivity.
+
+Where a deployment matches against an `application-security-domain` mapping, if that mapping references a `security-domain` then provided the web.xml is configured to use the mechanisms referenced in the servlet specification (BASIC, CLIENT_CERT, DIGEST, or FORM) then these will still be enabled for the application without requiring a `http-authentication-factory` reference.
```
mchoma: In this case Elytron mechanism implementation will be used or Undertow implementation?

undertow/WFLY-10614_Elytron_SecurityDomain_Referencing.adoc
```diff
@@ -0,0 +1,98 @@
...
+The existing `http-authentication-factory` will remain in the model without deprecation as a fully supported form of configuration, after this enhancement has been implemented it will still be appropriate to reference a `http-authentication-factory` for a number of reasons, examples include: -
+
+* Using custom HTTP authentication mechanisms.
+* Using mechanisms other than the four described in the Servlet specification.
+* Where the configuration in the deployment is to be overridden.
+* Where additional configuration needs to be defined for the mechanisms such as principal transformers, credential factories, mechanism realms etc...
+
+=== Nice-to-Have Requirements
+
+The default configuration will be cleaned up to remove the `/subsystem=elytron/http-authentication-factory=application-http-authentication` resource as by default this is only used for BASIC and FORM authentication anyway.
```
mchoma: I think we can't remove it in this release, but only deprecate. Removal can break users migrating from previous releases.

If user migrate app + standalone.xml it will be fine. But if user has app + CLI scripts (which does not create application-http-authentication because it was by default there) migration will fail

undertow/WFLY-10614_Elytron_SecurityDomain_Referencing.adoc
darranl: That is correct, this enhancement is predominantly a configuration simplification to cut our a part of the chain of configuration that would be redundant for many users.

undertow/WFLY-10614_Elytron_SecurityDomain_Referencing.adoc
darranl: @jmesnil / @bstansberry Where are we currently on the level of compatibility that must be maintained in relation to the default configuration.  I believe in the past we said it is actual configuration that must be moved from version to version not scripts.

Users that have written scripts that define their own hierarchy of resources will be unaffected, in this case it is only going to be the ones that refer to the existing named resource that would be affected.

We could consider adding some kind of 'restore-*' cli script that can add it back for the occasional few who did find a way to depend on it - or as a single resource adding it back themselves is only one command.

Outside of this we have plans to switch the default security implementation around WildFly 15 and then remove legacy entirely after that at a later point - both of these will impact the default configuration and any scripts built off the default configuration.

bstansberry: It's never been a requirement that users can take scripts written to update a standard config from one minor release, execute them against the standard config from another minor release, and be guaranteed that all will work fine.  I think users trying to do that are engaged in an anti-pattern, since our standard configs are a way we introduce latest features and best practices, and blindly jumping to those is a mistake. We support running old configs and introduced the config migration tool both as ways to allow users to move to our latest and greatest in a controlled manner.

bstansberry: @aloubyansky This is an interesting discussion in the context of galleon.

darranl:
@jfdenise Will this have any impact on the commands being supported by the CLI?  The resources can all be defined as before but the change that may affect you is we are taking away some of the default configuration so not sure if you rely on it?

jfdenise:
@darranl , yes we do rely on 'application-http-authentication' http-authentication-factory if the user just enable http-auth using "default" configuration.
We would have to evolve the CLI command to see if this resource exists, if it doesn't then fallback to offer the user a way to set a security-domain. BTW, what are the resources "security-domain", the one defined in security sub-system?

darranl:
@jfdenise The security-domain reference is to the security domain in the Elyron subsystem.

Unless the CLI requires mechanisms outside those specified in the servlet specification it probably would be good to evolve to reference the security domain instead of an authentication factory - the authentication factory then becomes something that would be used for advanced forms of configuration.

darranl:
@jfdenise Yes ApplicationDomain will remain, it is only the http-authentication-factory that has been made redundant.

jfdenise:
@darranl , you could add a note in the Analysis in order to document that some CLI changes are required, something like (@mchoma any feedback?):

=== Impact on CLI

The CLI command _security enable-http-auth-http-server_ should be modified to:

- Replace OOTB authentication factory usage by OOTB security domain for simplest authentication setup.
- Expose a new option _--referenced-security-domain_ that is exclusive with any other options that target http authentication factory (eg: _--mechanism_ option).
- The referenced security-domain must exist, CLI doesn't create the resource resource.
- _--referenced-security-domain_ completer exposes all existing elytron security-domains names.
- Existing options and command behavior are untouched.

Examples:

_security enable-http-auth-http-server --security-domain=foo_ ==> create foo undertow security domain and set "ApplicationDomain" as the security domain. +
_security enable-http-auth-http-server --security-domain=foo --referenced-security-domain=bar_ ==> create foo undertow security domain and set bar as the security domain. +

darranl:
Based on the questions and comments so far I have added some additional clarifications and the suggested text from @jfdenise.

@jfdenise is the CLI change something you will be able to look into?  Apart from testsuite clean up and documentation I think my changes are complete now - if I push my topic branches to our incubator project so we can collaborate will you be able to submit PRs there?

jfdenise:
@darranl , yes I can submit PR to your WIP repo. The test of this command are in wildfly Full, so I will have to open a PR against full too.
What about CLI community doc? I am wandering if the CLI shouldn't have its own EAP7 JIRA entry to track these changes? Would be aligned with the process.

darranl:
@kabir what do you think process wise here?  If @jfdenise is about to collaborate on these changes as they affect the CLI do we need a separate RFE?  IMO I think we can get away with one RFE and coordinate together.

On the coordination side I think we may need to break this down into a few steps to minimise the coordination needed on merging.

The changes to the Undertow subsystem go in quite cleanly so that can probably go to WildFly first.
We should then likely have a PR to WildFly Core for the CLI functional changes.

We will then need to wait for a WildFly Core tag.

Then the tests in WildFly can be updated in anticipation of the resource being dropped from the config.
Then we can drop the config in WildFly Core.

Then one more WildFly Core tag will sync everything.

Sounds like a lot of steps but it avoids a coordination multi merge, tag merge.

kabir:
@darranl @jfdenise collaborating sounds good. The one thing which jars me a bit is merging the undertow subsystem changes first. Can this all be done and reviewed + verified on feature branches before we start merging bits of the feature?

darranl:
+1 I am assuming that we will need that one reviewed and verified first - that one "is the change", the rest of this complexity is default config clean up and handling changes across two repositories.

jfdenise:
@darranl, @kabir, the PR with tests that I will open against wildfly full will have run results that show that the tests are passing with the feature branch. @darranl , if you could send me the link to the branch. Thanks.

darranl:
@jfdenise I will add the links for topic branches, TBH I think you should just send PRs against the topic branches and I will send PRs from the topic branches rather than having independent PRs in the upstream queue that will not pass CI

darranl:
The current proposal can be found on this branch, please send PRs here for modifications to the proposal: -

https://github.com/wildfly-security-incubator/wildfly-proposals/tree/WFLY-10614

@mchoma - if you have anything to add to the proposal you can also send a PR to this branch ^^

The main topic branch for the feature updating the Undertow subsystem can be found here: -

https://github.com/wildfly-security-incubator/wildfly/tree/WFLY-10614

The topic branch to drop the http-authentication-factory from the configuration is at: -

https://github.com/wildfly-security-incubator/wildfly-core/tree/WFCORE-3858

So I think looking at that regarding coordination the time to submit the CLI change to core will be after the first branch to WildFly has been merged.  We can create a new topic branch in the incubator with the CLI specific part of the change once implemented but I think we should hold off the upstream PR until a clean CI run is expected, otherwise we end up with unmergable PRs cluttering up the queue.

mchoma:
@jfdenise I think proposed changes to CLI are OK.

PR #104, opened 2018-07-19
[WFCORE-3947] OCSP trustmanager

hkalina:
https://issues.jboss.org/browse/WFCORE-3947
https://issues.jboss.org/browse/EAP7-715

### DIFF
--- a/elytron/WFCORE-3947-ocsp.adoc
+++ b/elytron/WFCORE-3947-ocsp.adoc
@@ -1,0 +1,79 @@
+= WFCORE-3947 OCSP support in Elytron SSL truststore
+:author:            Jan Kalina
+:email:             jkalina@redhat.com
+:toc:               left
+:icons:             font
+:keywords:          elytron
+:idprefix:
+:idseparator:       -
+
+== Overview
+
+Elytron provides trustmanager usable for SSL contexts used by other parts of WildFly.
+It supports certificate revocations using certificate revocations list (CRL), but support of Online Certificate Status Protocol (OCSP) is required too.
+This is required especially for client certificates validation.
+
+The advantages of OCSP in comparison with CRL are:
+* faster revocation (CRLs are usualy generated in constant periods)
+* smaller line bandwidth needed (CRLs requires to everytime download list of all revoked nonexpired certificates)
+
+== Issue Metadata
+
+=== Issue
+
+* https://issues.jboss.org/browse/EAP7-715[EAP7-715]
+
+=== Related Issues
+
+* https://issues.jboss.org/browse/WFCORE-3947[WFCORE-3947]
+
+=== Dev Contacts
+
+* mailto:jkalina@redhat.com[Jan Kalina]
+
+=== QE Contacts
+
+* mailto:jstourac@redhat.com[Jan Stourac]
+
+=== Affected Projects or Components
+
+* WildFly Core
+* WildFly Elytron
+
+//=== Other Interested Projects
+
+== Requirements
+
+=== Hard Requirements
+
+* OCSP will be supported as an alternative to CRL in Elytron subsystem trustmanager (for client certificates verification).
+* By default will be the OCSP responder URL obtained from the validated certificates.
+* Should be possible to specify the OCSP responder (URL and certificate) manually, ignoring responder defined in certificates - for example, for needs of creating network-level caching responders.  (`responder`)
+* OCSP responses should be cached. (`cache-enabled`, `cache-max-age`, `cache-max-entries`)
+* Configurable timeout preventing too long waiting for unavailable OCSP responder. (`connection-timeout`)
+* Configurable behavior (accept/reject) for unknown certificates. (OCSP response UNKNOWN) (boolean `accept-unknown`)
+* Configurable behavior (accept/reject) for unavailable OCSP responder. (boolean `accept-when-status-unavailable`)
+* If the OCSP responder is not defined in the trustmanager nor in the certificate (the certificate was issued without possibility to be revoked), the certificate has to be accepted.
+
+=== Nice-to-Have Requirements
+
+* OCSP would be nice on Elytron-client side too.
+* OCSP requests signing is not obligatory.
+* Multiple OCSP requests (to responders of different CAs) can be sent in parallel.
+
+=== Non-Requirements
+
+* Support of OCSP with CRL as fallback would be welcome, but is not required.
+
+== Test Plan
+
+* Test accepting/rejecting remoting client by good/revoked SSL certificate.
+* Test behavior when OCSP server is unavailable/does not respond correctly/in time.
+* Caching: test whether certificate status is cached (not requesting OCSP server again).
+* Caching: test whether certificate status is updated on timeout.
+* Test with invalid signature in the response.
+
+== Community Documentation
+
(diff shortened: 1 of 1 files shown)


elytron/WFCORE-3947-ocsp.adoc
```diff
@@ -0,0 +1,67 @@
...
+* mailto:jstourac@redhat.com[Jan Stourac]
+
+=== Affected Projects or Components
+
+* WildFly Core
+* WildFly Elytron
+
+//=== Other Interested Projects
+
+== Requirements
```
jstourac: Could you please describe what actually will be configurable and how? E.g. shall we support to configure only one OCSP responder at one moment? Or shall we allow to have some backup one (or even a list of backup responders) in case that the primary is not available? What are timeouts in such case? Is the timeout configured per every such responder or per whole request?

Maybe I complicate things too much, but we should probably have this described and clarified here. I guess that simplest solution should be good enough?

Also, what are the defaults? Is there gonna be some default OCSP responder provided such as [1]?

What is the default reaction when OCSP response is UNKNOWN or is not available at all? Is this still part of this RFE?

[1] http://wiki.cacert.org/OcspResponder

elytron/WFCORE-3947-ocsp.adoc
```diff
@@ -0,0 +1,67 @@
...
+
+=== Hard Requirements
+
+* OCSP will be supported as an alternative to CRL in Elytron subsystem trustmanager (for client certificates verification).
+* Should be possible to specify OCSP responder manually, ignoring responder defined in certificates, for needs of creating caching responders/mirrors.
```
jstourac: Do I understand correctly that in case we don't have an OCSP responder setup manually, the caching would not be available? Or maybe I just don't completely understand this bullet.

elytron/WFCORE-3947-ocsp.adoc
```diff
@@ -0,0 +1,67 @@
...
+* WildFly Core
+* WildFly Elytron
+
+//=== Other Interested Projects
+
+== Requirements
+
+=== Hard Requirements
+
+* OCSP will be supported as an alternative to CRL in Elytron subsystem trustmanager (for client certificates verification).
```
jstourac: Do we want to enforce exclusive use here? I mean - only CRL or OCSP can be configured and used exclusively at one moment?

elytron/WFCORE-3947-ocsp.adoc
```diff
@@ -0,0 +1,67 @@
...
+* Multiple OCSP requests (to responders of different CAs) should be sent in parallel.
+* OCSP responses should be cached.
+
+//=== Non-Requirements
+
+== Test Plan
+
+* Test accepting/rejecting remoting client by good/revoked SSL certificate.
+* Test behavior when OCSP server is unavailable/does not respond correctly/in time.
+* If caching implemented, test whether certificate status is updated on timeout.
```
jstourac: Hm, maybe we should also verify that cache is used in such case... although, I am not sure how to achieve this at the moment.

elytron/WFCORE-3947-ocsp.adoc
hkalina: This is related to network-level caching - OCSP server in the datacenter working as proxy for multiple servers. When not set, URL of responder defined directly in the certificate is used.

elytron/WFCORE-3947-ocsp.adoc
hkalina: We can support both in theory, but we need to define how to combine them:
* a) should be CRL used only when OCSP is unavailable?
* b) should be every certificate checked against both of them?

elytron/WFCORE-3947-ocsp.adoc
```diff
@@ -0,0 +1,67 @@
...
+* mailto:jkalina@redhat.com[Jan Kalina]
+
+=== QE Contacts
+
+* mailto:jstourac@redhat.com[Jan Stourac]
+
+=== Affected Projects or Components
+
+* WildFly Core
+* WildFly Elytron
```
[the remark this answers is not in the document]
hkalina: Undertow only consumes SSLContext from elytron.

elytron/WFCORE-3947-ocsp.adoc
hkalina: I has not find any backup OCSP requirement, so will not add it until it will be requested.
As it is not possible to specify multiple OCSP URIs in one certificate (HTTP level high-availability is used), I would not consider necessary to support it for local OCSP mirrors.

elytron/WFCORE-3947-ocsp.adoc
```diff
@@ -0,0 +1,67 @@
...
+:toc:               left
+:icons:             font
+:keywords:          elytron
+:idprefix:
+:idseparator:       -
+
+== Overview
+
+Elytron provides trustmanager usable for SSL contexts used by other parts of WildFly.
+It supports certificate revocations using certificate revocations list (CRL), but support of Online Certificate Status Protocol (OCSP) is required too.
```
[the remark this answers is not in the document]
hkalina: added

elytron/WFCORE-3947-ocsp.adoc
```diff
@@ -0,0 +1,67 @@
...
+* OCSP requests signing is not obligatory.
+* Multiple OCSP requests (to responders of different CAs) should be sent in parallel.
+* OCSP responses should be cached.
+
+//=== Non-Requirements
+
+== Test Plan
+
+* Test accepting/rejecting remoting client by good/revoked SSL certificate.
+* Test behavior when OCSP server is unavailable/does not respond correctly/in time.
```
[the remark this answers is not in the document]
hkalina: added

elytron/WFCORE-3947-ocsp.adoc
jstourac: To be honest, I have no idea whether it is common requirement to have both checks available and configured or not.

If we do want to support it, then I have found that some tool from Citrix is able to work like that [1] - both CRL and OCSP can be configured. Here is their decision/priority table:

>* If OCSPcheck parameter is set as mandatory, CRLcheck parameter will not be effective even if it is optionally enabled internally.
>* If CRL check is mandatory, then CRL check will be effective even if OCSP is configured.
>* If OCSP and CRL check are optional, then OCSP configuration is given first preference and then CRL check applies.
>* Both OCSP and CRL check cannot be set as mandatory.

We could probably mimic this somehow...

Also, if both CRL and OCSP are configured as optional, we should perform check request for both of them right away so we have result of the querry ASAP for both type of checks, not waiting e.g. timeout of the OCSP to respond...

[1] https://support.citrix.com/article/CTX138304

hkalina: * OCSP only is implemented
* CRL only is implemented
* OCSP with CRL fallback is not implemented yet 
@darranl do we want to support OCSP with CRL fallback?

darranl: Unless we have a strong argument it is required I would suggest we pull additional strategies into their own RFEs.

Once we have the core feature in place it will be a lot quicker to make minor enhancements.

jstourac:
@hkalina, @nekdozjam I think that this PR shall be closed in favor of https://github.com/wildfly/wildfly-proposals/pull/188, hm?

nekdozjam:
@jstourac Yes

PR #105, opened 2018-07-20
[WFLY-10711] Observability: Support MicroProfile Health

jmesnil:
In order to improve Observability of WildFly on OpenShift, WildFly needs
to expose health check probes
so that the container can query whether WildFly and its deployed
applications are healthy or not.

This will be implemented by supporting Eclipse MicroProfile Health 1.0
for applications deployed in WildFly.
The integration uses the smallrye-health component to provide the
MicroProfile Health implementation.

JIRA: https://issues.jboss.org/browse/WFLY-10711

### DIFF
--- a/microprofile/WFLY-10711_health_check_extension.adoc
+++ b/microprofile/WFLY-10711_health_check_extension.adoc
@@ -1,0 +1,151 @@
+= [WFLY-10711] Support Eclipse MicroProfile Health
+:author:            Jeff Mesnil
+:email:             jmesnil@redhat.com
+:toc:               left
+:icons:             font
+:idprefix:
+:idseparator:       -
+:keywords:          observability,microprofile,health,openshift
+
+== Overview
+
+In order to improve Observability of WildFly on OpenShift, WildFly needs to expose health check probes
+so that the container can query whether WildFly and its deployed applications are healthy or not.
+
+This will be implemented by supporting Eclipse MicroProfile Health 1.0 for applications deployed in WildFly.
+The integration uses the smallrye-health component to provide the MicroProfile Health implementation.
+
+== Issue Metadata
+
+=== Issue
+
+* https://issues.jboss.org/browse/WFLY-10711[WFLY-10711]
+
+=== Related Issues
+
+* https://issues.jboss.org/browse/EAP7-1029[EAP7-1029]
+* https://issues.jboss.org/browse/CLOUD-2730[CLOUD-2730] - Implement EAP Health Check so that it can be consumed by OpenShift Console
+* https://issues.jboss.org/browse/WFLY-10522[WFLY-10522] - MicroProfile Config extension for WildFly.
+** There is no direct relation between MicroProfile Health and MicroProfile Config API but there should be consistency in their extensions (subsystem names, etc.)
+
+=== Dev Contacts
+
+* mailto:{email}[{author}]
+
+=== QE Contacts
+
+* mailto:rsvoboda@redhat.com[Rostislav Svoboda] - WildFly
+* mailto:tremes@redhat.com[Tomas Remes] - OpenShift
+
+=== Affected Projects or Components
+
+The implementation of Eclipse MicroProfile Health is provided by https://github.com/smallrye/smallrye-health[smallrye-health].
+
+smallrye-health uses CDI to discover all health check probes and provides a `SmallryeHeathReporter`.
+WildFly extensions are not using CDI and smallrye-health will have to be enhanced so that its `SmallryeHeathReporter` can
+also be used by adding/removing health check probes manually.
+
+=== Other Interested Projects
+
+* A https://github.com/jmesnil/wildfly-microprofile-health[prototype of a WildFly extension] for MicroProfile Health was developed during the development of MicroProfile Health 1.0.
+
+== Requirements
+
+* Support Eclipse MicroProfile Health 1.0 for applications deployed in WildFly.
+* Provide a management runtime operation to check the health of the application server and its deployed applications. The result of the runtime operation
+  will be the DMR representation of the https://github.com/eclipse/microprofile-health/blob/master/spec/src/main/asciidoc/protocol-wireformat.adoc#appendix-b-json-payload-specification[MicroProfile Health JSON payload]
+* Provide an HTTP endpoint for the https://github.com/eclipse/microprofile-health/blob/master/spec/src/main/asciidoc/protocol-wireformat.adoc[health wireformat].
+  It requires the presence of the WildFly https://github.com/wildfly/wildfly-capabilities/blob/master/org/wildfly/management/http-interface/capability.adoc[HTTP management interface capability]
+* The HTTP endpoint is global to the server: it aggregates health check probes from *all deployments*.
+* Secure access to the HTTP endpoint must be configurable.
+
+== Nice-to-Have Requirements
+
+* Support drilldown of the HTTP endpoint (e.g. to be able to query the health of individual deployment, e.g. `/health/deploymentA`)
+* Provide a WildFly management API so that other extensions can define health check probes that taken into account
+  by the HTTP endpoint (this is out of scope of this proposal but the design of the new extension must not preclude to
+  provide such an API at a later point).
+
+== Non-Requirements
+
+* MicroProfile Health 1.0 only provides a "liveness" probe (as defined by https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-probes/#define-a-liveness-command[Kubernetes]).
+__A Readiness probe is not covered by this proposal__.
+
+== Implementation Plan
+
+* Add dependencies to `io.smallrye:smallrye-health` and `org.eclipse.microprofile.health:microprofile-health-api:1.0` artifacts
+* Add a new `microprofile-health-smallrye` extension to provide Health support for deployments.
+* Provide a Undertow handler that exposes the optional health HTTP endpoint `/health` on WildFly `management-http` interface.
+
+=== Subsystem description
+
+The subsystem will be named `microprofile-health-smallrye`.
+
+The subsystem contains child resources.
+
+The subsystem has 1 attribute:
+
+* `security-enabled` (`true` by default) - if security is enabled, the HTTP client must be authenticated to query the HTTP health endpoint.
+
+The subsystem defines 1 runtime operation:
+
+* `check` whose result is the DMR representation of the Health Check JSON payload.
+  The management operation outcome will always be `success`. The _outcome of the health check_ is provided
+  by the `outcome` property on the operation's `result`:
+
+----
+  /subsystem=microprofile-health-smallrye:check
+  {
+      "outcome" => "success",
+      "result" => {
+          "outcome" => "UP",
+          "checks" => [{
+              "name" => "random",
+              "state" => "UP",
+              "data" => {"foo" => "bar"}
+          }]
+      }
+  }
+----
+
+If the global outcome is `UP`, the result's `outcome` will be `UP`, otherwise it will be `DOWN`.
+
+The subsystem will install at runtime a HTTP endpoint `/health` on the `management-http` interface to provide
+the MicroProfile Health Check endpoint.
+
+If security is enabled, the HTTP client must be authenticated (otherwise, the server will reply with a
+  `401 NOT AUTHORIZED` response).
+
+----
+curl -v --digest  -u admin:adminpwd http://localhost:9990/health
+< HTTP/1.1 200 OK
+< Content-Length: 79
+< Content-Type: application/json
+...
+{"outcome":"UP","checks":[{"name":"random","state":"UP","data":{"foo":"bar"}}]}
+----
+
+If security is disabled, the HTTP client does not need to be authenticated:
+
+----
+curl -v http://localhost:9990/health
+< HTTP/1.1 200 OK
+< Content-Length: 79
+< Content-Type: application/json
+...
+{"outcome":"UP","checks":[{"name":"random","state":"UP","data":{"foo":"bar"}}]}
+----
+
+== Test Plan
+
+* smallrye-health component is passing the MicroProfile Health TCK during its release process.
+* WildFly integration test suite will be enhanced with tests that deploys Java archives with configurable
+  health checks probes and asserts that the individual outcome of the probes affects the global
+  outcome returned by the HTTP endpoint.
+
+== Community Documentation
+
+The feature will be documented in WildFly Admin Guide (in a new MicroProfile Health section).
+
+There will no developer guide for this feature as there is nothing specific to WildFly for users
+wanting to provide Health Check probes with their applications.


microprofile/WFLY-10711_health_check_extension.adoc
```diff
@@ -0,0 +1,82 @@
...
+The implementation of Eclipse MicroProfile Health is provided by https://github.com/smallrye/smallrye-health[smallrye-health].
+
+=== Other Interested Projects
+
+* A https://github.com/jmesnil/wildfly-microprofile-health[prototype of a WildFly extension] for MicroProfile Health was developed during the development of MicroProfile Health 1.0.
+
+== Requirements
+
+* Support Eclipse MicroProfile Health 1.0 for applications deployed in WildFly.
+* Provide a HTTP endpoint for the https://github.com/eclipse/microprofile-health/blob/master/spec/src/main/asciidoc/protocol-wireformat.adoc[health wireformat]
```
bstansberry: What are the requirements re: the configurability of that endpoint?

We've discussed exposing this over the management HTTP interface, so if that's all we are doing then there's probably no configurability at all, since it can just be integrated via a requirement for the static capability exposed by that resource.  But the proposal be clear that the exposure would only be over HTTP management and that having that interface in the config is required.

microprofile/WFLY-10711_health_check_extension.adoc
```diff
@@ -0,0 +1,82 @@
...
+=== Issue
+
+* https://issues.jboss.org/browse/WFLY-10711[WFLY-10711]
+
+=== Related Issues
+
+* https://issues.jboss.org/browse/EAP7-1029[EAP7-1029]
+* https://issues.jboss.org/browse/CLOUD-2730[CLOUD-2730] - Implement EAP Health Check so that it can be consumed by OpenShift Console
+* https://issues.jboss.org/browse/WFLY-10522[WFLY-10522] - MicroProfile Config extension for WildFly.
+** There is no direct relation between MicroProfile Health and MicroProfile Config but there should be consistency in their extensions (subsystem names, etc.)
```
rsvoboda: OpenLiberty example for MP Health uses MP Config stuff too, so I agree there is no direct relation, but in many cases MP Health and MP Config are used together.

microprofile/WFLY-10711_health_check_extension.adoc
```diff
@@ -0,0 +1,82 @@
...
+
+=== Dev Contacts
+
+* mailto:{email}[{author}]
+
+=== QE Contacts
+
+* mailto:tremes@redhat.com[Tomas Remes]
```
rsvoboda: Add me for "classical server" part, Tomas is focused on OpenShift part.

microprofile/WFLY-10711_health_check_extension.adoc
```diff
@@ -0,0 +1,82 @@
...
+
+== Requirements
+
+* Support Eclipse MicroProfile Health 1.0 for applications deployed in WildFly.
+* Provide a HTTP endpoint for the https://github.com/eclipse/microprofile-health/blob/master/spec/src/main/asciidoc/protocol-wireformat.adoc[health wireformat]
+* The HTTP endpoint is global to the server: it must aggregate health check probes from *all deployments*.
+
+== Nice-to-Have Requirements
+
+* Support drilldown of the HTTP endpoint (e.g. to be able to query the health of individual deployment, e.g. `/health/deploymentA`)
```
rsvoboda: Is this in scope of planned work for WildFly 14 ?

microprofile/WFLY-10711_health_check_extension.adoc
```diff
@@ -0,0 +1,82 @@
...
+
+== Non-Requirements
+
+* MicroProfile Health 1.0 only provides a "liveness" probe (as defined by https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-probes/#define-a-liveness-command[Kubernetes]).
+__A Readiness probe is not covered by this proposal__.
+
+== Implementation Plan
+
+* Add dependencies to `io.smallrye:smallrye-health` and `org.eclipse.microprofile.health:microprofile-health-api:1.0` artifacts
+* Add a new `microprofile-health-smallrye` extension to provide Health support for deployments.
```
rsvoboda: Will it be just extension or also management subsystem stuff ?

If there is subsystem introduces, what will be exposed there for configuration ?

microprofile/WFLY-10711_health_check_extension.adoc
```diff
@@ -0,0 +1,82 @@
...
+* Provide a Undertow handler that exposes the health HTTP endpoint `/health` on WildFly `management-https` socket binding (port `9090` of the `management` network interface).
```
rsvoboda: I would start with management-http, some tools can have troubles with https or with self signed certs for https.

Port `9090` is not there. Did you mean `9990` ?
`9993` is for "management-https.

microprofile/WFLY-10711_health_check_extension.adoc
```diff
@@ -0,0 +1,82 @@
...
+  ** Security of the HTTP endpoint must be addressed. Should it be also available under the non-secure `management-http` or should it redirect to the `management-https`?
```
rsvoboda: As I wrote above, I think we should start with management-http and maybe add option to use just management-https (when user configures it).

In our configs we have management-http enabled by default, not management-https.

rsvoboda:
1) Will be MP Health enabled by default in .zip/RPM distribution?
Will be MP Health enabled by default in OpenShift distribution?

2) @jmesnil how will microprofile-health-smallrye behave during boot and server shutdown ?
Asking because I noticed discussion on Payara server - https://github.com/payara/Payara/issues/2595

3) Another case is server boot "done", 5-10 applications are getting deployed, each of them implements at least one HealthCheck. Deploy of one of the application takes long time, `/health` is invoked before the last app is deployed. What should be returned? 
    * DOWN as the last HealthCheck wasn't deploy?
    * UP as all the other checks are fine?
      * What if the last deployment check fails, we would have inconsistency - first check would be reported as UP but in reality we should be saying DOWN  

4) Can be certain HealthCheck disabled server wide ? E.g. pulled in dependencies but we don't want that check to be counted in. 

5) Disabled per deployment for the above case?

microprofile/WFLY-10711_health_check_extension.adoc
jmesnil: that's right. I was talking about relation between Health and Config API. I'll clarify

microprofile/WFLY-10711_health_check_extension.adoc
jmesnil: No, I don't plan to work on it for WildFly 14. I put it in "nice-to-have" to keep a trace of further enhancements

microprofile/WFLY-10711_health_check_extension.adoc
jmesnil: a susbystem will be introduced. I'll add more details to the proposal.
Basically, I see the need for 2 configuration options and 1 operation:

* configuration for HTTP management (or nothing if we mandate its use)
* configuration for the HTTP endpoint (`/health` by default)
* runtime operation to check the health (`/subsystem=microprofile-health-smallrye:check`) that returns the DMR version of health check

microprofile/WFLY-10711_health_check_extension.adoc
```diff
@@ -0,0 +1,82 @@
...
+== Test Plan
+
+* smallrye-health component is passing the MicroProfile Health TCK during its release process.
+* WildFly integration test suite will be enhanced with tests that deploys Java archives with configurable
+  health checks probes and asserts that the individual outcome of the probes affects the global
+  outcome returned by the HTTP endpoint.
+
+== Community Documentation
+
+The feature will be documented in WildFly Admin Guide (in a new MicroProfile Health section).
```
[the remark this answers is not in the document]
jmesnil: I don't think so. There will be nothing specific to WildFly for users writing application with Health check probes.

microprofile/WFLY-10711_health_check_extension.adoc
jmesnil: done

microprofile/WFLY-10711_health_check_extension.adoc
jmesnil: @bstansberry I've clarified in the proposal.

Note that I've changed my mind:
* the subsystem will expose a `check` operation to query the healthiness of the application server (this way, we can locally query the healthiness without exposing anything more on HTTP(S)).
* HTTP endpoint is optional. If it is enabled, it requires the capability for `org.wildfly.management.http-interface`

microprofile/WFLY-10711_health_check_extension.adoc
jmesnil: fixed

jmesnil:
@rsvoboda 

1. I'd suggest to have it enabled on all
2. This proposal deals only with liveness of the application server, not readiness. So it will not behave differently during startup / shutdown sequence
3. that's a readiness issue, not a liveness. If we have 4 deployments that are fine, the health check will return UP. If we later deploy a 5th deployment that has issues, the health check will not be DOWN.
4. I don't plan to disable health checks per deployment. What's the benefit? If the user has provided them, I trust him that he wants them to participate in the overall healthiness of the application server. Do you have an use case?

rsvoboda:
> I don't plan to disable health checks per deployment. What's the benefit? If the user has provided them, I trust him that he wants them to participate in the overall healthiness of the application server. Do you have an use case?

a) I can have dependency containing class which implements HealthCheck. I don't want to repack that dependency to remove it or i don't have control over the dependency.

b) I can have HealthCheck which makes sense for testing environment but not yet for production. Something like TP / unsupported stuff we have in app server.

microprofile/WFLY-10711_health_check_extension.adoc
```diff
@@ -0,0 +1,99 @@
...
+
+The subsystem will be named `microprofile-health-smallrye`.
+It will define 1 runtime operation:
+
+* `check` whose result is the DRM representation of the Health Check JSON payload. If the global outcome is `UP`, the result's `outcome` will be `success` and the `result` will be the
+DMR representation of the HTTP JSON payload. Otherwise, the operation will fail.
+
+The subsystem will contains 2 attributes:
+* `http` - boolean to enable the HTTP endpoint on the HTTP management interface (false by default)
+* `http-endpoint` - String that represents the HTTP endpoint that exposes the health check (`/health` by default)
```
bstansberry: I'm concerned about the complexity these combinations introduce. We need clients to be able to access this endpoint, and it needs to be easy to tell the client how to do so. Plus we should be cautious about assuming we will have control over the long term over this kind of client. Every config permutation we introduce has the potential to bite us.

Now a client needs to deal with these scenarios:

1) Health check subsystem isn't present in the system.
2) Health check subsystem is present in the system, but http=false so not exposed over http management.
3) Health check subsystem is present and exposed over http management, but they chose a different context.

We can write *our* client to work around the 2 and 3 cases by just using the DMR op, but then our client is bound to DMR (non-spec). But no general purpose client is likely to ever use the DMR op. Maybe this is ok, but it makes me uneasy versus just always exposing  this over http management on the 'health' context.

(Note that is http management is not enabled, other liveness and readiness checks that already exist will fail, as they rely on access to it. So we can assume http management exists. OTOH this means our overall liveness checks already are tied to DMR anyway, making my concern about using DMR for this even more theoretical.)

Anyone using the DMR op will need to distinguish the "health check subsystem isn't present in the system" case from other failures. I'll comment more about the DMR response separately.

The 'http-endpoint' attribute I don't see any need for right now, so long as we check with the JDG folks in case they are using that context for their uses of ExtensibleHttpManagement. We don't support use of that hook by anyone else, so we can take ownership of /health and remove any current reason to make it configurable when the management interface is used.

Is the 'http' attribute the way we would go if we ever wanted this exposed on 8080?

bstansberry:
@rsvoboda Had this question earlier related to a section of the doc that's no longer present. But it's worth addressing:

"As I wrote above, I think we should start with management-http and maybe add option to use just management-https (when user configures it).

In our configs we have management-http enabled by default, not management-https."

I don't think it's particularly easy to make this kind of distinction. We have HTTP contexts associated with the management interface and they are available via whatever protocols are configured for that interface. I don't see a strong use case for rearchitecting that to allow mix-and-match. You either want management secured or you don't.  If you want normal management secured but this not, that seems like a future enhancement to let this run over a subsystem=undertow handler.

For the key use case for this RFE, doing checks on OpenShift, adding configuration options makes solving the use case harder, since it means the code doing the liveness probe potentially has to understand and adapt to a variety of options.

Our current probes on OS connect to the management interface over HTTP. They run via scripts executing in the same container as the server. The scripts use env vars to determine how to authenticate.

microprofile/WFLY-10711_health_check_extension.adoc
```diff
@@ -0,0 +1,99 @@
...
+* Add a new `microprofile-health-smallrye` extension to provide Health support for deployments.
+* Provide a Undertow handler that exposes the optional health HTTP endpoint `/health` on WildFly `management-http` socket binding (port `9990` of the `management` network interface).
+
+=== Subsystem description
+
+The subsystem will be named `microprofile-health-smallrye`.
+It will define 1 runtime operation:
+
+* `check` whose result is the DRM representation of the Health Check JSON payload. If the global outcome is `UP`, the result's `outcome` will be `success` and the `result` will be the
+DMR representation of the HTTP JSON payload. Otherwise, the operation will fail.
```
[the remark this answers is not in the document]
rsvoboda: Do we need to have such operation ? 

`curl http://SERVER:9990/health` or `firefox http://SERVER:9990/health` seems sufficient

microprofile/WFLY-10711_health_check_extension.adoc
rsvoboda: `/health` is defined by MP spec, maybe we can relax configuration of the endpoint address.
Didn't check other app servers if they enable configuration of the endpoint address.

microprofile/WFLY-10711_health_check_extension.adoc
```diff
@@ -0,0 +1,99 @@
...
+
+The subsystem will contains 2 attributes:
+* `http` - boolean to enable the HTTP endpoint on the HTTP management interface (false by default)
```
rsvoboda: I'd rather have it enabled by default, so `/health` is available on port 9990

microprofile/WFLY-10711_health_check_extension.adoc
rsvoboda: > Is the 'http' attribute the way we would go if we ever wanted this exposed on 8080?

To have `/health` exposed both on 8080 and 9990 is not bad idea.
Users from other MP impls can be expecting to have `/health` on the same port as app server.

For example Thorntail exposes `http://localhost:8080/health`
https://github.com/thorntail/thorntail-examples/tree/master/jaxrs/health

microprofile/WFLY-10711_health_check_extension.adoc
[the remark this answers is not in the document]
bstansberry: An operation like this might be useful in the openshift integration, yes.  All of our current liveness probes are based on management requests to the http interface, with the different requests done in a composite/batch. Adding another operation to the batch is an easy way to incorporate this with the overall liveness probe.

rsvoboda:
@tremes please check the proposal and start thinking about OpenShift scenarios for the test plan ;)

jmesnil:
> a) I can have dependency containing class which implements HealthCheck. I don't want to repack that dependency to remove it or i don't have control over the dependency.

@rsvoboda Can you elaborate? If you have a dependency to provides a health check, wouldn't you want to check this probe to determine the overall healthiness of the app server?

> b) I can have HealthCheck which makes sense for testing environment but not yet for production. Something like TP / unsupported stuff we have in app server.

It sounds sensible but I don't think this should be handled by letting deployment "disable" health check.
Instead the health check should have some flags to determine whether the check should be performed or bypassed e.g. using MicroProfile Config to determine the environment (test, staging, prod) etc.

That seems sensible but this

microprofile/WFLY-10711_health_check_extension.adoc
jmesnil: @bstansberry I see your points and we can make it simpler to configure.

* Let's mandate the presence of the http `management-interface` provided by [https://github.com/wildfly/wildfly-capabilities/blob/master/org/wildfly/management/http-interface/capability.adoc](org.wildfly.management.http-interface)  (so no more http enabled attributes or thing like that).
* Let's make the context always be `/health` (note that future version of the specification may add more context such as `/ready` but that's not decided at the moment). This context will be added to the http management interface by using the `org.wildfly.management.http.extensible` capability introduced by https://issues.jboss.org/browse/WFCORE-1742

I'm not sure if there is anything to do with regards to configuring http or https from the subsystem point of view.
It should only require the `org.wildfly.management.http-interface` capability. Whether the http management interface is exposed on HTTP or HTTPS is a concern of the [https://wildscribe.github.io/WildFly/13.0/core-service/management/management-interface/http-interface/index.html](core-service=management
management-interface=http-interface).

@bstansberry would that address your concern?

microprofile/WFLY-10711_health_check_extension.adoc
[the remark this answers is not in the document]
jmesnil: @bstansberry Having the management operation fails when the health check outcome is DOWN is consistent with the HTTP wireformat (200 for `UP` and 500/503 for `DOWN`) and [Kubernetes Liveness command](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-probes/#before-you-begin):

> If the command succeeds, it returns 0, and the kubelet considers the Container to be alive and healthy. If the command returns a non-zero value, the kubelet kills the Container and restarts it.

Making the `check-health` operation fails ensure that we do not report false positive to Kubernetes (as it is only interested by the return result of the command).

Data returned by the operation are informational and will help users analyse ongoing liveness issues but from operational point of view, they do not matter for OpenShift.
The check may fail because the subsystem is not installed but I'd argue this should also be part of the healthiness of the application server.

microprofile/WFLY-10711_health_check_extension.adoc
bstansberry: @jmesnil Yes, this addresses my concern.  This basically leaves us with a DMR op, plus knowing that the /health context will be present on the http management interface if that interface exists. That meets the spirit of the spec (a /health context) and gives two solid options for using this in a liveness probe on our OpenShift image, while still leaving us with flexibility for the future if we get real end user requirements for other options (turn off http, expose it on 8080 etc.)

+1 to this work not concerning itself with http vs https on the management interface. All contexts on that interface are available over whatever protocol the overall interface config supports.

microprofile/WFLY-10711_health_check_extension.adoc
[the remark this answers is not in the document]
bstansberry: I'm not convinced. :)

For sure I disagree that the presence of this subsystem is required for the server to be considered healthy.  Maybe some day if we start integrating other subsystem checks into the response we could say that, but right now it's only app checks and if the dev hasn't written any, then this subsystem is adding nothing.  We already have liveness and readiness checks; this is just adding another and shouldn't break what's already there. 

We'd ship it by default in our OS image configs, so if it's not present in an image version that supports it that means the user chose not to include it. Maybe that was a mistake on their part, as their app includes checks they've disabled, but I don't think we should fail liveness on the server because maybe they made the wrong choice. Maybe they knew what they were doing.

As a practical matter, having to use forked liveness probe code for images that support this feature vs those that do not is a pretty big hassle. It's much better to have the same probe code be able to discriminate the "not applicable" case from the DOWN case.

The outcome=success vs failure part I'm not convinced but I feel much less strongly about it. If we write our probes to use this DMR op I think we'd send a composite, a simple read-resource step and a check step. A failure in the r-r step would indicate "not applicable", so then the outcome of the check step is only meaningful if the r-r succeeds.

I don't think the HTTP wireformat requirement nor the liveness command return code control this. Particularly not the latter as there has to be non-HTTP-oriented logic between a DMR response and the return code of any command process that made the DMR request.

Still, I don't feel that strongly about it.

BTW https://issues.jboss.org/browse/WFCORE-185 would result in a 404 coming from the "REST" (i.e. HTTP, non-Upgrade) management interface for a read of a non-existent resource. This isn't directly relevant to this discussion though, it's just semi-related.

Related: if the /health context does not exist on the HTTP interface then I think the response should be a 404.  I believe that is what would happen and I think it's right.

tremes:
Maybe stupid questions and maybe I'm completely out but what does it exactly mean consumed by OS console? Or what is the exact main use case for OS user? Does it mean it will be new UI element somewhere in the UI of OS console? If so then how someone can get this info using "oc" (or this is not required?)?

Wrt the port to expose - so if it is only 9990 then OS user will need to define extra service and probably extra route to consume this over http? Am I right?

jmesnil:
@bstansberry @rsvoboda I have updated the proposals based on your feedback.

I've included examples of the management operation and HTTP endpoint that will be provided by the subsystem.

microprofile/WFLY-10711_health_check_extension.adoc
```diff
@@ -0,0 +1,132 @@
...
+
+The subsystem will be named `microprofile-health-smallrye`.
+
+The subsystem contains no attributes or child resources.
+
+The subsystem defines 1 runtime operation:
+
+* `check` whose result is the DMR representation of the Health Check JSON payload.
+  The management operation outcome will always be `success`. The _outcome of the health check_ is provided
+  by the `outcome` property on the operation's `result`:
```
bstansberry: Shouldn't this be the 'status' property on the operation's 'result'? That aligns with the JSON protocol wire format at https://github.com/eclipse/microprofile-health/blob/master/spec/src/main/asciidoc/protocol-wireformat.adoc#appendix-b-json-payload-specification and also helps avoid confusion between the top level 'outcome' and this field.

microprofile/WFLY-10711_health_check_extension.adoc
```diff
@@ -0,0 +1,132 @@
...
+
+----
+  /subsystem=microprofile-health-smallrye:check
+  {
+      "outcome" => "success",
+      "result" => {
+          "outcome" => "UP",
```
bstansberry: See previous comment re 'status' vs 'outcome'.

microprofile/WFLY-10711_health_check_extension.adoc
```diff
@@ -0,0 +1,132 @@
...
+          "checks" => [{
+              "name" => "random",
+              "state" => "UP",
```
bstansberry: In https://github.com/eclipse/microprofile-health/blob/master/spec/src/main/asciidoc/protocol-wireformat.adoc#appendix-b-json-payload-specification this is 'status' not 'state'.

microprofile/WFLY-10711_health_check_extension.adoc
```diff
@@ -0,0 +1,132 @@
...
+The subsystem will install at runtime a HTTP endpoint `/health` on the `management-http` interface to provide
+the MicroProfile Health Check endpoint.
+
+----
+curl -v -L --digest  -u admin:adminpwd http://localhost:9990/health/
+< HTTP/1.1 200 OK
+< Content-Length: 79
+< Content-Type: application/json
+...
+{"outcome":"UP","checks":[{"name":"random","state":"UP","data":{"foo":"bar"}}]}
```
bstansberry: This differs from https://github.com/eclipse/microprofile-health/blob/master/spec/src/main/asciidoc/protocol-wireformat.adoc#appendix-b-json-payload-specification

tremes:
What will be the outcome of the HTTP endpoint for an app (let's assume there is only one deployment), which doesn't implement any HealthCheck? I think it can't be DOWN nor undetermined.....next question is how this will fit into the logical AND of the health checks then?

rsvoboda:
Spec says `A producer without health check procedures installed MUST returns positive overall outcome (i.e. HTTP 200)`

Example response payloads - Without procedures installed into the runtime
Status 200 and the following payload:
```
{
  "outcome": "UP",
  "checks": []
}
```

tremes:
@rsvoboda Ahh OK makes sense. Thanks

bstansberry:
@tremes I will build up a test image today that has a mock impl of this subsystem along with my expected integration with the existing probes. (Mock subsystem doesn't do real health checks, just reports in different ways based on the value of a system property.  The point is to show how the probes will behave given the various possible responses.)

bstansberry:
I see this in the MP Health spec[1]

```
Security
Aspects regarding the secure access of health check information.

A producer MAY support security on all health check invocations (i.e. authentication)

A producer MUST not enforce security by default, it SHOULD be an opt-in feature (i.e. configuration change)
```

So to be compliant (a long run goal even if not required in first iteration) we need to allow authentication to be turned off.

I'm going to regard "by default" as not meaning the same thing as "in the standard configuration". IOW I think we're completely free to ship a standard config with xml that results in this context requiring authentication.

And I think we need to support that and it should perhaps be our standard behavior to turn it on.

But that means we need a config option to turn it on/off. I think that's part of the subsystem config, not http-management as it's specific to one context.

If authentication was required it would use the same mechanism as the /management context; i.e. it's not separately configurable.

[1] https://github.com/eclipse/microprofile-health/blob/master/spec/src/main/asciidoc/protocol-wireformat.adoc

rsvoboda:
> I'm going to regard "by default" as not meaning the same thing as "in the standard configuration". IOW I think we're completely free to ship a standard config with xml that results in this context requiring authentication

I slightly disagree, mp-health spec mandates that health point must be accessible w/o authentication

bstansberry:
TBH I'm not too concerned about the default. I think that requirement is silly, since if the spec allows both and the impl supports both and users presumably want both what business is it of the spec to declare which category of user has to change a setting. The fact both are allowed means clients need to cater to both. But whatever. :)

tremes:
@bstansberry OK thanks. Let me know once there is available some image and I'll try to experiment...

microprofile/WFLY-10711_health_check_extension.adoc
jmesnil: Health 1.0 uses `outcome` https://github.com/eclipse/microprofile-health/blob/1.0/spec/src/main/asciidoc/protocol-wireformat.adoc#json-schema

The link is from the in-progress specification that has not been released (or implemented)

microprofile/WFLY-10711_health_check_extension.adoc
jmesnil: idem https://github.com/eclipse/microprofile-health/blob/1.0/spec/src/main/asciidoc/protocol-wireformat.adoc#json-schema

jmesnil:
I have updated the proposal about security.

The subsystem will have a `security-enabled` attribute to control whether authentication is required to access the HTTP endpoint.
I proposed to have it as `true` by default (most secure) and explicitly set it to `false` in standalone configuration (for ease of use). The `management` interface is bound to `127.0.0.1` in our standalone configuration and it seems appropriate to allow unauthenticated access to the health probe on that local address.

microprofile/WFLY-10711_health_check_extension.adoc
jmesnil: idem

microprofile/WFLY-10711_health_check_extension.adoc
jmesnil: idem https://github.com/eclipse/microprofile-health/blob/1.0/spec/src/main/asciidoc/protocol-wireformat.adoc#json-schema

microprofile/WFLY-10711_health_check_extension.adoc
```diff
@@ -0,0 +1,137 @@
...
+  }
+----
+
+If the global outcome is `UP`, the result's `outcome` will be `UP`, otherwise it will be `DOWN`.
+
+The subsystem will install at runtime a HTTP endpoint `/health` on the `management-http` interface to provide
+the MicroProfile Health Check endpoint.
+
+----
+curl -v -L --digest  -u admin:adminpwd http://localhost:9990/health/
```
rsvoboda: Stress out that this example is for `security-enabled=true` case

`curl -v http://localhost:9990/health/` example should be also mentioned

rsvoboda: hmm, `-L` is no longer needed I think

jmesnil: I updated the cURL snippet.

microprofile/WFLY-10711_health_check_extension.adoc
bstansberry: Thanks!

PR #106, opened 2018-07-20
[WFLY-10709] Observability: Support MicroProfile OpenTracing

jpkrohling:
In order to improve Observability of applications deployed in WildFly it would be beneficial to provide
transparent tracing of Java EE components via automatic initialization of a tracer (Jaeger) plus a set
of instrumentation libraries for components commonly used in Java EE applications, such as Servlets, JAX-RS, CDI, EJB, ...

This will be implemented by supporting Eclipse MicroProfile OpenTracing 1.0 for applications deployed in WildFly. The integration uses the smallrye-opentracing component to provide the MicroProfile OpenTracing implementation.

JIRA: https://issues.jboss.org/browse/WFLY-10709

Signed-off-by: Juraci Paixão Kröhling <juraci@kroehling.de>

### DIFF
--- a/microprofile/WFLY-10709-MicroProfile-Tracing-Extension.adoc
+++ b/microprofile/WFLY-10709-MicroProfile-Tracing-Extension.adoc
@@ -1,0 +1,93 @@
+= [WFLY-10709] Support Eclipse MicroProfile OpenTracing
+:author:            Juraci Paixão Kröhling
+:email:             jpkroehling@redhat.com
+:toc:               left
+:icons:             font
+:idprefix:
+:idseparator:       -
+:keywords:          observability,microprofile,tracing,openshift
+
+== Overview
+
+In order to improve Observability of applications deployed in WildFly it would be beneficial to provide
+transparent tracing of Java EE components via automatic initialization of a tracer (Jaeger) plus a set
+of instrumentation libraries for components commonly used in Java EE applications, such as JAX-RS and CDI.
+
+This will be implemented by supporting Eclipse MicroProfile OpenTracing 1.1 for applications deployed in WildFly.
+The integration uses the `smallrye-opentracing` component to provide the MicroProfile OpenTracing implementation.
+
+For WildFly 14, this shall be marked as "Tech Preview", to get community feedback before committing to a stable API.
+Notably, this first version lacks any configuration option following WildFly's standard management model, as this is
+something that will likely change before a stable version. 
+
+== Issue Metadata
+
+=== Issue
+
+* https://issues.jboss.org/browse/WFLY-10709[WFLY-10709]
+
+=== Related Issues
+
+* https://issues.jboss.org/browse/EAP7-1027[EAP7-1027]
+* https://issues.jboss.org/browse/CLOUD-2728[CLOUD-2728] - Expose EAP "Tracing" information so that it can be consumed by OpenShift Console
+* https://issues.jboss.org/browse/TRACING-85[TRACING-85] - Create jboss-modules for OpenTracing and Jaeger
+* https://issues.jboss.org/browse/TRACING-87[TRACING-87] - OpenTracing integration for JAX-RS in Wildfly
+
+=== Dev Contacts
+
+* mailto:{email}[{author}]
+
+=== QE Contacts
+
+* mailto:tremes@redhat.com[Tomas Remes]
+* mailto:jstourac@redhat.com[Jan Stourac]
+
+=== Affected Projects or Components
+
+The implementation of Eclipse MicroProfile OpenTracing is provided by https://github.com/smallrye/smallrye-opentracing[smallrye-opentracing].
+
+=== Other Interested Projects
+
+* None
+
+== Requirements
+
+* Support https://github.com/eclipse/microprofile-opentracing/blob/1.1/spec/src/main/asciidoc/microprofile-opentracing.asciidoc[Eclipse MicroProfile OpenTracing 1.1] for applications deployed in WildFly, including JAX-RS and CDI integration. Passing the TCK is required.
+* Support a concrete tracer, such as Jaeger's
+* Allow applications to supply their own tracers via `TracerResolver`
+* Each individual application deployed (WAR) should have its own `Tracer` instance
+* Each WAR within an EAR should be treated as an individual WAR, having their own `Tracer` instance
+
+== Nice-to-Have Requirements
+
+Items that are currently being considered for the next version of this feature:
+
+* Enable/disable tracing on the entire server
+* Enable/disable tracing on a per-deployment basis
+* Configuration following WildFly's standard management model
+
+== Non-Requirements
+
+* None at this point
+
+== Implementation Plan
+
+* Add dependencies to `io.smallrye:smallrye-opentracing` artifacts, including its transitive dependencies (notably OpenTracing)
+* Add dependencies to `io.jaegertracing:jaeger-client` artifacts, including its transitive dependencies
+* Add a new `microprofile-opentracing-smallrye` extension to provide OpenTracing support for deployments.
+* Provide a servlet context initializer to initialize Jaeger Tracer
+* Provide the required filters to get SmallRye's JAX-RS integration working
+* Provide the required CDI extensions and producers to get SmallRye's CDI integration working
+
+== Test Plan
+
+* `smallrye-opentracing` component is passing the MicroProfile OpenTracing TCK during its release process.
+* WildFly integration test suite will be enhanced with tests that ensure applications getting deployed are able 
+to be traced. Among possibly others, tests should cover that:
+** A `Tracer` can be injected into managed beans
+** Tracing can be disabled for specific endpoints/CDI beans by using the `@Traced` annotation
+** Operation names can be customized by using the `@Traced` annotation
+
+== Community Documentation
+
+The feature will be documented in WildFly Admin Guide (in a new MicroProfile OpenTracing section).


jpkrohling:
cc @jmesnil

microprofile/WFLY-10709-MicroProfile-Tracing-Extension.adoc
```diff
@@ -0,0 +1,77 @@
...
+:keywords:          observability,microprofile,tracing,openshift
+
+== Overview
+
+In order to improve Observability of applications deployed in WildFly it would be beneficial to provide
+transparent tracing of Java EE components via automatic initialization of a tracer (Jaeger) plus a set
+of instrumentation libraries for components commonly used in Java EE applications, such as Servlets, JAX-RS,
+CDI, EJB, ...
+
+This will be implemented by supporting Eclipse MicroProfile OpenTracing 1.0 for applications deployed in WildFly.
```
pavolloffay: SmallRye-OpenTracing implements on MP-OT 1.1.

microprofile/WFLY-10709-MicroProfile-Tracing-Extension.adoc
```diff
@@ -0,0 +1,77 @@
...
+
+* Provide an option for operators to specify other OpenTracing-compatible tracer implementations than Jaeger's.
+
+== Non-Requirements
+
+* None at this point
+
+== Implementation Plan
+
+* Add dependencies to `io.smallrye:smallrye-opentracing` artifacts, including its transitive dependencies (notably OpenTracing and Jaeger)
```
pavolloffay: `io.smallrye:smallrye-opentracing`

does not depend on Jaeger

microprofile/WFLY-10709-MicroProfile-Tracing-Extension.adoc
```diff
@@ -0,0 +1,79 @@
...
+
+== Requirements
+
+* Support Eclipse MicroProfile OpenTracing 1.1 for applications deployed in WildFly.
+* Support a concrete tracer, such as Jaeger's
+
+== Nice-to-Have Requirements
+
+* Provide an option for operators to specify other OpenTracing-compatible tracer implementations than Jaeger's.
+
```
bstansberry: How do the TRACING-86 and TRACING-88 issues relate to this?  The spec at [1] looks focused on JAX-RS, so TRACING-87 is covered by the "Support Eclipse MicroProfile OpenTracing 1.1" requirement. Are TRACING-86/88 Nice-to-have (i.e. we'll probably try but it's not a blocker)? Non-Requirements (i.e. explicitly out of scope for this work, but that doesn't mean it can never happen)?

[1] https://github.com/eclipse/microprofile-opentracing/blob/master/spec/src/main/asciidoc/microprofile-opentracing.asciidoc

microprofile/WFLY-10709-MicroProfile-Tracing-Extension.adoc
```diff
@@ -0,0 +1,79 @@
...
+The implementation of Eclipse MicroProfile OpenTracing is provided by https://github.com/smallrye/smallrye-opentracing[smallrye-opentracing].
+
+=== Other Interested Projects
+
+* A https://github.com/jpkrohling/java-wildfly-tracing[prototype of a WildFly extension] for OpenTracing was developed while exploring a solution for TRACING-85/.../TRACING-88.
+
+== Requirements
+
+* Support Eclipse MicroProfile OpenTracing 1.1 for applications deployed in WildFly.
+* Support a concrete tracer, such as Jaeger's
```
bstansberry: I think it's required that the concrete tracer must be configurable in a way consistent with general WildFly configuration standards. Things like proper use of paths, [outbound]-socket-bindings.

Details on what that config would look like would be good. A lot of times a draft those details surfaces requirements.

microprofile/WFLY-10709-MicroProfile-Tracing-Extension.adoc
jpkrohling: Would it be OK to leave them out for now and, once we have this feature working properly, we add these extra libraries? Or is it preferable to list it all now?

microprofile/WFLY-10709-MicroProfile-Tracing-Extension.adoc
jpkrohling: Added some details. I'm probably a bit outdated regarding WF configuration, but would would the outbound socket-binding you mentioned be something relevant for `JAEGER_AGENT_HOST`, `JAEGER_AGENT_PORT` and `JAEGER_COLLECTOR_ENDPOINT` ?

microprofile/WFLY-10709-MicroProfile-Tracing-Extension.adoc
bstansberry: It's fine to leave them out, but you should remove the JIRA links so there's no confusion. 

It's no problem doing > 1 of these over time as new functionality gets added over multiple releases.

Generally "Non-Requirements" is used for things an interested reader might assume were required, so you're clearly saying "No it's not." That's not the case here; I doubt a typical reader would assume integrations not part of the spec.

"Nice-to-have" is ambiguous (will it be in or not?) so it's best avoided except for stuff where there's a real good chance of it getting into this iteration of a feature, but maybe not.

microprofile/WFLY-10709-MicroProfile-Tracing-Extension.adoc
bstansberry: Sounds like it, yes. A remote-outbound-socket-binding is used for declaring the host and port of a remote service to which an in-server client will connect.

This is something we should chat about as I can see how this might be a pain in a cloud use case. On OpenShift we provide a standard standalone-openshift.xml and then based on env var the user provides (e.g. when create their app from an application template) scripts modify the config.  The more complex the script modifications, the harder it is.  We'll have to work out how the scripts will work regardless.

jpkrohling: Currently, the Jaeger Java Client can be almost completely configured via env vars and system properties. The cases where it cannot (like, when there's a custom `RemoteReporter` or `Sender`) would also (probably) not be covered by configuring it via WildFly's standard mechanism.

Additionally, this configuration is Jaeger-specific and I remember that it was said during the kick-off meeting that it would be a good idea to think twice about the configuration for this first version, as whatever we decide now is harder to change later, especially considering that we might support other tracers in the future (with a different set of configuration options).

That all said, could we then just document that the tracer can be configured by setting env vars/system properties, either via:

* real env vars (for Docker, Kubernetes, ...)
* `-u JAEGER_SERVICE_NAME=service-name`
* `-P file:///path/to/configuration.properties`
* `<server><system-properties><property name="JAEGER_SERVICE_NAME" value="service-name"/></system-properties>...</server>`

We could probably also have a `<properties/>` as a child of the subsystem, like `management/security-realms/security-realm/authentication/properties`, but I don't see how it would bring any benefits over the methods above.

microprofile/WFLY-10709-MicroProfile-Tracing-Extension.adoc
jpkrohling: Got it. I removed those, as well as the nice-to-have requirement.

microprofile/WFLY-10709-MicroProfile-Tracing-Extension.adoc
```diff
@@ -0,0 +1,80 @@
...
+* Add dependencies to `io.jaegertracing:jaeger-tracerresolver` and `jaeger-client` artifacts, including its transitive dependencies
+* Add a new `microprofile-opentracing-smallrye` extension to provide OpenTracing support for deployments.
+* Provide a servlet context initializer to initialize Jaeger Tracer
+* Provide the required filters to get SmallRye's JAX-RS integration working
+* Provide the required CDI extensions and producers to get SmallRye's CDI integration working
+
+== Test Plan
+
+* `smallrye-opentracing` component is passing the MicroProfile OpenTracing TCK during its release process.
+* WildFly integration test suite will be enhanced with tests that deploys Java archives with servlets and other components that should be traced during a reques.
```
jstourac: typo - `reques -> request`.

microprofile/WFLY-10709-MicroProfile-Tracing-Extension.adoc
bstansberry: @jpkrohling Sorry, I didn't see this response before.

Configuring WildFly via env vars / system properties is quite contrary to how things are supposed to be done.

Re: thinking hard about the configuration because it's hard to change later, that applies to configuring via env vars and properties as well. Once this is in, the specific env vars that are used now have to work forever. Of course we could say things in standalone.xml take precedence over the env vars. But the default behavior with no config beyond having the subsystem would have to remain as we do it now.

In the future I don't much like a <properties/> child of the subsystem. As a way of configuring a plugin provider, sure. But our standard provider should be configurable via fully specified attributes.

@jmesnil FYI.

My inclination is this should be a Tech Preview feature until the config model is clarified. But I think providing it configured via env vars is ok. The TP basically gives us freedom to change the config model in the future if we decide we want to

jpkrohling: > Once this is in, the specific env vars that are used now have to work forever

This is already true for Jaeger and how it's been documented to work on other stacks.

> The TP basically gives us freedom to change the config model in the future if we decide we want to

+1

jstourac: @bstansberry Even though we mark this as a TP, we need to consider adding proper means of configuration for WildFly. While for CD, this is no problem, for EAP zip stream, this has to be done properly. It is not realistic to include new subsystem out of blue in CP patches in proper stream.

Also, even though we mark this as a TP, this feature will be adopted by all layered products based on EAP (this will be their default go-to solution for Kiali), thus any changes in its configuration model will have to be done carefully and with backward compatibility in mind. So we should try to make this as much final as possible.

Configuration of tracing agent via WildFly management is therefore necessary. We also need to consider that on OpenShift, the management will not be too useful, so default configuration of tracing agent should utilise expression, like how it is currently done e.g. for:
```
/socket-binding-group=standard-sockets/socket-binding=http[port]

ls /socket-binding-group=standard-sockets/socket-binding=http:read-resource-description
bound=true                    bound-port=8080               fixed-port=false              multicast-address=undefined   name=http                     
bound-address=127.0.0.1       client-mappings=undefined     interface=undefined           multicast-port=undefined      port=${jboss.http.port:8080}
```
Also, looking at the list of [JAEGER properties](https://github.com/jaegertracing/jaeger-client-java/tree/master/jaeger-core#configuration-via-environment), when we implement those to be configurable via management model, we need to use existing infrastructure in WildFly - namely use socket bindings for information about connections, Elytron to hold security, etc.

jpkrohling: > we need to consider adding proper means of configuration for WildFly

Other than tracer-specific configuration, I'm not sure what else there is to configure. Perhaps a switch to turn tracing on/off? This would probably require a re-deploy of the affected applications, as filters/interceptors would need to be removed.

> any changes in its configuration model will have to be done carefully and with backward compatibility in mind

That's exactly why I'm a bit reluctant in adding non-mandatory configuration options now. In general, it's easier to keep backwards compatibility when we add options than when we move/remove them. I believe this is true for WildFly as well.

> Also, looking at the list of JAEGER properties, when we implement those to be configurable via management model, we need to use existing infrastructure in WildFly - namely use socket bindings for information about connections, Elytron to hold security, etc.

None of those properties are required, except for the service name, which we can derive from the deployment unit's name.

Not sure how Elytron fits here, would you mind expanding on that? 

About the socket bindings: those are Jaeger-specific settings and we discussed in the past that we *could* envision users wanting to make use of their own tracers (from a third-party vendor or home-made), either as part of their deployments or as a custom JBoss Module (like a JDBC driver), which is why I wanted to avoid having Jaeger-specific information in the configuration model. If we do want to have Jaeger-specific configuration under WildFly's management model, I'm OK with that. I would just need a suggestion on how to best do that in a way that we can change it in the future.

microprofile/WFLY-10709-MicroProfile-Tracing-Extension.adoc
```diff
@@ -0,0 +1,80 @@
...
+
+The implementation of Eclipse MicroProfile OpenTracing is provided by https://github.com/smallrye/smallrye-opentracing[smallrye-opentracing].
+
+=== Other Interested Projects
+
+* A https://github.com/jpkrohling/java-wildfly-tracing[prototype of a WildFly extension] for OpenTracing was developed while exploring a solution for TRACING-85/.../TRACING-88.
+
+== Requirements
+
+* Support Eclipse MicroProfile OpenTracing 1.1 for applications deployed in WildFly.
```
mjurc: @jpkrohling Are we aiming towards the supporting the tracing of JAX-RS services running inside of WildFly now? The scope is kinda unclear for now.

jpkrohling: I'll update the doc to make it clearer, but tracing of JAX-RS endpoints is indeed part of MicroProfile OpenTracing 1.1.

microprofile/WFLY-10709-MicroProfile-Tracing-Extension.adoc
jstourac: Yes, on/off switch is a good thing to have. It is easier to control than adding/removing the whole extension IMHO - at least from the UX point of view.

Our point is to have a possibility to configure tracing agent via management model. I believe there is some common set of attributes that can be used to configure various tracing agents? E.g. user/password (that was the reason why I mentioned elytron integration in previous comment). Such set should be determined and added into the `microprofile-opentracing-smallrye` subsystem. Maybe @bstansberry can provide any opinion?

Regarding Elytron - please see `authentication-confguration` and `authentication-context` server configuration elements. These are used when server is authenticating against remote security provider and it would be useful if tracing agents were able to use the same mechanisms as the rest of endpoints on the server. cc @darranl

jpkrohling: > Yes, on/off switch is a good thing to have. It is easier to control than adding/removing the whole extension IMHO - at least from the UX point of view.

Would you say this is required for the Tech Preview? My understand is that the final PR has to be ready by this Friday (two days from now), and the proposal hasn't been merged yet.

If this is required, I'll start working on it right away, but I'd really, really appreciate to have a list of things that are required to be ready by this Friday.

> E.g. user/password

There's no auth/z in Jaeger. So far, the current approach in protecting Jaeger's UI is to place an OAuth proxy in front of it, which is why we have a token and username/password env vars, but I don't think this setup will be supported in the product.

mjurc:
@tremes Kindly take a look :)

microprofile/WFLY-10709-MicroProfile-Tracing-Extension.adoc
```diff
@@ -0,0 +1,80 @@
...
+
+== Non-Requirements
+
+* None at this point
+
+== Implementation Plan
+
+* Add dependencies to `io.smallrye:smallrye-opentracing` artifacts, including its transitive dependencies (notably OpenTracing)
+* Add dependencies to `io.jaegertracing:jaeger-tracerresolver` and `jaeger-client` artifacts, including its transitive dependencies
+* Add a new `microprofile-opentracing-smallrye` extension to provide OpenTracing support for deployments.
```
jstourac: Is it possible or does it make sense to enable OpenTracing not only globally for whole server but also per deployment?

jpkrohling: Absolutely! How could that work, in your view? Something similar to Keycloak's `secure-deployment` configuration option?

https://www.keycloak.org/docs/3.2/getting_started/topics/secure-jboss-app/subsystem.html

Like:

```xml
<subsystem...>
  <deployment name="acme.war">
    <enabled>false</enabled>
  </deployment>
</subsystem>
```

jstourac:
@jpkrohling, I think we should summarize this proposal a little bit, so we can approve this for merge and finalise our test plan for WF14.

As MP OpenTracing has been approved to be marked as a Technical preview only in EAP7.2.0, we should mention this here too with some brief disclamer that behaviour and configuraiton of the feature may change in the upcoming releases. We also should have the minimal requirements for TP in the hard requirements section:

* MicroProfile OpenTracing compliance
* enabling/disabling tracing globally per server
* container with multiple deployments

Any additional work to finalize this from TP to a supported feature shall be described in a followup  RFE. It would be useful to add the rest of the discussed features (the  ones that will have to be implemented when the feature is fully  supported) to a nice-to-have section so it can be easily searchable in  the future.

Regarding  to the possibility to disable particular deployment from tracing - will this be  already part of this MR? If not let's postpone this to a followup RFE  then.

Also, I can see you have implemented possibility to have a custom Tracer included in a deployment so it is used instead of the build-in one. Do you want to include that in TP? it should be mentioned in hard requirements, then.

Is the list of supported EE components final? Servlets, JAX-RS,￼ CDI, EJB? Do we go with just spec required JAX-RS for WF14 or other components aswell?

Otherwise all other content seems to reflect current reality of the MR.

jpkrohling:
@jstourac I just updated this PR (and squashed) based on your comment and other discussions here. Would you please review?

jstourac:
Thank you, I think it is good now. @mjurc may wanna have a look too.

microprofile/WFLY-10709-MicroProfile-Tracing-Extension.adoc
```diff
@@ -0,0 +1,93 @@
...
+
+* None
+
+== Requirements
+
+* Support https://github.com/eclipse/microprofile-opentracing/blob/1.1/spec/src/main/asciidoc/microprofile-opentracing.asciidoc[Eclipse MicroProfile OpenTracing 1.1] for applications deployed in WildFly, including JAX-RS and CDI integration. Passing the TCK is required.
+* Support a concrete tracer, such as Jaeger's
+* Allow applications to supply their own tracers via `TracerResolver`
+* Each individual application deployed (WAR) should have its own `Tracer` instance
+* Each WAR within an EAR should be treated as an individual WAR, having their own `Tracer` instance
```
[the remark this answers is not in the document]
jpkrohling: Did you mean JARs inside WARs ? If so, I don't think this is desirable, as this would mean that each JAR would be a "service".

cc @objectiser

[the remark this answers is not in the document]
mjurc: I meant JARs in EARs actually.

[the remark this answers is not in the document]
jpkrohling: My Java EE is a bit rusty: can JAX-RS endpoints be deployed from JARs inside EARs ? If not, then the only way I can think of to have those beans exposed to external consumers is via EJBs, right?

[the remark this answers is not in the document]
mjurc: Maybe I am misunderstanding myself - if we're tracing only JAX-RS endpoints, then just wars are fine. I thought that the invocations between other components (CDI beans for example) may be traced in the scope of this aswell - if we're tracing just JAX-RS endpoints, then WARs are sufficient :)

microprofile/WFLY-10709-MicroProfile-Tracing-Extension.adoc
```diff
@@ -0,0 +1,93 @@
...
+
+== Nice-to-Have Requirements
+
+Items that are currently being considered for the next version of this feature:
+
+* Enable/disable tracing on the entire server
+* Enable/disable tracing on a per-deployment basis
+* Configuration following WildFly's standard management model
```
jstourac: @jpkrohling, there are some other things that should be considered for full support (and not just from the WildFly side). Even in case they don't make sense from the point of Jaeger server, we need to discuss them and clarify what will happen in such situation:
* What happens when Jaeger server is not available to push messages to (shall we log some warning?), is it possible to provide some kind of probe for administrator to check that connection between client and server is working and messages can be successfully delivered to Jaeger server?
* Is there any kind of HA from the Jaeger server available (this is not a WildFly concern, but we still need a fully featured integration)? If so are we able to detect fail-over and redirect messages to different server?
* Will there be a possibility for the tracing client to use TCP instead of UDP?

Could you please update this analysis with these bullets (feel free to reformulate) or put some statements regarding to those? Thank you!

jpkrohling: I wasn't sure where this content would be suitable, so, I added to the overview. Could you please check? Let me know if further clarification is needed.

jstourac: Well, my first thought was to add a bullets here in the "Nice to have" section (nice to though about), but generally I have no strong opinion. Anyway, I've just checked and I am not sure whether I found updated content in the overview section?

jpkrohling: I'm sorry, looks like I didn't push the new commit... I just opened a new PR to add this: https://github.com/wildfly/wildfly-proposals/pull/123

jstourac: Thank you, I'll check there.

PR #109, opened 2018-07-24
[EAP7-1063] Support patching EAP with Galleon

aloubyansky:
https://issues.jboss.org/browse/EAP7-1063

### DIFF
--- a/patching/EAP7-1063_EAP_Patching_With_Galleon.adoc
+++ b/patching/EAP7-1063_EAP_Patching_With_Galleon.adoc
@@ -19,7 +19,7 @@ Galleon is a separate project and an external tool, in itself not a part of Wild
 
 Galleon is a repository-based mechanism. That means the final distribution (of WildFly, EAP, LP, etc) is described in a configuration file as a set of artifacts (feature-packs) that contain the info and the content from which the final distribution is going to be built. Those feature-pack artifacts must be available at provisioning time in the artifact repository (or repositories) (Maven in this case, although other repositories could work as well). The repository could be local and/or remote. This is another difference compared to the current patching mechanism where a patch has to be downloaded and provided to the tool by the user directly.
 
-Advatages of the repository-based approach are:
+Advantages of the repository-based approach are:
 
 * user doesn't have to download patches and version updates manually, typing "update"/"check-updates" commands instead to review and apply the available in the repository patches and updates;
 
@@ -32,7 +32,7 @@ When Galleon installs a feature-pack that is provided directly by the user, by d
 
 ==== Patch feature-pack
 
-Patch file itself is basically a feature-pack. The only difference comparing to regular feature-packs is it contains an element in its XML description indicating that it is a patch that applies to a specific version of another feature-pack. A patch will be accepted only if the installation includes the feature-pack the patch applies to. It is possible to install multiple different patches for the same feature-pack in the installation. When the feature-pack that was patched is installed or a different (lower or higher) version of it is installed, all the installed patches for that feature-pack will automatically be removed from the installation.
+Patch file itself is basically a feature-pack. The only difference comparing to regular feature-packs is it contains an element in its XML description indicating that it is a patch that applies to a specific version of another feature-pack. A patch will be accepted only if the installation includes the feature-pack the patch applies to. It is possible to install multiple different patches for the same feature-pack in the installation. When the feature-pack that was patched is uninstalled or a different (lower or higher) version of it is installed, all the installed patches for that feature-pack will automatically be removed from the installation.
 
 A patch declare dependency on another patch(es) that applies to the same feature-pack version as the patch declaring the dependency. When a user installs a patch that depends on other patches, the dependencies of the patch will be installed automatically and applied before the patch declaring the dependencies.
 
--- a/patching/EAP7-1063_EAP_Patching_With_Galleon.adoc
+++ b/patching/EAP7-1063_EAP_Patching_With_Galleon.adoc
@@ -1,0 +1,121 @@
+= EAP7-1063 Support patching EAP with Galleon
+:author:            Alexey Loubyansky
+:email:             alexey.loubyansky@redhat.com
+:toc:               left
+:icons:             font
+:keywords:          comma,separated,tags
+:idprefix:
+:idseparator:       -
+:issue-base-url:    https://issues.jboss.org/browse/
+
+== Overview
+
+Current and our first ever patching mechanism has a few serious usability issues (such as reliance on overlays and not obvious actual JAR file locations, increased distribution size with more and more patches applied and limited history cleaning, not flexible in terms of skipping content that was removed or not originally installed but included in the patch, patching mechanism being a part of the server creates a problem of self-patching, configuration patching is not supported, poor support for patch streams, etc). The issues are fundamental to a degree that fixing them would effectively mean a complete re-write of the mechanism.
+
+The intention of this proposal is to replace the current patching mechanism with the one based on the provisioning mechanism (Galleon) which is currently used to build WildFly and EAP distributions.
+
+Galleon is a separate project and an external tool, in itself not a part of WildFly or EAP. It is a generic provisioning and patching tool that allows project/product speicific patching/provisioning plugins. So the first immediate difference with the current patching mechanism is the user would have to obtain the tool before applying a patch.
+(BTW, Galleon also serves as an installer, meaning it could be used to install WildFly/EAP in the first place and then to apply patches and version updates.)
+
+Galleon is a repository-based mechanism. That means the final distribution (of WildFly, EAP, LP, etc) is described in a configuration file as a set of artifacts (feature-packs) that contain the info and the content from which the final distribution is going to be built. Those feature-pack artifacts must be available at provisioning time in the artifact repository (or repositories) (Maven in this case, although other repositories could work as well). The repository could be local and/or remote. This is another difference compared to the current patching mechanism where a patch has to be downloaded and provided to the tool by the user directly.
+
+Advatages of the repository-based approach are:
+
+* user doesn't have to download patches and version updates manually, typing "update"/"check-updates" commands instead to review and apply the available in the repository patches and updates;
+
+* not necessary to store content that has been patched or replaced with the version update in a hidden directory in the distribution itself (as the current patching mechanism does), leaving the distribution directory looking like a freshly installed one. (there is still .galleon directory containing the installation provisioning configuration for future reference and a few other text files but not actual heavy JAR content or something of that kind).
+
+While a user still has the possibility to download the feature-pack as a ZIP file and install it locally using Galleon (a la current patching) the repository still remains a general requirement. The reason for that is Galleon does not keep installed feature-packs in the distribution itself but when it's time to re-provision the distribution (and that is happening on every provisioning operation including install feature-pack/patch, uninstall feature-pack/patch, provision specific state, apply update) it assumes the neccessary feature-packs are available in the repository.
+
+Here is a use-case which demonstrates the limitation. User installs EAP (that, btw, currently means Galleon will install three feature-packs: core, servlet and full). Then the user downloads a patch (which is also a kind of feature-pack) and installs it directly, not going through the repository. First of all, what has been installed before the patch (core, servlet and full feature-packs) has to be available in the repository at patch application time. If those feature-packs are not available, patch application will fail. (It is still possible to directly provide core, servlet, full and the patch feature-packs to Galleon for provisioning, although that seems like an impractical way to promote the use of the tool).
+When Galleon installs a feature-pack that is provided directly by the user, by default, besides provisioning that feature-pack, it also installs the directly provided feature-pack into the repository, so it is available for future use (in our case it's a local Maven repository). The user may instruct not to install the feature-pack into the repository but it is strongly discouraged to do that.
+
+==== Patch feature-pack
+
+Patch file itself is basically a feature-pack. The only difference comparing to regular feature-packs is it contains an element in its XML description indicating that it is a patch that applies to a specific version of another feature-pack.
+
+Patch feature-pack basically overwrites or potentially re-defines certain content of the feature-pack it applies to. That includes:
+
+* feature-pack packages (units of FS content such as jboss modules and other content);
+
+* feature specs and feature groups (units of configuration);
+
+* resources (various resources included into feature-pack, including module artifact versions);
+
+* provisioning plugin (it is possible to patch a provisioning plugin included in the original feature-pack).
+
+Patch in this proposal is always assumed to be a one-off patch. There is no notion of a cumulative patch. A CP is assumed to be a version update in Galleon mechanism.
+
+==== Atomic provisioning state transition
+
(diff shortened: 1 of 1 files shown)


LittleJohnII:
Thanks for the proposal, I had a quick look and it looks good so far. I need to think a little more about it, then I'll put up a review for this.

Cc @bmaxwell @dosoudil

@aloubyansky do you think there's enough time this can make it in WF14? If not, the sooner we start communicating that, the better.

(rjanik)

LittleJohnII:
I went over the proposal in more detail. I have a question: Brad mentioned the ability to `ageout-history` - to remove old layers after cumulative patch - what about this?

AIUI, a cumulative patch is now considered a version upgrade - therefore after upgrade, the installation will be provisioned anew and there will be no cumulative jars appearing. This might make it a non-issue and `ageout-history` is not necessary. One-off patches are not cumulative by principle and are also not affected by this size growth. So, `ageout-history` is not needed. Is this correct?

aloubyansky:
Right, https://github.com/wildfly/wildfly-proposals/pull/109/files#diff-d5d88ec6113566e386dcffebb4a19515R26
And there is still a way (cli command and an api) to clean the history or disable it completely.

patching/EAP7-1063_EAP_Patching_With_Galleon.adoc
```diff
@@ -0,0 +1,121 @@
...
+
+Next, Galleon provisions an updated installation at a staged location and if that went ok, it applies the user changes identified previously to the newly provisioned state.
+
+User changes to the files that haven't changed in the updated installation are re-applied to the updated installation.
+
+If a user changed a file in the installation that has changed in the updated installation then if the changes can be merged (text files, property files) they are merged. If the changes can't be merged, the file with the user changes will appear next to the updated file with a suffix ".prev".
+
+Standalone configuration changes are identified as diffs in terms of feature configurations and then re-applied to the corresponding updated configurations.
+
+Host and domain configuration changes are not currently merged as the standalone ones and *not* merged as text files. If user modified host or domain configurations and the corresponding configurations have changed in the updated installation, the user-modified configuration files will appear in the installation with suffix ".prev".
```
bstansberry: This is a regression versus current patching. Current patching ignores the config files, so the user changes are preserved.

patching/EAP7-1063_EAP_Patching_With_Galleon.adoc
```diff
@@ -0,0 +1,121 @@
...
+* Uninstall installed feature-packs, including patches;
+* Maintain provisioning state history with the ability to undo going back to previously provisioned state;
+* Preserve user changes during patching and version updates.
+
+=== Soft requirements
+
+== Limitations
+
+Requirement to have access to the repository (local or remote).
+
```
bstansberry: In the limitations section we need a discussion of whether the server can be / cannot be running when its FS is being patched.

I'm pretty sure it's "cannot be". Which is an improvement disguised as a limitation, as trying to patch a running server was probably not necessary yet is the source of many of the problems of the current patching.

Related: not being able to patch a running server would disallow using the HAL tool to apply patches.

bstansberry: Another limitation that follows from not allowing a running server is that remote patching is not supported. The galleon tool is designed to make local FS changes; it is not sending commands to some remote process that is making the FS changes.

patching/EAP7-1063_EAP_Patching_With_Galleon.adoc
```diff
@@ -0,0 +1,121 @@
...
+=== Affected Projects or Components:
+
+Any project using the tool or its API for patching and/or provisioning is going to be affected.
+
+== Requirements
+
+=== Hard requirements
+
+* Install feature-packs, including patches (directly as a file and/or from a repository providing coordinates or using the discovery mechanism);
```
bstansberry: I think the ability to install multiple patches (i.e. one-offs) needs to be stated. I know you mean for it to be included.

Of course one-offs can conflict with each other.

The relationship of one-offs to subsequent micro-version updates also should be stated. I believe the version update discards the one-offs, same as the current mechanism discards one-offs when a CP is applied.

patching/EAP7-1063_EAP_Patching_With_Galleon.adoc
aloubyansky: It can be viewed as a regression. On the other hand, keeping the previous strategy will be wrong in the new implementation simply because the new mechanism allows the user to re-configure the installation, extend it by including additional modules or installing new products, reduce it by excluding installed modules or uninstalling products, etc. Each of those operations may significantly affect the current configuration. Keeping the current config as the active one in this case may result in a broken state.

patching/EAP7-1063_EAP_Patching_With_Galleon.adoc
aloubyansky: Right. We could imagine developing an agent for both use-cases above. For the running server though, it would be picking the time to shut it down and reboot the updated version.

patching/EAP7-1063_EAP_Patching_With_Galleon.adoc
aloubyansky: I added this to the proposal. Thanks!

patching/EAP7-1063_EAP_Patching_With_Galleon.adoc
bstansberry: I don't mean to advocate keeping the previous strategy; sorry if I was unclear about that. The concern is dropping the user modifications; how/why they didn't get dropped before isn't important.

The proposed strategy for standalone is fine. My concern is not applying a similar strategy to domain/host means user modifications get dropped (i.e. stored to .prev, but not in the primary file.)

aloubyansky: I share your concern. All we need to do to support the diff and merging is to export domain and host configs as a list of ops or preferably features. And I guess that's doable.

aloubyansky: What I think we need here is read-resource(recursive) that would return features. Doing the diff we are doing for standalone currently is too expensive (feature-diff operation). If we had that read-features operation that would have been enough and easier to do for domain as well.

aloubyansky: Actually, more like describe-model but in features.

aloubyansky:
@LittleJohnII let's first clarify what our intention wrt to `patch apply` would actually be if we were to keep it. I mean what would our message be doing that?
1) you can continue using it;
2) we guess you hit it because you are used to the command and we let it through this time but next time please use the proper tool?

I am pretty much against 1. We are not going to be maintaining competing patching CLI interfaces designed for two completely different patching mechanisms.

About 2. I am thinking, how would a customer end up doing 2:
a) I know there is a new tool but I want to keep using the old interface;
b) I wasn't aware that the patching mechanism has been replaced with a new one.

A. First of all some of the command options (especially if we are considering other commands) might not be supported anymore which can be viewed as a regression in the interface. What do we do if the customer includes those in a command? Or the whole command doesn't make sense? Warn or fail? We are going to end up with a half-dead CLI interface, which is going to be quite frustrating.
The output of the commands will certainly be affected as well because in many cases we won't be able to generate the output of the previous implementation.
If those are non-interactive use-cases, i.e. scripts. That is especially interesting. Because the mechanisms work differently (on the input and output), how do we know that the result of the updated command will not break the script? How do we communicate that to the customer? You can try using it but be aware that it works in a completely different way and if you have certain expectations wrt to the output of the command or its effects on the filesystem state from your previous experience, in many cases your expectation won't be met.

B *If the client is actually unaware about the patching mechanism replacement* the state of the filesystem content of the installation after the command (although being valid) can be a huge surprise which is really not good. That simply can't happen. Which means we have to clearly communicate what has changed and what to expect from the new mechanism. So B doesn't happen.

Given all that, I don't really see a decent gradual transition. If somebody can suggest such a plan, I'll be happy to discuss it.

BTW, the version where it all happens should also be a fresh install (not a patch upgrading from the previous version):
1) unzip of a galleon-built distribution;
2) installed with galleon.
Either will work. But 2 should probably be encouraged. If the customer intends on patching and updating the new version, galleon will have to be obtained anyway. And starting from the installation might be the first simple experience.

aloubyansky:
I think we should be able to support existing customer configs though. Meaning customers should be able to copy their configs over to the freshly installed distribution. We should be ok with that. Assuming configs are backward compatible (which is not something Galleon itself has to do with).

rsvoboda:
@bmaxwell we need list of `patch xxx` commands which are suggested to customers in GSS documents / KB articles

These commands are supposed to "work" in EAP 7.2 because of backward compatibility requirements.

jfdenise:
@rsvoboda , if we go at removing the CLI patch commands in 7.3 (seems that we will), shouldn't we tag in some way the CLI patch commands now to advertise that they will be removed in the next EAP release? It is not really deprecating because we are not offering an alternative in 7.2. Seems that we can't really  remove them in EAP 7.3 without having said so in 7.2.

patching/EAP7-1063_EAP_Patching_With_Galleon.adoc
```diff
@@ -0,0 +1,124 @@
...
+(BTW, Galleon also serves as an installer, meaning it could be used to install WildFly/EAP in the first place and then to apply patches and version updates.)
+
+Galleon is a repository-based mechanism. That means the final distribution (of WildFly, EAP, LP, etc) is described in a configuration file as a set of artifacts (feature-packs) that contain the info and the content from which the final distribution is going to be built. Those feature-pack artifacts must be available at provisioning time in the artifact repository (or repositories) (Maven in this case, although other repositories could work as well). The repository could be local and/or remote. This is another difference compared to the current patching mechanism where a patch has to be downloaded and provided to the tool by the user directly.
+
+Advatages of the repository-based approach are:
```
wolfc: Advantages

patching/EAP7-1063_EAP_Patching_With_Galleon.adoc
```diff
@@ -0,0 +1,124 @@
...
+When Galleon installs a feature-pack that is provided directly by the user, by default, besides provisioning that feature-pack, it also installs the directly provided feature-pack into the repository, so it is available for future use (in our case it's a local Maven repository). The user may instruct not to install the feature-pack into the repository but it is strongly discouraged to do that.
+
+==== Patch feature-pack
+
+Patch file itself is basically a feature-pack. The only difference comparing to regular feature-packs is it contains an element in its XML description indicating that it is a patch that applies to a specific version of another feature-pack. A patch will be accepted only if the installation includes the feature-pack the patch applies to. It is possible to install multiple different patches for the same feature-pack in the installation. When the feature-pack that was patched is installed or a different (lower or higher) version of it is installed, all the installed patches for that feature-pack will automatically be removed from the installation.
```
wolfc: Do not confuse one-off patches (unsupported) by async z-stream delivery.

patching/EAP7-1063_EAP_Patching_With_Galleon.adoc
wolfc: Might be better to have .new files as opposed to .prev.
See also http://people.ds.cam.ac.uk/jw35/docs/rpm_config.html

Note that 3 way merges have been tried in the past, but so far I'm not aware of great success.

patching/EAP7-1063_EAP_Patching_With_Galleon.adoc
```diff
@@ -0,0 +1,124 @@
...
+
+The documentation is not yet covering all the latest features and additions in details.
+
+http://docs.wildfly.org/galleon/[Galleon Documentation]
+
+http://docs.wildfly.org/galleon-plugins/[WildFly Galleon Plug-ins Documentation]
+
+== Design Notes
+
+
```
wolfc: I'm missing bits on how a Galleon pack is build and what the process is to get from any source base to one that is usable with Galleon.

RPM uses a tarball, patch and spec files to bring it all together, which I think is a bad process.
http://ftp.rpm.org/max-rpm/ch-rpm-basics.html

wolfc: Maybe something like https://github.com/hibernate/lucene-jbossmodules/

bmaxwell:
@rsvoboda here are what users have been using.  
From some of the other comments, it sounds like Galleon is going to be a separate utility, so at the point that the 'patch apply' is removed, it looks like there will be questions about how domain mode will work at that point, as users would expect to be able to apply patches to the domain hosts via the DC.

https://access.redhat.com/solutions/625683
https://access.redhat.com/solutions/1269123

// patching via the Web Console which would be calling the management commands

// patching domain hosts and standalone instances
patch apply /<patch_download_url>/<patch_name>.zip
patch --host=$HOST apply /<patch_download_url>/<patch_name>.zip
$JBOSS_HOME/bin/jboss-cli.sh --command="patch apply /<patch_download_url>/<patch_name>.zip"
// patch apply with --override=... --override-all --override-modules=...

// rollback a patch:
patch rollback --patch-id=<patch-id> --reset-configuration=<true|false>

// ageout history to remove the old layers
/core-service=patching:ageout-history
/host=master/core-service=patching:ageout-history

patching/EAP7-1063_EAP_Patching_With_Galleon.adoc
aloubyansky: Do I? Z-stream is a micro version update, right? So to me, technically, it falls into the version update category. Am I wrong?

patching/EAP7-1063_EAP_Patching_With_Galleon.adoc
```diff
@@ -0,0 +1,124 @@
...
+== Requirements
+
+=== Hard requirements
+
+* Install feature-packs, including patches (directly as a file and/or from a repository providing coordinates or using the discovery mechanism);
+** Install multiple patches targeting the same feature-pack version present in the installation;
+* Uninstall installed feature-packs, including patches;
+* Maintain provisioning state history with the ability to undo going back to previously provisioned state;
+* Preserve user changes during patching and version updates.
```
jstourac: What about version downgrades or one-off/patch uninstall? Do we want to preserve user changes as well? I assume that at least in case of one-off uninstall the answer is yes?

aloubyansky: It doesn't matter whether it's upgrade or downgrade, we preserve user changes when the version changes. More generally, we preserve user changes for any provisioning operation be it install, uninstall or a provisioning plan (a set of install/uninstall, patch, version updates) applied as an atomic installation state transition.
I think we should add an operation to also reset user changes to go back to the last provisioned state. Then we can also add it as an option to any other provisioning operation such as install/uninstall, so the user can choose whether the changes since the last provisioned state should preserved or not.

LittleJohnII:
Bumping a previous unanswered question from @jfdenise 

> @jfdenise : if we go at removing the CLI patch commands in 7.3 (seems that we  will), shouldn't we tag in some way the CLI patch commands now to  advertise that they will be removed in the next EAP release? It is not  really deprecating because we are not offering an alternative in 7.2.  Seems that we can't really  remove them in EAP 7.3 without having said  so in 7.2.
> @rsvoboda : I think this must be discussed and agreed with product management, @n1hility what's your plan with this ?

Now that the patching got postponed to 7.3, it still might be a good idea to mark the patch commands as "being phased out" or something. @n1hility any plans with this?

bstansberry:
I'm closing this one as there are other documents related to updating Galleon-based installations.

PR #110, opened 2018-07-25
[WFCORE-3970] Add Support for Elytron's ldap-key-store to be dynamically reloaded

justinmcook:
Jira: https://issues.jboss.org/browse/EAP7-1052

### DIFF
--- a/elytron/WFCORE-3970-reload-ldap-key-store.adoc
+++ b/elytron/WFCORE-3970-reload-ldap-key-store.adoc
@@ -1,0 +1,58 @@
+= WFCORE-3970 Add support for Elytron's ldap-key-store to be dynamically reloaded
+:author:            Justin Michael Cook
+:email:             jucook@redhat.com
+:toc:               left
+:icons:             font
+:idprefix:
+:idseparator:       -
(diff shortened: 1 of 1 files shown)


elytron/WFCORE-3970-reload-ldap-key-store.adoc
```diff
@@ -0,0 +1,58 @@
...
+
+* WildFly, Security
+
+=== Other Interested Projects
+
+== Requirements
+
+=== Hard Requirements
+
+* When changes are made to certificates in the LDAP directory, it should be possible to reload these certificates dynamically without having to restart the server. This should be done by adding a load operation to the Elytron subsystem's ldap-key-store resource. This new management operation will not take any parameters.
```
fjuma: Just a minor note, it would be good to mention that the new management operation will be a runtime-only operation like the existing key-store load operation.

justinmcook: Updated.

justinmcook:
Need to analysis further so will close for now and reopen later

PR #111, opened 2018-07-25
WFLY-5550 Distributable Web subsystem

pferraro:
https://issues.jboss.org/browse/WFLY-5550

### DIFF
--- a/clustering/web/WFLY-5550_Distributable_Web_Subsystem.adoc
+++ b/clustering/web/WFLY-5550_Distributable_Web_Subsystem.adoc
@@ -1,0 +1,142 @@
+= Distributable Web Subsystem
+:author:            Paul Ferraro
+:email:             paul.ferraro@redhat.com
+:toc:               left
+:icons:             font
+:idprefix:
+:idseparator:       -
+
+== Overview
+
+Distributable web applications are indicated by the use of the <distributable/> element within web.xml.
+A distributable web application is one that is capable of passivating state to support the following scenarios:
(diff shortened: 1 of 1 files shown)


tommaso-borgato:
@pferraro I have one question: in this doc the routing option "RANKED" is mentioned but I cannot find it in [this|https://github.com/pferraro/wildfly/blob/web/docs/src/main/asciidoc/_high-availability/subsystem-support/Distributable_Web_Applications.adoc]; is it intentional?

pferraro:
@tommaso-borgato I was originally going to include the RANKED routing option within the scope of this feature, but decided to move it into a separate body of work that is dependent on this one.  Thus, it should not be mentioned in this analysis document, as it will have its own.   I'll fix that.
The documentation link you pasted only reflects features that are already committed.  You should never see uncommitted features in committed documentation.

clustering/web/WFLY-5550_Distributable_Web_Subsystem.adoc
```diff
@@ -0,0 +1,142 @@
+= <INSERT TITLE HERE>
```
jamezp: Could you add a title?

pferraro: So picky! :)

pferraro: Done.

jamezp:
@tommaso-borgato Does this now seem reasonable to you? The feature freeze is 14.Nov.

PR #113, opened 2018-07-27
[ELY-816] Support for masked passwords in client XML config

hkalina:
https://issues.jboss.org/browse/ELY-816
https://issues.jboss.org/browse/EAP7-1065

Elytron client should support masked passwords - passwords encrypted using predefined key, as it is supported in PicketBox, in any place in Elytron Client XML where clear passwords are currently used.

### DIFF
--- a/elytron/ELY-816-masked-password.adoc
+++ b/elytron/ELY-816-masked-password.adoc
@@ -1,0 +1,74 @@
+= ELY-816 Support for masked passwords in client XML config
+:author:            Jan Kalina
+:email:             jkalina@redhat.com
+:toc:               left
+:icons:             font
+:keywords:          comma,separated,tags
+:idprefix:
+:idseparator:       -
+:issue-base-url:    https://issues.jboss.org/browse
+
+== Overview
+
+Elytron client should support masked passwords - passwords encrypted using predefined key, as it is supported in PicketBox.
+Masked passwords should be possible to use in place of any clear password written in Elytron Client XML configuration.
+
+For example, instead of:
+[source,xml]
+----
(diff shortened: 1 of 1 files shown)


mchoma:
If I recall correctly there is a way to use masked password in credential reference for master password [1].

"it is supported in PicketBox" - what you mean we do not have nothing like wildfly-config.xml in legacy security
"Should be possible to migrate passwords in PicketBox-format into Elytron"  you mean from server configuration to client configuration. They can use credential-reference masking feature

Given that shouldn't it be more about extending credential-reference to other places, rather than introducing another way of hide clear passwords?

Probably this should be added to bigger discussion about Credential Stores RFEs. 

CC: @darranl 

[1] https://access.redhat.com/documentation/en-us/red_hat_jboss_enterprise_application_platform/7.1/html-single/how_to_configure_server_security/#obtain_the_master_password_for_the_credential_store_from_an_external_source

mchoma:
@darranl we should decide if this is really way we should go. Seems to me as duplication to Credential Store. What do you think?

dmlloyd:
It _does_, strictly speaking, but on the other hand a credential store should not be necessary for simple use cases.  It's a usability enhancement IMO.

mchoma:
I would prefer one robust solution. If credential store has some user experience issues, lets rather address these.

dmlloyd:
On the surface that's reasonable, but the user experience issue in this case *is* the credential store: it's two files instead of one.

ashley-abdelsayed98:
This proposal is superseded by wildfly/wildfly-proposals#214 and can be closed

PR #115, opened 2018-07-30
[WFCORE-3970] Add Support for Elytron's trust-manager to be dynamically reloaded

justinmcook:
Jira: https://issues.jboss.org/browse/WFCORE-3970


msve:
The analysis looks good from QE point of view.

PR #116, opened 2018-08-01
EAP7-1075 read-provisioning-config management operation

aloubyansky:
https://issues.jboss.org/browse/EAP7-1075
https://issues.jboss.org/browse/WFCORE-3991

### DIFF
--- a/server/EAP7-1075_read-provisioning-config.adoc
+++ b/server/EAP7-1075_read-provisioning-config.adoc
@@ -1,0 +1,57 @@
+= EAP7-1075 read-provisioning-config management operation
+:author:            Alexey Loubyansky
+:email:             alexey.loubyansky@redhat.com
+:toc:               left
+:icons:             font
+:idprefix:
+:idseparator:       -
+
+== Overview
+
+To be able to preserve user configuration changes when applying patches, version updates, installing and uninstalling products and components (basically pretty much any provisioning operation that may affect the configuration of the installation) we need to identify what the user has changed (manually or using any of the management interfaces) since the last provisioned state. That can be done by comparing the last provisioned state to the currrent state of the installation. The last provisioned state is recorded by Galleon in terms of configured features. What is currently missing is a way to represent the current state of the management model in terms of configured provisioning features.
+
+We already have operation `read-feature-description` that generates provisioning feature specifications. The intent of this proposal is to add operation `read-provisioning-config` to the standalone server and domain that will describe the current state of the management model in terms of configured provisioning features.
+
+== Issue Metadata
+
+=== Issue
+
+* {issue-base-url}EAP7-1075[EAP7-1075]
+
+=== Related Issues
+
+* {issue-base-url}EAP7-1063[EAP7-1063]
+* {issue-base-url}WFCORE-3991[WFCORE-3991]
+
+=== Dev Contacts
+
+* mailto:alexey.loubyansky@redhat.com[Alexey Loubyansky]
+
+=== QE Contacts
+
+* mailto:rjanik@redhat.com[Richard Janík]
+
+=== Affected Projects or Components
+
+* the operation handlers will be added to wildfly-core and used by Galleon provisioning plugins for any project distribution including wildfly-core.
+
+== Requirements
+
+=== Hard Requirements
+
+* the result of the operation has to describe the management model in terms of configured provisioning features in DMR format.
+
+=== Nice-to-Have Requirements
+
+=== Non-Requirements
+
+== Implementation Plan
+
+Operation handlers will be added to server and host-controller modules in wildfly-core.
+
+== Test Plan
+
+== Community Documentation
+
+This has to be implemented in wildfly-core. So the documentation will have to be added as a follow up to wildfly.
+
--- a/server/EAP7-1075_read-provisioning-config.adoc
+++ b/server/EAP7-1075_read-provisioning-config.adoc
@@ -1,4 +1,4 @@
-= EAP7-1075 read-provisioning-config management operation
+= WFCORE-3991 read-config-as-features management operation
 :author:            Alexey Loubyansky
 :email:             alexey.loubyansky@redhat.com
 :toc:               left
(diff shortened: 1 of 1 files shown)


rsvoboda:
@LittleJohnII please take a look

server/EAP7-1075_read-provisioning-config.adoc
```diff
@@ -0,0 +1,57 @@
...
+
+=== Affected Projects or Components
+
+* the operation handlers will be added to wildfly-core and used by Galleon provisioning plugins for any project distribution including wildfly-core.
+
+== Requirements
+
+=== Hard Requirements
+
+* the result of the operation has to describe the management model in terms of configured provisioning features in DMR format.
```
jmesnil: Does its output has to be similar to `read-feature-description` or Galleon will have all the logics to make a "diff" from the 2 operation outputs?

aloubyansky: `read-feature-description` returns different information - the description of a feature (which parameters it exposes, capability requirements, etc), how a feature can be configured. This operation is used during feature spec generation, when wildfly feature-packs are built.

`read-provisioning-config` returns a set of configured provisioning features that represent the current configuration.

server/EAP7-1075_read-provisioning-config.adoc
```diff
@@ -0,0 +1,57 @@
...
+:toc:               left
+:icons:             font
+:idprefix:
+:idseparator:       -
+
+== Overview
+
+To be able to preserve user configuration changes when applying patches, version updates, installing and uninstalling products and components (basically pretty much any provisioning operation that may affect the configuration of the installation) we need to identify what the user has changed (manually or using any of the management interfaces) since the last provisioned state. That can be done by comparing the last provisioned state to the currrent state of the installation. The last provisioned state is recorded by Galleon in terms of configured features. What is currently missing is a way to represent the current state of the management model in terms of configured provisioning features.
+
+We already have operation `read-feature-description` that generates provisioning feature specifications. The intent of this proposal is to add operation `read-provisioning-config` to the standalone server and domain that will describe the current state of the management model in terms of configured provisioning features.
```
bstansberry: On what resources would this operation be registered? My guess is the root resource for both a server and an HC, but on the HC the result would not descend into the /host=* tree, making the result a description of the domain-wide model. Then for host-specific data, the op would be registered on /host=x where x is the local HC name. Perhaps not on /host=*, i.e. it is not wildcard op where you can get the results from all hosts in 1 call. (TBH I think registering it on /host=* should be a 'nice to have', not because it's so nice to have, but because it _may_ be easier to implement it that way. But maybe not.)

My assumption is that, unlike :read-feature-description, this is not meant to be an inherited operation, i.e. available on all child resources of the resource type against which it is registered.

It seems to be analogous to "read-config-as-xml" and perhaps should be called "read-config-as-features".  That's not such a wonderful name that I'm insisting on it. :) I'm just commenting in general about how this best fits into the overall API.

server/EAP7-1075_read-provisioning-config.adoc
```diff
@@ -0,0 +1,57 @@
+= EAP7-1075 read-provisioning-config management operation
```
bstansberry: Please rename and retitle this to use the WFCORE.  EAP7 JIRAs are secondary paperwork.

server/EAP7-1075_read-provisioning-config.adoc
bstansberry: A simple example output would be good.

server/EAP7-1075_read-provisioning-config.adoc
```diff
@@ -0,0 +1,57 @@
...
+== Requirements
+
+=== Hard Requirements
+
+* the result of the operation has to describe the management model in terms of configured provisioning features in DMR format.
+
+=== Nice-to-Have Requirements
+
+=== Non-Requirements
+
```
bstansberry: Any representation of feature groups is not required.

server/EAP7-1075_read-provisioning-config.adoc
aloubyansky: I was not actually aware of `read-config-as-xml`. Yes, having it a root op will be enough.
I have a prototype of this op for standalone. I might need some help for the HC equivalent. And yes, I would like to get the host config as well.

I am sure we'll agree on the name :) `read-config-as-features` sounds ok for me.

server/EAP7-1075_read-provisioning-config.adoc
```diff
@@ -0,0 +1,82 @@
...
+        {
+            "spec" => "core-service.management.management-interface.http-interface.http-upgrade",
+            "id" => {
+                "core-service" => "management",
+                "management-interface" => "http-interface"
+            },
+            "params" => {"enabled" => true}
+        },
+----
+
```
bstansberry: The operation must be registered with OperationEntry.Flag.HIDDEN, which marks it as an internal-use only operatoin.  This has the following effects:
1) the operation is not visible in the publicly described managment API, e.g. in read-operation-names, read-operation-description, read-resource-description
2) As a result of 1) the operation cannot be invoked from the CLI unless jboss-cli.xml has validation turned off.  The CLI also won't suggest it via tab completion.
3) However the operation can be invoked remotely.

server/EAP7-1075_read-provisioning-config.adoc
```diff
@@ -0,0 +1,82 @@
...
+== Implementation Plan
+
+Operation handlers will be added to server and host-controller modules in wildfly-core.
+
+== Test Plan
+
+== Community Documentation
+
+This has to be implemented in wildfly-core. So the documentation will have to be added as a follow up to wildfly.
```
bstansberry: As this is a hidden internal use operation there is no documentation requirement.

server/EAP7-1075_read-provisioning-config.adoc
bstansberry: The operation must have the same RBAC SensitivityClassification as the analogous "read-config-as-xml" operation -- SensitivityClassification.READ_WHOLE_CONFIG.

bstansberry:
A general comment on this is it's not really an end user feature as it's an internal use only API. (WildFly Galleon plugin counting as internal.)  It's good to write down the proposal but in terms of supportability this is basically an internal detail.

jmesnil:
PR for the feature 
https://github.com/wildfly/wildfly-core/pull/3438 has been merged

PR #117, opened 2018-08-08
[WFCORE-3999] Allow registering a custom HTTP handler for management interface

jmesnil:
Issue: https://issues.jboss.org/browse/WFCORE-3999

### DIFF
--- a/management/WFCORE-3999_HttpHandler_ExtensibleHttpManagement.adoc
+++ b/management/WFCORE-3999_HttpHandler_ExtensibleHttpManagement.adoc
@@ -1,0 +1,100 @@
+= [WFCORE-3999] Allow registering a custom HTTP handler for management interface
+:author:            Jeff Mesnil
+:email:             jmesnil@redhat.com
+:toc:               left
+:icons:             font
+:idprefix:
+:idseparator:       -
+
+== Overview
+
+== Issue Metadata
+
+=== Issue
+
+* https://issues.jboss.org/browse/WFCORE-3999[WFCORE-3999]
+
+=== Related Issues
+
+* https://issues.jboss.org/browse/EAP7-1079[EAP7-1079]
+* https://issues.jboss.org/browse/WFLY-10711[WFLY-10711] - Implement Health Check so that it can be consumed by OpenShift Console
+** The HTTP endpoint provided by the MicroProfile Health Check will be exposed by a HTTP handler on
+    WildFly HTTP management interface
+* https://issues.jboss.org/browse/WFCORE-1742[WFCORE-1742] - Allow registering a custom handler for management interface
+** This feature allows to add ResourceManager to the WildFly HTTP management interface. It is targeting static resources such
+    as files and have limitations that make it not a good fit to provide a HTTP API endpoint.
+
+=== Dev Contacts
+
+* mailto:{email}[{author}]
+
+=== QE Contacts
+
+* mailto:rsvoboda@redhat.com[Rostislav Svoboda] - WildFly
+
+=== Affected Projects or Components
+
+* WildFly Core (to expose the new API)
+* WildFly and the `microprofile-smallrye-health` extension that will be added in https://issues.jboss.org/browse/WFLY-10711[WFLY-10711]
+
+=== Other Interested Projects
+
+== Requirements
+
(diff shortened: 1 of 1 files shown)


jmesnil:
@rsvoboda will you be the QE contact for this one (as it relates to the Health Check feature)?

rsvoboda:
@jmesnil yes

rsvoboda:
Will this feature also solve current issue that the `/health` endpoint requires authentication ATM ?

bstansberry:
This needs some clarification re: security integration.  I don't think we want the caller of this new method to take responsibility for standard security integration, but OTOH we need an option to tell the kernel to require authentication for a context, and also need to be able to have a context not require authentication.

MP Health spec says this:

```
Security
It must be possible to secure the endpoints via the usual means. The definition of 'usual means' is in this version of the specification implementation specific.

In case of a secured endpoint, accessing /metrics without valid credentials must return a 401 Unauthorized header.

A server SHOULD implement TLS encryption by default.

It is allowed to ignore security for trusted origins (e.g. localhost)
```

So, we have to be able to turn security on/off on the /health context.

MP Metrics says this:

```
Security
It must be possible to secure the endpoints via the usual means. The definition of 'usual means' is in this version of the specification implementation specific.

In case of a secured endpoint, accessing /metrics without valid credentials must return a 401 Unauthorized header.

A server SHOULD implement TLS encryption by default.

It is allowed to ignore security for trusted origins (e.g. localhost)
```

To me in our case the 'usual means' means you get the normal security configured on the management interface.  And the spec says nothing about default behavior, beyond TLS, but IMO the default should be to require authentication. Metrics can expose all sorts of information.

rsvoboda:
For sure MP Metrics will be more tricky about the data exposure and supported security configuration

jmesnil:
@rsvoboda what I told you previously was incorrect, access to the /health HTTP endpoint does not require authentication.

It requires that a security realm is present (and a management user must exist). If that criteria is met, it is possible to access the endpoint without client authentication.

rsvoboda:
unzip the serve => start the server => curl http://HOST:POST/heath
This works for servers like OpenLiberty

What would be the workflow on WildFly ?

jmesnil:
@rsvoboda current workflow is:

* unzip the server => start the server | enable management user => curl http://HOST:POST/heath

I am working with updating the proposal so that added management handler could bypass having a security realm.
I am also checking that *if* the context wants to secure access, it can do so.

rsvoboda:
enable management user == running add-user.sh script ?

jmesnil:
yes

jmesnil:
I have updated the proposal. The new method has now a `requiresSecurity` boolean to specify how it integrates the management handler with the app server security

management/WFCORE-3999_HttpHandler_ExtensibleHttpManagement.adoc
```diff
@@ -0,0 +1,100 @@
...
+     * Add a context for a HTTP management handler.
+     * @param contextName the name of the context. Cannot be {@code null} or empty
+     * @param requiresSecurity {@code true} if the management handler requires security.
+     * @param managementHandler HTTP management handler. Cannot be {@code null}
+     *
+     * @throws IllegalArgumentException if either parameter is invalid
+     * @throws IllegalStateException if there is already a context present named the same as {@code contextName}, either
+     *                               added via this interface or otherwise
+     */
+    void addManagementHandler(String contextName, HttpHandler managementHandler);
```
rsvoboda: add `boolean requiresSecurity`

jmesnil: fixed

jmesnil:
> Does it mean add-user.sh script invocation won't be needed ? Or it will be needed, but the endpoint won't require authentication?

if security is disabled, add-user.sh script invocation is not needed

management/WFCORE-3999_HttpHandler_ExtensibleHttpManagement.adoc
```diff
@@ -0,0 +1,100 @@
...
+[code,java]
+====
+    /**
+     * Add a context for a HTTP management handler.
+     * @param contextName the name of the context. Cannot be {@code null} or empty
+     * @param requiresSecurity {@code true} if the management handler requires security.
+     * @param managementHandler HTTP management handler. Cannot be {@code null}
+     *
+     * @throws IllegalArgumentException if either parameter is invalid
+     * @throws IllegalStateException if there is already a context present named the same as {@code contextName}, either
```
rsvoboda: just minor thing, this line could be a bit shorter and following a bit longer

PR #124, opened 2018-08-23
[WFCORE-4058] Automatic detection of file based KeyStore types.

nekdozjam:
https://issues.jboss.org/browse/EAP7-675[EAP7-675]
https://issues.jboss.org/browse/ELY-1650[ELY-1650]
https://issues.jboss.org/browse/WFCORE-4058[WFCORE-4058]

### DIFF
--- a/elytron/WFCORE-4058-automatic-keystore-detection.adoc
+++ b/elytron/WFCORE-4058-automatic-keystore-detection.adoc
@@ -1,0 +1,96 @@
+= Automatic detection of file based KeyStore types.
+:author:            Maritn Mazanek
+:email:             mmazanek@redhat.com
+:toc:               left
+:icons:             font
+:idprefix:
+:idseparator:       -
+
+== Overview
+
+* If a user does not specify the type of a file based KeyStore can we auto detect the type?
+
+* There is most likely a finite set of types we would want to automatically recognise, possibly we could inspect the
+contents ourselves to detect different types - alternative attempt initialisation of different KeyStore objects until one succeeds. With the latter we would need to be careful one doesn't silently succeed when it should have failed.
+
+== Issue Metadata
+
+=== Issue
+
+* https://issues.jboss.org/browse/EAP7-675[EAP7-675]
+* https://issues.jboss.org/browse/ELY-1650[ELY-1650]
+* https://issues.jboss.org/browse/WFCORE-4058[WFCORE-4058]
+
+//=== Related Issues
+
+//* https://issues.jboss.org/browse/WFLY[WFLY-XXXX]
+
+=== Dev Contacts
+
+* mailto:mmazanek@redhat.com[Martin Mazanek]
+
+=== QE Contacts
+
+* TBD
+
+=== Affected Projects or Components
+
+* WildFly, Security
+
+== Requirements
+
+=== Hard Requirements
+
+* Point of this change is to be able to load keystore without specifying its type.
+
+== Implementation Plan
+
+* Finite set of supported keystore types for automatic detection: JKS, JCEKS, PKCS12, BKS, BCFKS, UBER
+* Some types are fairly easily recognizable, as they start with magic number: JKS (0xfeedfeed), JCEKS (0xcececece) - although if a keystore starts with such number it doesn't necessarily mean that it is that type
+* Detect types of encrypted keystores? They may be a good candidate for try until success (let the provider do the decryption for us)
+
+* The idea is to prioritize different techniques with a priority
+** Magic number or different recognizeable pattern - check if the store contains one of known magic numbers - if yes, then try to load it and see if it is really that keystore
+** Try remaining supported types in a specific order / don't try anything and fail - the order will be determined using file extension
+
+* Recognizable patterns:
+
+** JKS
+*** Starts with magic number 0xfeedfeed
+*** Version number - 0x00000002 or 0x00000001
+
+** JCEKS
+*** Starts with magic number 0xcececece
+*** Version number - 0x00000002 or 0x00000001
+
+** BKS
+*** Starts with version number 0x00000002 - older version (BKS-v1) has version number 0x00000001
+*** Salt length - looking at the BC provider source I found out its 20, hopefully it't the only case - 0x00000014
+*** Salt - 20 bytes
+*** Iteration count - int with value 1024 - 2047 - 0x00000400 - 0x000007ff - top 21 bits are 0
+**** 32 bytes until now
+*** First entry type - byte of value 0-5 - first different pattern from UBER
+
+
+** BCFKS
+*** ASN.1 encoding - starts with SEQUENCE - tag 0x30
+
+** PKCS12
+*** ASN.1 encoding - starts with SEQUENCE - tag 0x30
+
+** UBER
+*** Starts with version number 0x00000001 or 0x00000000
+*** Salt length - looking at the BC provider source I found out its 20, hopefully it't the only case - 0x00000014
+*** Salt - 20 bytes
+*** Iteration count - int with value 1024 - 2047 - 0x00000400 - 0x000007ff - top 21 bits are 0
+**** 32 bytes until now (same as BKS)
+*** encrypted data
+
+== Test Plan
+
+* TBD based on final implementation
+
+== Community Documentation
+////
+Generally a feature should have documentation as part of the PR to wildfly master, or as a follow up PR if the feature is in wildfly-core. In some cases though the documentation belongs more in a component, or does not need any documentation. Indicate which of these will happen.
+////


darranl:
Looking good so far, I think we will still discuss precedence and decide if we want to give the providers a chance - I would be more inclined to just detect the ones we can detect and leave the others.

One thing we should mention within the analysis is where this will be supported so I would suggest specifying we will support this within authentication client configurations and for KeyStore definitions within the application server.  Any other places that still define their own KeyStores should be listed as out of scope for this feature.

Also as I mentioned in the Jira issue, we could also consider if this makes sense to add to our command line tool, although the tool presently only supports credential store the intent was for it to become a general utility so we could add some form of detect-keystore type option to it.

elytron/WFCORE-4058-automatic-keystore-detection.adoc
```diff
@@ -0,0 +1,92 @@
...
+:icons:             font
+:idprefix:
+:idseparator:       -
+
+== Overview
+
+* If a user does not specify the type of a file based KeyStore can we auto detect the type?
+
+* There is most likely a finite set of types we would want to automatically recognise, possibly we could inspect the
+contents ourselves to detect different types - alternative attempt initialisation of different KeyStore objects until one succeeds. With the latter we would need to be careful one doesn't silently succeed when it should have failed.
```
mchoma: Personally I consider trying loading of keystore more robust approach for this. To be independent of implementation details of keystore types. Also there can be optimalisation try to load only types which are discoverable by JCA. It means try only keystores which we have a chance to load.

Also this detection should be performed ideally only once.

darranl: The load at the end will become the final accurate check - deliberately attempting to initialise a KeyStore multiple times seems quite inefficient - each implementation may have it's own initialisation to perform before the file is read which may be preparation to decrypt an encrypted stream, we then need to read the file in once for each KeyStore we try.  We then need to handle the failure conditions leaving the failed attempt to be garbage collected.

If we reduced the scope to just JKS, JCEKS, and PKCS12 they seem to be coming out with the more predicable signature.

elytron/WFCORE-4058-automatic-keystore-detection.adoc
```diff
@@ -0,0 +1,92 @@
...
+*** Starts with version number 0x00000002 - older version (BKS-v1) has version number 0x00000001
+*** Salt length - looking at the BC provider source I found out its 20, hopefully it't the only case - 0x00000014
+*** Salt - 20 bytes
+*** Iteration count - int with value 1024 - 2047 - 0x00000400 - 0x000007ff - top 21 bits are 0
+
+** BCFKS
+*** Whole file encrypted - needs decryption
+
+** PKCS12
+*** ASN.1 encoding - starts with SEQUENCE - tag 0x30
```
hkalina: If this should be in "specific order", you should probably move PKCS12 between JCEKS and BKS.

elytron/WFCORE-4058-automatic-keystore-detection.adoc
```diff
@@ -0,0 +1,92 @@
...
+:author:            Maritn Mazanek
+:email:             mmazanek@redhat.com
+:toc:               left
+:icons:             font
+:idprefix:
+:idseparator:       -
+
+== Overview
+
+* If a user does not specify the type of a file based KeyStore can we auto detect the type?
```
mchoma: Idea: What about making this opt-in feature. E.g. provide custom type value, e.g.  "AUTO" or "DETECT". I am still worrying about making this code branch default, so everybody must go through by default.

darranl: Maybe this should be clarified in the analysis but this should definitely be opt-in - i.e. if no type specified we use discovery of the type.  If they specify a type that should be the only type we try so this will not impact any existing default configurations out there.

elytron/WFCORE-4058-automatic-keystore-detection.adoc
mchoma: Given PKCS12 is default in JDK9+ and JCEKS is deprecated, I vote for putting it even higher, maybe between JKS and JCEKS :)

darranl: @nekdozjam To detect PKCS12 do you need to decrypt it first or was it detectable from the raw file?

I would propose those that don't need decryption come first in the list, if anything needs decryption it should then follow.

nekdozjam: Yeah, this list was just to make notes about each type. And as the JKS, JCEKS and PKCS12 all have some unique patterns at the start of the file so they don't need to be blindly tried out.

darranl: So maybe the two resolution strategies become: -

1. Defined Type
2. Signature Test -> Extension -> Brute Force

In the second resolution strategy if at any stage the load fails it can fail over to the next in the stage.

elytron/WFCORE-4058-automatic-keystore-detection.adoc
nekdozjam: Yeah, it would probably not be so good to try to discover encrypted keystores, so leave the distinguishable and select the type based on that? If that wouldn't succeed we can simply throw exception and there would be no need for another discovery code.

elytron/WFCORE-4058-automatic-keystore-detection.adoc
nekdozjam: This would be only for the keystores without defined type, as is required before this change. You would basically opt-in by not specifying the type. (thought is that a good idea? - maybe put keystore type as "auto" or something like that, with the hope that it wouldn't collide with actual keystore type name)

nekdozjam: by specifying the keystore type as some hardcoded string (let's say "auto") we would not need to make a model change, only issue would be the collision with probably not ever existing keystore type

darranl: One of the motivations for an RFE like this is to reduce the configuration required - IMO if you don't define a value at all it should be expected we would have to work it out.  

If we start using strings that could collide with custom keystore implementations we will not have reduced the configuration required and now we will have introduced a new problem.

mchoma: You are looking from migration point of view. I am looking from starting from scratch point of view.

Once we change model to be key-store.type optional. This will be possible to create keystore
`/subsystem=elytron/key-store=a:add(credential-reference={clear-text=a},path=/path/to/keystore)`
So by default autodetection will be turned on. And to disable autodetection user action will be needed - type has to be specified. 

What I have been proposing is by default turn autodetection off and turn it on after user action - e.g. type="AUTO".

darranl: The two proposals however are the same: -

1. don't define a type, we default to 'auto' and we detect the keystore type.
2. don't define a type, we take it as undefined so we detect the keystore type.

The only difference is we have now given the string 'auto' special meaning,

elytron/WFCORE-4058-automatic-keystore-detection.adoc
darranl: IMO I don't like brute force unless absolutely necessary - we can still implement signature checks in a way that allows us to fall back to brute force if the signature detection was invalid - the signature check could identify the most likely file type and use that as the highest priority in the brute force loop.

But we should also keep in mind the brute force loop could fail for other reasons such as an invalid password being specified.

elytron/WFCORE-4058-automatic-keystore-detection.adoc
mchoma: No I meant add 'auto' to enum of option. Not to default to 'auto'
1. don't define a type, type is mandatory, it will fail
2. define auto, detect the keystore type

If detection become default, keep in mind startup time of server. Could detection be performed on first usage instead of startup?

darranl: Yes I do not like the idea of using a type of 'auto' as having a special meaning as it would mean we could never support a custom KeyStore of type 'auto'.

elytron/WFCORE-4058-automatic-keystore-detection.adoc
```diff
@@ -0,0 +1,92 @@
...
+* Finite set of supported keystore types for automatic detection: JKS, JCEKS, PKCS12, BKS, BCFKS, UBER
+* Some types are fairly easily recognizable, as they start with magic number: JKS (0xfeedfeed), JCEKS (0xcececece) - although if a keystore starts with such number it doesn't necessarily mean that it is that type
+* Detect types of encrypted keystores? They may be a good candidate for try until success (let the provider do the decryption for us)
+
+* The idea is to prioritize different techniques with a priority
+** Magic number - check if the store contains one of known magic numbers - if yes, then try to load it and see if it is really that keystore
+** File extension - try to parse as a keystore based on the file extension
+** Try remaining supported types in a specific order / don't try anything and fail
+
+* Recognizable patterns:
```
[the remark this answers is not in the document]
darranl: A more suitable question would be - what mechanism to the existing providers use to verify if the file is suitable for them to load?  If they are also using these same signature tests it will not be possible for them to be changed in the future.

However if we use the signature checks to adjust the priority order for a brute force check a subsequent change changes very little.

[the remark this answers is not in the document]
nekdozjam: The providers do similar checks - with keystores that have magic numbers they check that, they usualy have version number after that, so they checkt that and the rest is just parsing, so if something is corrupted, it will simply fail parsing (+ there is usually some kind of digest at the end). Others don't perform any verification at all, they start parsing and if the data is not in the right format then you have parsing fail.

Anyways the JKS, JCEKS and PKCS12 should not change its recognizable patterns, as JKS and JCEKS have magic numbers (which definitely won't change) and PKCS12 is purely ASN.1 DER data, which will always start with SEQUENCE tag 0x30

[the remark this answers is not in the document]
nekdozjam: What could change is the version numbers of the keystores, but that can be left out for the providers to figure out, as we can figure out the type without it

elytron/WFCORE-4058-automatic-keystore-detection.adoc
```diff
@@ -0,0 +1,92 @@
...
+
+=== Affected Projects or Components
+
+* WildFly, Security
+
+== Requirements
+
+=== Hard Requirements
+
+* Point of this change is to be able to load keystore without specifying its type.
```
mchoma: What about detect type during management operation and persist detected type into standalone.xml?

elytron/WFCORE-4058-automatic-keystore-detection.adoc
```diff
@@ -0,0 +1,92 @@
...
+
+== Implementation Plan
+
+* Finite set of supported keystore types for automatic detection: JKS, JCEKS, PKCS12, BKS, BCFKS, UBER
+* Some types are fairly easily recognizable, as they start with magic number: JKS (0xfeedfeed), JCEKS (0xcececece) - although if a keystore starts with such number it doesn't necessarily mean that it is that type
+* Detect types of encrypted keystores? They may be a good candidate for try until success (let the provider do the decryption for us)
+
+* The idea is to prioritize different techniques with a priority
+** Magic number - check if the store contains one of known magic numbers - if yes, then try to load it and see if it is really that keystore
+** File extension - try to parse as a keystore based on the file extension
```
mchoma: This step is erlated only to keystore types, which weren't tried in previous step, right?

elytron/WFCORE-4058-automatic-keystore-detection.adoc
```diff
@@ -0,0 +1,92 @@
...
+** Try remaining supported types in a specific order / don't try anything and fail
```
mchoma: These are opposite ways. So which we will take? 

My concern is performance. You said it is intended to try only on server start (what about keystore reloads?) I assume biggest time consumption will be with trying to load keystores. How many keystores in worst case server will load? Could you approximate how many miliseconds it could take?

elytron/WFCORE-4058-automatic-keystore-detection.adoc
nekdozjam: yes, this is only to optimize the bruteforce approach

elytron/WFCORE-4058-automatic-keystore-detection.adoc
nekdozjam: Luckily I managed to make it 2 keystore loads as the worst case scenario, so performance shouldn't be a big problem. About time I can't guess, but the failed try will take less time than success, so in theory it should take less time than loading 2 keystores with defined type.

It will be implemented in the KeyStoreService, reaload should work fine and will use the recognized type - only one keystore load (or should I make it to recognize again, if someone decided to change the keystore format on running server?)

mchoma: I would say, changing keystore format on running server will be marginal use case. But when we say it will be automatic, I would say we should cover that scenario as well.

Can you measure duration of worst case detection? To know exactly what performance lost are we talking here about? So from what was said, I assume it will be just couple of milliseconds.

So again. You should choose in analysis form 
* Try remaining supported types in a specific order 
* Don't try anything and fail

PR #126, opened 2018-08-26
[WFLY-10810] Add JASPI support to WildFly Elytron.

darranl:
This replaces the analysis previously being tracked at https://developer.jboss.org/wiki/AnalysisDesign-JASPICIntegrationWithWildFlyElytron

### DIFF
--- a/elytron/WFLY-10810_Elytron_JASPI_Servlet_Profile_Support.adoc
+++ b/elytron/WFLY-10810_Elytron_JASPI_Servlet_Profile_Support.adoc
@@ -80,8 +80,8 @@ Our `ServerAuthConfig` implementation will include the following features: -
 * Each will be associated with a flag to control the behaviour in the event of success / failure (The result of these flags will be specified in the documentation)
 ** Required
 ** Requisite
-** SUFFICIENT
-** OPTIONAL
+** Sufficient
+** Optional
 * Each `ServerAuthModule` can be associated with it's own properties that will be passed in for initialisation.
 
 The WildFly Elytron Subsystem will have a new resource `jaspi-configuration` defined, as JASPI configuration providers are associated with the `AuthConfigFactory` using both a `layer` and an `appContext` those two values will be defined as attributes on the resource - this means the name of the resource will only be used to allow it to be referenced in the management model and will bear no relationship to the registration.  The new resource will support an ordered list of auth modules being specified, each with their own control flag and configuration properties.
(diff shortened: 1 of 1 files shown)


elytron/WFLY-10810_Elytron_JASPI_Servlet_Profile_Support.adoc
```diff
@@ -0,0 +1,160 @@
...
+The JASPI specification defines a very flexible API which allows for programatic manipulation by deployments, however it does not provide any API for actually instantiating an `AuthConfigProvider` or related classes to be able to register them.
+
+This RFE is focused on support for the Servlet profile which defines server side request handling only, WildFly Elytron implementations of `javax.security.auth.message.config.AuthConfigProvider`, `javax.security.auth.message.config.ServerAuthConfig`, and `javax.security.auth.message.config.ServerAuthContext`.  Again these will be implemented according to the requirements of the API however we are given an opportunity to define our own behaviour in relation to how we support multiple authentication modules.
+
+Our `ServerAuthConfig` implementation will include the following features: -
+* The `ServerAuthModule` instances will be strictly ordered.
+* Each will be associated with a flag to control the behaviour in the event of success / failure (The result of these flags will be specified in the documentation)
+** Required
+** Requisite
+** SUFFICIENT
```
OndrejKotek: Are these two in upper case intentionally?

darranl: No I will correct that tomorrow, this is only supposed to be an illustration of the flags that will be supported not specify an exact API.

elytron/WFLY-10810_Elytron_JASPI_Servlet_Profile_Support.adoc
```diff
@@ -0,0 +1,160 @@
...
+
+The overall implementation of this will affect four different projects, however no special steps should be required provided they are merged in dependency order: -
+
+. WildFly Elytron
+. Elytron Web
+. WildFly Core
+. WildFly 
+
+== Test Plan
+
```
OndrejKotek: Which scenarios do you plan to add/have you added to Elytron and WildFly Core?

TCK also cover some cases which is mapped by QE test plan (draft).

darranl: I will add a lit tomorrow of the test coverage currently planned, we can then discuss if we want to get any additional scenarios added.

darranl:
@OndrejKotek Have just added the test scenarios and updated the case on those other flags.

OndrejKotek:
@darranl Thanks a lot.

The analysis covers the RFE comprehensibly. OK for QE.

PR #129, opened 2018-09-13
[WFLY-7186] Add analysis for supporting the Security API defined under JSR-375 with WildFly Elytron.

darranl:
https://issues.jboss.org/browse/WFLY-7186
https://issues.jboss.org/browse/EAP7-993


OndrejKotek:
OK for QE.

PR #131, opened 2018-09-26
[EAP7-1113] [WFCORE-4127] Disable CLI paging with an option

MMarus:
Issue: https://issues.jboss.org/browse/EAP7-1113

### DIFF
--- a/cli/EAP7-1113_Disable_CLI_paging_with_an_option.adoc
+++ b/cli/EAP7-1113_Disable_CLI_paging_with_an_option.adoc
@@ -1,0 +1,69 @@
+= EAP7-1113 Disable CLI paging with an option
+:author:            Marek Marusic
+:email:             mmarusic@redhat.com
+:toc:               left
+:icons:             font
+:keywords:          cli,disable,option,output,scrolling
+:idprefix:
+:idseparator:       -
+
+== Overview
+
+The output paging is not desired in some cases e.g. when copy/pasting the full output of a command.
+The paging could be disabled by a simple option.
+
+=== Workaround
+It is currently possible to override the paging of a single command by redirecting the output to /dev/stdout.
+
+==== Example of the workaround:
+----
+/subsystem=elytron:read-resource-description(recursive=true) > /dev/stdout
+----
+
+
+== Issue Metadata
+
+=== Issue
+
+* https://issues.jboss.org/browse/EAP7-1113[EAP7-1113]
+
+=== Related Issues
+
+* https://issues.jboss.org/browse/WFCORE-4127[WFCORE-4127]
+* https://issues.jboss.org/browse/WFCORE-4096[WFCORE-4096]
+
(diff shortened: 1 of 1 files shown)


cli/EAP7-1113_Disable_CLI_paging_with_an_option.adoc
```diff
@@ -0,0 +1,66 @@
...
+=== Affected Projects or Components
+
+* CLI
+
+=== Other Interested Projects
+
+== Requirements
+
+=== Hard Requirements
+* Ability to disable output paging via CLI argument (--no-output-paging)
```
mnovak1: Here should be provided example of CLI command where this option will be used. 

--no-output-paging does not seem to be intuitive. At the first look i'm not sure what paging means. What about some other name? Maybe  --no-output-scrolling wdyt?

simkam: I vote for "paging" :)

jfdenise: I would vote for --no-paging, paging seems the right word.

mnovak1: Ok, let's have "--no-paging"

MMarus: Thank you everyone I have made the requested changes.

cli/EAP7-1113_Disable_CLI_paging_with_an_option.adoc
```diff
@@ -0,0 +1,69 @@
...
+* CLI
+
+=== Other Interested Projects
+
+== Requirements
+
+=== Hard Requirements
+* Ability to disable output paging via CLI argument e.g.:
+----
+./jboss-cli.sh --no-paging
```
mnovak1: @MMarus  @jfdenise I think users will need to change this behaviour per command basis. Like:
```
[standalone@localhost:9990 /] /subsystem=elytron:read-resource-description(recursive=true) --no-paging
```
or 
```
./jboss-cli.sh -c 
[standalone@localhost:9990 /] disable-paging
[standalone@localhost:9990 /] /subsystem=elytron:read-resource-description(recursive=true)
```
At moment of starting CLI, I don't think that users plans to disable paging. At least I don't :-) wdyt?

MMarus: There is a workaround for this:
```/subsystem=elytron:read-resource-description(recursive=true) > /dev/stdout```
however, it might not be the best solution for this. Let's wait for @jfdenise 's opinion about this :).

jfdenise: The RFE is really to "start" the cli with paging ON or OFF, not to enable/disable per command.

PR #133, opened 2018-10-01
[WFCORE-4144] Add support for registering a default SSLContext within the server.

darranl:
https://issues.jboss.org/browse/WFCORE-4144

### DIFF
--- a/elytron/WFCORE-4144-default-ssl_context-server.adoc
+++ b/elytron/WFCORE-4144-default-ssl_context-server.adoc
@@ -1,0 +1,59 @@
+= WFCORE-4144 Default SSLContext Registration (Within Application Server)
+:author:            Darran Lofthouse
+:email:             darran.lofthouse@redhat.com
+:toc:               left
+:icons:             font
+:idprefix:
+:idseparator:       -
+
+== Overview
+
+This is a relatively simple enhancement to ensure that as the server is started a JVM wide default SSLContext is registered for the use of any libraries within the application server that support the use of the default context.
+
+== Issue Metadata
+
+=== Issue
+
+* https://issues.jboss.org/browse/WFCORE-4144[WFCORE-4144]
+
+=== Related Issues
+
+=== Dev Contacts
+
+* mailto:{email}[{author}]
+
+=== QE Contacts
+
+=== Affected Projects or Components
+
+WildFly Core
+
+=== Other Interested Projects
+
+None
+
+== Requirements
+
+=== Hard Requirements
+
+At the root of the WildFly Elytron subsystem a new attribute `default-ssl-context` will be registered.  This will be a reference to resources providing the `org.wildfly.security.ssl-context` capability.
+
+As the services for the subsystem start a reference to the `SSLContext` instance will be obtained and will be set globally using the API https://download.java.net/java/early_access/jdk11/docs/api/java.base/javax/net/ssl/SSLContext.html#setDefault(javax.net.ssl.SSLContext)[SSLContext.setDefault(SSLContext)].
+
+=== Nice-to-Have Requirements
+
+None
+
+=== Non-Requirements
+
+Handling of the registration of a default SSLContext in stand alone clients will not be covered by this RFE.
+
+This RFE is concerned with the registration of a single SSLContext, this RFE will not cover selecting alternative configurations based on the peer being accessed.
+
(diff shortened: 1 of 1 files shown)


elytron/WFCORE-4144-default-ssl_context-server.adoc
```diff
@@ -0,0 +1,59 @@
...
+
+None
+
+== Requirements
+
+=== Hard Requirements
+
+At the root of the WildFly Elytron subsystem a new attribute `default-ssl-context` will be registered.  This will be a reference to resources providing the `org.wildfly.security.ssl-context` capability.
+
+As the services for the subsystem start a reference to the `SSLContext` instance will be obtained and will be set globally using the API https://download.java.net/java/early_access/jdk11/docs/api/java.base/javax/net/ssl/SSLContext.html#setDefault(javax.net.ssl.SSLContext)[SSLContext.setDefault(SSLContext)].
```
mchoma: Just to make sure. There is no option deployment could start before this service will start, right? In other words there is not possible couple of calls will use old SSLContext and then new one will be used.

elytron/WFCORE-4144-default-ssl_context-server.adoc
```diff
@@ -0,0 +1,59 @@
...
+
+None
+
+=== Non-Requirements
+
+Handling of the registration of a default SSLContext in stand alone clients will not be covered by this RFE.
+
+This RFE is concerned with the registration of a single SSLContext, this RFE will not cover selecting alternative configurations based on the peer being accessed.
+
+It will not be possible to cache a current default `SSLContext` as this would lead to redundant initialisation that could fail, additionally it is not possible to set the default to `null` to trigger initialisation on next access so the `SSLContext` will not be cleared if the subsystem is shut down.  Unsetting the `default-ssl-context` attribute would put the process in a `restart-required` state instead of `reload-required`.
```
mchoma: +1. 

But could there happen SSLContext change (SSLContext service is restarted) but subsystem is not restarted?
E.g. key/trust-manager reload() management operation? Therefore shouldnt be default flag somewhere on ssl-context definition?

elytron/WFCORE-4144-default-ssl_context-server.adoc
darranl: I still need to work a little on that area to check, I was trying to avoid using a service dependency so I could guarantee registration happens during Stage.RUNTIME.

I may still be able to use a delegate to avoid any race conditions.

elytron/WFCORE-4144-default-ssl_context-server.adoc
darranl: In those cases we use wrappers so the SSLContext itself remains the same.

The reason I wanted to avoid the 'default' flag on the management model was because there is not way to define a constraint that ensures there is only one default until we add a manual verification check to the model.

Also using the default flag it would mean only our SSLContext could be registered, we couldn't pull in an SSLContext from another subsystem.

Finally at some point in the future I would like to look at adding a dynamic SSLContext that can actually map peer information to different SSLContext instances, using this single referencing attribute will make it much easier to select the dynamic context without having different resources all with a 'default' attribute.

elytron/WFCORE-4144-default-ssl_context-server.adoc
```diff
@@ -0,0 +1,59 @@
...
+
+At the root of the WildFly Elytron subsystem a new attribute `default-ssl-context` will be registered.  This will be a reference to resources providing the `org.wildfly.security.ssl-context` capability.
+
+As the services for the subsystem start a reference to the `SSLContext` instance will be obtained and will be set globally using the API https://download.java.net/java/early_access/jdk11/docs/api/java.base/javax/net/ssl/SSLContext.html#setDefault(javax.net.ssl.SSLContext)[SSLContext.setDefault(SSLContext)].
+
+=== Nice-to-Have Requirements
+
+None
+
+=== Non-Requirements
```
mchoma: How about domain mode? I assume we are not able to register default SSLContext on server jvms, right? So could you explicitly mention this new attribute will be available only in standalone mode?

darranl: Each process in domain mode (host controller and server) will install it's own copy of the subsystem so it will be possible to register a default SSLContext in each of those processes in exactly the same way as we do for standalone mode.

PR #136, opened 2018-10-04
[ELY-1686] Utility to port legacy properties file into a FileSystemRealm

justinmcook:
Jira: https://issues.jboss.org/browse/ELY-1686

### DIFF
--- a/elytron/ELY-1686-legacy-properties-files-into-FileSystemRealm-utility.adoc
+++ b/elytron/ELY-1686-legacy-properties-files-into-FileSystemRealm-utility.adoc
@@ -1,0 +1,170 @@
+= ELY-1686 Add utility to port legacy properties files into a FileSystemRealm
+:author:            Justin Michael Cook
+:email:             jucook@redhat.com
+:toc:               left
+:icons:             font
+:idprefix:
+:idseparator:       -
+:issue-base-url:    https://issues.jboss.org/browse
+
+== Overview
+
+Currently, converting legacy security properties files into an Elytron FileSystemRealm
+must be done manually, there should be a migration tool that automates this conversion.
+This task is to add such a tool to WildFly Elytron Tool.
+
+== Issue Metadata
+
+=== Issue
+
+* https://issues.jboss.org/browse/ELY-1686[ELY-1686]
+
+=== Related Issues
+
+* https://issues.jboss.org/browse/EAP7-1093[EAP7-1093]
+
+=== Dev Contacts
+
+* mailto:{email}[{author}]
+
+=== QE Contacts
+
+* mailto:msvehla@redhat.com[Martin Švehla]
+
+=== Affected Projects or Components
+
+* WildFly Elytron Tool
+
+=== Other Interested Projects
+
+== Requirements
+
+=== Hard Requirements
+
+* It should be possible to have legacy properties file converted into an Elytron
+FileSystemRealm automatically. This should be done by adding a command named
+FileSystemRealmCommand to the Elytron-Tool. This command will create a users directory from
+a given path and then use that directory to create a FileSystemRealm. The command will provide
+a script file containing any additional WildFly CLI commands needed to fully register the
+FileSystemRealm in Elytron, such as those needed to configure the security-domain.
+The command will not provide any WildFly CLI commands in the output for any optional parameters
+that are not specified.
+
+* This command will require a users properties file, roles properties file, and output
+directory. The paths to these files and directory should be able to be
+relative to the administrator's current directory or should be able to be absolute.
+Optionally, the name of the filesystem-realm and security-domain can be provided.
+If the name of the filesystem-realm or security-domain are not provided, then the command
+will use default values for these options. The default value for filesystem-realm will be
+converted-properties-filesystem-realm and the default value for security-domain will be
+converted-properties-security-domain.
+
+** Command with only required options
+[source,bash]
+./bin/elytron-tool.sh filesystem-realm --users-file/-u <usersFile> --roles-file/-r <rolesFile>
+--output-location/-o <outDirectory>
+
+** Command with all possible options
+[source,bash]
+./bin/elytron-tool.sh filesystem-realm --users-file/-u <usersFile> --roles-file/-r <rolesFile>
+--output-location/-o <outDirectory> [--filesystem-realm-name/-f] <name>
+[--security-domain-name/-s] <name>
+
+* There should be a text descriptor file that allows multiple inputs to be specified. The
+descriptor file should only accept the long-form option names for readability. Using the
+text descriptor file will have its own option, --bulk-convert/-b
+
+** CLI Command
+[source,bash]
+./bin/elytron-tool.sh filesystem-realm --bulk-convert/-b <file>
+
+** Text descriptor file
+[source,text]
+----
+user-file:/relative/path/to/usersFile1
+roles-file:/relative/path/to/rolesFile1
+output-location:/relative/path/to/output/outDirectory1
+filesystem-realm-name:nameOfFileSystemRealm1
+security-domain-name:nameOfSecurityDomain1
+
+user-file:/absolute/path/to/usersFile2
+roles-file:/absolute/path/to/rolesFile2
+output-location:/absolute/path/to/output/outDirectory2
+filesystem-realm-name:nameOfFileSystemRealm2
+security-domain-name:nameOfSecurityDomain2
+----
+
+* Summary of options and their description
+|===
+|Long-form |Short-form |Description
+
+| --users-file
+| -u
+| The relative or absolute path to the users file
+
+| --roles-file
+| -r
+| The relative or absolute path to the roles file
+
+| --output-location
+| -o
+| The relative or absolute path to the output directory
+
+| --filesystem-realm-name
+| -f
+| Name of the filesystem-realm to be configured
+
+| --security-domain-name
+| -s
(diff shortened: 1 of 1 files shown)


elytron/ELY-1686-legacy-properties-files-into-FileSystemRealm-utility.adoc
```diff
@@ -0,0 +1,109 @@
...
+== Requirements
+
+=== Hard Requirements
+
+* It should be possible to have legacy properties files converted into an Elytron
+FileSystemRealm automatically. This should be done by adding a command named
+FileSystemRealmCommand to the Elytron-Tool. This command will take in the name of
+the file or files to be converted as the only parameter. The command will parse the given
+legacy properties file to configure the new security domain, appropriate factories,
+appropriate subsystems, and http-interface if necessary. There should be two possible
```
fjuma: It would be good to give some more details on what specifically the command will do (e.g., it will parse a legacy properties file and create a ```FileSystemSecurityRealm``` with users/passwords from the legacy properties file, and then it will output Elytron subsystem commands that can be used to configure a ```filesystem-realm```, ```security-domain```, etc.).

elytron/ELY-1686-legacy-properties-files-into-FileSystemRealm-utility.adoc
```diff
@@ -0,0 +1,109 @@
...
+ways to specify the file(s): a singular file, or a comma-separated list of files.
+When specifying multiple files, the command should ensure each entry in the list is unique.
+
+** Singular file
+[source,bash]
+./bin/elytron-tool.sh filesystem-realm --file/-f <file>
```
fjuma: A parameter will be needed to specify the location where the new ```FileSystemSecurityRealm``` will be stored.

elytron/ELY-1686-legacy-properties-files-into-FileSystemRealm-utility.adoc
```diff
@@ -0,0 +1,109 @@
...
+
+** List of files
+[source,bash]
+./bin/elytron-tool.sh filesystem-realm --file/-f <file1,file2,...>
```
fjuma: For determining the Elytron subsystem commands that can be used to configure things like a ```filesystem-realm``` and ```security-domain```, names for these resources will be required. These could be specified using parameters.

When attempting to convert multiple legacy properties file, multiple path parameters and multiple resource name parameters would be needed. To simplify things, perhaps you could make use of a descriptor file to do this, similar to what the Elytron Tool already does to bulk convert multiple vaults.

elytron/ELY-1686-legacy-properties-files-into-FileSystemRealm-utility.adoc
```diff
@@ -0,0 +1,109 @@
...
+
+=== Nice-to-Have Requirements
+
+* The command should have the ability to specify the file(s) from a regex pattern or a
+list of regex patterns
+[source,bash]
+./bin/elytron-tool.sh filesystem-realm --match/-m <regex>
```
fjuma: If allowing regular expressions, something to consider is how the required parameter values will be specified in this case.

elytron/ELY-1686-legacy-properties-files-into-FileSystemRealm-utility.adoc
```diff
@@ -0,0 +1,109 @@
...
+./bin/elytron-tool.sh filesystem-realm --match/-m <regex1,regex2,...>
+
+* The command should have the ability to search directories, converting all found
```
fjuma: Same thing here, would need to consider how the required parameter values will be specified in this case.

elytron/ELY-1686-legacy-properties-files-into-FileSystemRealm-utility.adoc
fjuma: This analysis should also mention how roles properties files will be handled.

justinmcook:
Updated the doc.

elytron/ELY-1686-legacy-properties-files-into-FileSystemRealm-utility.adoc
```diff
@@ -0,0 +1,198 @@
...
+relative to the administrator's current directory or should be able to be absolute.
+Optionally, it will require the name of the filesystem-realm, security-domain,
+http-authentication-factory, and application-security domain to be specified. If these
+options are not specified, then they will be generated using the inputted file names.
+It should be possible to specify the password using any supported mechanism.
+
+** Command with only required options
+[source,bash]
+./bin/elytron-tool.sh filesystem-realm --user-file/-u <userFile> --roles-file/-r <rolesFile>
+--output-location/-o <outDirectory> --password/-p <password>
```
fjuma: What is the password for?

elytron/ELY-1686-legacy-properties-files-into-FileSystemRealm-utility.adoc
```diff
@@ -0,0 +1,199 @@
...
+
+** Command with all possible options
+[source,bash]
+./bin/elytron-tool.sh filesystem-realm --user-file/-u <userFile> --roles-file/-r <rolesFile>
+--output-location/-o <outDirectory> --password/-p <password> [--filesystem-realm/-f] <name>
```
fjuma: To make it a bit more clear that the ```filesystem-realm``` parameter requires a name, it may be better to call it ```filesystem-realm-name```. Same with other parameters that take a name.

elytron/ELY-1686-legacy-properties-files-into-FileSystemRealm-utility.adoc
```diff
@@ -0,0 +1,199 @@
...
+| --bulk-convert
+| -b
+| Tells the command to use a descriptor file, for multiple users-roles file combinations
+can be provided.
+|===
+
+=== Nice-to-Have Requirements
+
+* The command should have an interactive option, similar to the -i option in the unix rm
+command, where the user is prompted for confirmation prior to each file conversion
```
fjuma: Since the user is explicitly specifying the files they want to convert in the descriptor file, I wonder if it makes sense to still prompt them for each conversion. May be good to check the behaviour of the existing Vault command to see what it does since its bulk conversion process is similar.

elytron/ELY-1686-legacy-properties-files-into-FileSystemRealm-utility.adoc
justinmcook: It is for the set-password part of filesystem-realm, for example: /subsystem=elytron/filesystem-realm=exampleFsRealm/identity=user1:set-password( clear={password="password123"}) in https://docs.jboss.org/author/display/WFLY/WildFly+Elytron+Security#WildFlyElytronSecurity-ConfigureAuthenticationwithaFilesystemBasedIdentityStore

fjuma: The usernames/passwords will be in the ```users-file```.

elytron/ELY-1686-legacy-properties-files-into-FileSystemRealm-utility.adoc
```diff
@@ -0,0 +1,174 @@
...
+|===
+|Long-form |Short-form |Description
+
+| --silent
+| N/A
+| Suppresses all output except errors and prompts
+
+| --summary
+| N/A
+| Provides a detailed summary of all operations performed, once the command finishes.
```
msve: Are --silent and --summary mutually exclusive?

justinmcook: I think it would make sense to have them mutually exclusive, but I see no reason they have to be. --silent would suppress the runtime output, while --summary would give an output of everything that happened, after the command finishes converting all files.

msve: Yeah both ways make sense here, hence my question. Do we have any precedent from other similar commands in WF or other JBoss projects?

justinmcook: WF has a silent option and some commands in WildFly-Elytron-Tool have a summary option, but I'm not aware of anywhere that they're both available.

elytron/ELY-1686-legacy-properties-files-into-FileSystemRealm-utility.adoc
```diff
@@ -0,0 +1,170 @@
...
+relative to the administrator's current directory or should be able to be absolute.
+Optionally, the name of the filesystem-realm and security-domain can be provided.
+If the name of the security-domain is not provided, then the output WildFly script will not
+include configuring the security-domain. If the name of the filesystem-realm is not provided,
+then no output WildFly script will be generated, due to the security-domain requiring the
+name of the filesystem-realm.
+
+** Command with only required options
+[source,bash]
+./bin/elytron-tool.sh filesystem-realm --user-file/-u <userFile> --roles-file/-r <rolesFile>
```
fjuma: I think ```--user-file``` should be ```--users-file``` to make it consistent with the roles parameter.

justinmcook: Fixed

kabir:
@fjuma @msve Please approve and flick the AD PT

PR #142, opened 2018-10-16
EAP7-1132 Replace 'list-log-files' with using the 'log-file' resource

claudio4j:
https://issues.jboss.org/browse/EAP7-1132

### DIFF
--- a/console/EAP7-1132_Replace_list-log-files.adoc
+++ b/console/EAP7-1132_Replace_list-log-files.adoc
@@ -1,0 +1,38 @@
+= EAP7-1132 Replace 'list-log-files' with using the 'log-file' resource
+:author:            Claudio Miranda
+:email:             claudio@redhat.com
+:toc:               left
+:icons:             font
+:idprefix:
+:idseparator:       -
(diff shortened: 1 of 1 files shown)


console/EAP7-1132_Replace_list-log-files.adoc
```diff
@@ -0,0 +1,38 @@
...
+
+* mailto:claudio@redhat.com[Claudio Miranda]
+
+=== QE Contacts
+
+* mailto:pjelinek@redhat.com[Pavel Jelinek]
+
+== Requirements
+
+* Read the log files with :read-resource operation on /subsystem=logging/log-file=* address
```
jamezp: One minor note that the attributes are all runtime-only attributes. You may need to add `include-runtime=true`. However that also attaches the log to the `OperationResponse` as a stream.

claudio4j: We use the following operation to read all log file names 
```/subsystem=logging:read-children-resources(child-type=log-file,recursive,include-runtime)```

PR #147, opened 2018-10-18
[WFLY-10712] Observability: MicroProfile Metrics

jmesnil:
Proposal to improve WildFly observability by exposing its metrics
through the MicroProfile Metrics specification and HTTP endpoints.

JIRA: https://issues.jboss.org/browse/WFLY-10712


microprofile/WFLY-10712_microprofile_metrics.asciidoc
```diff
@@ -0,0 +1,230 @@
...
+:toc:               left
+:icons:             font
+:idprefix:
+:idseparator:       -
+:keywords:          observability,microprofile,metrics,openshift
+
+== Overview
+
+The goal of this feature is to increase observability of WildFly by exposing its metrics to monitoring tools such as https://prometheus.io[Promotheus]
+and the https://docs.openshift.com/container-platform/3.11/architecture/infrastructure_components/web_console.html[OpenShift Console].
```
rsvoboda: Is the console really consumer of these metrics?

What about https://github.com/kiali/kiali ?

microprofile/WFLY-10712_microprofile_metrics.asciidoc
```diff
@@ -0,0 +1,230 @@
...
+
+This feature adds support for the https://github.com/eclipse/microprofile-metrics/releases/download/1.1.1/metrics_spec-1.1.1.pdf[Eclipse MicroProfile Metrics 1.1.1 specification].
```
rsvoboda: Any idea when 1.2 will be available ? @pilhuhn

https://github.com/eclipse/microprofile-metrics/compare/1.1.1...master shows decent amount of commits

microprofile/WFLY-10712_microprofile_metrics.asciidoc
```diff
@@ -0,0 +1,230 @@
...
+
+=== Affected Projects or Components
+
+The implementation of Eclipse MicroProfile Metrics 1.1.1 is provided by https://github.com/smallrye/smallrye-metrics[smallrye-metrics].
+
+=== Other Interested Projects
+
+* OpenShift Web Console MUST be able to consume metrics exposed by this extension.
+* Prometheus MUST be able to consume metrics exposed by this extension.
+* HAL, WildFly Web Console, will expose the new extension to configure it and potentially display the metrics exposed by the extension.
```
rsvoboda: Where is the RFE defined for HAL ?

microprofile/WFLY-10712_microprofile_metrics.asciidoc
```diff
@@ -0,0 +1,230 @@
...
+* https://www.kiali.io/[Kiali] SHOULD be able to consume metrics exposed by this extension.
+* https://github.com/thorntail/thorntail/tree/master/fractions/microprofile/microprofile-metrics/src/main/java/org/wildfly/swarm/microprofile/metrics/runtime[Thorntail metrics fraction] - Some WildFly-based code to activate and install metrics in WildFly have been developed for Thorntail 2.x and could be
+reused in the proposed extension.
```
rsvoboda: +1 to reuse these. Can it get into smallrye-metrics or new smallrye-metric-xxx project will be needed ?
e.g. smallrye-metrics-activator

rsvoboda:
@maschmid please take a look, you will probably need to cover openshift part

microprofile/WFLY-10712_microprofile_metrics.asciidoc
```diff
@@ -0,0 +1,230 @@
...
+
+== Requirements
+
+=== External Requirements
+
+The external requirements corresponds to the way metrics are discovered and consumed *from* WildFly by monitoring tools such as Promotheus and the OpenShift Web Console.
+
+* Provide the MicroProfile Metrics HTTP endpoints (under `/metrics`) on WildFly management HTTP interface (on port `9990`).
+* Support both JSON and https://prometheus.io/docs/instrumenting/exposition_formats/#text-format-details[Prometheus format] from the Metrics HTTP endpoints
```
rsvoboda: Maybe stress out that Prometheus format is the default ?

microprofile/WFLY-10712_microprofile_metrics.asciidoc
```diff
@@ -0,0 +1,230 @@
...
+** Current and peak usages of the memory pools
+
+WildFly management model has a notion of metrics for resource attributes (as described in http://docs.wildfly.org/14/Admin_Guide.html#Description_of_the_Management_Model[WildFly Admin Guide]).
+This extension MUST be able to expose these metrics from WildFly management model, such as:
+
+* Transaction metrics (from the http://wildscribe.github.io/WildFly/14.0/subsystem/transactions/index.html[transactions subsystem])
+* HTTP usage (from the http://wildscribe.github.io/WildFly/14.0/subsystem/undertow/server/http-listener/index.html[undertow subsystem])
+* JDBC Pool usage (from the http://wildscribe.github.io/WildFly/14.0/subsystem/datasources/data-source/ExampleDS/statistics/pool/index.html[datasources subsystem])
+
+Application metrics from deployments are also supported.
```
rsvoboda: MP Metrics - scope application ?

microprofile/WFLY-10712_microprofile_metrics.asciidoc
```diff
@@ -0,0 +1,230 @@
...
+
+=== Non-Requirements
+
+* The MicroProfile 1.1.1 does not support multiple deployments. It is out of scope of this feature to support it
+in WildFly. Work will have to be done in a future version of the specification to support it. This adds constraints
+on the application deployments that must coordinate to avoid exposing the same metrics (especially absolute ones whose name
+can clash between deployments).
```
rsvoboda: Deployment name as a prefix for these application based metrics ?
e.g. stock-service-war.foo.bar=baz

microprofile/WFLY-10712_microprofile_metrics.asciidoc
```diff
@@ -0,0 +1,230 @@
...
+
+== Implementation Plan
+
+* Add dependencies to `smallrye-metrics` and `microprofile-metrics-api` artifacts
+* Add a new `microprofile-metrics-smallrye` extension to provide Metrics support (including its HTTP endpoints)
+  * Expose Metrics HTTP endpoints on WildFly HTTP management interface under the `/metrics` context.
+* Add mechanism to register metrics from WildFly management model into the `MetricRegistry` for vendor metrics.
+
```
rsvoboda: HTTP Verbs: GET, OPTIONS
In spec it's 2.3. Exposing metrics via REST API

we should also disallow other HTTP verbs

microprofile/WFLY-10712_microprofile_metrics.asciidoc
```diff
@@ -0,0 +1,230 @@
...
+# HELP base:cpu_system_load_average Displays the system load average for the last minute. The system load average is the sum of the number of runnable entities queued to the available processors and the number of runnable entities running on the available processors averaged over a period of time. The way in which the load average is calculated is operating system specific but is typically a damped time-dependent average. If the load average is not available, a negative value is displayed. This attribute is designed to provide a hint about the system load and may be queried frequently. The load average may be unavailable on some platform where it is expensive to implement this method.
+# TYPE base:cpu_system_load_average gauge
+base:cpu_system_load_average 1.9658203125
+...
+----
+
+If security is disabled in the subsystem configuration, the HTTP client does not require authentication (and it does not require a WildFly management user either).
+
+Note that the _standalone profiles will explicitly disable authentication_ as the HTTP management interface is bound to `127.0.0.1`.
+
```
rsvoboda: something special for https ?

microprofile/WFLY-10712_microprofile_metrics.asciidoc
```diff
@@ -0,0 +1,230 @@
...
+----
+
+For example, the properties to expose the _Total Loaded Class Count_ of the JVM are:
+
+----
+classloader.totalLoadedClass.count.displayName: Total Loaded Class Count
+classloader.totalLoadedClass.count.type: counter
+classloader.totalLoadedClass.count.unit: none
+classloader.totalLoadedClass.count.description: Displays the total number of classes that have been loaded since the Java virtual machine has started execution.
+classloader.totalLoadedClass.count.mbean: java.lang:type=ClassLoading/TotalLoadedClassCount
```
rsvoboda: uff, now I'm confused when `:` and `.` should be used as separators
e.g. this vs. base:cpu_system_load_average

microprofile/WFLY-10712_microprofile_metrics.asciidoc
jmesnil: I listed both OpenShift Console and Kiali in the interested projects but I am still figuring out which does what and which will actually consume metrics exposed by WildFly

microprofile/WFLY-10712_microprofile_metrics.asciidoc
jmesnil: https://issues.jboss.org/browse/EAP7-1060 I'll add it to the issue metadata

jmesnil: it's already in the proposal :)

microprofile/WFLY-10712_microprofile_metrics.asciidoc
jmesnil: Right. I need to check the limitations we will face with MP Metrics (that does not support multiple deployment afaict).

microprofile/WFLY-10712_microprofile_metrics.asciidoc
rsvoboda: OK, probably not worth extra library for WF/Thornlail usage

microprofile/WFLY-10712_microprofile_metrics.asciidoc
pilhuhn: We will not do a 1.2, but rather jump to 2.0 with many changes and also support for multiple applications per server. This is for the ~Feb 2019 release of MicroProfile.

microprofile/WFLY-10712_microprofile_metrics.asciidoc
```diff
@@ -0,0 +1,230 @@
...
+
+* Transaction metrics (from the http://wildscribe.github.io/WildFly/14.0/subsystem/transactions/index.html[transactions subsystem])
+* HTTP usage (from the http://wildscribe.github.io/WildFly/14.0/subsystem/undertow/server/http-listener/index.html[undertow subsystem])
+* JDBC Pool usage (from the http://wildscribe.github.io/WildFly/14.0/subsystem/datasources/data-source/ExampleDS/statistics/pool/index.html[datasources subsystem])
+
+Application metrics from deployments are also supported.
+
+=== Non-Requirements
+
+* The MicroProfile 1.1.1 does not support multiple deployments. It is out of scope of this feature to support it
```
rsvoboda: so the initial impl will support one deployment?
Or also multiple deployments where the exposed matrices are unique ?

What would happen is there are 2 deployments exposing matrix with the same name? Error, warning, silence, random pick ?

microprofile/WFLY-10712_microprofile_metrics.asciidoc
jmesnil: that's already handled by smallrye-metrics: https://github.com/smallrye/smallrye-metrics/blob/1.1.1/implementation/src/main/java/io/smallrye/metrics/MetricsRequestHandler.java#L152

Allowed HTTP verbs:
* `GET`, `OPTIONS` for JSON format
* Only `Get` for Prometheus format

jmesnil: that's already handled by smallrye-metrics: https://github.com/smallrye/smallrye-metrics/blob/1.1.1/implementation/src/main/java/io/smallrye/metrics/MetricsRequestHandler.java#L152

Allowed HTTP verbs:
* `GET`, `OPTIONS` for JSON format
* Only `Get` for Prometheus format

microprofile/WFLY-10712_microprofile_metrics.asciidoc
jmesnil: I mention that in the "HTTP Endpoint formats" section below

jmesnil: I mention that in the "HTTP Endpoint formats" section below

jmesnil: I mention that in the "HTTP Endpoint formats" section below

jmesnil: I mention that in the "HTTP Endpoint formats" section below

jmesnil: I mention that in the "HTTP Endpoint formats" section below

microprofile/WFLY-10712_microprofile_metrics.asciidoc
jmesnil: It's handled by smallrye-metrics: https://github.com/smallrye/smallrye-metrics/blob/1.1.1/implementation/src/main/java/io/smallrye/metrics/MetricsRequestHandler.java#L152

microprofile/WFLY-10712_microprofile_metrics.asciidoc
jmesnil: No. 
The extension requires only the http management interface and it is not possible to bind it to the https management interface (that's the design already chosen for MicroProfile Health Check)

jmesnil: No. 
The extension requires only the http management interface and it is not possible to bind it to the https management interface (that's the design already chosen for MicroProfile Health Check)

microprofile/WFLY-10712_microprofile_metrics.asciidoc
jmesnil: that's mentioned in the "HTTP Endpoint formats" section below

microprofile/WFLY-10712_microprofile_metrics.asciidoc
```diff
@@ -0,0 +1,304 @@
...
+
+Note that the _standalone profiles will explicitly disable authentication_ as the HTTP management interface is bound to `127.0.0.1`.
+
+==== Exposed Metrics
+
+===== Base Metrics
+
+By default, the subsystem will expose all required base metrics specified in the chapter 4 of the MicroProfile Metrics specification.
+All these base metrics are gathered from the JVM MBeans and uses smallrye-metrics JmxRegistrar to bridge from JMX to the MicroProfile Metrics API.
+The required base metrics are explicitly specified in a property file (named `base-metrics.properties`) using a set of properties for each metric:
```
bstansberry: I'm unclear on the significance of this part. Is this just an explanation of an internal implementation detail? Or is it an explanation of a configuration mechanism?

microprofile/WFLY-10712_microprofile_metrics.asciidoc
```diff
@@ -0,0 +1,304 @@
...
+
+Examples of such metrics are:
+
+* Number of modules loaded by JBoss Module
+* Transaction statistics from Narayana Transaction Manager
+* Bytes throughput from Undertow
+* Etc.
+
+Some of these metrics can be obtained by JMX and can rely on smallrye-metrics that loads these metrics from a property file named `vendor-metrics.properties`
+that works similarly to the `base-metrics.properties` as explained in the section above.
```
bstansberry: Same question my one above about base-metrics.properties.

Overall we need to work through the requirements re: configuration.

microprofile/WFLY-10712_microprofile_metrics.asciidoc
bstansberry: As work on this proceeds the discussion of what console to support needs to shift to focus on explicit technology requirements, with particular consumers just being examples of consumers who consume whatever way is spelled out. So how the WF server will be discovered, what protocol, what authentication etc. That's already touched on here to a large degree.

microprofile/WFLY-10712_microprofile_metrics.asciidoc
jmesnil: You're right, I've removed the justification being that the management interface is bound to `127.0.0.1`.

microprofile/WFLY-10712_microprofile_metrics.asciidoc
```diff
@@ -0,0 +1,365 @@
...
+To solve this, the `:read-attribute` operation used to fetch the metric value will be enhanced with a "include-undefined-metric".
+By default, this flag is `false`. If it is true, the `:read-attribute` operation handler will not use the "undefined metric value" if the metric read handler returns an undefined value.
+The `micrprofile-smallrye-metrics` will set this flag to `false` and remove the metric from the list returned by its endpoints.
+
+This will requires additional fixes in the various metric handler that *must not* return a defined value (e.g. `0`) if the metric can not be computed.
+
+====== Management Resources Added After Server Boot Will Not Expose Their Metrics
+
+The `microprofile-smallrye-metrics` extension will register any valid metric from WildFly Management Model when it is installed.
+However if other management resources are added afterwards, the extension will not be aware of them and will not register their metrics.
```
bstansberry: FWIW, this does not concern me at all.

It's good to note it though.

microprofile/WFLY-10712_microprofile_metrics.asciidoc
```diff
@@ -0,0 +1,365 @@
...
+** `io` - 5 metrics (per worker)
+** `jca` - 8 metrics (per workmanager)
+** `request-controller` - 1 metric (overall)
+** `transactions` - 11 metrics (overall)
+** `undertow` - 12 metrics (per http/https listener)
+* Standalone Full Profile (10 additional metrics):
+** `messaging-activemq` - 5 metrics (per destinations)
+* Standalone Full HA Profile (186 additional metrics):
+** `undertow` - 6 metrics (for 1 additional `ajp-listener`)
+** `jgroups` - 180 metrics (for 1 ee channel)
```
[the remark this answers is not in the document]
jmesnil: I've updated the proposal with a `exposed-subsystems` to white-list the subsystems that expose their metrics

microprofile/WFLY-10712_microprofile_metrics.asciidoc
jmesnil: I've updated the spec to mention the names of the metrics exposed by the subsystem in the base and vendor spec.

microprofile/WFLY-10712_microprofile_metrics.asciidoc
```diff
@@ -267,8 +281,10 @@ For example the transaction metrics will be retrieved from the http://wildscribe
 * number-of-aborted-transactions
 * etc.
 
-When the `micrprofile-smallrye-metrics` is installed, it will browse WildFly Resource Model and find every metrics registered by resources from the model.
+When the `micrprofile-smallrye-metrics` is installed, it will browse WildFly Resource Model and find every metrics registered by subsystem resources from the model.
+If will only expose metrics from the subsystems specified by the `exposed-subsystems` attribute (or all if the wildcard `*` is used).
```
rsvoboda: Probably typo: If -> It

microprofile/WFLY-10712_microprofile_metrics.asciidoc
```diff
@@ -267,8 +281,10 @@ For example the transaction metrics will be retrieved from the http://wildscribe
...
 All those WildFly metrics will be translated to MicroProfile Metrics (with corresponding metadata) and registered in the MicrProfile Metrics' Vendor registry.
+Note that metrics from other parts of WildFly resource model (e.g. below `/core-service` resources are __not__ registered).
```
rsvoboda: (e.g. below `/core-service` resources) are __not__ registered.

microprofile/WFLY-10712_microprofile_metrics.asciidoc
bstansberry: How will this be utilized in our OpenShift images?  Will this form of configurability allow us to expose the desired metrics without exposing way more than is desired?

This style may be ok. Looking at the metrics declared in the hawkular agent config that this subsystem will replace, I don't see JGroups metrics, which AIUI are the problematic ones. There are some runtime attributes mentioned from JGroups, but not metrics.

https://github.com/jboss-openshift/cct_module/blob/master/os-eap70-hawkular/added/hawkular-javaagent-config.yaml

microprofile/WFLY-10712_microprofile_metrics.asciidoc
```diff
@@ -0,0 +1,407 @@
...
+====== Complex WildFly Metrics Are Not Supported
+
+Resources may return arbitrarily complex return types for metrics. Most complex example I found in the codebase is http://wildscribe.github.io/WildFly/14.0/core-service/platform-mbean/type/memory/index.html#attr-heap-memory-usage[ core-service=platform-mbean
+type=memory#heap-memory-usage].
+The registration code in `micrprofile-smallrye-metrics` will only register MicroProfile Metrics for simple numerical `ModelType`.
+
+====== WildFly Metrics Are Registered As Gauge
+
+As there is no way to determine whether a WildFly metric corresponds to a https://github.com/eclipse/microprofile-metrics/blob/1.1.1/api/src/main/java/org/eclipse/microprofile/metrics/Gauge.java[Gauge] or a https://github.com/eclipse/microprofile-metrics/blob/1.1.1/api/src/main/java/org/eclipse/microprofile/metrics/Counter.java[Counter], they
+will all be registered in the MicroProfile vendor registry as *Gauge*.
```
bstansberry: Perhaps as a future enhancement an arbitrary attribute descriptor can be used to allow metrics to be classified, defaulting to 'gauge'.

I'm referring to "Arbitrary Descriptors" in http://docs.wildfly.org/14/Admin_Guide.html#Description_of_the_Management_Model

PR #148, opened 2018-10-23
[WFCORE-4172] Add support for TLS 1.3

fjuma:
https://issues.jboss.org/browse/WFCORE-4172

### DIFF
--- a/elytron/WFCORE-4172-tls-1.3.adoc
+++ b/elytron/WFCORE-4172-tls-1.3.adoc
@@ -1,0 +1,116 @@
+= WFCORE-4172 Add support for TLS 1.3
+:author:            Farah Juma
+:email:             fjuma@redhat.com
+:toc:               left
+:icons:             font
+:keywords:          comma,separated,tags
+:idprefix:
+:idseparator:       -
+:issue-base-url:    https://issues.jboss.org/browse
+
+== Overview
+
+JDK 11 has added support for TLS 1.3. This task is to update Elytron to be able to make use of TLS 1.3. This RFE assumes
+that the TLS 1.3 protocol is provided by the default JSSE provider.
+
+== Issue Metadata
+
+=== Issue
+
+* https://issues.jboss.org/browse/WFCORE-4172[WFCORE-4172]
+
+=== Related Issues
+
+* https://issues.jboss.org/browse/EAP7-1022[EAP7-1022]
+
+=== Dev Contacts
+
+* mailto:{email}[{author}]
+
+=== QE Contacts
+
+* mailto:okotek@redhat.com[Ondrej Kotek]
+
+=== Affected Projects or Components
+
+* WildFly, Security
+
+=== Other Interested Projects
+
+== Requirements
+
+=== Hard Requirements
+
+* It should be possible to configure a ```server-ssl-context``` and ```client-ssl-context``` in the Elytron subsystem
+using the TLS 1.3 protocol and TLS 1.3 https://tools.ietf.org/html/rfc8446#appendix-B.4[cipher suites]. (Note
+that TLS 1.2 cipher suites cannot be used with TLS 1.3 and vice and versa.)
+
+** A new attribute ```cipher-suite-names``` will be added to both ```server-ssl-context``` and ```client-ssl-context```
+   to specify the TLS 1.3 cipher suites that should be enabled. The format will be a colon (:) separated list of
+   cipher suite names (similar to the new OpenSSL-style TLS 1.3 ```ciphersuites``` option).
+
+** The default value for the ```cipher-suite-names``` attribute will be the same as the OpenSSL default cipher suites for
+   TLS 1.3:
+   ```"TLS_AES_256_GCM_SHA384:TLS_CHACHA20_POLY1305_SHA256:TLS_AES_128_GCM_SHA256"```
+
+** Example configuration:
+
+[source,xml]
+----
+   <server-ssl-context name="myTLS13ServerContext" ... cipher-suite-names="TLS_AES_128_CCM_8_SHA256:TLS_AES_256_GCM_SHA384"/>
+----
+[source,xml]
+----
+   <client-ssl-context name="myTLS13ClientContext" ... cipher-suite-names="TLS_AES_128_CCM_8_SHA256:TLS_AES_256_GCM_SHA384:TLS_CHACHA20_POLY1305_SHA256:TLS_AES_128_GCM_SHA256"/>
+----
+
+* It should be possible to configure subsystems that reference an ```ssl-context``` to make use of this TLS 1.3 ```server-ssl-context```.
+
+* It should be possible to configure an ```ssl-context``` using the Elytron authentication client using the TLS 1.3
+protocol and TLS 1.3 https://tools.ietf.org/html/rfc8446#appendix-B.4[cipher suites].
+
+** A new attribute ```names``` will be added to the ```cipher-suite``` element to specify the TLS 1.3 cipher suites that
+   should be enabled. The format will be a colon (:) separated list of cipher suite names (similar to the new OpenSSL-style
+   TLS 1.3 ```ciphersuites``` option).
+** The default value for the ```cipher-suite-names``` option will be the same as the OpenSSL default cipher suites for
+   TLS 1.3:
+   ```"TLS_AES_256_GCM_SHA384:TLS_CHACHA20_POLY1305_SHA256:TLS_AES_128_GCM_SHA256"```
+
+** Example configuration:
+[source,xml]
+----
+   <ssl-context name="myTLS13Context">
+   ...
+     <cipher-suite names="TLS_AES_128_CCM_8_SHA256:TLS_AES_256_GCM_SHA384"/>
+   ...
+   </ssl-context>
+----
+
+*
+
+=== Nice-to-Have Requirements
+
+=== Non-Requirements
+
+* The ability to use TLS providers other than the default JSSE provider is a non-requirement.
+
+== Test Plan
+
+All testing should be done using the default JSSE provider. The TLS 1.3 cipher suites that are actually available will depend
+on this provider.
+
+One-way and two-way SSL tests will be added that make use of the TLS 1.3 protocol and TLS 1.3 cipher suites.
+The ```java.specification.version``` system property will be used to make sure that these new tests only get run
+when JDK 11 is in use.
+
+One thing that is important to note is that TLS 1.3 has made session IDs essentially
+http://openjdk.5641.n7.nabble.com/SSLSocket-session-resumption-not-working-with-TLS-1-3-and-11-27-td347431.html#a347563[obsolete].
+This means that the session ID can no longer reliably be used to test if a session was resumed. Instead, creation time
+can be used to test if a session was resumed. Currently, clients need to read from server after the first handshake in
+order to receive the ```NewSessionTicket``` that can be used for resumption. However, this may change once
+https://bugs.openjdk.java.net/browse/JDK-8209953[JDK-8209953] is resolved.
+
+== Community Documentation
+
+Documentation will be added in the "Using the Elytron Subsystem" section in the WildFly documentation to indicate
+that the TLS 1.3 protocol can be used with JDK 11. The documentation should mention the limitations around session resumption.
--- a/elytron/WFCORE-4172-tls-1.3.adoc
+++ b/elytron/WFCORE-4172-tls-1.3.adoc
@@ -49,9 +49,12 @@ that TLS 1.2 cipher suites cannot be used with TLS 1.3 and vice and versa.)
    to specify the TLS 1.3 cipher suites that should be enabled. The format will be a colon (:) separated list of
    cipher suite names (similar to the new OpenSSL-style TLS 1.3 ```ciphersuites``` option).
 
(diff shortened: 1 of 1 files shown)


elytron/WFCORE-4172-tls-1.3.adoc
```diff
@@ -0,0 +1,75 @@
...
+One-way and two-way SSL tests will be added that make use of the TLS 1.3 protocol and TLS 1.3 cipher suites.
+The ```java.specification.version``` system property will be used to make sure that these new tests only get run
+when JDK 11 is in use.
+
+One thing that is important to note is that TLS 1.3 has made session IDs essentially
+http://openjdk.5641.n7.nabble.com/SSLSocket-session-resumption-not-working-with-TLS-1-3-and-11-27-td347431.html#a347563[obsolete].
+This means that the session ID can no longer reliably be used to test if a session was resumed. Instead, creation time
+can be used to test if a session was resumed. Currently, clients need to read from server after the first handshake in
+order to receive the ```NewSessionTicket``` that can be used for resumption. However, this may change once
+https://bugs.openjdk.java.net/browse/[JDK-8209953] is resolved.
```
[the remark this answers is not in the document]
fjuma: Currently, they can no longer be reliably used.

OndrejKotek:
```suggestion
https://bugs.openjdk.java.net/browse/JDK-8209953[JDK-8209953] is resolved.
```

elytron/WFCORE-4172-tls-1.3.adoc
```diff
@@ -0,0 +1,75 @@
...
+
+* It should be possible to configure an ```ssl-context``` using the Elytron authentication client using the TLS 1.3
+protocol and TLS 1.3 https://tools.ietf.org/html/rfc8446#appendix-B.4[cipher suites].
+
+* The default cipher suites for TLS 1.3 should be the same as the openssl default cipher suites for TLS 1.3
+(TLS_AES_256_GCM_SHA384, TLS_CHACHA20_POLY1305_SHA256, TLS_AES_128_GCM_SHA256).
+
+=== Nice-to-Have Requirements
+
+=== Non-Requirements
```
OndrejKotek: Do I understand it well that this RFE is related just to the Elytron and no implementation for the legacy security will be provided?

elytron/WFCORE-4172-tls-1.3.adoc
fjuma: Fixed the link.

elytron/WFCORE-4172-tls-1.3.adoc
fjuma: Correct, this RFE is only related to Elytron.

elytron/WFCORE-4172-tls-1.3.adoc
```diff
@@ -0,0 +1,75 @@
...
+
+== Test Plan
+
+One-way and two-way SSL tests will be added that make use of the TLS 1.3 protocol and TLS 1.3 cipher suites.
+The ```java.specification.version``` system property will be used to make sure that these new tests only get run
+when JDK 11 is in use.
+
+One thing that is important to note is that TLS 1.3 has made session IDs essentially
+http://openjdk.5641.n7.nabble.com/SSLSocket-session-resumption-not-working-with-TLS-1-3-and-11-27-td347431.html#a347563[obsolete].
+This means that the session ID can no longer reliably be used to test if a session was resumed. Instead, creation time
```
OndrejKotek: Could this affect any parts where SSLSession is used? For example in CLIENT_CERT SSO, CachedIdentity objects are stored in SSLSession.

fjuma: If the identity cache is retrieved after attempting to resume an SSL session, then any CachedIdentity objects that were stored on the SSL session before resumption won't be available and so authentication would just get attempted again.

OndrejKotek:
OK for QE.

elytron/WFCORE-4172-tls-1.3.adoc
```diff
@@ -0,0 +1,75 @@
...
+* It should be possible to configure a ```server-ssl-context``` and ```client-ssl-context``` in the Elytron subsystem
+using the TLS 1.3 protocol and TLS 1.3 https://tools.ietf.org/html/rfc8446#appendix-B.4[cipher suites]. (Note
+that TLS 1.2 cipher suites cannot be used with TLS 1.3 and vice and versa.)
+
+* It should be possible to configure subsystems that reference an ```ssl-context``` to make use of this TLS 1.3 ```server-ssl-context```.
+
+* It should be possible to configure an ```ssl-context``` using the Elytron authentication client using the TLS 1.3
+protocol and TLS 1.3 https://tools.ietf.org/html/rfc8446#appendix-B.4[cipher suites].
+
+* The default cipher suites for TLS 1.3 should be the same as the openssl default cipher suites for TLS 1.3
```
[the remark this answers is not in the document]
fjuma: I've added more details about the TLS 1.3 default cipher suites and how the TLS 1.3 cipher suites are configured.

elytron/WFCORE-4172-tls-1.3.adoc
```diff
@@ -0,0 +1,108 @@
...
+
+== Requirements
+
+=== Hard Requirements
+
+* It should be possible to configure a ```server-ssl-context``` and ```client-ssl-context``` in the Elytron subsystem
+using the TLS 1.3 protocol and TLS 1.3 https://tools.ietf.org/html/rfc8446#appendix-B.4[cipher suites]. (Note
+that TLS 1.2 cipher suites cannot be used with TLS 1.3 and vice and versa.)
+
+** A new attribute ```cipher-suite-names``` will be added to both ```server-ssl-context``` and ```client-ssl-context```
```
mchoma: Why is necessary to add new attribute? There is already ```cipher-suite-filter``` attribute on ```*-ssl-context``` which has similar meaning?

fjuma: This is similar to the approach OpenSSL has taken. Because ciphersuites work differently in TLS 1.3, they introduced a new option to specify them and they no longer allow the special characters that were allowed for specifying TLS 1.2 ciphersuites (""+", "!", "-", etc.) for TLS 1.3. See https://wiki.openssl.org/index.php/TLS1.3#Ciphersuites for more details on the separate options.

mchoma: I must admit I don't fully get why they introduced distinct attribute for similar thing. Question is if we should really repeat they decision. 

From user perspective it become confusing. Is there any technical limitation (API requirement) I am missing which prevents of using one attribute? Is it because of scenario when I specify TLSv1.2 in conjuction with TLSv1.3, but cipher suites are disjunct?

What about the future when there will be another TLSv1.3 cipher suites and we will begin to need selector just as on old attribute cipher-suite-filter?

Maybe their parsing logic does not allow them to reuse one attribute, but it seems to me in Elytron case this filter is just to prepare string which will be passed to jdk. I think selector in Elytron is written generic enough to work with TLSv1.3 cipher suites just fine.

@dmlloyd what do you think?

fjuma: The format and parsing for our ```cipher-suite-filter``` attribute is based on the OpenSSL-style pre TLS 1.3 format (see ```CipherSuiteSelector#fromString```). Having two separate attributes allows appropriate default values to be set (e.g., If ```cipher-suite-filter="ECDHE"``` and ```cipher-suite-names``` is left unspecified, any pre-TLS 1.3 cipher suites that support ECDHE will be considered as well as the default cipher suites for TLS 1.3. With only one attribute, we wouldn't know to additionally include the default cipher suites for TLS 1.3.).

Another thing to note is that the format of the cipher suite names has changed significantly. Instead of a cipher suite name being made up of a protocol, key exchange type, authentication type, AEAD cipher mode, and PRF hash algorithm, a TLS 1.3 cipher suite name is now made up of just the protocol, AEAD cipher mode, and HKDF hash algorithm. So some keywords that were previously used to match pre-TLS 1.3 cipher suite names no longer make sense for TLS 1.3.

Also since there are only 5 cipher suites for TLS 1.3 and since TLS 1.2 cipher suites cannot be used with TLS 1.3 and vice versa, I think it makes sense to have the additional attribute. A user likely would not even need to set this attribute explicitly since it already defaults to 3 out of the 5 TLS 1.3 cipher suites.

elytron/WFCORE-4172-tls-1.3.adoc
```diff
@@ -0,0 +1,108 @@
...
+   to specify the TLS 1.3 cipher suites that should be enabled. The format will be a colon (:) separated list of
```
mchoma: In ```cipher-suite-filter``` separator is ```:``` as well as ```,``` . Our documentation use mainly ```,```

elytron/WFCORE-4172-tls-1.3.adoc
OndrejKotek: From user perspective, having two attributes is confusing. What about an alternative approach with just one attribute? The TLS 1.3 cipher suites could be part of the original mechanism database (the format is the same). Any special treatment with a TLS 1.3 cipher suite can be enabled by the suite affiliation with the protocol (which is set in the database) and by using new TLS 1.3 related keywords. For example `cipher-suite-filter="TLS1.3_DEFAULT:ECDHE"` would pick the 3 default TLS 1.3 cipher suites and any pre-TLS 1.3 cipher suites that support ECDHE. Are there any obstacles/disadvantages?

@fjuma @darranl WDYT, please?

fjuma: @OndrejKotek One thing to note is that the keywords we currently use in our ```cipher-suite-filter``` are OpenSSL keywords. So introducing new keywords like ```TLS1.3_DEFAULT``` may be confusing. Another thing to note is that for ```cipher-suite-filter``` configuration that doesn't match the TLS 1.3 cipher suites, TLS 1.3 would not get automatically used by default if available without explicitly specifying a special keyword as you mentioned.

dmlloyd: I think mixing TLS 1.2 and 1.3 selectors might get very confusing given how different they are.  I agree with Farah and the OpenSSL authors that it is best to keep them separate for now.

darranl: I think the second attribute is going to make sense and be mutually exclusive with the original attribute, by the sounds of it once we are ready to push all users to use TLS 1.3 it sounds like the original attribute would end up being deprecated.

The problem we have here is the two attributes have completely different formats, this really means tooling will need to treat each of them differently and each will be documented differently - validation for each would also be different.

If these two attributes were combined into one we don't have any way to detect the format in use to validate and make use of the value, we can't even rely on the protocol being specified on the SSLContext as alternative implementations may use different names.

I can see where the comments are coming from related to having two related attributes but in this case it does seem it will be a necessity.

OndrejKotek: Thank you for your clarifications and confirmations. I have no more suggestions or objections.

elytron/WFCORE-4172-tls-1.3.adoc
jstourac: @fjuma, apologies, maybe you already discussed that with Martin or Adam, but doesn't it make sense to use both `:` and `,` separators for suites separation in newly introduced `cipher-suite-names` to conform current form in already existing `cipher-suite-filter` attribute? This makes it a little bit user-friendly, though.

dmlloyd: Using just `:` appears to be consistent with OpenSSL: https://wiki.openssl.org/index.php/TLS1.3#Ciphersuites

fjuma: @jstourac As David mentioned, allowing only ```:``` is consistent with the format OpenSSL allows for TLS 1.3 cipher suites.

jstourac: @fjuma, @dmlloyd Thank you. Yes, I know it is consistent with OpenSSL, it is also mentioned in the analysis text. I think that OpenSSL always had just `:` as a separator for cipher-suites even before TSL 1.3 ciphers. But we have always allowed both `:` and `,` in `cipher-suite-filter` although only `:` in newly introduced attribute... which is simply weird... although, if this is the desired approach, then okay. Doc update should be sufficient.

fjuma: @jstourac Prior to TLS 1.3, OpenSSL allowed commas in cipher suite names:

> The cipher list consists of one or more cipher strings separated by colons. Commas or spaces are also acceptable separators but colons are normally used.

Our documentation for the pre-TLS 1.3 ```cipher-suite-filter``` which allows OpenSSL-style cipher list strings is as follows:

> Each item is separated from the other items by a colon (":"), though for compatibility, commas (",") or spaces (" ") are allowed delimiters as well.


For the TLS 1.3 ```cipher-suite-names``` attribute, I think we should stick to ```:``` only. This doesn't affect the pre-TLS 1.3 ```cipher-suite-filter``` attribute.

jstourac: @fjuma ah, sorry I missed that bit. Now I see it in the [manpage](https://www.openssl.org/docs/man1.1.1/man1/ciphers.html) too. Thanks for clarification on this :)

PR #157, opened 2018-12-05
EAP7-1156 Display open network ports for a managed server

claudio4j:
https://issues.jboss.org/browse/EAP7-1156

### DIFF
--- a/console/EAP7-1156_Display_open_network_ports_for_a_managed_server.adoc
+++ b/console/EAP7-1156_Display_open_network_ports_for_a_managed_server.adoc
@@ -1,0 +1,49 @@
+= EAP7-1156 Display open network ports for a managed server
+:author:            Claudio Miranda
+:email:             claudio@redhat.com
+:toc:               left
+:icons:             font
+:idprefix:
+:idseparator:       -
+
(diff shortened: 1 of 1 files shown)


console/EAP7-1156_Display_open_network_ports_for_a_managed_server.adoc
```diff
@@ -0,0 +1,49 @@
...
+
+* Show all bound ports of each socket-binding and its name.
+* Show them in the server preview pane.
+* Show the open ports when the server is in the following status: running, reload-required, restart-required
+* Do not show while the server is starting, stopping or in suspended mode.
+* Should work in domain and standalone mode.
+
+=== Non-Requirements
+
+There are services that doesn't register themselves with the socket-binding, as a consequence the socket is open but it is not displayed as bound, they will not be displayed in the "Opent Ports" preview. Currently `modcluster` is one service that doesn't register the socket-binding.
```
bstansberry: If a consumer of a socket-binding doesn't record that it has bound the socket, that's a bug. I agree though that it should not be a requirement of the implementors of this RFE to correct such bugs.

PR #159, opened 2018-12-10
[WFCORE-4251] List modules on deployments classpath via a Management Operation

yersan:
EAP issue: https://issues.jboss.org/browse/EAP7-521
Upstream issue: https://issues.jboss.org/browse/WFCORE-4251

### DIFF
--- a/server/WFCORE-4251_list_modules_in_deployment_classpath.adoc
+++ b/server/WFCORE-4251_list_modules_in_deployment_classpath.adoc
@@ -1,0 +1,118 @@
+= Provide ability to list modules which are on deployment's classpath via a Management Operation.
+:author:            Yeray Borges
+:email:             yborgess@redhat.com
+:toc:               left
+:icons:             font
+:idprefix:
+:idseparator:       -
+:issue-base-url:    https://issues.jboss.org/browse/
+
+== Overview
+Currently, the list of modules available at classpath by each deployment is only available via jconsole:
+
+jboss.modules -> ModuleLoader -> ServiceModuleLoader -> Operations -> getDependencies( "deployment.<application_war_ear_name>" )
+
+This document describes the RFE to provide the ability to expose this information via a management operation.
+
+When optimizing the application footprint, this information becomes important:
+
+* The user could use this list to identify which dependencies are added implicitly by the server, and then, exclude any unneeded dependency from the deployment.
+* The user could use this information to discover any duplicated artifact bundled together with the deployment and included in one of the modules that the server is adding to the application classpath.
+* This information could facilitate discovering unneeded subsystems in the installation of his servers.
+
+
+The proposed solution is to include a new operation `list-modules` available under `/deployment` and, in case of ears, `/deployment=*/subdeployment` resource:
+
+ [standalone@localhost:9990 /] /deployment=ejb-in-ear.ear:list-modules
+
+ [standalone@localhost:9990 /] /deployment=ejb-in-ear.ear/subdeployment=ejb-in-ear-web.war:list-modules
+
+An example of the information returned by this method could be:
+
+  {
+      "outcome" => "success",
+      "result" => {
+          "ejb-in-ear.ear" => [
+              {
+                  "name" => "com.fasterxml.jackson.datatype.jackson-datatype-jdk8",
+                  "slot" => "main",
+                  "optional" => true,
+                  "export" => false,
+                  "import-services" => true,
+                  "duplicated" => false
+              },
+              {
+                  "name" => "com.fasterxml.jackson.datatype.jackson-datatype-jsr310",
+                  "slot" => "main",
+                  "optional" => true,
+                  "export" => false,
+                  "import-services" => true,
+                  "duplicated" => true
+              },
+              ...
+          ]
+      }
+  }
+
+Where:
+
+* name: The module name.
+* slot: The slot where this module is.
+* optional: If the dependency was added as an optional dependency.
+* export: If the dependency is being exported to other modules.
+* import-services: If this module is importing services from other dependencies.
+* duplicated: If it is a module defined by the user and the server is adding a module implicitly with the same name.
+
+
+== Issue Metadata
+
+=== Issue
+
+* {issue-base-url}WFCORE-4251[WFCORE-4251]
+
+=== Related Issues
+
+* {issue-base-url}EAP7-521[EAP7-521]
+
+=== Dev Contacts
+
+* mailto:{email}[{author}]
+
+=== QE Contacts
+
+* mailto:[TBD]
+
+=== Affected Projects or Components
+
+* Wildfly Core
+* JBoss Modules (optional, it could help if we finally show the artifacts included on each module, see Nice-to-Have requirements)
+
+=== Other Interested Projects
+
+== Requirements
+
+=== Hard Requirements
+
+* It must be possible to list the implicit modules added to the classpath for a specific deployment and, in case of ears, for a particular sub-deployment.
+* List of modules should be sorted alphabetically by module name.
+* The list of modules will be only available if the deployment is enabled.
+* The new operation will be available only in standalone mode. Domain mode changes are not going to be covered by this RFE.
+* The output of the operation will include an attribute named `duplicated` which will be true if the module was added by the user in the deployment, via manifest file or deployment-structure.xml, and added implicitly by the server.
+
+
+=== Nice-to-Have Requirements
+
+* The possibility to show together with the module name the artifact name(s), if there is one, that the module is exposing.
+** This possibility could facilitate to the user the task to check which specific artifact(s) and version is added by a module in order to find any duplicate jar bundled by its own application.
+
+
+=== Non-Requirements
+
+== Test Plan
+
+A test case deploying an application and getting the expected list of modules using the new operation.
+The application should include variants, adding and excluding modules or subsystems via in the deployment deployment-structure.xml file.
+
+== Community Documentation
+
+This has to be implemented in wildfly-core. So the documentation will have to be added as a follow up to wildfly.
--- a/server/WFCORE-4251_list_modules_in_deployment_classpath.adoc
+++ b/server/WFCORE-4251_list_modules_in_deployment_classpath.adoc
@@ -27,34 +27,74 @@ The proposed solution is to include a new operation `list-modules` available und
 
  [standalone@localhost:9990 /] /deployment=ejb-in-ear.ear/subdeployment=ejb-in-ear-web.war:list-modules
 
+In domain mode this operation will be available via server resource at host level:
+
+ [domain@localhost:9990 /] /host=master/server=server-one/deployment=ejb-in-ear.ear:list-modules
+
+ [domain@localhost:9990 /] /host=master/server=server-one//deployment=ejb-in-ear.ear/subdeployment=ejb-in-ear-web.war:list-modules
+
 An example of the information returned by this method could be:
 
+  [standalone@localhost:9990 /] /deployment=ejb-in-ear.ear:list-modules
   {
       "outcome" => "success",
       "result" => {
-          "ejb-in-ear.ear" => [
+          "system-dependencies" => [
               {
                   "name" => "com.fasterxml.jackson.datatype.jackson-datatype-jdk8",
                   "slot" => "main",
                   "optional" => true,
                   "export" => false,
                   "import-services" => true,
-                  "duplicated" => false
+                  "duplicated" => true
               },
               {
-                  "name" => "com.fasterxml.jackson.datatype.jackson-datatype-jsr310",
+                  "identifier" => "com.fasterxml.jackson.datatype.jackson-datatype-jsr310",
                   "slot" => "main",
                   "optional" => true,
                   "export" => false,
                   "import-services" => true,
+                  "duplicated" => false
+              },
+              ...
+          ],
+          "user-dependencies" => [
+              {
+                  "name" => "com.fasterxml.jackson.datatype.jackson-datatype-jdk8",
+                  "slot" => "main",
+                  "optional" => false,
+                  "export" => false,
+                  "import-services" => false,
                   "duplicated" => true
               },
               ...
-          ]
+          ],
+          "local-dependencies" => [
+              {
+                "name" => "deployment.ejb-in-ear.ear.ejb-in-ear-ejb.jar",
+                "slot" => "main",
+                "optional" => false,
+                "export" => false,
+                "import-services" => true,
+                "duplicated" => false
+              },
+              ...
+          ],
+          "resources" => [
+                 {"resource" => "vfs:/content/ejb-in-ear.ear/ejb-in-ear-ejb.jar/"}
+                 {"resource" => "vfs:/content/ejb-in-ear.ear/lib/commons-logging-commons-logging-1.2.jar/"},
+                 ...
+              ]
       }
   }
 
-Where:
+The information is presented in three different categories:
+
+* system-dependencies: These are the dependencies added implicitly by the server.
+* user-dependencies: These are the dependencies defined by the user via manifest file or deployment-structure.xml.
+* local-dependencies: These are dependencies on other parts of the deployment.
+
+For each category, the following information will be shown:
 
 * name: The module name.
 * slot: The slot where this module is.
@@ -96,7 +136,7 @@ Where:
 * It must be possible to list the implicit modules added to the classpath for a specific deployment and, in case of ears, for a particular sub-deployment.
 * List of modules should be sorted alphabetically by module name.
 * The list of modules will be only available if the deployment is enabled.
-* The new operation will be available only in standalone mode. Domain mode changes are not going to be covered by this RFE.
+* The `list-modules` operation will not be available in domain mode under /deployment=xyz resource. Domain mode changes for this resource are not going to be covered by this RFE.
 * The output of the operation will include an attribute named `duplicated` which will be true if the module was added by the user in the deployment, via manifest file or deployment-structure.xml, and added implicitly by the server.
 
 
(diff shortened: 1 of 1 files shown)


server/WFCORE-4251_list_modules_in_deployment_classpath.adoc
```diff
@@ -0,0 +1,118 @@
...
+
+* It must be possible to list the implicit modules added to the classpath for a specific deployment and, in case of ears, for a particular sub-deployment.
+* List of modules should be sorted alphabetically by module name.
+* The list of modules will be only available if the deployment is enabled.
+* The new operation will be available only in standalone mode. Domain mode changes are not going to be covered by this RFE.
+
+
+=== Nice-to-Have Requirements
+
+* The possibility to show together with the module name the artifact name(s) that the module is exposing. Essentially in Wildfly this will be the GAV of the artifact in the module.xml file.
```
kabir: In the thick distribution (i.e. the dist/target directory) the modules don't have the GAV information, but rather the name of the jar.

yersan: Thanks for noticing that @kabir, I've just removed the incorrect afirmation.

server/WFCORE-4251_list_modules_in_deployment_classpath.adoc
```diff
@@ -0,0 +1,118 @@
...
+=== Other Interested Projects
+
+== Requirements
+
+=== Hard Requirements
+
+* It must be possible to list the implicit modules added to the classpath for a specific deployment and, in case of ears, for a particular sub-deployment.
+* List of modules should be sorted alphabetically by module name.
+* The list of modules will be only available if the deployment is enabled.
+* The new operation will be available only in standalone mode. Domain mode changes are not going to be covered by this RFE.
```
bstansberry: I agree that there shouldn't be anything done on the HCs for this, but I don't see why this should not work for a domain mode server. The operation just has to addressed to /host=x/server=y/...

I suspect it would be more work to have it not work on a domain mode server than it would be to have it work.

server/WFCORE-4251_list_modules_in_deployment_classpath.adoc
```diff
@@ -0,0 +1,118 @@
...
+* The output of the operation will include an attribute named `duplicated` which will be true if the module was added by the user in the deployment, via manifest file or deployment-structure.xml, and added implicitly by the server.
+
+
+=== Nice-to-Have Requirements
+
+* The possibility to show together with the module name the artifact name(s), if there is one, that the module is exposing.
+** This possibility could facilitate to the user the task to check which specific artifact(s) and version is added by a module in order to find any duplicate jar bundled by its own application.
```
bstansberry: We already have /core-service=module-loading:list-resource-loader-paths(module=themodulename). I don't think we should try for anything more than optional incorporating the same data. If the jar name incorporates a version, which is often the case, then the resource loader path will provide the version.

server/WFCORE-4251_list_modules_in_deployment_classpath.adoc
```diff
@@ -0,0 +1,118 @@
...
+
+== Requirements
+
+=== Hard Requirements
+
+* It must be possible to list the implicit modules added to the classpath for a specific deployment and, in case of ears, for a particular sub-deployment.
+* List of modules should be sorted alphabetically by module name.
+* The list of modules will be only available if the deployment is enabled.
+* The new operation will be available only in standalone mode. Domain mode changes are not going to be covered by this RFE.
+* The output of the operation will include an attribute named `duplicated` which will be true if the module was added by the user in the deployment, via manifest file or deployment-structure.xml, and added implicitly by the server.
```
bstansberry: If we have the data to know if it was duplicate, how about an implicit=true|false?  Probably some better name than implicit, more descriptive.

TBH, 'duplicated' sounds like a nice-to-have. I doubt that it's a common occurrence and AFAIK it's harmless if it occurs. This check would not cover cases where the deployment includes a jar that is also available via a module dependency. I expect that is much more common and is more likely to be a problem, as it means classes with the same name are available from two different classloaders. (I don't think this RFE should try and address that though!)

server/WFCORE-4251_list_modules_in_deployment_classpath.adoc
```diff
@@ -0,0 +1,118 @@
...
+              },
+              {
+                  "name" => "com.fasterxml.jackson.datatype.jackson-datatype-jsr310",
+                  "slot" => "main",
+                  "optional" => true,
+                  "export" => false,
+                  "import-services" => true,
+                  "duplicated" => true
+              },
+              ...
```
bstansberry: Does this list include the deployment module itself -- i.e. the one created for it by WildFly?

server/WFCORE-4251_list_modules_in_deployment_classpath.adoc
```diff
@@ -0,0 +1,118 @@
...
+This document describes the RFE to provide the ability to expose this information via a management operation.
+
+When optimizing the application footprint, this information becomes important:
+
+* The user could use this list to identify which dependencies are added implicitly by the server, and then, exclude any unneeded dependency from the deployment.
+* The user could use this information to discover any duplicated artifact bundled together with the deployment and included in one of the modules that the server is adding to the application classpath.
+* This information could facilitate discovering unneeded subsystems in the installation of his servers.
+
+
+The proposed solution is to include a new operation `list-modules` available under `/deployment` and, in case of ears, `/deployment=*/subdeployment` resource:
```
bstansberry: Re: ears for each of the deployment or subdeployment resources, what is shown re: the other? IOW there are module dependencies between the elements of these kinds of deployments -- are these shown?

I don't think there should be any special magic handling for these, i.e. working out data that isn't available in a similar way to how you get other module dependency data. My question is more one of clarifying what can / cannot be displayed given the data that's available.

server/WFCORE-4251_list_modules_in_deployment_classpath.adoc
yersan: Yes, it does.

server/WFCORE-4251_list_modules_in_deployment_classpath.adoc
yersan: > Re: ears for each of the deployment or sub-deployment resources, what is shown re: the other? IOW there are module dependencies between the elements of these kinds of deployments -- are these shown?

Yes, we will see the dependencies added implicitly by the server and the shared modules between them. E.g, if we deploy an ear that contains a war and an ejb-jar, when we are listing the modules of the war, we will see the modules added implicitly by the server plus the ear module and the ejb-jar module, these latter ones being modules created by the server at deployment time.

> I don't think there should be any special magic handling for these, i.e. working out data that isn't available in a similar way to how you get other module dependency data. My question is more one of clarifying what can / cannot be displayed given the data that's available.

We have access to the list of final modules created by each deployment and sub-deployment. We could even have the list split by modules added by the user, modules added by the server and the modules created by the server for the deployment file deployed by the user.

server/WFCORE-4251_list_modules_in_deployment_classpath.adoc
yersan: > If we have the data to know if it was duplicate, how about an implicit=true|false? Probably some better name than implicit, more descriptive.

Do you mean to add additionally a new attribute implicit=true|false so if 'duplicate=false', then to know if it was a dependency added by the server or by the user?, if yes, we can also use a user-defined=true|false

> 
> Separate point:
> 
> TBH, 'duplicated' sounds like a nice-to-have. I doubt that it's a common occurrence and AFAIK it's harmless if it occurs. 

Agreed.

> This check would not cover cases where the deployment includes a jar that is also available via a module dependency. I expect that is much more common and is more likely to be a problem, as it means classes with the same name are available from two different classloaders. (I don't think this RFE should try and address that though!)

Yes, the idea is to be able to know if the user is including modules explicitly via manifest file or deployment-structure.xml when they are not required because the server is including them implicitly.

server/WFCORE-4251_list_modules_in_deployment_classpath.adoc
bstansberry: I think we need to be very careful about trying to do complex analysis as it becomes a maintenance and compatibility burden. If we try and do things and we are wrong it's our problem to fix. And we are talking about exposing internals here so we need to be careful that we do not turn impl details into API.

Note that even exposing the resource loader paths is "deprecated" which really means "experimental". We shouldn't make accessing that data via another op non-experimental. (Perhaps the original op shouldn't be experimental any more; it's quite old. It depends on what JBoss Modules intends to support.)

Input from @bmaxwell on data that would be most commonly useful in handling problems would be good.

My guess is what would be most useful is the module list plus the resource loader data from the deployment modules themselves. That's enough for a support person or someone providing help on a forum to see "aha, they are packaging foo.jar and I know foo.jar is provided by the foobar module, so that's suspicious."  Showing all resource loader paths is also useful, but it's a lot of data. And then showing none may be useful for compact output. Maybe not very useful though. 

I suspect limiting resource loader data to the synthetic deployment modules may not be feasible though; it requires being able to distinguish those modules from other modules. I don't know if that's easily doable.

server/WFCORE-4251_list_modules_in_deployment_classpath.adoc
bstansberry: I like 'user-defined'. 👍

server/WFCORE-4251_list_modules_in_deployment_classpath.adoc
bstansberry: Can you spell out these details in the proposal? What you say sounds good.

server/WFCORE-4251_list_modules_in_deployment_classpath.adoc
yersan: > And we are talking about exposing internals here so we need to be careful that we do not turn impl details into API.

Keep in mind the user can get this information is already exposed by jboss module using the jconsole: jconsole -> jboss.modules -> ModuleLoader -> LocalModuleLoader -> getResourceloaders operation. 

> My guess is what would be most useful is the module list plus the resource loader data from the deployment modules themselves. That's enough for a support person or someone providing help on a forum to see "aha, they are packaging foo.jar and I know foo.jar is provided by the foobar module, so that's suspicious." Showing all resource loader paths is also useful, but it's a lot of data. And then showing none may be useful for compact output. Maybe not very useful though.

I think it would be fine for an administrator, but not from the point of view of the developer. The developer already knows which jars are packaged within his application, I guess what is useful for a developer is to know which resources are added to the application by the container. 

Listing the jboss modules added implicitly to the application classpath could be enough to know this information. However, the developer has to check each module.xml and understand how they are interrelated. I was trying to facilitate this task to the developer, but coming back to your first point, we run in the risk of a complex feature that exposes internals. So I suppose we should forget this posibility at the moment.

> I suspect limiting resource loader data to the synthetic deployment modules may not be feasible though; it requires being able to distinguish those modules from other modules. I don't know if that's easily doable.

I think we could distinguish them. Those modules are intrinsically related with the deployment or subdeployment name, so we can get and list the resources from those names. That means if we are listing the modules using:

`/depoyment=ejb-in-ear.ear`
We will list the resources used by deployment.ejb-in-ear.ear

`/deployment=ejb-in-ear.ear/subdeployment=ejb-in-ear-web.war`
We will list the resources used by deployment.ejb-in-ear.ear.ejb-in-ear-web.war

Do we agree adding this infomration to the output? Do you see it useful at all?

server/WFCORE-4251_list_modules_in_deployment_classpath.adoc
yersan: If we are displaying the information divided into categories like system dependencies, user dependencies, and local dependencies, the user-defined attribute is no longer needed.

server/WFCORE-4251_list_modules_in_deployment_classpath.adoc
yersan: I've just updated the PR with the output divided into dependency categories; system dependencies, user dependencies, and local dependencies

server/WFCORE-4251_list_modules_in_deployment_classpath.adoc
bstansberry: > Keep in mind the user can get this information. It's already exposed by jboss module using the jconsole: jconsole -> jboss.modules -> ModuleLoader -> LocalModuleLoader -> getResourceloaders operation.

Any JMX API except for the jboss.as and jboss.as.expr domains is not supported WildFly management API.  Granted, if things change in the other JMX APIs, particularly 'core' ones like modules, and people complain we would likely react. But in the end if we can't react we would just say "sorry, we don't support that and we can't do what you want". Adding something to the primary management API brings a different level of requirements.

I think the main thing about this part of the conversation is to get clarification from @dmlloyd about what JBoss Modules will support long term and limit anything we do to what that allows. And be conservative in general too, just because. ;)  And if the things we already have are supported we can get rid of the deprecation notice too. (Separate task from this of course.)

> I think it would be fine for an administrator, but not from the point of view of the developer. The developer already knows which jars are packaged within his application, I guess what is useful for a developer is to know which resources are added to the application by the container.

True. 

A bunch of different varieties of output is probably not a good idea. If a non-expert is asking for help, the expert can easily enough ask for the verbose output and ignore data they don't need. At most I can see a use for a verbose and a non-verbose option (since the verbose would likely be a ton of data; we add a lot of modules.) The question then becomes what the non-verbose option would include. 

> I mean, display the classloader resources (probably limiting them only to .jar files) of the synthetic deployment modules, and not adding the resources exposed by the modules added by the server/user. Do you see it useful at all? We can get them via module.getClassLoader().getResources("/")

What does the output of module.getClassLoader().getResources("/")
 look like for a non-deployment module whose resources include a jar and a dir? Skipping the 'limiting them only to .jar files' bit.

server/WFCORE-4251_list_modules_in_deployment_classpath.adoc
bstansberry: Be cautious about exposing implementation details.

jamezp:
This is kind of minor, but I wonder if we should refer to the "classpath" as "dependencies" instead since it's not a typical class path.

server/WFCORE-4251_list_modules_in_deployment_classpath.adoc
yersan: My fault, module.getClassLoader().getResources("/") is not the correct method to get the resources; indeed it doesn't work as I was expecting for non-deployment modules, and, after looking deeply, even for deployment modules. We must forget it.

It looks like we can only get this information via the MXBean, and even using it, we will need some minor tweaks in jboss modules to be able to get the jar file name of non-deployment modules.

That means, showing the .jar files available to the application presents some problems:
1. We will be exposing internals to the application.
2. It requires the use of, initially, experimental features.
3. If we were to show what is available coming from non-deployment modules, we will need to show all the resources of the dependency tree that has exports=true.

Having said that, I suppose we should abandon this path in favor of a minimum feature, which will show the list of jboss modules added to the deployment split into three categories, system-dependencies, user-dependencies, and local-dependencies. This information sorted and printed in these categories already helps to identify problems and to understand what is available to the user deployment. This is useful, since this information is not easy to retrieve in Wildfly, the jconsole output is the only possibility. We could address additional functionalities in future RFEs. I'm going to update the PR to this minimal version.

Do we agree to stop the analysis at this point?

yersan: > At most I can see a use for a verbose and a non-verbose option (since the verbose would likely be a ton of data; we add a lot of modules.) The question then becomes what the non-verbose option would include.

After playing with a POC project, yes, the list is a bit long, so makes sense to me the use of verbose and non-verbose output even for a minimal version. Non-verbose will include only the name of the module, whereas a detailed information will include additional attributes as slot, optional, export and import-services

server/WFCORE-4251_list_modules_in_deployment_classpath.adoc
yersan: @dmlloyd confirmed that the intention is to have a long term support for the following attributes exposed by this operation:
- name
- slot
- optional
- export
- import-services

Additionally he suggested to collapse "name" and "slot" in a single "name" property. A module with a slot of "main" would just be named by the module name; other slots would be concatenated to the end of the module name after the ":" separator.

I'm going to update the proposal with this last point.

server/WFCORE-4251_list_modules_in_deployment_classpath.adoc
```diff
@@ -0,0 +1,170 @@
...
+
+* system-dependencies: These are the dependencies added implicitly by the server.
+* local-dependencies: These are dependencies on other parts of the deployment.
+* user-dependencies: These are the dependencies defined by the user via manifest file or deployment-structure.xml.
+
+
+For each category, the meaning of each attribute is:
+
+* name: The module name and, if the slot name is not the default "main" slot, the slot name is concatenated after ":" character separator.
+* optional: If the dependency was added as an optional dependency.
```
bstansberry: I interpret 'optional' as meaning the module might not be present, but either the user (for user dependencies) or the server (for system and local) recorded dependency info in case it is present. But then, if optional=true, can we reliably report whether the dependency is actually present?  If yes, that's something to consider for any future iteration on this feature. I don't think it's necessary for this first iteration though

yersan: > I interpret 'optional' as meaning the module might not be present

Correct.

> But then, if optional=true, can we reliably report whether the dependency is actually present?

Yes, we can detect if the dependency, being optional, is not included in the server distribution, and then, somehow inform about this situation in the report. I can see some cases that have to be handled, e.g. a deployment depends optionally in another deployment and that second deployment is undeploy, we must deal with these cases as well to reflect correctly those runtime changes. But, yes, we could deal with it. It is left for a second iteration in case of be requested.

yersan:
This is just for the record, I've just forced-pushed the branch adding a minor note describing that in domain mode the operation list-modules is not available also under /server-group=*/deployment=xyz. We have no plans to make them available there at this stage.

jmesnil:
https://github.com/wildfly/wildfly-core/pull/3661 has been merged

PR #161, opened 2018-12-12
WFLY-11404 Add journal-file-open-timeout attribute to messaging server

TomasHofman:
Jira: https://issues.jboss.org/browse/WFLY-11404


TomasHofman:
@ehsavoie @mnovak1 I was checking docs in wildfly docs/ directory, and there isn't practically anything about related journal attributes, so I changed the part about documentation to "Covered by attribute description in management model." It would the same text as in attribute description anyway...

jamezp:
https://github.com/wildfly/wildfly/pull/11918 was merged

PR #163, opened 2018-12-12
Add a Release Note section to the proposal template

bstansberry:
Ease the WildFly release process by encouraging proposal authors to provide draft short text for an entry in the Release Note blog article.


bstansberry:
@n1hility This is a follow-up to our conversation yesterday.

darranl:
+1 we do have a separate wildfly-security-incubator organisation for more collaborative work - maybe should try using it a lot more.

A lot of the feedback submitted against proposals could be better in a PR such as please add me as a contact or typo updates.  Also I think once discussions begin others on the feature team should be able to add to the analysis without it always coming back to the engineer.

PR #167, opened 2019-01-13
EAP7-1153: JBoss EAP 7.3 Server Migration Tool

emmartins:
EAP7: https://issues.jboss.org/browse/EAP7-1153


darranl:
Is there a community issue that can be referenced in this PR instead of the private RFE?

OndrejKotek:
@emmartins Can you update the analysis with changes in the model and their impact on implementation please? It would enable me to plan tests to be implemented.

emmartins:
@OndrejKotek I have updated the doc, pretty much just added mp metrics subsystem compared with what we had for 7.2 so you should be able to reuse all that existed for 7.2

bstansberry:
Let's use Google docs or something for things that solely relate to EAP.

PR #170, opened 2019-01-18
[WFLY-11529] Expose WildFly metrics

jmesnil:
Proposal to expose metrics from WildFly subsystem and deployments to the
`/metrics` HTTP endpoint in the Prometheus format to be able to be
consumer and aggregate by Prometheus.

JIRA: https://issues.jboss.org/browse/WFLY-11529

### DIFF
--- a/microprofile/WFLY-11529_subsystem_metrics.asciidoc
+++ b/microprofile/WFLY-11529_subsystem_metrics.asciidoc
@@ -1,0 +1,158 @@
+= [WFLY-11529] Expose WildFly Metrics
+:author:            Jeff Mesnil
+:email:             jmesnil@redhat.com
+:toc:               left
+:icons:             font
+:idprefix:
+:idseparator:       -
+:keywords:          observability,microprofile,metrics,openshift
+
+== Overview
+
+The goal of this feature is to improve observability of WildFly by exposing its own metrics to monitoring tools such as https://prometheus.io[Prometheus]
+and the https://docs.openshift.com/container-platform/3.11/architecture/infrastructure_components/web_console.html[OpenShift Console].
+
+This feature exposes the WildFly metrics (from the subsystem and deployment management models) to the `/metrics` HTTP endpoint in the Prometheus format.
+
+== Issue Metadata
+
+=== Issue
+
+* https://issues.jboss.org/browse/WFLY-11529[WFLY-11529] - Expose WildFly metrics in /metrics endpoints
+
+=== Related Issues
+
+* https://issues.jboss.org/browse/EAP7-1169[EAP7-1169] - Expose EAP subsystem Metrics
+* https://issues.jboss.org/browse/EAP7-1026[EAP7-1026] - EAP7-1026 Expose EAP Metrics so that it can be consumed by Kiali (Istio) Console
+** in this feature, base metrics (e.g. JVM) and application metrics (from MicroProfile Metrics API) were exposed but there were none coming
+  from WildFly subsystems and deployments
+
+=== Dev Contacts
+
+* mailto:{email}[{author}]
+
+=== QE Contacts
+
+* mailto:rsvoboda@redhat.com[Rostislav Svoboda]
+
+=== Other Interested Projects
+
+* Prometheus MUST be able to consume and aggregate WildFly metrics.
+
+== Requirements
+
+* Expose WildFly metrics from subsystem and deployment mananagement model to the `/metrics` HTTP endpoint in the Prometheus format.
+* Provide mechanism to be _compatible_ with the WildFly metrics exposed by EAP images in OpenShift (that uses Prometheux JMX exporter as configured in https://github.com/jboss-container-images/jboss-eap-modules/blob/master/jboss/container/eap/prometheus/config/7.2/artifacts/opt/jboss/container/prometheus/etc/jmx-exporter-config.yaml[jmx-exporter-config.yaml])
+** compatibility means that the metric names and labels should be identical (or, if identity is not achievable, similar).
+
+=== Non-requirements
+
+* WildFly metrics are *NOT* exposed to the `/metrics` HTTP endpoint in the JSON format.
+* WildFly metrics are *NOT* exposed to other metrics HTTP endpoints (such as `/metrics/vendor`).
+
+=== Internal requirements
+
+WildFly management model has a notion of metrics for resource attributes (as described in http://docs.wildfly.org/15/Admin_Guide.html#Description_of_the_Management_Model[WildFly Admin Guide]).
+This feature expose these metrics from WildFly management model, such as:
+
+* Transaction metrics (from the http://wildscribe.github.io/WildFly/15.0/subsystem/transactions/index.html[transactions subsystem])
+* HTTP usage (from the http://wildscribe.github.io/WildFly/15.0/subsystem/undertow/server/http-listener/index.html[undertow subsystem])
+* JDBC Pool usage (from the http://wildscribe.github.io/WildFly/15.0/subsystem/datasources/data-source/ExampleDS/statistics/pool/index.html[datasources subsystem])
+
+
+To be exposed, a WildFly metric must satisfy the following requirements:
+
+* it must have been registered as a metric (its `AccessType` is `METRIC` and its `Storage` is `RUNTIME`)
+* its `ModelType` MUST be a single numerical value
+* its `:read-attribute` operation MUST return an actual numerical value (metrics that returns `undefined` are not exposed)
+
+[NOTE]
+====
+The `:read-attribute` operation is called internally with `include-undefined-metric` set to `true` (as explained in https://issues.jboss.org/browse/WFCORE-4190[WFCORE-4190])
+====
+
+=== Subsystem description
+
+This feature adds 2 attributes to the `microprofile-metrics-smallrye` subsystem resource:
+
+* `exposed-subsystems` - a list of strings corresponding the name of subsystems that exposes their metrics in the HTTP endpoints.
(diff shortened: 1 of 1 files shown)


microprofile/WFLY-11529_subsystem_metrics.asciidoc
```diff
@@ -0,0 +1,145 @@
...
+
+== Issue Metadata
+
+=== Issue
+
+* https://issues.jboss.org/browse/WFLY-11529[WFLY-11529]
+
+=== Related Issues
+
+* https://issues.jboss.org/browse/EAP7-1169[EAP7-1169]
```
rsvoboda: change this line to
* https://issues.jboss.org/browse/EAP7-1169[EAP7-1169] - EAP7-1169 Expose EAP subsystem Metrics

to be in sync with the line 26

microprofile/WFLY-11529_subsystem_metrics.asciidoc
```diff
@@ -0,0 +1,145 @@
...
+
+== Requirements
+
+* Expose WildFly metrics from subsystem and deployment mananagement model to the `/metrics` HTTP endpoint in the Prometheus format.
+* Provide mechanism to be _compatible_ with the WildFly metrics exposed by EAP images in OpenShift (that uses Prometheux JMX exporter as configured in https://github.com/jboss-container-images/jboss-eap-modules/blob/master/jboss/container/eap/prometheus/config/7.2/artifacts/opt/jboss/container/prometheus/etc/jmx-exporter-config.yaml[jmx-exporter-config.yaml])
+** compatibility means that the metric names and labels should be identical (or, if identity is not achievable, similar).
+
+=== Non-requirements
+
+* WildFly metrics are *NOT* exposed to the `/metrics` HTTP endpoint in the JSON format.
```
rsvoboda: I would prefer to have both Prometeus and JSON so everything under `/metrics` is consistent.

How complex would it be to have JSON enabled?

microprofile/WFLY-11529_subsystem_metrics.asciidoc
```diff
@@ -0,0 +1,145 @@
...
+* WildFly metrics are *NOT* exposed to other metrics HTTP endpoints (such as `/metrics/vendor`).
+
+=== Internal requirements
+
+WildFly management model has a notion of metrics for resource attributes (as described in http://docs.wildfly.org/14/Admin_Guide.html#Description_of_the_Management_Model[WildFly Admin Guide]).
+This feature expose these metrics from WildFly management model, such as:
+
```
rsvoboda: 14.0 => 15.0 in the following links

microprofile/WFLY-11529_subsystem_metrics.asciidoc
```diff
@@ -0,0 +1,145 @@
...
+* Transaction metrics (from the http://wildscribe.github.io/WildFly/14.0/subsystem/transactions/index.html[transactions subsystem])
+* HTTP usage (from the http://wildscribe.github.io/WildFly/14.0/subsystem/undertow/server/http-listener/index.html[undertow subsystem])
+* JDBC Pool usage (from the http://wildscribe.github.io/WildFly/14.0/subsystem/datasources/data-source/ExampleDS/statistics/pool/index.html[datasources subsystem])
+
```
rsvoboda: Mention `-Dwildfly.statistics-enabled=true` in the analysis document or not ?

microprofile/WFLY-11529_subsystem_metrics.asciidoc
```diff
@@ -0,0 +1,145 @@
...
+wildfly_undertow_request_count_total{deployment="helloworld-rs.war",servlet="org.jboss.as.quickstarts.rshelloworld.JAXActivator",subdeployment="helloworld-rs.war",} 6.0
+...
+----
+
+===== Implementation Issues
+
+====== Complex WildFly Metrics Are Not Supported
+
+Resources may return arbitrarily complex return types for metrics.
+The registration code in `microprofile-smallrye-metrics` will only register WildFly metrics for simple numerical `ModelType`.
```
rsvoboda: Is there a warning when such complex type is encountered ?

microprofile/WFLY-11529_subsystem_metrics.asciidoc
```diff
@@ -0,0 +1,145 @@
...
+
+====== Management Resources Added After Server Boot Will Not Expose Their Metrics
+
+The `microprofile-smallrye-metrics` extension will register any valid metric from WildFly Management Model when it is installed.
+However if other management resources are added afterwards, the extension will not be aware of them and will not register their metrics.
```
rsvoboda: `management resources are added afterwards` ... what can be an example of this ?

I'm expecting server reload would make them accessible by the extension.  Maybe it could be mentioned here too.
If there is less intrusive method than server reload it could be mentioned too.

microprofile/WFLY-11529_subsystem_metrics.asciidoc
jmesnil: no, there is no warning. We do not need to warn the user as there is nothing he could do about it.
We have some metrics that returns STRING (e.g. enum of server state) which is correct from WildFly point of view.
However Prometheus only consumes numerical metric.

If/When we decide to support more complex metrics (eg histogram), we would have to create corresponding RFEs for it.

microprofile/WFLY-11529_subsystem_metrics.asciidoc
jmesnil: Right, if the server reloads, the new management resources would be taken into account.

However if you add a new management resource that does not require to reload the server (e.g. a new JMS Queue), the metrics for it would not be available until the next server reload.

jmesnil:
@rsvoboda JSON format is out of scope of this RFE (as the JSON format is specified in MicroProfile Metrics and is incompatible with having the same metric name for different metrics associated to different labels). As long as we stay with MP Metric 1.x, this will not be possible.

This RFE is all about exposing WildFly metrics to Prometheus. Having these metrics in the text format in `/metrics` endpoint is enough to provide the feature.

rsvoboda:
**1.** Prometheus exposure is one aspect of the "Expose WF/EAP subsystem Metrics" RFE.
Front-end application world is based on JavaScript and TypeScript where JSON is primary format.
Frameworks like Angular, React, Vue.js. Use-case example: status page in the web application.

My current feeling is that for EAP7-1169 this could go in almost as is now (see point 4.)
For EAP7-1146 I hesitate to proceed, I'd rather wait for MP Metrics 2.0 which is in progress and almost done.

**2.** Your point about MP Metric 1.x limitations for JSON format is valid.
Is it going to be solved with MP Metric 2.0 ?

**3.** MP Metric 1.x limitations for JSON format should be described more in the analysis document so it's more understandable why just Prometheus format is available at the moment for subsystem metrics.

**4.** `curl -X OPTIONS -H "Accept: application/json" http://127.0.0.1:999/metrics` should inform about presence of WildFly subsystem metrics with message saying something like: "accessible only via text / Prometheus format". Or is there other suggested way to get "help" about WildFly subsystem metrics ?

jmesnil:
Your point (1) about Prometheus exposure is *the* feature provided by this RFE (and the previous 
https://issues.jboss.org/browse/WFLY-10712).
This feature is only about providing WildFly metrics to Prometheus. Any other hypothetical use case (such as Web client) is not covered by this RFE. We agreed to use MP metrics as a tool to achieve this feature but as we realised that MP metrics 1.x would not meet our requirements, we pivoted to another approach.

Regarding EAP7-1146, we comply to MicroProfile Metrics 1.1. However as this version does not provide the feature we need to expose our own "vendor" metrics in a sensible fashion, we chose to not expose them (and still be compliant with Metrics 1.1). 

(2) Yes, MP Metrics 2.0 should fix the issue related to metric registration but we may not be able to use MicroProfile Metrics 2.0 before some time and this should not block our ability to deliver this feature.

(3) You're right, I'll edit the document to describe it in the implementation issues

(4) There is no place in the JSON format to put such information. And I don't see the point to specify what's not there. WildFly metrics are not exposed in JSON. The absence of metadata is consistent with the absence of metrics.

rsvoboda:
OK, I will wait for the update for (3) before approval, have no other concerns.

jmesnil:
@rsvoboda I've amended the document regarding (3).

microprofile/WFLY-11529_subsystem_metrics.asciidoc
```diff
@@ -0,0 +1,158 @@
...
+
+==== Metric name and labels
+
+The name of a WildFly metric is based on the subsystem that provides it as well as the name of the attribute from the management model.
+The name can also be prepended with a `prefix` (specified on the `/subsystem=microprofile-metrics` resource).
+The name can also be appended with the `_total` suffix if it corresponds to a Counter metrics (as explained in https://issues.jboss.org/browse/WFCORE-4238[WFCORE-4238])
+Other information are stored using labels.
+
+For example Undertow exposes a metric attribute `request-count` (which is a counter) for every Servlet in an application deployment.
+This attribute will be exposed to Prometheus with the name `wildfly_undertow_request_count_total`.
```
rsvoboda: replace `wildfly_undertow_request_count_total` with `wildfly_undertow_request_count`

Drop trailing `_total`

```
# HELP wildfly_undertow_request_count The number of requests this listener has served
# TYPE wildfly_undertow_request_count gauge
```

rsvoboda: this is valid for multiple places in the document

jmesnil: adding `_total` is the convention for counters in Prometheus. We have https://issues.jboss.org/browse/WFLY-11578 to be able to distinguish them from gauges (which is the default metric type used by WildFly)

PR #174, opened 2019-02-12
[WFLY-11697] WS integration with Elytron on the client side

Skyllarr:
JIRA: 
https://issues.jboss.org/browse/WFLY-11697

### DIFF
--- a/elytron/WFLY-11697_WS_client_integration_with_Elytron.adoc
+++ b/elytron/WFLY-11697_WS_client_integration_with_Elytron.adoc
@@ -1,0 +1,114 @@
+= WFLY-11697 WS integration with WildFly Elytron - AuthenticationClient for Authentication / SSL
+:author:            Diana Vilkolakova
+:email:             dvilkola@redhat.com
+:toc:               left
+:icons:             font
+:keywords:          elytron, client, webservices
+:idprefix:
+:idseparator:       -
+
+== Overview
+
+WildFly Elytron uses the Elytron Client project to enable remote clients to authenticate using Elytron.
+Elytron client is already used by other WildFly clients allowing for both use of an Elytron API and pre-defined configurations in the xml file (https://docs.jboss.org/author/display/WFLY/WildFly+Client+Configuration#WildFlyClientConfiguration-wildflyconfig.xmlDiscovery[wildfly-config.xml]).
+
+This requested feature is about making use of Elytron's configuration on the client side of WS deployment so that the WS client supports the common framework available within the application server.
+
+The implementation will involve updating Elytron client configuration with additional settings that will specify the policy (mechanisms) the webservices client will use to authenticate with the server.
+
+The implementation will involve loading of this authentication configuration from Elytron client, specifically credentials, mechanisms and SSL context, and making those available to the Webservices client.
+
+== Issue Metadata
+
+=== Issue
+
+* https://issues.jboss.org/browse/WFLY-11697[WFLY-11697]
+
+=== Related Issues
+
+* https://issues.jboss.org/browse/EAP7-1097[EAP7-1097]
+
+=== Dev Contacts
+
+* mailto:dvilkola@redhat.com[Diana Vilkolakova]
+
+=== QE Contacts
+
+* mailto:jbliznak@redhat.com[Jan Bliznak]
+
+=== Testing By
+// Put an x in the relevant field to indicate if testing will be done by Engineering or QE.
+// Discuss with QE during the Kickoff state to decide this
+[x] Engineering
+
+[ ] QE
+
+=== Affected Projects or Components
+
+* Security
+* Web Services
+
+=== Other Interested Projects
+
+N/A
+
+== Requirements
+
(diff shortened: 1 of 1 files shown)


elytron/WFLY-11697_WS_REST_client_integration_with_Elytron.adoc
```diff
@@ -0,0 +1,60 @@
...
+If wildfly-config.xml file is present when using RESTEasy or Web Services on the client side, this authentication configuration will be parsed and made available for this client.
+
+=== Nice-to-Have Requirements
+
+=== Non-Requirements
+
+== Implementation Plan
+
+SPI will be exposed in RESTEasy and Web Services for client's authentication configuration. This SPI will be implemented in Elytron.
+
```
Skyllarr: Alternatively, this SPI will be in its own module that both RestEasy and WS depend on and Elytron will implement.

Skyllarr: @asoldano So RestEasy will have SPI exposed that Elytron client implements and the loading of service providers could happen during build of RestEasyClientBuilder. Does this sound reasonable?

asoldano: @Skyllarr I suggest dealing with JBossWS and RESTEasy independently. JBossWS has its own SPI component, jbossws-spi (https://github.com/jbossws/jbossws-spi), which really looks like where the additional spi should be defined.
RESTEasy doesn't have such clearly defined spi component, so I believe a new module will have to be added to the project.

Skyllarr: @asoldano Thank you for your reply! We will split this issue and will deal with JBossWS and RESTEasy separately.

Skyllarr:
Hello @asoldano . This PR is now only for the WS side. I force-pushed new analysis document. Can you please take a look? Thank you in advance.

asoldano:
Thanks @Skyllarr . I'm adding @jimma to the discussion here, as he's now leading JBossWS / WebServices.
Jim, Diana here is designing the Elytron-WS integration that will have soon to be implemented. Please check if the proposal makes sense.

jimma:
@Skyllarr  @asoldano  Is there background thing fot me to understand the purpose of add wildfly-config.xml in ws client ?

Skyllarr:
Hello @jimma. This proposal is about being able to make use of elytron on the client side of the web services deployment. [Wildfly-config.xml](https://docs.jboss.org/author/display/WFLY/WildFly+Client+Configuration) contains authentication configuration for the client to use.

darranl:
The backround really is the WildFly Elytron authentication client is already used by a number of our clients allowing for both use of an API and pre-defined configurations.  As well as configuration this set of APIs is also fundamental for features such as identity propagation.  

At the moment WS and Rest clients are not supporting the common framework available within the application server.

elytron/WFLY-11697_WS_client_integration_with_Elytron.adoc
```diff
@@ -0,0 +1,75 @@
...
+
+The method setConfigHandlers takes client config file as a parameter.
+This file adheres to the schema
+http://anonsvn.jboss.org/repos/jbossws/spi/tags/jbossws-spi-2.1.0.Beta1/src/main/resources/schema/jbossws-jaxws-config_4_0.xsd[jbossws-jaxws-config schema] defined in jbossws-spi.
+Possible solutions are:
+
+* schema of wildfly-config.xml file and jaxws-config file could be combined and user can specify everything in single file
+* these configs will stay separate and user will provide paths to both files (if some information is conflicting, wildfly-config.xml setting takes precedence?)
+* the existence of wildfly-config.xml and its processing will be done by default in setConfigHandlers method
+* new method for this config file will be added (the use could look like this: `configurerBuilder.setConfigHandlers().setAuthHandlers().build()`)
```
Skyllarr: @darranl @jimma I am guessing the last two options are the most viable? We can schedule a call if preferred. Thanks!

darranl: jaxws-config could also make use of the wildfly-config.xml - but I think for this RFE the priority is we can already define client side authentication policies in the wildfly-config.xml so find a way to make use of them.

elytron/WFLY-11697_WS_client_integration_with_Elytron.adoc
```diff
@@ -0,0 +1,86 @@
...
+N/A
+
+== Requirements
+
+=== Hard Requirements
+
+The WS-Security standard offers three methods of authentication. Those methods are:
+
+* UsernameToken Profile (adds credentials to header in the SOAP requests)
+* X 508 Certificates Token Profile (to use with TLS protocol)
```
jbliznak: Minor one - should this be X.509?

Skyllarr: Fixed.

PR #176, opened 2019-02-14
[WFLY-11031] Add proposal for JSON access logging.

jamezp:
https://issues.jboss.org/browse/WFLY-11031

### DIFF
--- a/undertow/WFLY-11031_Formatted_Access_Logs.adoc
+++ b/undertow/WFLY-11031_Formatted_Access_Logs.adoc
@@ -1,0 +1,76 @@
+= JSON Formatted Access Logs
+:author:            James Perkins
+:email:             jperkins@redhat.com
+:toc:               left
+:icons:             font
+:idprefix:
+:idseparator:       -
+
+== Overview
+
+Currently access logs can only be written as plain text. This plain text will be written to either a specific file or
+logged via the normal logging facilities. It's desirable for a log aggregator to accept formatted log messages. Using
+JSON seems to be the most common format.
+
+== Issue Metadata
+
+=== Issue
+
+* https://issues.jboss.org/browse/WFLY-11031[WFLY-11031]
+* https://issues.jboss.org/browse/EAP7-1183[EAP7-1183]
+
+=== Related Issues
+
+=== Dev Contacts
+
+* mailto:{email}[{author}]
+
+=== QE Contacts
+
+* mailto:jstourac@redhat.com[Jan Stourac]
+
+=== Affected Projects or Components
+
+* WildFly
+** Undertow subsystem
+
+=== Other Interested Projects
+
+* OpenShift
+
+== Requirements
+
+=== Hard Requirements
+
+* The output must be well formed JSON.
+* The formatted message must be on a single line.
+* The output must be able to write to standard out. This is required for OpenShift.
+* A label of some sort, whether it's hard-coded or user defined, must be added to the output. This allows filtering
+  log messages of different types.
+
+=== Nice-to-Have Requirements
+
+* Allow the user to choose which attributes are added to the JSON.
+* The ability to add custom key/value (meta-data) pairs to the output.
+    ** This would be something like `{ "@version" : "1" }`
+
+=== Non-Requirements
+
+* Messages will only be written to `stdout` at this point.
+* Messages cannot go through a log manager. The message itself needs to be structured
+  which would not be the case if going through a log manager.
+* Messages written to the console for domain servers will not remove the
+  [server-name] prefix from standard out.
+
+== Implementation Plan
+
+The implementation will be done in WildFly and we should not need to do anything to Undertow itself.
+
+A new `setting` resource will be introduced. The resource will be
+`/subsystem=undertow/server=default-server/host=default-host/setting=console-access-log`. This allows
+no changes to the current access logging and allows the console access logging to be used independently
+from the file or standard logging access logging.
+
+== Test Plan
+
+Standard unit and integration tests will be used.


undertow/WFLY-11031_Formatted_Access_Logs.adoc
```diff
@@ -0,0 +1,71 @@
...
+
+=== Related Issues
+
+=== Dev Contacts
+
+* mailto:{email}[{author}]
+
+=== QE Contacts
+
+TBD
```
jstourac: feel free to put me in

undertow/WFLY-11031_Formatted_Access_Logs.adoc
```diff
@@ -0,0 +1,71 @@
...
+* The output must be able to write to standard out. This is required for OpenShift.
+* A label of some sort, whether it's hard-coded or user defined, must be added to the output. This allows filtering
+  log messages of different types.
+
+=== Nice-to-Have Requirements
+
+* Allow the user to choose which attributes are added to the JSON.
+* The ability to override the key name.
+* The ability to add custom key/value pairs to the output.
+    ** This would be something like `{ "@version" : "1" }`
```
jstourac: Do I understand it correctly that 'value' is gonna be just some constant, right?

undertow/WFLY-11031_Formatted_Access_Logs.adoc
```diff
@@ -0,0 +1,71 @@
...
+
+=== Non-Requirements
+
+== Implementation Plan
+
+The implementation will be done in WildFly and we should not need to do anything to Undertow itself.
+
+This is not a complete implementation plan. However it might be desirable to deprecate the current
+`/subsystem=undertow/server=default-server/host=default-host/setting=access-log`. Adding new attributes for a JSON
+formatter and the ability to write to standard out might make the model confusing. This is yet to be determined though.
```
jstourac: Has this been determined somehow already? Could you please describe how the configuration may look like?

undertow/WFLY-11031_Formatted_Access_Logs.adoc
jamezp: That is correct. There will be the ability to use expressions for the value however it would be somewhat constant in that it would be set once the service was started.

undertow/WFLY-11031_Formatted_Access_Logs.adoc
jamezp: Will do. I've got it working locally so I can put down what I did which will hopefully explain it better.

undertow/WFLY-11031_Formatted_Access_Logs.adoc
```diff
@@ -0,0 +1,70 @@
...
+=== Non-Requirements
+
+== Implementation Plan
+
+The implementation will be done in WildFly and we should not need to do anything to Undertow itself.
+
+A new `setting` resource will be introduced. The resource will be
+`/subsystem=undertow/server=default-server/host=default-host/setting=console-access-log`. This allows
+no changes to the current access logging and allows the console access logging to be used independently
+from the file or standard logging access logging.
```
jstourac: @jamezp Just want to note that there is already `use-server-log` attribute for `access-log` resource (false by default):
```
/subsystem=undertow/server=default-server/host=default-host/setting=access-log:write-attribute(name=use-server-log,value=true)
```
which effectively cease logging into a separate access-log file and emits access-log messages to the `server.log` and `console` instead. In a combination of JSON_FORMATTER used on `console-handler`, the access-log messages can be formatted into JSON with following result:
```
{"timestamp":"2019-04-15T10:11:14.258+02:00","sequence":129,"loggerClassName":"org.jboss.logging.Logger","loggerName":"io.undertow.accesslog","level":"INFO","message":"127.0.0.1 - - [15/Apr/2019:10:11:14 +0200] \"GET / HTTP/1.1\" 200 1496","threadName":"default task-2","threadId":255,"mdc":{},"ndc":"","hostName":"dhcp-10-40-4-237.brq.redhat.com","processName":"jboss-modules.jar","processId":17364}
```
This although cannot be separated for `access-log` messages only but for whole `console` output.

What is gonna be exact difference between what I have just described and what you propose here with new `console-access-log` resource? Shall we think about `use-server-log` attribute to be marked as deprecated as new resource will be effectively replacing it with more powerful configuration options?

jamezp: When the access-log writes logs with a standard logger it's just a flat message:
```
9:33:28,530 INFO  [io.undertow.accesslog] (default task-2) 127.0.0.1 - - [16/Apr/2019:09:33:28 -0700] "GET /bkg.gif HTTP/1.1" 200 51660
```
vs the JSON structured data
```
{"eventSource":"web-access","timestamp":"2019-04-16T09:55:15.109-07:00","remoteHost":"127.0.0.1","identUsername":null,"remoteUser":null,"dateTime":"[16/Apr/2019:09:55:15 -0700]","requestLine":["GET /bkg.gif HTTP/1.1","GET /bkg.gif HTTP/1.1"],"responseCode":"200","bytesSent":"51660"}
```

With the structured data something like the EFK stack (ElasticSearch, Fluentd, Kibana) can easily filter messages based on the structured data. When the data is flat it makes it a lot more difficult, if not impossible, to filter correctly.

jstourac: Thanks @jamezp, I can see the difference now - in my example the 'message' part was a single, not structured string.

jamezp:
@jstourac @fl4via I've got a question on how we should handle the fields/pattern for this. For the file it currently just uses a pattern like `%h %l %u %t %r %s %b %{i,Referer} %{i,User-Agent}`. The problem is that doesn't really provide a key for the value. What I'm currently using is that same pattern and using the `ExchangeAttributeBuilder.getName()` as the key. However if the value of `getName()` changes at some point the data would end up with a different key.

One option would be to switch to some kind of property format type of thing like:
```
attributes={"remoteHost"="%h", "identityUsername"="%l"}
```

I'm not sure how user friendly that is though.

The other option would be to limit it to known exchange attributes, but that would require a change every time a new attribute is added to Undertow itself or we'd have to allow for some kind of custom attribute.

undertow/WFLY-11031_Formatted_Access_Logs.adoc
```diff
@@ -0,0 +1,70 @@
...
+
+* OpenShift
+
+== Requirements
+
+=== Hard Requirements
+
+* The output must be well formed JSON.
+* The formatted message must be on a single line.
+* The output must be able to write to standard out. This is required for OpenShift.
```
bstansberry: Is stdout all it can write to? The "setting=console-access-log" resource name implies that.

If that's all it can write to, perhaps add a Non-Requirement re writing to the access log file or piping output through the logging subsystem to whatever handlers are there.

jamezp: That's not a bad idea. I'll add that thanks!

jamezp:
I've added a non-requirement to *not* remove the server name prefix from standard out for domain servers. This specific feature will likely not be used for domain servers. Unless a specific requirement comes in for this the change would be too invasive and not worth the backwards compatibility risk.

PR #180, opened 2019-02-24
[WFLY-6143] Ability to add a server-side configured Interceptor for t…

tadamski:
…he applications without changing the deployments

https://issues.jboss.org/browse/WFLY-6143

### DIFF
--- a/ejb/WFLY-6143-Ability-to-add-a-server-side-configured-Interceptor-for-the-applications-without-changing-the-deployments.adoc
+++ b/ejb/WFLY-6143-Ability-to-add-a-server-side-configured-Interceptor-for-the-applications-without-changing-the-deployments.adoc
@@ -1,0 +1,60 @@
+= Ability to add a server-side configured Interceptor for the applications without changing the deployments
+:author:            Tomasz Adamski
+:email:             tomasz.adamski@redhat.com
+:toc:               left
+:icons:             font
+:idprefix:
+:idseparator:       -
+
+== Overview
+The goal of this issue is to implement server side EJB interceptors that can be configured system-wide through ejb3 subsystem configuration.
+
+=== Issue
+* https://issues.jboss.org/browse/WFLY-6143
+
+=== Dev Contacts
+
+* mailto:tadamski@redhat.com
+
+=== QE Contacts
+
+=== Testing By
+[ ] Engineering
+
+[x] QE
+
+=== Affected Projects or Components
+EJB
+
+== Requirements
+* Interceptors are implemented and packed inside the custom modules
+* Interceptors should be configured in the subsystem configuration and not in deployments
+* There should be a number of binding points in which the interceptors should be executed (f.e. after security)
+
+== Solution draft
+
+EJB3 subsystem is extended with <server-interceptors> tags, which specifies modules that include the interceptors implementation, interceptors that are going to be plugged and a binding point for each of them.
+Sample XML fragment:
+
+[source,xml]
+----
+<server-interceptors>
+	<modules>
+		<module>org.abccorp:tracing-interceptors:1.0</module>
+		<module>org.abccorp:timers-interceptors:1.1</module>
+	<modules>
+	<interceptors>
+		<interceptor class="org.abccorp.TracingInterceptor" binding="post-security"/>
+		<interceptor class="org.abccorp.TimerInterceptors"/>
+	</interceptors>
+</server-interceptors>
+----
+
+In order to use the extension module must be added to application server (that is moved to $JBOSS_HOME/modules/system/layers/base) and server extension similiar to the one above must be added the the server.
+
+
+
+== Test Plan
+
+== Community Documentation
+Part of the PR.
--- a/ejb/WFLY-6143-Ability-to-add-a-server-side-configured-Interceptor-for-the-applications-without-changing-the-deployments.adoc
+++ b/ejb/WFLY-6143-Ability-to-add-a-server-side-configured-Interceptor-for-the-applications-without-changing-the-deployments.adoc
@@ -26,32 +26,102 @@ The goal of this issue is to implement server side EJB interceptors that can be 
 === Affected Projects or Components
 EJB
 
+=== WildFLY custom interceptors
+
+WildFly allows users to implement their own intereceptors as a part of the deployment (https://docs.wildfly.org/17/Developer_Guide.html#Container_interceptors). Most important characteristics of those are as follows:
+
+* interceptors are POJO classes marked whose interceptor methods are marked by annotations
+* interceptor provider can specify in the descriptor the rules to match interceptors with specific beans
+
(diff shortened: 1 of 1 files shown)


ejb/WFLY-6143-Ability-to-add-a-server-side-configured-Interceptor-for-the-applications-without-changing-the-deployments.adoc
```diff
@@ -0,0 +1,57 @@
...
+EJB3 subsystem is extended with <server-interceptors> tags, which specifies modules that include the interceptors implementation, interceptors that are going to be plugged and a binding point for each of them.
+Sample XML fragment:
+
+<server-interceptors>
+	<modules>
+		<module>org.abccorp:tracing-interceptors:1.0</module>
+		<module>org.abccorp:timers-interceptors:1.1</module>
+	<modules>
+	<interceptors>
+		<interceptor class="org.abccorp.TracingInterceptor" binding="post-security"/>
```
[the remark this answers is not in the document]
tadamski: @brianstansberry I'm OK with changing the so that the module is repeated in each interceptor tag. The management model would be as follows: I will add another child, server-interceptor, to ejb3 subsytem resource and each interceptor will be represented by it's own child resource.

Regarding other questions, I'm planning the following implementation: For each binding point create an interceptor (== org.jboss.invocation.Interceptor) implementation that would be responsible for executing customer interceptors for this particular binding. In the context of that:
1. Binding points: There can be arbitrary number of binding points. otoh when the above implementation is ready adding another binding point should be simple. As a result of that, my initial idea was to start with one binding point (f.e. "default", at the beginning of invocation chain) and later add binding points based on customer demand. This is not an ideal solution, but it seems to me better than choosing arbitrary group of points by myself.
2. Restarting the services: I'm still reserching the implementation but my initial preference would be to make the changes to server interceptors induce reload-required. Changing interceptors in the running server would imo induce subystantial implementation complexity (f.e. modifing deployment module dependcies in the flight, modifing components view in the flight with all the possible race conditions etc.) so imo we would need strong customer demand for that.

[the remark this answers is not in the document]
zhantemirov: @tadamski any update on this? Should I consider your last commentary as final implementation draft?

ejb/WFLY-6143-Ability-to-add-a-server-side-configured-Interceptor-for-the-applications-without-changing-the-deployments.adoc
```diff
@@ -0,0 +1,57 @@
...
+
+[x] QE
+
+=== Affected Projects or Components
+EJB
+
+== Requirements
+* Interceptors are implemented and packed inside the custom modules
+* Interceptors should be configured in the subsystem configuration and not in deployments
+* There should be a number of binding points in which the interceptors should be executed (f.e. after security)
```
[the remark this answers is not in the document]
bstansberry: Resolving: below it now says "Any change to server-interceptors attribute should move server to the reload-required state."

ejb/WFLY-6143-Ability-to-add-a-server-side-configured-Interceptor-for-the-applications-without-changing-the-deployments.adoc
```diff
@@ -0,0 +1,129 @@
...
+[ ] Engineering
+
+[x] QE
+
+=== Affected Projects or Components
+EJB
+
+=== WildFLY custom interceptors
+
+WildFly allows users to implement their own intereceptors as a part of the deployment (https://docs.jboss.org/author/display/WFLY10/Container+interceptors). Most important characteristics of those are as follows:
```
bstansberry: Please use https://docs.wildfly.org/17/Developer_Guide.html#Container_interceptors as the docs URL.

ejb/WFLY-6143-Ability-to-add-a-server-side-configured-Interceptor-for-the-applications-without-changing-the-deployments.adoc
```diff
@@ -0,0 +1,129 @@
...
+
+In order to keep programming model consistent server-side interceptors would use the same APIs (POJO with javax.interceptor annotation's). Contrarily to customer interceptors, server-interceptors must be decoupled from the deployments. As a result, server-interceptors would not provide the ability to match them with specific beans.
+
+
+== Requirements
+* Interceptors are implemented and packed inside the custom modules
+* Interceptors should be configured in the subsystem configuration and not in deployments
+
+== Nice to have requirements
+* There should be a number of binding points in which the interceptors should be executed (f.e. after security)
```
bstansberry: Per the "Binding points" section below, this should now be in a "Non-Requirement" section. IOW you are no longer considering this as a requirement for this iteration of this feature; you've decided against it.

ejb/WFLY-6143-Ability-to-add-a-server-side-configured-Interceptor-for-the-applications-without-changing-the-deployments.adoc
```diff
@@ -0,0 +1,129 @@
...
+        return invocationContext.proceed();
+    }
+}
+----
+
+=== Interceptor module
+Server interceptor classes should be located in a module and place in $JBOSS_HOME/modules directory.
+
+== Binding points
+I suggest not to implement binding points currently. Adding binding points is a simple fix but I would prefer not to overengineer here and and add them based on customer demand instead of arbitrary selection. Furthermore, custom interceptors do not implement them as well and if there is a customer demand for specific binding point then we may consider implementing them both in server and customer interceptors to keep models consistent.
```
bstansberry: There should be a hard requirement then specifying where these interceptors will be placed in the interceptor chain.

My assumption is the behavior will be consistent with Section 9.6.3 in https://docs.wildfly.org/17/Developer_Guide.html#Container_interceptors., i.e. before the WildFly provided interceptors. However, please clarify where these interceptors will be placed relative to the jboss-ejb3.xml Container interceptors described in that document.

ejb/WFLY-6143-Ability-to-add-a-server-side-configured-Interceptor-for-the-applications-without-changing-the-deployments.adoc
tadamski: fixed

ejb/WFLY-6143-Ability-to-add-a-server-side-configured-Interceptor-for-the-applications-without-changing-the-deployments.adoc
tadamski: fixed

tadamski:
@bstansberry I have fixed the document. Could you please accept this plan or is it something more that should be fixed?

PR #181, opened 2019-03-05
[WFCORE-4355] Add proposal for event-logging API.

jamezp:
https://issues.jboss.org/browse/WFCORE-4355

### DIFF
--- a/logging/WFCORE-4355-event-logger.adoc
+++ b/logging/WFCORE-4355-event-logger.adoc
@@ -1,0 +1,80 @@
+= Event Logger
+:author:            James R. Perkins
+:email:             jperkins@redhat.com
+:toc:               left
+:icons:             font
+:idprefix:
+:idseparator:       -
+
+== Overview
+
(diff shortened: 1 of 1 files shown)


jmesnil:
@jamezp I looked at https://github.com/wildfly/wildfly-core/pull/3700 and I was wondering if this belongs to wildfly-core codebase or should be a separate project (under the wildfly organization).

darranl:
+1 for the comment from @jmesnil - if this lives within WildFly Core then components that come in to WildFly Core would not be able to depend on it.

On the other hand maybe that is correct and the subsystem responsible for a component should then be required to handle wiring the two together.  If this did not live within WildFly Core would it have any meaning on the outside?

jamezp:
So far I'm not aware of anything outside of WildFly that would use this. The only component I can think of would be Elytron. We can move it to it's own project outside core, but I figured we could start in core and if needed we refactor it out later. I'm good with either option though.

jamezp:
I'm merging this because the WildFly Core PR was merged.

PR #184, opened 2019-03-06
[WFCORE-4394] Enhanced audit logging

justinmcook:
Jira: https://issues.jboss.org/browse/EAP7-699

### DIFF
--- a/elytron/ELY-1712-enhanced-audit-logging.adoc
+++ b/elytron/ELY-1712-enhanced-audit-logging.adoc
@@ -1,0 +1,84 @@
+= [WFCORE-4394] Enhanced Audit Logging - RFC support and Configuring Reconnects
+:author:            Justin Michael Cook
+:email:             jucook@redhat.com
+:toc:               left
+:icons:             font
+:idprefix:
+:idseparator:       -
+:issue-base-url:    https://issues.jboss.org/browse
+
+== Overview
+
+Currently, Audit Logging in Elytron does not have the same capabilities as Legacy Security. This task is to enhance Elytron Audit Logging to have support for RFC5424/RFC3164 and add the additional ability to configure reconnect attempts.
+
+== Issue Metadata
+
+=== Issue
+
+* https://issues.jboss.org/browse/WFCORE-4394[WFCORE-4394]
+
+=== Related Issues
+
+* https://issues.jboss.org/browse/EAP7-699[EAP7-699]
+
+=== Dev Contacts
+
+* mailto:{email}[{author}]
+
+=== QE Contacts
+
+* mailto:jkasik@redhat.com[Jan Kašík]
+
+=== Testing By
+
+[X] Engineering
+
+[ ] QE
+
+=== Affected Projects or Components
+
+* Security
+
+=== Other Interested Projects
+
+== Requirements
+
+=== Hard Requirements
+
+* Support for https://tools.ietf.org/html/rfc5424[RFC 5424] should be enhanced, according to the following parameter,
+which should be added to ```syslog-audit-log``` in the Elytron subsystem
+** An optional ```syslog-format``` attribute should be added which supports a value of ```RFC5424```
+** This attribute will also have ```RFC5424``` as the default value
+** A Rfc5424SyslogEvent should be added as an audit event, which will have the ```RFC5424``` format as a property
+
+* Support for https://tools.ietf.org/html/rfc3164[RFC 3164] should be added to ```syslog-audit-log``` in the Elytron
+subsystem
+** The ```syslog-format``` attribute described in the RFC 5424 section above should also support a value of
+ ```RFC3164```
+** A Rfc3164SyslogEvent should be added as an audit event, which will have the ```RFC3164``` format as a property
+
+* An initial message should be sent to the syslog server when the Elytron Subsystem syslog endpoint is created
+** This message will indicate that Elytron audit logging is enabled and the format it is enabled with
+
+* The ability to configure the amount of reconnect attempts should be added
+** This should be done by configuring the number of reconnect attempts to a syslog-server
+** This should be available as an attribute ```reconnect-attempts``` under ```syslog-audit-log``` in the Elytron
+subsystem
+** The possible values for this parameter would be: -1 - signifying indefinite attempts,
+0 - signifying no attempts, and any positive integer - signifying that nunmber of attempts
+** This would be an optional parameter and have a default value of 0
+** This parameter will only apply for logging to a syslog server with UDP, since TCP already has reconnections
+configured in the jboss-logmanager project
+** This parameter will only apply if an IOException, such as the destination being unreachable, occurs while sending.
+
+=== Nice-to-Have Requirements
+
+=== Non-Requirements
+
+== Test Plan
+
+Subsystem parsing and transformer tests will be added
+
+== Community Documentation
+
+These new Audit Logging capabilities will be documented in the WildFly documentation under Audit.


justinmcook:
Hi @jamezp please review, in particular about the integration with [WFCORE-4355](https://issues.jboss.org/browse/WFCORE-4355)

jamezp:
Looks go do me @justinmcook. I just want to note too if you think the API for WFLY-4355 will make this more difficult it's not required.

honza-kasik:
Hi @justinmcook , thank you! I took a look at this, and I think this is _a lot_ of features for a single RFE. We have to ensure, that all aspects mentioned in this analysis will be implemented and what is even more important: that, they will be tested. From a brief look at this I could identify several areas this could be split into:

- RFC 5424/RFC 3164 accordance
- User formatted strings
- `@Audit` annotation
- Criteria filtering
- ... (each point basically describes enough features for one RFE)

I just want to express my concerns and in my opinion, finer granularity would be beneficial here.

@jamezp @justinmcook  Is it a good idea to present one RFE this big? Is this just an analysis for one big umbrella RFE and there will be more smaller ones? Am I missing something? :)

justinmcook:
@jamezp I think using the API won't make it more difficult, or, at worst, just slightly more difficult

justinmcook:
@honza-kasik It is somewhat large, and I think it was briefly discussed about splitting the RFE into smaller ones, if you think that would be beneficial then I can look into doing so and keeping this/ELY-1712 as the umbrella issue/analysis

jamezp:
@justinmcook Just an FYI I've submitted a [PR](https://github.com/wildfly/wildfly-core/pull/3700) for WFLY-4355. Hopefully it will end up looking like this and will be usable for you. If you feel there needs to be any changes please let me know.

elytron/ELY-1712-enhanced-audit-logging.adoc
```diff
@@ -0,0 +1,283 @@
...
+*** message format
+*** hostname (?)
+*** ssl-context
+** Additional attributes which will need to be added are:
+*** include timeQuality
+*** include origin
+*** include meta
+*** facility values
+*** priority mapping from facility and severity values, in accordance with https://tools.ietf.org/html/rfc5424#page-9[RFC 5424 Section 6.2.1]
+** The following attributes will also be available, but should be discarded if timeQuality is not specified:
```
honza-kasik: How this will be done? Using _requires_ relationship in model?

elytron/ELY-1712-enhanced-audit-logging.adoc
```diff
@@ -0,0 +1,283 @@
...
+*** setting hostname
+** The current attributes available in the Elytron schema, which will be used for RFC 5424, are:
+*** server-address
+*** port
+*** transport type
+*** message format
+*** hostname (?)
+*** ssl-context
+** Additional attributes which will need to be added are:
+*** include timeQuality
```
honza-kasik: I see `timeQuality` is name used in RFC 5424. What will be the name of this attribute in WF model?

elytron/ELY-1712-enhanced-audit-logging.adoc
```diff
@@ -0,0 +1,283 @@
...
+*** setting TCP
+*** setting SSL
+*** setting socket factory
+*** setting hostname
+** The current attributes available in the Elytron schema, which will be used for RFC 5424, are:
+*** server-address
+*** port
+*** transport type
+*** message format
+*** hostname (?)
```
honza-kasik: Why the question mark? I see, that we already have `host-name` attribute in model (`/subsystem=elytron/syslog-audit-log=*`)

elytron/ELY-1712-enhanced-audit-logging.adoc
```diff
@@ -0,0 +1,283 @@
...
+=== Other Interested Projects
+
+== Requirements
+
+=== Hard Requirements
+
+* Support for RFC 5424 should be enhanced, according to the following parameters, all of which should be added to the ```syslogAuditLogType``` in the Elytron schema
+// Could not find information about this in legacy
+** A ```syslog-format``` attribute should be added which supports a value of ```RFC5424```
+** Some of the parameters needed for RFC 5424 already exist in the Elytron codebase, which are the following:
```
honza-kasik: Do they also exists in WildFly model?

elytron/ELY-1712-enhanced-audit-logging.adoc
```diff
@@ -0,0 +1,283 @@
...
+
+* Security
+
+=== Other Interested Projects
+
+== Requirements
+
+=== Hard Requirements
+
+* Support for RFC 5424 should be enhanced, according to the following parameters, all of which should be added to the ```syslogAuditLogType``` in the Elytron schema
```
honza-kasik: This RFE targets Wildfly. Where in the WildFly will be those parameters added? By parameters you mean attributes?

elytron/ELY-1712-enhanced-audit-logging.adoc
```diff
@@ -0,0 +1,283 @@
...
+*** ip
+*** enterpriseId
+*** software
+*** swVersion
+** The following attributes will also be available, but should be discarded if meta is not specified:
+*** sequenceId
+*** sysUpTime
+*** language
+
+* Support for RFC 3164 should be added, based off of the implementation for RFC 5424 above, as such should also be added to the ```syslogAuditLogType``` in the Elytron schema
```
honza-kasik: I see, that RFC 5424 obsoletes RFC 3164. Do we want to support both? Also, again, where in WildFly (from user POV) will this be added?

elytron/ELY-1712-enhanced-audit-logging.adoc
```diff
@@ -0,0 +1,283 @@
...
+** The ```syslog-format``` attribute described in the RFC 5424 section above should also support a value of ```RFC3164```
+** All parts of RFC 3164 are included in the RFC 5424 specification, as such the support for RFC 3164 should be only
+changing the message format and not require any additional implementation
+** The implementation of the RFC 3164 message format can be implemented based on the https://github.com/kabir/tamper-detecting-audit-log/blob/master/src/main/java/org/jboss/audit/log/TcpSyslogHandler.java[tamper-detecting-audit-log]
+
+* Support for user formatted strings should be added
+** Currently, this feature does not have a way to be implemented
```
honza-kasik: Will be this somehow improved? Can you please elaborate?

elytron/ELY-1712-enhanced-audit-logging.adoc
```diff
@@ -0,0 +1,283 @@
...
+
+* There should be the ability to specify a Java class should be logged by using an @Audit Java annotation, similarly to what was available in Picketbox
```
honza-kasik: Can you please elaborate on this? This analysis will be also used as a resource for documentation and it is not clear for me how the `@Audit` annotation should be used.

elytron/ELY-1712-enhanced-audit-logging.adoc
```diff
@@ -0,0 +1,283 @@
...
+
+* There should be the ability to configure routing, and create a custom filter allowing only a selection of events to be logged, based on several criteria, listed in the below table
```
honza-kasik: This is about custom message format or introducing some filtering system into what will be logged?

elytron/ELY-1712-enhanced-audit-logging.adoc
```diff
@@ -0,0 +1,283 @@
...
+** In each case, the user should be able to specify the criteria, the value(s), and a list of configuration changes from the default values, such as a different destination
+** Specifying the criteria, the value(s), and the configuration changes should be done as follows:
+*** When the programmatic approach is used, a method call should be available to the endpoint which will accept a given criteria, the corresponding values, and the filtering options
+*** When the markup approach is used, the inputs should be specified via attributes ```criteria="supported_criteria"```, ```values="respective_value_input"```, ```filtering="option1=change,option2=change,..."```.
```
honza-kasik: Where will be the markup approach used? How can user use it?

elytron/ELY-1712-enhanced-audit-logging.adoc
```diff
@@ -0,0 +1,283 @@
...
+
+* Support for user formatted strings should be added
+** Currently, this feature does not have a way to be implemented
+
+* There should be the ability to specify a Java class should be logged by using an @Audit Java annotation, similarly to what was available in Picketbox
+
+* There should be the ability to configure routing, and create a custom filter allowing only a selection of events to be logged, based on several criteria, listed in the below table
+** In each case, the user should be able to specify the criteria, the value(s), and a list of configuration changes from the default values, such as a different destination
+** Specifying the criteria, the value(s), and the configuration changes should be done as follows:
+*** When the programmatic approach is used, a method call should be available to the endpoint which will accept a given criteria, the corresponding values, and the filtering options
```
honza-kasik: When can user use programmatic approach?

elytron/ELY-1712-enhanced-audit-logging.adoc
```diff
@@ -0,0 +1,283 @@
...
+| Values ranging from 0 to 191, calculated by multiplying the Facility value by 8 and adding the Severity value, as described in https://tools.ietf.org/html/rfc5424#page-9[RFC 5424 Section 6.2.1]
+| Two int values, first being the lower value and second being the higher value, for programmatically, lower value and higher value separated by a comma for markup
+
+| Level
+| FATAL, ERROR, WARN, INFO, DEBUG, TRACE
+| Java String List for programmatically, comma separated list for markup
+
+|===
+
+* Additional audit events should be added for feature parity with Legacy Security Audit Logging. The following tables makes a comparison of Legacy Security Audit Logging features included in, not included in, and if the feature should be added to Elytron Audit Logging:
```
honza-kasik: What do you mean by adding event? Where they will be added?

elytron/ELY-1712-enhanced-audit-logging.adoc
```diff
@@ -0,0 +1,283 @@
...
+
+|===
+* Additional customization attributes should be added based on community feedback, listed in the following table:
+|===
+|Feature |Available in Legacy Auditing |Support Case if Applicable |Implementation Required
+
+| "hostname" attribute
+| No
+| https://access.redhat.com/support/cases/#/case/02188481[Case 02188481]
+| Currently supported in Elytron Audit Logging resulting in just needing the attribute to be added to the management configuration
```
honza-kasik: I already see `host-name` attribute in syslog configuration in Elytron model. How this will be different?

elytron/ELY-1712-enhanced-audit-logging.adoc
```diff
@@ -0,0 +1,283 @@
...
+
+|===
+
+=== Nice-to-Have Requirements
+
+=== Non-Requirements
+
+== Test Plan
+
+Tests for the new Audit Logging capabilities will be added.
```
honza-kasik: Which tests and where they should be added?

elytron/ELY-1712-enhanced-audit-logging.adoc
```diff
@@ -0,0 +1,283 @@
...
+----
+... criteria="facility" values="0,15" filtering="location='/home/audit/logs',suffix='.yyyy-MM-dd_HHmm'" ...
+----
+
+* Logging should be done by using the proposal API as described in https://github.com/wildfly/wildfly-proposals/pull/181[WFCORE-4355]
+** A StdoutAuditEndpoint should be added to WildFly-Core that uses this API to log an event to Stdout
+** An AysncAuditEndpoint should be added to WildFly-Core that uses this API to log events to a queue before they're writing to Stdout
+** JsonSecurityEventFormatter should be moved to WildFly-Core and modified to use the JsonFormatter in this API
+** The current FileAuditEndpoints, FileAuditEndpoint, PeriodicRotatingFileAuditEndpoint, and SizeRotatingFileAuditEndpoint
+should be moved to WildFly-Core
```
honza-kasik: What changes will this introduce to model (if any)?

elytron/ELY-1712-enhanced-audit-logging.adoc
```diff
@@ -0,0 +1,283 @@
...
+*** setting socket factory
+*** setting hostname
+** The current attributes available in the Elytron schema, which will be used for RFC 5424, are:
+*** server-address
+*** port
+*** transport type
+*** message format
+*** hostname (?)
+*** ssl-context
+** Additional attributes which will need to be added are:
```
honza-kasik: Please include type of attributes which will be added (`String`, `int`...).

honza-kasik:
@justinmcook Hi Justin! I have made a small review of your analysis. WDYT?

justinmcook:
@honza-kasik Hi! Sorry I haven't responded to your comments yet, I plan to reply to them by the end of the day

elytron/ELY-1712-enhanced-audit-logging.adoc
justinmcook: I was thinking to have these available as methods within the parent method, such that they would only ever be available if the parent is included. I.e. (...).timeQuality(...).setTimeZone(...).build().(...)

elytron/ELY-1712-enhanced-audit-logging.adoc
justinmcook: I plan to use the same names as within the RFC, so, for example, timeQuality will be called timeQuality within the model.

elytron/ELY-1712-enhanced-audit-logging.adoc
justinmcook: Mistake on my part, I added the question as a reminder to look into this more, relating to a support issue, but forgot to remove the question mark and so will do so in the next update

elytron/ELY-1712-enhanced-audit-logging.adoc
justinmcook: Yes, they are currently available in the Elytron subsystem AuditResourceDefinitions

elytron/ELY-1712-enhanced-audit-logging.adoc
justinmcook: I added this based on other audit logging in WildFly, such as [tamper-detecting-audit-log](https://github.com/kabir/tamper-detecting-audit-log/blob/1db363a0abb5599a88d7043eb728e8bf0791757c/src/main/java/org/jboss/audit/log/TcpSyslogHandler.java#L163), having support for both. I wasn't sure however on if this should be ultimately included, what are your thoughts on including it vs not including it? As above, it would be added to the Elytron subsystem in WildFly-Core

elytron/ELY-1712-enhanced-audit-logging.adoc
justinmcook: This is something that was requested in the RFC, but I'm unsure what it would refer to exactly. I included it in the proposal for feedback on what should be added for this capability

elytron/ELY-1712-enhanced-audit-logging.adoc
justinmcook: This is one of the feature parity with legacy auditing parts. As far as I was able to tell, in PicketBox, the @Audit annotation can be added onto a class to signify that instances of that class being called should be logged. For the resource in Picketbox, see: https://developer.jboss.org/wiki/PicketBoxSecurityAnnotations#jive_content_id_SecurityAudit_Annotation

elytron/ELY-1712-enhanced-audit-logging.adoc
justinmcook: This refers to a filtering system

elytron/ELY-1712-enhanced-audit-logging.adoc
justinmcook: By markup, I refer to using the XML files, i.e. standalone.xml. This would be similiar to the management audit logging described in https://docs.jboss.org/author/display/WFLY8/Audit+logging and also similar to the following code snippet:
`<audit-logging><file-audit-log format="JSON" name="local-audit" path="audit.log" relative-to="jboss.server.log.dir"/></audit-logging>`

elytron/ELY-1712-enhanced-audit-logging.adoc
justinmcook: This would be available to any Java program that is extending wildfly-elytron-audit, similarly to the Picketbox example in https://developer.jboss.org/wiki/PicketBoxAudit

elytron/ELY-1712-enhanced-audit-logging.adoc
justinmcook: By adding events I mean adding additional options for events that can be logged and they would coexist with the other event types in org.wildfly.security.auth.server.event and be added to the formatters

elytron/ELY-1712-enhanced-audit-logging.adoc
justinmcook: Originally I included this because of the support case, but, looking back at it, I think this is out of scope for this RFE and so I'll remove this section.

elytron/ELY-1712-enhanced-audit-logging.adoc
justinmcook: The tests will be added in the same package as where the code is added and will verify that the added Endpoints/formatters/etc work correctly, I will put additional detail for this in the next update

elytron/ELY-1712-enhanced-audit-logging.adoc
justinmcook: I don't think this would introduce any changes to users other than additional Endpoints being available

elytron/ELY-1712-enhanced-audit-logging.adoc
justinmcook: Some of the attributes I won't know what class they would be until I start implementing it, but I'll define the ones I can

elytron/ELY-1712-enhanced-audit-logging.adoc
honza-kasik: How it will be done from users point of view? I don't think, that user will manipulate with source code. Or am I wrong? Sorry if I misunderstood you.

elytron/ELY-1712-enhanced-audit-logging.adoc
honza-kasik: What I am looking for is description how can user benefit from this new "feature". IMHO it doesn't matter how it will be implemented, rather than how the user can benefit from it.

elytron/ELY-1712-enhanced-audit-logging.adoc
honza-kasik: Thank you. I know, it seems clear to you and it might seemingly be clear to me, but this analysis must be comprehensible for a new comer. For someone who have never used features provided in legacy audit log configuration. For example, it is possible, that technical writers who will be processing this document will not know anything about it, because they have no experience.

elytron/ELY-1712-enhanced-audit-logging.adoc
honza-kasik: Thank you for explanation. However, editing of XML files no longer recommended way how to configure the server. Could you please instead of XML reference the model changes?

elytron/ELY-1712-enhanced-audit-logging.adoc
honza-kasik: Thanks! Where/How can user access them?

honza-kasik:
@justinmcook Thank you again, Justin. I replied to some of your comments, yet I need to discuss few things which I do not know about so much with Martin (who has sickday today).

elytron/ELY-1712-enhanced-audit-logging.adoc
honza-kasik: From what I know (and the check-up with Martin), the preferred names for model are not camel case variants but names which has words separated with dashes. Like this: `time-quality`.

elytron/ELY-1712-enhanced-audit-logging.adoc
honza-kasik: Thanks, and are they available to user?

elytron/ELY-1712-enhanced-audit-logging.adoc
honza-kasik: How and where can users access them?

elytron/ELY-1712-enhanced-audit-logging.adoc
honza-kasik: IMHO this needs further investigation. Do customers use syslog defined by one or another RFC in their infrastructure? Would they welcome both formats?

elytron/ELY-1712-enhanced-audit-logging.adoc
honza-kasik: In that case, can you please explain in this analysis how the user will configure it and how it will be introduced to the WildFly model?

elytron/ELY-1712-enhanced-audit-logging.adoc
honza-kasik: I am sorry, but I still don't understand what does that mean from users point of view.

elytron/ELY-1712-enhanced-audit-logging.adoc
justinmcook: Sorry, I misunderstood originally. Yes, I would use the requires relationship

elytron/ELY-1712-enhanced-audit-logging.adoc
justinmcook: I was thinking about within the code. Within the model, I would use the dashes version of the word like in your example

elytron/ELY-1712-enhanced-audit-logging.adoc
justinmcook: The attributes set by the user are available, I will clarify in the proposal which ones are available to the user (i.e. TCP) and which ones are internal/based off of events (i.e. severity)

elytron/ELY-1712-enhanced-audit-logging.adoc
justinmcook: These attributes will be available under the 'syslog-audit-log' on the CLI

elytron/ELY-1712-enhanced-audit-logging.adoc
justinmcook: I will discuss this with support

elytron/ELY-1712-enhanced-audit-logging.adoc
justinmcook: I will discuss this with support and @darranl

elytron/ELY-1712-enhanced-audit-logging.adoc
justinmcook: I'll add the details from this comment to the proposal for it is better clarified

elytron/ELY-1712-enhanced-audit-logging.adoc
justinmcook: Sure, I'll add it with the next update

elytron/ELY-1712-enhanced-audit-logging.adoc
justinmcook: Ah okay, I didn't realize this was no longer recommended (this is my first time modifying the model in detail). I'll change this section to reference the model changes

elytron/ELY-1712-enhanced-audit-logging.adoc
justinmcook: For the user, it would just affect the types of events that get logged when they look at their log. It would also tie into the filtering system above in having more options for the types of events that the user wishes to log/not log

elytron/ELY-1712-enhanced-audit-logging.adoc
justinmcook: These would be additional attributes in the model, I'll add in their names to the proposal

justinmcook:
Updated the PR to reflect the splitting of the task

honza-kasik:
@justinmcook Thank you for the update. I would prefer talking about changes introduced to the  elytron subsystem in the model -- newly introduced attributes and  operations which I can use to configure anything I need. Talking about model also allows us to talk about whether the attribute is required, its default value, capabilities and so on.

Also, I, as a  user use only supported ways of configuration such as CLI and/or web console, not editing XML files directly (this is an implementation detail which can change anytime). This analysis should describe changes made to the model. Changes which are available to use by user. Not implementation details and/or changes in private API.

elytron/ELY-1712-enhanced-audit-logging.adoc
```diff
@@ -0,0 +1,72 @@
...
+** This attribute will also have ```RFC5424``` as the default value
+** A Rfc5424SyslogEvent should be added as an audit event, which will trigger when the Elytron Subsystem attempts to log to a syslog server with the format ```RFC5424```
+
+* Support for RFC 3164 should be added to ```syslog-audit-log``` in the Elytron subsystem
+** The ```syslog-format``` attribute described in the RFC 5424 section above should also support a value of ```RFC3164```
+** A Rfc3164SyslogEvent should be added as an audit event, which will trigger when the Elytron Subsystem attempts to log to a syslog server with the format ```RFC3164```
+
+* The ability to configure for less reliability for increased speed should be added
+** This should be done by configuring the number of reconnect attempts to a syslog-server
+** This should be available as an attribute ```reconnect-attempts``` under ```syslog-audit-log``` in the Elytron subsystem
```
fjuma: Will this new attribute have a default value?

elytron/ELY-1712-enhanced-audit-logging.adoc
```diff
@@ -0,0 +1,72 @@
...
+
+* Security
+
+=== Other Interested Projects
+
+== Requirements
+
+=== Hard Requirements
+
+* Support for RFC 5424 should be enhanced, according to the following parameter,
```
fjuma: Just a minor comment, it would be good to add links to the RFCs.

elytron/ELY-1712-enhanced-audit-logging.adoc
justinmcook: Updated the document to reference this

elytron/ELY-1712-enhanced-audit-logging.adoc
justinmcook: Added the links

justinmcook:
@honza-kasik Updated the document

darranl:
As the analysis documents are analysis documents and not design documents I don't think it is always true that model changes can be fully defined, an engineer might have an idea how this will be represented in the model but this can evolve during design and implementation.  Regarding documentation by producing community documentation as the feature is merged we have then provided an account describing the end result of the work - at that stage the changes being made are definitely known.

I would also say the implementation detail if it is known is a valuable part of the analysis documents, as these pull requests are sent in these analysis documents are available to the wider engineering team who may also need an input on how a feature is to be developed - by omitting implementation details those engineers don't see anything to comment on the approach until the final pull requests are sent in which is too late in the process.

honza-kasik:
@justinmcook Thank you very much for the updated analysis! This look good from the QA point of view!

elytron/ELY-1712-enhanced-audit-logging.adoc
```diff
@@ -0,0 +1,73 @@
...
+:email:             jucook@redhat.com
+:toc:               left
+:icons:             font
+:idprefix:
+:idseparator:       -
+:issue-base-url:    https://issues.jboss.org/browse
+
+== Overview
+
+Currently, Audit Logging in Elytron does not have the same capabilities as Legacy Security. This task is to enhance Elytron Audit Logging to have the same capabilities as Legacy Security Audit Logging.
```
honza-kasik: @justinmcook Just one last change. Can you please edit the overview so it reflects current state of this analysis?

honza-kasik:
Sorry @justinmcook , I missed the overview - can you please edit it? Otherwise, this looks good!

elytron/ELY-1712-enhanced-audit-logging.adoc
```diff
@@ -0,0 +1,73 @@
...
+** A Rfc5424SyslogEvent should be added as an audit event, which will trigger when the Elytron Subsystem attempts to log to a syslog server with the format ```RFC5424```
+
+* Support for https://tools.ietf.org/html/rfc3164[RFC 3164] should be added to ```syslog-audit-log``` in the Elytron subsystem
+** The ```syslog-format``` attribute described in the RFC 5424 section above should also support a value of ```RFC3164```
+** A Rfc3164SyslogEvent should be added as an audit event, which will trigger when the Elytron Subsystem attempts to log to a syslog server with the format ```RFC3164```
+
+* The ability to configure for less reliability for increased speed should be added
+** This should be done by configuring the number of reconnect attempts to a syslog-server
+** This should be available as an attribute ```reconnect-attempts``` under ```syslog-audit-log``` in the Elytron subsystem
+** This would be an optional parameter and not have a default value
```
honza-kasik: Just a very last thing. Wouldn't it be better if this has a default value of current count of reconnect attempts and making this a required attribute?

honza-kasik:
@justinmcook I am sorry for last moment edits, but can you please edit the overview and WDYT about `recconect-attempts` comment?

justinmcook:
@honza-kasik Updating the overview now. For the reconnect-attempts: I think updating it to still be an optional attribute but with a default value would be best, because there would always have to be a specified number of reconnect attempts, but the attribute would only be needed if a user wished to customize that. On the same note, I will update syslog-format to also be an optional attribute, since it would default to RFC5424 and the attribute would only need to be added by a user if they wished to use RFC3164 (or any other RFC if any new ones are created in the future)

elytron/ELY-1712-enhanced-audit-logging.adoc
justinmcook: Edited the overview

elytron/ELY-1712-enhanced-audit-logging.adoc
justinmcook: Updated this section to still be optional but have a default value of 0

honza-kasik: Thank you. Sorry, I meant "required" in sense, that it will have default value presenting current amount of reconnect attempts. Users won't have to set anything if they don't want to. If it will have default value of `0`, does it mean, that there will be no reconnect attempts made, or there will be infinite amount of them?

justinmcook: By 0 I mean there would be no reconnect attempts

honza-kasik: So how many reconnect attempts is done now?

justinmcook: Currently it indefinitely attempts to reconnect

justinmcook:
@darranl @jamezp I was looking further into configuring the amount of reconnect attempts, and it looks like to be able to do this would require modifying the [TcpOutputStream class in jboss-logmanager](https://github.com/jboss-logging/jboss-logmanager/blob/master/ext/src/main/java/org/jboss/logmanager/ext/handlers/TcpOutputStream.java) for that the [RetryConnector](https://github.com/jboss-logging/jboss-logmanager/blob/master/ext/src/main/java/org/jboss/logmanager/ext/handlers/TcpOutputStream.java#L336) can take in the number of maximum attempts as an argument, rather than it currently being indefinite. Do you guys think making those modifications to jboss-logmanager would be worth the ability to configure the maximum number of reconnect attempts?

elytron/ELY-1712-enhanced-audit-logging.adoc
honza-kasik: In that case, wouldn't be the value of `-1` best? If user wishes to change the behavior, he can set any positive number (thus swapping reliability for speed).

justinmcook: Ya that would be better, I'll change it to -1

justinmcook: Updated the description

elytron/ELY-1712-enhanced-audit-logging.adoc
```diff
@@ -0,0 +1,76 @@
...
+** A Rfc5424SyslogEvent should be added as an audit event, which will trigger when the Elytron Subsystem attempts to log to a syslog server with the format ```RFC5424```
+
+* Support for https://tools.ietf.org/html/rfc3164[RFC 3164] should be added to ```syslog-audit-log``` in the Elytron subsystem
+** The ```syslog-format``` attribute described in the RFC 5424 section above should also support a value of ```RFC3164```
+** A Rfc3164SyslogEvent should be added as an audit event, which will trigger when the Elytron Subsystem attempts to log to a syslog server with the format ```RFC3164```
+
+* The ability to configure for less reliability for increased speed should be added
+** This should be done by configuring the number of reconnect attempts to a syslog-server
+** This should be available as an attribute ```reconnect-attempts``` under ```syslog-audit-log``` in the Elytron subsystem
+** The possible values for this parameter would be: -1 - signifying indefinite reconnect attempts,
```
honza-kasik: @justinmcook From our discussion I understood that currently there is only on reconnect attempt made (and error is ignored). From my point of view we should keep backward compatibility and keep this as is by default. Can you please find out if current behavior is intended (if it's not a bug)?

cc @fjuma

justinmcook:
Updated the analysis document to reflect changes/discussions. In particular, the change of the default value for ```reconnect-attempts``` to 0 and noting that it only applies to UDP.

elytron/ELY-1712-enhanced-audit-logging.adoc
```diff
@@ -0,0 +1,80 @@
...
+** This message will indicate that Elytron audit logging is enabled and the format it is enabled with
+
+* The ability to configure for less reliability for increased speed should be added
+** This should be done by configuring the number of reconnect attempts to a syslog-server
+** This should be available as an attribute ```reconnect-attempts``` under ```syslog-audit-log``` in the Elytron subsystem
+** The possible values for this parameter would be: -1 - signifying indefinite reconnect attempts,
+0 - signifying no reconnect attempts, and any positive integer - signifying that nunmber of
+reconnect attempts
+** This would be an optional parameter and have a default value of 0
+** This parameter will only apply for logging to a syslog server with UDP, since TCP already has reconnections configured in the jboss-logmanager project
```
honza-kasik: Thanks for the update @justinmcook! Is there a way how user can configure the reconnect attempts for TCP connection?

justinmcook:
Updated the analysis document to reflect discussions with @honza-kasik, in particularly changing the attribute ```reconnect-attempts``` to ```icmp-error-reconnect-attempts``` and updating its description

honza-kasik:
Thanks for the update @justinmcook. Please also add information why the attribute was named like this (ICMP error).

justinmcook:
Updated the analysis document explaining that reconnection will be attempted if there is an ```IOException``` and changed the attribute back to ```reconnect-attempts``` from ```icmp-error-reconnect-attempts``` to reflect there being no guarantee that the ```IOException``` being due to an ICMP error

PR #187, opened 2019-03-08
[WFLY-11824] WildFly Operator for Kubernetes/OpenShift

jmesnil:
JIRA: https://issues.jboss.org/browse/WFLY-11824

### DIFF
--- a/cloud/WFLY-11824_wildfly_operator_kubernetes.adoc
+++ b/cloud/WFLY-11824_wildfly_operator_kubernetes.adoc
@@ -17,8 +17,8 @@ Managing an application running on WildFly in a cloud environment can be a compl
 
 === Related Issues
 
-* https://issues.jboss.org/browse/EAP7-1214[EAP7-1214] - Tech Preview Operator for handling already built EAP/WildFly based images
-
+* https://issues.jboss.org/browse/EAP7-1149[EAP7-1149] - Tech Preview Operator for handling already built EAP/WildFly based images
+* https://issues.jboss.org/browse/CLOUD-2262[CLOUD-2262] - Provide support for transaction recovery for transactions which contain subordinate transactions propagated over JBoss Remoting
 === Dev Contacts
 
 * mailto:{email}[{author}]
@@ -35,6 +35,7 @@ Managing an application running on WildFly in a cloud environment can be a compl
 
  * https://github.com/wildfly/wildfly[WildFly]
  * https://github.com/openshift-s2i/s2i-wildfly[WildFly s2i]
+ * https://www.github.com/jboss-dockerfiles/wildfly[WildFly Docker image]
 
 
 === Other Interested Projects
@@ -56,19 +57,20 @@ The CRD will allow the user to:
 
 * Specify the name an application image that incorporates a WildFly server and an deployment that will be deployed by that server.
 ** The name can include either a fixed or floating tag.
-* Specify an optional ConfigMap where the WildFly standalone configuration can be read from.
+** The application image is built on top of the https://www.github.com/jboss-dockerfiles/wildfly[WildFly Docker image]
 * Specify the desired number of replicas of that image
+* Specify an optional ConfigMap where the WildFly standalone configuration can be read from.
 * Specify whether the replicas are intended to form a High Availability cluster (default is true).
-* Specify whether the containers are intended to be 'stateful'
-** default is to be determined (might be inferred from the server configuration)
+* Specify the type of storage it needs
+  * it must define a `PersistentVolumeClaim` if it requires persistent storage.
+  * otherwise a `EmptyDir` wil be associated to the container (without persistence)
 * Specify information for interfaces for which a Service and Route should be established
-** perhaps expose 8080 by default with an option to disable that
-** same for 9990 to expose management interface (including health check and metrics endpoints)
+** expose http port (8080) by default with an option to disable that
 * Specify information for interfaces for which a headless service should be established.
 
 ### Operator requirements
 
-The WildFly operation will provide the following features:
+The WildFly operator will provide the following features:
 
 * In reaction to deployment of a Custom Resource with the CRD's type start up a cluster of containers running the specified image, with the desired number of replicas.
 * If the containers are meant to form a High Availability cluster, perform scale up and scale down in a manner designed to optimize cluster efficiency
@@ -78,9 +80,6 @@ The WildFly operation will provide the following features:
 ** If the necessary facilities to allow this are not available, container start should be aborted
 ** Appropriate transaction recovery activities should be performed
 * Establish the appropriate services and routes
-* Detect new versions of the image in the container catalog and take appropriate action. Possible actions include
-** Scale down to a min number of nodes, begin replacing old nodes with new until all old gone
-** Leave old and new both running in some sort of A/B scenario
 * Incorporate information about active sessions in all of this (this requires LB integration that is out of scope unless already present)
 
 Another requirement for this operator is to provide a better Observability of EAP such as:
@@ -97,8 +96,8 @@ Another requirement for this operator is to provide a better Observability of EA
 
 === Non-Requirements
 
- * Orchestration of the building of images or the creation of Custom Resource instances. The images are available in the container catalog; how they get there is out of scope for this operator.
- * Facilitating operation of a container that embeds a messaging broker within the WildFly server (e.g. by ensuring it has access to a persistent volume). Running an embedded broker within WildFly in the cloud is not recommended. Use an external messaging broker.
+* Orchestration of the building of images or the creation of Custom Resource instances. The images are available in the container catalog; how they get there is out of scope for this operator.
+* Facilitating operation of a container that embeds a messaging broker within the WildFly server (e.g. by ensuring it has access to a persistent volume). Running an embedded broker within WildFly in the cloud is not recommended. Use an external messaging broker.
 
 == Implementation Plan
 
@@ -106,6 +105,28 @@ Another requirement for this operator is to provide a better Observability of EA
 * Develop a WildFly operator based on the https://github.com/operator-framework[Operator Framework].
 
 There is an existing WildFly operator at https://github.com/banzaicloud/wildfly-operator but the codebase for it will not be used.
+
+== Implementation Details
+
+This operator defines the expected behaviour of a WildFly application deployed on Kubernetes and OpenShift.
+
+One of the key decision is the underlying kind of resources to represent the application.
+Currently, the https://github.com/jboss-container-images/jboss-eap-7-openshift-image/blob/89f045c458e1d196c6c4288b34ae5aae4ab58934/templates/eap-cd-basic-s2i.json#L298[EAP S2I templates] uses https://kubernetes.io/docs/concepts/workloads/controllers/deployment/[`DeploymentConfig`].
+However this introduces issues for transaction recovery (such as https://issues.jboss.org/browse/CLOUD-2262[CLOUD-2262]) as the pod names are not stable.
+
+The operator will instead use a https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/[`StatefulSet`] to manage the WildFly applications.
+Compared to using `DeploymentConfig`, this ensures:
+
+* stable, unique network identifiers.
+* stable, persistent storage.
+* ordered, graceful deployment and scaling.
+* ordered, automated rolling updates.
+
+These properties are prerequisites for transaction recovery and ensuring an efficient cluster formation if the application requires clustering with JGroups.
+
+The Operator will let the user configure a https://kubernetes.io/docs/concepts/storage/volumes/#persistentvolumeclaim[`PersistentVolumeClaim`] to define its requirements in terms of persistent storage.
+
+In the case where the application is not stateful and do not requires persistent storage, the Operator will allow to use an https://kubernetes.io/docs/concepts/storage/volumes/#emptydir[`EmptyDir`] that will not be persisted across pod restart.
 
 == Test Plan
 
@@ -121,4 +142,4 @@ There is an existing WildFly operator at https://github.com/banzaicloud/wildfly-
 
 == Community Documentation
 
-The wildfly-operator project will have dedicated documentation about installing and using the WildFly operator with use cases for all the different features (configuration, HA, transaction recovery, etc.)
+The wildfly-operator project will have dedicated https://github.com/wildfly/wildfly-operator/tree/master/doc[documentation] about installing and using the WildFly operator with use cases for all the different features (configuration, HA, transaction recovery, etc.)
--- a/cloud/WFLY-11824_wildfly_operator_kubernetes.adoc
+++ b/cloud/WFLY-11824_wildfly_operator_kubernetes.adoc
@@ -57,7 +57,9 @@ The CRD will allow the user to:
 
 * Specify the name an application image that incorporates a WildFly server and an deployment that will be deployed by that server.
 ** The name can include either a fixed or floating tag.
-** The application image is built on top of the https://www.github.com/jboss-dockerfiles/wildfly[WildFly Docker image]
+** Application images must be built on top of [WildFly S2I](https://github.com/wildfly/wildfly-s2i) starting with WildFly 17.
+** The Operator should be able to manage operatios built with the [legacy WildFly S2I for WildFly 16](https://github.com/openshift-s2i/s2i-wildfly)
+** Application images built on top of older WildFly versions are not supported.
 * Specify the desired number of replicas of that image
 * Specify an optional ConfigMap where the WildFly standalone configuration can be read from.
 * Specify whether the replicas are intended to form a High Availability cluster (default is true).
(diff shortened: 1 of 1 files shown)


cloud/WFLY-11824_wildfly_operator_kubernetes.adoc
```diff
@@ -0,0 +1,126 @@
...
+** same for 9990 to expose management interface (including health check and metrics endpoints)
+* Specify information for interfaces for which a headless service should be established.
+
+### Operator requirements
+
+The WildFly operation will provide the following features:
+
+* In reaction to deployment of a Custom Resource with the CRD's type start up a cluster of containers running the specified image, with the desired number of replicas.
+* If the containers are meant to form a High Availability cluster, perform scale up and scale down in a manner designed to optimize cluster efficiency
+** Start one container and then start the others, thus ensuring the later nodes see the first as the group coordinator, avoiding mergers.
```
luck3y: This might be addressed if we use stateful sets if the config is deemed to be clustered / HA etc. Probably needs some testing.

cloud/WFLY-11824_wildfly_operator_kubernetes.adoc
```diff
@@ -0,0 +1,126 @@
...
+
+* Codebase will be hosted in a new GitHub repository at https://github.com/wildfly/wildfly-operator.
+* Develop a WildFly operator based on the https://github.com/operator-framework[Operator Framework].
+
+There is an existing WildFly operator at https://github.com/banzaicloud/wildfly-operator but the codebase for it will not be used.
+
+== Test Plan
+
+* Simple scenario by using an image based on a JAX-RS web application
+** create a customer resource for WildFly CRD type
```
scottmarlow: Should customer be custom?

cloud/WFLY-11824_wildfly_operator_kubernetes.adoc
```diff
@@ -0,0 +1,126 @@
...
+
+==== Create a Custom Resource Definition (CRD) for WildFly.
+
+Provide a Custom Resource Definition (CRD) for an application based on a WildFly application server.
+
+The CRD will allow the user to:
+
+* Specify the name an application image that incorporates a WildFly server and an deployment that will be deployed by that server.
+** The name can include either a fixed or floating tag.
+* Specify an optional ConfigMap where the WildFly standalone configuration can be read from.
```
jorgemoralespou: Should configuration be in one single configmap? If in a single configmap, should it be multiple entries or just one single entry? IMHO when the configuration is moved to configmaps (one or multiple, one entry per CM or multiple) it needs to be simplified and written in an easy user editable form (e.g. yaml, key=value), not xml.

cloud/WFLY-11824_wildfly_operator_kubernetes.adoc
```diff
@@ -0,0 +1,126 @@
...
+* Specify the desired number of replicas of that image
+* Specify whether the replicas are intended to form a High Availability cluster (default is true).
+* Specify whether the containers are intended to be 'stateful'
+** default is to be determined (might be inferred from the server configuration)
+* Specify information for interfaces for which a Service and Route should be established
+** perhaps expose 8080 by default with an option to disable that
```
jorgemoralespou: In this case, if only allowing secure access, certificate management need to be considered as part of the operator.

cloud/WFLY-11824_wildfly_operator_kubernetes.adoc
```diff
@@ -0,0 +1,126 @@
...
+** same for 9990 to expose management interface (including health check and metrics endpoints)
```
jorgemoralespou: Should management be accessible from outside the cluster? IMHO the wildfly operator should contribute a HAL console to the OpenShift as an extension and to a kubernetes cluster as a standalone management interface, so that users can manage the cluster in an easy/centralized/aggregated way. HAL should be the one providing proxied/aggregated APIs from the outside world? Why, because this way you can provide aggregated information from wildfly clusters, and not having to aggregate information on the client side, if every instance exposes its own management endpoint.

cloud/WFLY-11824_wildfly_operator_kubernetes.adoc
```diff
@@ -0,0 +1,126 @@
...
+
+* In reaction to deployment of a Custom Resource with the CRD's type start up a cluster of containers running the specified image, with the desired number of replicas.
+* If the containers are meant to form a High Availability cluster, perform scale up and scale down in a manner designed to optimize cluster efficiency
+** Start one container and then start the others, thus ensuring the later nodes see the first as the group coordinator, avoiding mergers.
+* If the resources are configured as stateful
+** Ensure the existence of the necessary persistent volume(s) and persistent volume claims necessary for each container to have access to its own writable persistence
+** If the necessary facilities to allow this are not available, container start should be aborted
+** Appropriate transaction recovery activities should be performed
+* Establish the appropriate services and routes
+* Detect new versions of the image in the container catalog and take appropriate action. Possible actions include
```
jorgemoralespou: This should not happen automatically, but should only happen by the user triggering a change in the version of the CR defining the application. Automatic deployments based on an image available on a registry are only acceptable in certain development scenarios.

cloud/WFLY-11824_wildfly_operator_kubernetes.adoc
```diff
@@ -0,0 +1,126 @@
...
+** Scale down to a min number of nodes, begin replacing old nodes with new until all old gone
+** Leave old and new both running in some sort of A/B scenario
+* Incorporate information about active sessions in all of this (this requires LB integration that is out of scope unless already present)
+
+Another requirement for this operator is to provide a better Observability of EAP such as:
```
jorgemoralespou: Observability, as mentioned before should be provided by a HAL console that aggregates cluster information and provides proper RBAC controls. This is not exclusive to what's described below.

cloud/WFLY-11824_wildfly_operator_kubernetes.adoc
```diff
@@ -0,0 +1,124 @@
...
+
+### Operator requirements
+
+The WildFly operation will provide the following features:
+
+* In reaction to deployment of a Custom Resource with the CRD's type start up a cluster of containers running the specified image, with the desired number of replicas.
+* If the containers are meant to form a High Availability cluster, perform scale up and scale down in a manner designed to optimize cluster efficiency
+** Start one container and then start the others, thus ensuring the later nodes see the first as the group coordinator, avoiding mergers.
+* If the resources are configured as stateful
+** Ensure the existence of the necessary persistent volume(s) and persistent volume claims necessary for each container to have access to its own writable persistence
```
ochaloup: do we consider the JDBC storage option too?

david-rh:
I would like to propose an option where a JBoss CLI script may be added via Config Map and that the script would configure the standalone.xml. This way the developer doesn't have to provide a complete hardcoded standalone.xml doc.

cloud/WFLY-11824_wildfly_operator_kubernetes.adoc
jmesnil: @hpehl fiy. Depending on how HAL can be integrated with OpenShift, the operator would have to be enhanced/configured to take HAL into account.

hpehl: I think the operator should have such an option. The way how HAL is integrated into OpenShift is still wip, but the operator should definitely make it possible to include HAL and make it accessible from outside the cluster.

cloud/WFLY-11824_wildfly_operator_kubernetes.adoc
hpehl: This will require a new HAL console which is able to aggregate such cluster information and take RBAC into account.

bstansberry: We need to be clear about what RBAC we are talking about.

cloud/WFLY-11824_wildfly_operator_kubernetes.adoc
bstansberry: Good point.

fmarchioni: +1
In terms of configuration, a ConfigMap can be represented also as YAML file.
Could we adopt the same format used by Thorntail (project-defaults.yml) to wrap configuration bits in YAML format? Ideally a sort of Profile cloner (like https://github.com/tfonteyn/profilecloner but XML->YAML) could be re-used both by Thorntail applications and for bootstrapping the Operator.

fmarchioni:
I'm conveying here some concerns raised by the QA Team about the WildFly/EAP Operator:
- How is CP team going to maintain the operator for CP streams?
- Are operators extendable? E.g. can a customer extend the operator provided by Red Hat, in order to add/modify functionality? This ties in to the first question - our customers could expect us to have the operators written in java.
- How many operators are there going to be in the operator registry? One operator for EAP, or one operator per EAP image stream?
- What's the support policy for operators? Related to questions 1, and 3.
- How abstract are operators going to be? E.g. if I want to configure HA, do I need to specify the configuration, or will the operator take care of the configuration for me?
- Do the operators depend on the environment? E.g. if I want to configure persistent volumes, do I need to worry if NFS or Cinder is available, or does the operator take care of that for me?
- What versions of OpenShift do the operators target? I assume this is 4.1, but it's good to be certain.

fmarchioni:
Here is what I've figured out (inline answers). Please outline any point that needs to be revised: 

_1, Why is the EAP operator being written in Go? We are supposedly one of the few, or possibly the only team in middleware, that does not use java API to write the operator._

Some Operators are written in Java like this one:
https://github.com/strimzi/strimzi-kafka-operator - this one is Java based.
It might be that the SDK to write operators is not the same as the SDK to write tests for operators. We should probably get a clear description of what exists and what's supported.

_2, How is CP team going to maintain the operator for CP streams?_
Currently Operators are not defined as a deliverable for any CP stream (6.4.z, 7.2.z).

_3, Are operators extendable? E.g. can a customer extend the operator provided by Red Hat, in order to add/modify functionality? This ties in to the first question - our customers could expect us to have the operators written in java._

Looking at the OperatorHub, I cannot see any Extendable Operator. I can see that you can instead define Custom Resources in it to customize the Operator (See for example https://operatorhub.io/operator/camel-k-operator)

_4, How many operators are there going to be in the operator registry? One operator for EAP, or one operator per EAP image stream?_

In other Operators, there’s one Container Image for each Operator (for WildFly: quay.io/wildfly/wildfly-operator:0.1.0 )

_5, What's the support policy for operators?_ 
As said, there is no support for CP streams and it’s tech preview for EAP 7.3 

_6, How abstract are operators going to be? E.g. if I want to configure HA, do I need to specify the configuration, or will the operator take care of the configuration for me?_

In the Clustering example provided, it is specified both the application (Image) and the standalone-openshift.xml configuration in a ConfigMap:


apiVersion: wildfly.org/v1alpha1
kind: WildFlyServer
metadata:
  name: clusterbench
spec:
  applicationImage: "quay.io/jmesnil/clusterbench-ee7:16.0"
  size: 2
  standaloneConfigMap:
    name: clusterbench-config-map
    key: standalone-openshift.xml



_7, Do the operators depend on the environment? E.g. if I want to configure persistent volumes, do I need to worry if NFS or Cinder is available, or does the operator take care of that for me?_

I don’t believe that the operator depends at all on the environment as it can only create a PVC. You can then allocate space with kubectl/oc


_8, What versions of OpenShift do the operators target? I assume this is 4.1, but it's good to be certain._
Operators are fully supported in Openshift 4 and beyond

fmarchioni:
Also, after discussing with CSA (thanks clichybi@redhat.com) to have also pointers from the Customer side, I'd like to stress the importance to group requirements into the Kubernates Operator Maturity Model.

Phase I Basic Install
Automated application provisioning and configuration management

Phase II Seamless Upgrades
Patch and minor version upgrades supported

Phase III Full Lifecycle
App lifecycle, storage lifecycle (backup, failure recovery)

Phase IV Deep Insights
Metrics, alerts, log processing and workload analysis

Phase V Auto Pilot
Horizontal/Vertical scaling, Auto config Tuning, abnormal detection, Scheduling tuning

If we map the requirements with this schema, it will be easier for the Customer to understand at first glance what the Operator can do for you, simpler for us to create a Test Plan and we won't risk to miss features on the go. For example, if we now include these features in the plan:
_App lifecycle, storage lifecycle (backup, failure recovery)_
Then I'd expect we target the "Phase III Full Lifecycle". On the other hand I couldn't find in the Requirements information about the earlier phase (Phase II Seamless Upgrades)

jmesnil:
> Here is what I've figured out (inline answers). Please outline any point that needs to be revised:
> 
> _1, Why is the EAP operator being written in Go? We are supposedly one of the few, or possibly the only team in middleware, that does not use java API to write the operator._

The [https://github.com/operator-framework/operator-sdk](operator-sdk) is the most mature way to write operators.
There are pros and cons to use this SDK as opposed to a home-made Java API but the main advantage is that it is bound to the k8s API directly which significantly reduces the time to market to write and maintain an operator.

The language of the operator itself is an implementation detail. The users will install it from a Docker image from operatorhub.io and it make no difference to the behaviour whether it is based on Go, Ansible or something else.

> _3, Are operators extendable? E.g. can a customer extend the operator provided by Red Hat, in order to add/modify functionality? This ties in to the first question - our customers could expect us to have the operators written in java._

Operators are not meant to be extendable.
If users requires new features from the operator, we will add them to the operator (provided they kale sense).

> _6, How abstract are operators going to be? E.g. if I want to configure HA, do I need to specify the configuration, or will the operator take care of the configuration for me?_

The operator is closely tied to the WildFly S2I (source-to-image) to provide its features.
WildFly S2I is responsible to start WildFly with the expected features according to the env vars provided by the container.
For the time being, the WildFly Operator requires the users to set most of these properties by hand (e.g. he has to specify [4 env vars to use a PostgreSQL database](https://github.com/wildfly/wildfly-operator/blob/6c7bef2a6c543ba2bf974a435fd51222503262bb/examples/postgresql/crds/taskrs-app.yaml#L8-L27))

Some of these often used features (e.g. clustering and datasource) could benefit from a higher level API from the Operator specification.
But for the time being, we decided to expose env vars directly from the operator to the S2I images.
(the only exception is some properties for KUBE_PING that are setup by the operator based on the CRD metadata)
 
> In the Clustering example provided, it is specified both the application (Image) and the standalone-openshift.xml configuration in a ConfigMap:
> 
> apiVersion: wildfly.org/v1alpha1
> kind: WildFlyServer
> metadata:
> name: clusterbench
> spec:
> applicationImage: "quay.io/jmesnil/clusterbench-ee7:16.0"
> size: 2
> standaloneConfigMap:
> name: clusterbench-config-map
> key: standalone-openshift.xml

WildFly S2I for WildFly 16 uses the standalone.xml configuration by default. Any application requires additional subsystems (such as clustering) must bring its own standalone configuration.
This will change with WildFly S2I for WildFly 17 and the clustering example will be significantly simpler (the user would have to specify a few env vars to configure JGroups with KUBE_PING for example).

> _7, Do the operators depend on the environment? E.g. if I want to configure persistent volumes, do I need to worry if NFS or Cinder is available, or does the operator take care of that for me?_

The operator will not install any persistent volumes. They must be provided by the container platform.
However, CRDs managed by the operator can specify their [storage requirements](https://github.com/wildfly/wildfly-operator/blob/master/doc/apis.adoc#storagespec) to make sure that the persistent volumes provided by the platform meet them.

jmesnil:
> Also, after discussing with CSA (thanks [clichybi@redhat.com](mailto:clichybi@redhat.com)) to have also pointers from the Customer side, I'd like to stress the importance to group requirements into the Kubernates Operator Maturity Model.
> 
> Phase I Basic Install
> Automated application provisioning and configuration management
> 
> Phase II Seamless Upgrades
> Patch and minor version upgrades supported
> 
> Phase III Full Lifecycle
> App lifecycle, storage lifecycle (backup, failure recovery)
> 
> Phase IV Deep Insights
> Metrics, alerts, log processing and workload analysis
> 
> Phase V Auto Pilot
> Horizontal/Vertical scaling, Auto config Tuning, abnormal detection, Scheduling tuning

The Operator documentation implies that phases are linear but that does not fit properly the WildFly operator.

The WildFly operator actually manages an application image built on top of WildFly.
That means that Phase II might not be as seamless as expected the users will have to patch its application image on top of a patched WildFly image before the operator can upgrade it.

On the other hand, the WildFly operator already provide features from Phase IV (e.g. metrics).

@fmarchioni I agree with you that each feature of the operator should state the phase it belongs to.

cloud/WFLY-11824_wildfly_operator_kubernetes.adoc
```diff
@@ -0,0 +1,147 @@
...
+The CRD will allow the user to:
+
+* Specify the name an application image that incorporates a WildFly server and an deployment that will be deployed by that server.
+** The name can include either a fixed or floating tag.
+** Application images must be built on top of [WildFly S2I](https://github.com/wildfly/wildfly-s2i) starting with WildFly 17.
+** The Operator should be able to manage operatios built with the [legacy WildFly S2I for WildFly 16](https://github.com/openshift-s2i/s2i-wildfly)
+** Application images built on top of older WildFly versions are not supported.
+* Specify the desired number of replicas of that image
+* Specify an optional ConfigMap where the WildFly standalone configuration can be read from.
+* Specify whether the replicas are intended to form a High Availability cluster (default is true).
```
mchoma: Which attribute from https://github.com/wildfly/wildfly-operator/blob/master/doc/apis.adoc is that?

PR #188, opened 2019-03-15
[WFCORE-3947] Support SSL Certificate revocation using OCSP

nekdozjam:
https://issues.jboss.org/browse/WFCORE-3947
https://issues.jboss.org/browse/EAP7-715

### DIFF
--- a/elytron/WFCORE-3947-ocsp-ssl-revocation.adoc
+++ b/elytron/WFCORE-3947-ocsp-ssl-revocation.adoc
@@ -1,0 +1,123 @@
+= OCSP certificate revocation
+:author:            Martin Mazanek
+:email:             mmazanek@redhat.com
+:toc:               left
+:icons:             font
+:idprefix:
+:idseparator:       -
+
+== Overview
+
+== Issue Metadata
+
+=== Issue
+
+* https://issues.jboss.org/browse/WFCORE-3947[WFCORE-3947]
+
+=== Related Issues
+
+* https://issues.jboss.org/browse/ELY-1617[ELY-1617]
+
+=== Dev Contacts
+
+* mailto:mmazanek@redhat.com[Martin Mazanek]
+
+=== QE Contacts
+
+* mailto:jstourac@redhat.com[Jan Stourac]
+
+=== Affected Projects or Components
+
+* Security
+
+=== Other Interested Projects
+
+== Requirements
+
+=== Hard Requirements
+
+* New class X509RevocationTrustManager
+* Configurable certificate revocation using OCSP. This will be implemented using PKIXRevocationChecker.
+* Configuration will be done using Builder class and will include:
+** Current configuration in X509CRLExtendedTrustManager
+** OCSP Responder URI for overriding default, where the default is retrieved from AIA extension (RFC5280) or from `ocsp.responderURL` security property
+** Default revocation mechanism (OCSP/CRL) in a form of prefer-crls (by default OCSP if preferred)
+** Fallback behaviour - cases for unavailable CRL / OCSP responder services
+** Checking of leaf certificates only
+* Deprecating X509CRLExtendedTrustManager, both mechanisms will share one class for fallback purposes and easier configuration
+
+Current subsystem CRL configuration:
+
+```xml
+<trust-manager name="trust-with-ocsp-crl" algorithm="PKIX" key-store="ElytronCaTruststore">
+    <certificate-revocation-list path="crl.pem" maximum-cert-path="2"/>
+</trust-manager>
+```
+
+Proposed subsystem configuration:
+
+```xml
+<trust-manager name="trust-with-ocsp-crl" algorithm="PKIX" key-store="ElytronCaTruststore" soft-fail="true" only-leaf-cert="true" maximum-cert-path="2">
+    <certificate-revocation-list path="crl.pem"/>
+    <ocsp responder="responderUri" prefer-crls="true" responder-certificate="certalias" responder-keystore="responder-store"/>
+</trust-manager>
+```
+
+
+New subsystem configuration elements:
+
+ * `trust-manager` elements:
+ ** `soft-fail` can pass certificates, where revocation status could not be obtained due to network or internal responder error. This is optional false by default.
+ ** `only-leaf-cert` will check revocation status of only the leaf certificate. This is optional and false by default.
+ ** `maximum-cert-path` moved from `certificate-revocation-list` to `trust-manager`, as this can be used by OCSP without CRL configured.
+ * `ocsp` element:
+ ** `responder` this overrides OCSP Responder URI resolved from certificate.
+ ** `prefer-crls` when both OCSP and CRL is configured, OCSP will be executed first by default. When `prefer-crls` is true, CRL will be checked first instead.
+ ** `responder-keystore` alternative keystore for responder certificate. `responder-certificate` must be defined.
+ ** `responder-certificate` alias for responder certificate located int `responder-keystore` or `trust-manager` keystore if responder-keystore is not defined
+Order of revocation mechanisms (if both are defined) will be based on attribute of `ocsp` - `prefer-crls`, defaults to false (OCSP prioritized by default).
+
+Other configuration is done using security properties of elytron subsystem, see: https://docs.oracle.com/javase/8/docs/technotes/guides/security/certpath/CertPathProgGuide.html#AppC
+
+
+==== Elytron client configuration
+
+Current elytron-client configuration:
+
+```xml
+<ssl-context name="ssl-crl-context">
+    <trust-store key-store-name="ca"/>
+    <certificate-revocation-list path="crl.pem" maximum-cert-path=4/>
+</ssl-context>
+```
+
+
+Proposed elytron-client configuration:
+
+```xml
+<ssl-context name="ssl-crl-context">
+    <trust-store key-store-name="ca"/>
+    <trust-manager algorithm="PKIX" maximum-cert-path=4 soft-fail="true" only-leaf-cert="true"/>
+    <certificate-revocation-list path="crl.pem"/>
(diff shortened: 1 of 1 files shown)


elytron/WFCORE-3947-ocsp-ssl-revocation.adoc
```diff
@@ -0,0 +1,81 @@
...
+Proposed configuration:
+
+```xml
+<trust-manager name="trust-with-ocsp-crl" algorithm="PKIX" key-store="ElytronCaTruststore">
+    <certificate-revocation-list path="crl.pem" maximum-cert-path="2"/>
+    <ocsp responder="responderUrl"/>
+    <soft-fail/>
+    <only-end-entity/>
+    <prefer-crls/>
+</trust-manager>
```
jstourac: What is the behavior here? I am a little bit confused ATM. Both OCSP and CRL are set - thanks to `perfer-crl`, the CRL check is performed first, in case of failure then the OCSP and in case of another failure thanks to the 'soft-fail' element the request passes, correct?

Is it a `prefer-crl` or a `prefer-crls`?

elytron/WFCORE-3947-ocsp-ssl-revocation.adoc
```diff
@@ -0,0 +1,81 @@
...
+
+* Security
+
+=== Other Interested Projects
+
+== Requirements
+
+=== Hard Requirements
+
+* New class RevocatingTrustManager
```
fjuma: Just minor, perhaps ```RevocationTrustManager``` is a better name here?

elytron/WFCORE-3947-ocsp-ssl-revocation.adoc
```diff
@@ -0,0 +1,75 @@
...
+
+```xml
+<trust-manager name="trust-with-ocsp-crl" algorithm="PKIX" key-store="ElytronCaTruststore">
+    <certificate-revocation crl-path="crl.pem" maximum-cert-path="2", key-store=trustedKS, soft-fail="true", prefer-crl="false", ocsp-responder="responderUrl"/>
+</trust-manager>
+```
+
+This would include both CRL and OCSP configuration in one thing, as they can't be configured separately due to sharing some parameters (eg. trusted key-store).
+Note: it is basically just refactoring and expanding existing CRL configuration, so we could stick to using element certificate-revocation-list for both CRL and OCSP to remain existing config,
+but I think refactoring it with just certificate-revocation will be less confusing. On top of that the behavior will slightly change, because certificate-revocation will enable OCSP by default.
```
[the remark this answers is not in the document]
nekdozjam: This is from outdated AD, at first I didn't realize that trust-store is shared within the trust-manager element, so no existing configuration needs to change

elytron/WFCORE-3947-ocsp-ssl-revocation.adoc
```diff
@@ -0,0 +1,75 @@
...
+
+Other configuration is done globally on JVM scope, as documented in http://openjdk.java.net/jeps/249[JEP-249]
```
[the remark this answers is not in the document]
nekdozjam: It would make sense, but as it is not in JDK8 it could be confusing, as it would not work as configured. So I am not really sure with this one

elytron/WFCORE-3947-ocsp-ssl-revocation.adoc
nekdozjam: Yes, that is the scenario. Basically if you can't get a response with status of the certificate, the request passes. This is usually the default behaviour of browsers, if they have revocation checking enabled. It is not really secure, but that is why default behaviour would be hard fail, where you will deny the request if you cannot get the status.

prefer-crls, so it reflects the Java API

elytron/WFCORE-3947-ocsp-ssl-revocation.adoc
nekdozjam: Probably yes, will change

elytron/WFCORE-3947-ocsp-ssl-revocation.adoc
```diff
@@ -0,0 +1,75 @@
...
+== Requirements
+
+=== Hard Requirements
+
+* New class RevocatingTrustManager
+* Configurable certificate revocation using OCSP. This will be implemented using PKIXRevocationChecker.
+* Configuration will be done using Builder class and will include:
+** Current configuration in X509CRLExtendedTrustManager
+** OCSP Responder URI for overriding default, where the default is retrieved from AIA extension (RFC5280) or from 'ocsp.responderURL' security property
+** Default revocation mechanism (OCSP/CRL) in a form of prefer crl (default false)
```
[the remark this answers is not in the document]
nekdozjam: By default OCSP is preffered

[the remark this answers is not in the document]
jstourac: okay, could you please alternate the sentence a little bit then? e.g. something like
`Default revocation mechanism (OCSP/CRL) can be set via an prefer-crls resource (false by default)`

elytron/WFCORE-3947-ocsp-ssl-revocation.adoc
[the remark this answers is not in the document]
jstourac: I think that in case we properly document this behaviour (in doc and also in an attribute/element description), it is not a problem. On JDK8 those values are simply ignored (optionally some warning may be generated for customer to be aware). There is already similar such thing for TLSv1.3 `cipher-suite-names` attribute - as TLSv1.3 is available since JDK-11 (well, TLSv1.3 will probably available with JDK8 in case when appropriate OpenSSL provider is used instead of default JSSE, but it's a different story, though).

Also, I can see that there is already this feature request https://issues.jboss.org/browse/ELY-434 to implement OCSP stampling support into WildFly. So maybe we can keep this proposal simply on a base OCSP support basis and 'official' stampling feature being implemented by another proposal - ELY-434. In such case this decision has to be stated in this analysis document. WDYT?

elytron/WFCORE-3947-ocsp-ssl-revocation.adoc
```diff
@@ -0,0 +1,81 @@
...
+
+== Requirements
+
+=== Hard Requirements
+
+* New class RevocatingTrustManager
+* Configurable certificate revocation using OCSP. This will be implemented using PKIXRevocationChecker.
+* Configuration will be done using Builder class and will include:
+** Current configuration in X509CRLExtendedTrustManager
+** OCSP Responder URI for overriding default, where the default is retrieved from AIA extension (RFC5280) or from 'ocsp.responderURL' security property
```
jstourac: Speaking about possibility of configuring `ocsp.responderURL`, I think we should make possible to configure also other available JVM properties via our own configuration that are described in [Appendix C: On-Line Certificate Status Protocol (OCSP) Support](https://docs.oracle.com/javase/8/docs/technotes/guides/security/certpath/CertPathProgGuide.html#AppC).

nekdozjam: This is configurable using security-properties element of elytron subsystem. Or do you want a specific thing of configuring those related to ocsp?

jstourac: I see, still I think it makes sense to put it into the configuration directly under the introduced `ocsp` element. My motivation is driven by following two assumptions:
1. having it directly under the `ocsp` element is easier for user to realize what all can be configured actually and is more readable than in a global set of properties
1. `security-properties` under the `elytron` subsystem as I understand them, means global configuration, although here we are talking about local, per `trust-manager` configuration

elytron/WFCORE-3947-ocsp-ssl-revocation.adoc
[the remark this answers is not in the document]
nekdozjam: This is mean as a base OCSP support RFE, stamping has its own issues. I would like to have them separate, as the stamping is JDK9+ and everything proposed here is in JDK8. The stamping support would consist mostly of tests and docs, as the configuration can be made through System and Security properties, which require no code change. So it will be possible to use this JKD8 targeted RFE with stamping released in JDK9 JEP249,

elytron/WFCORE-3947-ocsp-ssl-revocation.adoc
nekdozjam: Sorry, my bad, I thought that these were global, but looking at the code it seems that they are loaded from the properties only at init, so it can be set up differently for each trust-manager. Will change, thanks for pointing this out

nekdozjam: However, after a second look it is basically implemented, as the API introduced in JDK8 [JEP124](http://openjdk.java.net/jeps/124) overrides these properties. There is a little change, where the `responderCert*` properties are passed as one certificate, so the certificate will be selected by alias instead of subject/issuer name or serial number to isolate it from the keystore, which is more user friendly. So the only reason why to use the properties would be to set up global defaults

rmartinc: Just a minor comment, I think it would be nice to give an option to configure the `responderCert`. This is usually something that is always requested by OCSP users (sometimes the certificate used by the responder is not the one that issued the final certificate being checked, and that certificate should be configured). In the java API is the method [setOcspResponderCert](https://docs.oracle.com/javase/8/docs/api/java/security/cert/PKIXRevocationChecker.html#setOcspResponderCert-java.security.cert.X509Certificate-). It could be just the alias in the trust-store configured (at EAP level I mean). For example in apache httpd the option is this one [SSLOCSPResponderCertificateFile](https://httpd.apache.org/docs/2.4/mod/mod_ssl.html#sslocsprespondercertificatefile) It's something that is always exposed by the configuration.

nekdozjam: Thanks for noticing that, I somehow missed it. It should be included now :)

PR #189, opened 2019-03-20
[WFLY-11868] RESTEasy integration with Elytron on the client side

Skyllarr:
JIRA: https://issues.jboss.org/browse/WFLY-11868

### DIFF
--- a/elytron/WFLY-11868_RESTEasy_client_integration_with_Elytron.adoc
+++ b/elytron/WFLY-11868_RESTEasy_client_integration_with_Elytron.adoc
@@ -1,0 +1,102 @@
+= WFLY-11868 REST integration with WildFly Elytron - AuthenticationClient for Authentication / SSL
+:author:            Diana Vilkolakova
+:email:             dvilkola@redhat.com
+:toc:               left
+:icons:             font
+:keywords:          elytron, client, resteasy
+:idprefix:
+:idseparator:       -
+
+== Overview
+
+WildFly Elytron uses the Elytron Client project to enable remote clients to authenticate using Elytron.
+Elytron client is already used by other WildFly clients allowing for both use of an Elytron API and pre-defined configurations in the xml file (https://docs.jboss.org/author/display/WFLY/WildFly+Client+Configuration#WildFlyClientConfiguration-wildflyconfig.xmlDiscovery[wildfly-config.xml]).
+
+This requested feature is about making use of Elytron's configuration on the client side of REST deployment so that the RESTEasy client supports the common framework available within the application server.
+
+The implementation will involve loading of configuration from Elytron client, specifically `credentials`, `bearer token` and `SSL context`, and making those available to the RESTEasy client. Credentials will be used for `HTTP Basic authentication` and bearer token for `Bearer Authentication`.
+
+Elytron configuration should be ignored if different configuration was already set. So if there was already Bearer token or credentials available this implementation will not overwrite it.
+
+== Issue Metadata
+
+=== Issue
+
+* https://issues.jboss.org/browse/WFLY-11868[WFLY-11868]
+
+=== Related Issues
+
+* https://issues.redhat.com/browse/EAP7-1219[EAP7-1219] - EAP issue tracker
+* https://issues.redhat.com/browse/ELY-1891[EAP7-1891] - Elytron project issue tracker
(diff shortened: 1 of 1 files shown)


elytron/WFLY-11868_RESTEasy_client_integration_with_Elytron.adoc
```diff
@@ -0,0 +1,86 @@
...
+
+== Issue Metadata
+
+=== Issue
+
+* https://issues.jboss.org/browse/WFLY-11697[WFLY-11868]
+
+=== Related Issues
+
+* https://issues.jboss.org/browse/EAP7-1097[EAP7-1219]
```
asoldano: Please fix the link, right one is likely https://issues.jboss.org/browse/EAP7-1219

elytron/WFLY-11868_RESTEasy_client_integration_with_Elytron.adoc
```diff
@@ -0,0 +1,86 @@
...
+
+Wildfly-config.xml file is used by Elytron client and it contains the authentication configuration that should be used
+when attempting to connect to the server (username, password, ssl config, mechanism specification etc).
+This requested feature is about making use of Elytron and wildfly-config.xml file on the client side of REST deployment.
+
+== Issue Metadata
+
+=== Issue
+
+* https://issues.jboss.org/browse/WFLY-11697[WFLY-11868]
```
asoldano: Please fix the link, right one is likely https://issues.jboss.org/browse/WFLY-11868

elytron/WFLY-11868_RESTEasy_client_integration_with_Elytron.adoc
```diff
@@ -0,0 +1,86 @@
...
+* mailto:dvilkola@redhat.com[Diana Vilkolakova]
+
+=== QE Contacts
+
+* TBD
+
+=== Affected Projects or Components
+
+* Security
+* Web Services
```
asoldano: Please replace Web Services with REST / RESTEasy

Skyllarr:
@asoldano I have fixed the issues you mentioned. I'll keep the backward compatibility in mind, thank you for the feedback.

marekkopecky:
@Skyllarr With current RESTEasy client approach, I can have two clients with different credentials and I can use these both clients for the requests for one server. Is this possible with wildfly-config.xml approach?

elytron/WFLY-11868_RESTEasy_client_integration_with_Elytron.adoc
```diff
@@ -0,0 +1,86 @@
...
+
+* https://issues.jboss.org/browse/EAP7-1219[EAP7-1219]
+
+=== Dev Contacts
+
+* mailto:dvilkola@redhat.com[Diana Vilkolakova]
+
+=== QE Contacts
+
+* TBD
```
marekkopecky: It's me -> mkopecky@redhat.com

marekkopecky:
`Testing By` section is missing.

Skyllarr:
> @Skyllarr With current RESTEasy client approach, I can have two clients with different credentials and I can use these both clients for the requests for one server. Will this be possible with wildfly-config.xml approach?

@marekkopecky wildfly-config.xml can have multiple configurations (with different credentials) defined. Rules are then applied to each request to conclude which config to use, and the rule can take eg. username into account as well ([relevant wildfly-config doc section](https://docs.jboss.org/author/display/WFLY/WildFly+Client+Configuration#WildFlyClientConfiguration-%3Cauthenticationrules%2F%3Eand%3Csslcontextrules%2F%3E)). I think this should be possible as long as there is a way to decide which client config to use when making a request.

Skyllarr:
Hello @ronsigal @darranl @marekkopecky , I have updated the analysis document. Please review. Thanks in advance!

Skyllarr:
> @Skyllarr With current RESTEasy client approach, I can have two clients with different credentials and I can use these both clients for the requests for one server. Will this be possible with wildfly-config.xml approach?

Hello @marekkopecky , I have read your question again and maybe can provide clearer answear. You can set different credentials in Elytron client for the same URL. But if there is no additional rule (eg. port) that specifies which credentials to choose ( I think is what you were asking), then the first credentials will be applied always (in order that they were configured). But I don't think this will result in backwards incompatibility, because user can specify their own credentials for every RESTEasy client like before. This will overwrite credentials that were set by Elytron. But I will keep this in mind during implementation.

PR #191, opened 2019-03-25
[WFLY-1160] Provide ability to easily apply certain JBoss module libraries to all deployments running in a server

yersan:
An RFE to allow the configuration of external folders with libraries shared across all deployments without specifying global modules.

Jira issue: https://issues.jboss.org/browse/WFLY-1160

### DIFF
--- a/ee/WFLY-1160-Provide_ability_to_configure_a_global_directory_which_puts_the_contents_in_the_classpath.adoc
+++ b/ee/WFLY-1160-Provide_ability_to_configure_a_global_directory_which_puts_the_contents_in_the_classpath.adoc
@@ -1,0 +1,151 @@
+= Provide ability to easily apply certain JBoss module libraries to all deployments running in a server
+:author:            Yeray Borges
+:email:             yborgess@redhat.com
+:toc:               left
+:icons:             font
+:idprefix:
+:idseparator:       -
+
+== Overview
+Wildfly users can share libraries across all deployed applications without having to modify the packaging of their applications creating a custom module and making it global using the EE subsystem:
+
+  [standalone@localhost:9990 /] module add --name=com.test --absolute-resources=/Users/me/libs:/Users/me/libs/my-utils1.jar:/Users/me/libs/my-utils2.jar
+  [standalone@localhost:9990 /] /subsystem=ee:list-add(name=global-modules,value={name=com.test})
+
+This configuration implies the following:
+
+  * Users have to specify all required resource roots using --absolute-resources argument.
+  * It is necessary to modify the generated module.xml, or at least re-create the global module if we want to add/remove any additional resource-root for this global module.
+  * This configuration allows any customization of the generated module.xml, making it very flexible to cover broader cases, for example, it is possible to declare module dependencies, mark them as exported, declare the main class and so on.
+  * This command is not appropriate for use in a managed domain or when connecting to the management CLI remotely. Modules should be added and removed manually in a production environment.
+
+This RFE is opened to have an alternative new operation to address simpler cases, where it is not required any additional configuration beyond on the default module.xml configurations, and supported in production environments and domain mode.
+
+The simplest way to accomplished this is to use the EE subsystem to configure a path to the designated directory that will contain the libraries we want to be shared across all deployed applications. A new resource named `global-directory` will be added to the EE subsystem. This resource will use two attributes path and relative-to to configure a global directory:
+
+  [standalone@localhost:9990 /] /subsystem=ee/global-directory=my-common-libs:add(path=lib, relative-to=jboss.home.dir)
+
+  [domain@localhost:9990 /] /profile=default/subsystem=ee/global-directory=my-common-libs:add(path=lib, relative-to=jboss.server.data.dir)
+
+The following attributes will be available on the global-directory resource:
+
+* path: The path of the directory to scan. (Mandatory)
+* relative-to: The name of another previously named path, or of one of the standard paths provided by the system. (Optional)
+
+When a new `global-directory` is created, the server will create a module that will have a https://jboss-modules.github.io/jboss-modules/manual/#path-resource-loader[Path Resource Loader] using 'path' and 'relative-to' attributes and one https://jboss-modules.github.io/jboss-modules/manual/#jar-resource-loader[Jar Resource loader] for each jar file included in this directory and subdirectories.
+
+The 'Path Resource Loader' will make available any file as a resource to the applications.
+The 'Jar Resource loader' will make available any class inside of the jar file to the applications.
+
+The jar resources of this module will be created iterating over all jar files found in the directory tree. Each directory level is scanned alphabetically starting from the root, and on each level each subdirectory is also explored alphabetically until visiting all the branch. Files found on each level are also added in alphabetical order. The resource root order declaration will govern the class loading order for the classes available from the jar resources of this module.
+
+To illustrate how the server will load the jar files, suppose you have the following directory tree:
+
+    /lib/Z/a-lib.jar
+    /lib/A/A/z-lib.jar
+    /lib/A/a-lib.jar
+    /lib/A/b-lib.jar
+    /lib/a-lib.jar
+    /lib/A/B/a-lib.jar
+
+The module generated internally by the global-directory operation will be equivalent to declare a module.xml with the following resources:
+
+  <resources>
+    <resource-root path="/lib"/>
+    <resource-root path="/lib/a-lib.jar"/>
+    <resource-root path="/lib/A/a-lib.jar"/>
+    <resource-root path="/lib/A/b-lib.jar"/>
+    <resource-root path="/lib/A/A/z-lib.jar"/>
+    <resource-root path="/lib/A/B/a-lib.jar"/>
+    <resource-root path="/lib/Z/a-lib.jar"/>
+  </resources>
+
+The module created from the shared directory will be loaded as soon as the first application is deployed in the server. That means, if the server is started/restarted and there are no applications deployed, then the shared directory is neither scanned nor the module loaded.
+
+In case of domain mode or distributed environments, it is the user responsibility to make the content of the configured global directory consistent across all the server instances, as well as distribute the jar files that they contain.
+
+Removal of a global-directory could be done using the following operation:
+
+  [standalone@localhost:9990 /] /subsystem=ee/global-directory=my-common-libs:remove()
+
+  [domain@localhost:9990 /] /profile=default/subsystem=ee/global-directory=my-common-libs:remove()
+
+As any other resources, users can read the current configuration of a global-directory using resource-resource operation:
+
+ [standalone@localhost:9990 /] /subsystem=ee/global-directory=my-common-libs:read-resource
+ {
+      "outcome" => "success",
+      "result" => {
+          "path" => "lib",
+          "relative-to" => "jboss.home.dir"
+      }
+ }
+
+ [domain@localhost:9990 /] /subsystem=ee/global-directory=my-common-libs:read-resource
+ {
+      "outcome" => "success",
+      "result" => {
+          "path" => "lib",
+          "relative-to" => "jboss.server.data.dir"
+      }
+ }
+
+This RFE will not cover the availability of hot JBoss modules redeployments, that means, modifications of existing jar files as well as adding a new jar or removal an existing one will require a complete server restart, however, changes in plain resources as property files will be available immediately to the applications.
+
+It will be allowed only one global directory configuration.
+
+== Issue Metadata
+
+=== Issue
+* https://issues.jboss.org/browse/EAP7-1124[EAP7-1124]
+* https://issues.jboss.org/browse/WFLY-1160[WFLY-1160]
+* https://issues.jboss.org/browse/WFCORE-4375[WFCORE-4375]
+
+=== Related Issues
+
+* https://issues.jboss.org/browse/MODULES-99[MODULES-99]
+
+=== Dev Contacts
+
+* mailto:{email}[{author}]
+
+=== QE Contacts
+
+* mailto:vmarek@redhat.com[Vratislav Marek]
+* mailto:tterem@redhat.com[Tomas Terem]
+
+
+=== Testing By
+[ ] Engineering
+
+[X] QE
+
+=== Affected Projects or Components
+Wildfly
+Wildfly-core
+
+=== Other Interested Projects
+
+== Requirements
+
+=== Hard Requirements
+- The new module created from the global-directory configuration will get automatically a dependency on `javaee.api` module name.
+- Hot JBoss modules redeployments will require a server restart to make them available to all existing and new applications.
+- New modules created from the content of the shared directory will be added to the deployed applications as a system dependencies.
+- Modifications of existing jar files as well as adding a new jar or removal an existing one will require a complete server restart.
+- It will not allowed the configuration of multiple global directories.
+
+=== Nice-to-Have Requirements
+- We could allow the possibility of adding existing module dependencies to the new module created from the global-directory configuration, and, if there is no dependencies configured, use by default `javaee.api`.
+
+=== Non-Requirements
+- Putting the module content on the filesystem of any machine is not a requirement; it is completely a user responsibility to distribute and make available the jar resources in the target machines.
+- Path validation on stage.MODEL beyond validating the 'relative-to' value references a valid path capability is not a requirement.
+
+== Test Plan
+
+Tests will be added to verify that an application gets in its class path resources from a shared directory.
+
+== Community Documentation
+
+The feature will be documented in WildFly Admin Guide, under Java EE Application Deployment section of EE Subsystem Configuration.


ee/WFLY-1160-Provide_ability_to_configure_a_global_directory_which_puts_the_contents_in_the_classpath.adoc
```diff
@@ -0,0 +1,86 @@
...
+:email:             yborgess@redhat.com
+:toc:               left
+:icons:             font
+:idprefix:
+:idseparator:       -
+
+== Overview
+Wildfly users can share libraries across all deployed applications without having to modify the packaging of their applications creating a custom module and making it global using the EE subsystem:
+
+  [standalone@localhost:9990 /] module add --name=com.test --absolute-resources=/Users/me/libs:/Users/me/libs/my-utils1.jar:/Users/me/libs/my-utils2.jar
```
vmarekrh: Can you please more highlights that "method add" not will be used in the final solution?

ee/WFLY-1160-Provide_ability_to_configure_a_global_directory_which_puts_the_contents_in_the_classpath.adoc
```diff
@@ -0,0 +1,86 @@
...
+
+The name and path attributes will be mandatory.
+Internally, this operation will create a module with the name `my-common-libs` that will have as resource-roots the folder `/Users/me/libs/` and all the jars files included in this folder. Subdirectories will not be scanned, but if it is required, users can create a new global-folder using the subdirectory path.
+
+Removal of a global-folder could be done using the following operation:
+
+  [standalone@localhost:9990 /] /subsystem=ee:list-remove(name=global-folders,value={name=my-common-libs})
+
+This RFE will not cover the availability of hot JBoss modules redeployments, that means, modifications of existing jar files as well as adding a new jar or removal an existing one will require a complete server restart, however changes in plain resources as property files will be available immediately to the applications.
+
```
vmarekrh: Can you please describe when while server starting it will be shared library loaded?

ee/WFLY-1160-Provide_ability_to_configure_a_global_directory_which_puts_the_contents_in_the_classpath.adoc
yersan: Added:

> This command is not appropriate for use in a managed domain or when connecting to the management CLI remotely. Modules should be added and removed manually in a production environment.

> This RFE is opened to have an alternative new operation to address simpler cases

Hi @vmarekrh , would you mind to read it again and give me feedback? I think with those sentences is clear we are not going to use module add for this feature.

ee/WFLY-1160-Provide_ability_to_configure_a_global_directory_which_puts_the_contents_in_the_classpath.adoc
yersan: Added:

> The module created from the shared folder will be loaded as soon as the first application is deployed in the server

ee/WFLY-1160-Provide_ability_to_configure_a_global_directory_which_puts_the_contents_in_the_classpath.adoc
vmarekrh: Hi, @yersan I was discussing with my colleague and it is only my point of view.
It is ok now.

ee/WFLY-1160-Provide_ability_to_configure_a_global_directory_which_puts_the_contents_in_the_classpath.adoc
vmarekrh: Look good.
Thank you @yersan

ee/WFLY-1160-Provide_ability_to_configure_a_global_directory_which_puts_the_contents_in_the_classpath.adoc
```diff
@@ -0,0 +1,94 @@
...
+  * This configuration allows any customization of the generated module.xml, making it very flexible to cover broader cases, for example, it is possible to declare module dependencies, mark them as exported, declare the main class and so on.
+  * This command is not appropriate for use in a managed domain or when connecting to the management CLI remotely. Modules should be added and removed manually in a production environment.
+
+This RFE is opened to have an alternative new operation to address simpler cases, where it is not required any additional configuration beyond on the default module.xml configurations, and supported in production environments and domain mode.
+
+The simplest way to accomplished this is to use the EE subsystem to configure a path to the shared folder. An example of this configuration could be:
+
+  [standalone@localhost:9990 /] /subsystem=ee:list-add(name=global-folders,value={name=my-common-libs, path=/Users/me/libs})
+
+  [domain@localhost:9990 /] /profile=default/subsystem=ee:list-add(name=global-folders,value={name=my-common-libs, path=/Users/me/libs})
```
bstansberry: The path specification must follow the standard WildFly conventions for declaring paths; i.e. 'path' and and 'relative-to' attributes with 'relative-to' optional.

ee/WFLY-1160-Provide_ability_to_configure_a_global_directory_which_puts_the_contents_in_the_classpath.adoc
```diff
@@ -0,0 +1,94 @@
...
+
+== Requirements
+
+=== Hard Requirements
+- The new module created from the global-folder configuration will get automatically a dependency on `javaee.api` module name.
+- Hot JBoss modules redeployments will require a server restart to make them available to all existing and new applications.
+- New modules created from the content of the shared folder will be added to the deployed applications as a system dependencies.
+
+=== Nice-to-Have Requirements
+- We could have a deployment pattern name to filter which deployment is affected by this global shared folder.
```
bstansberry: Let's not. :)

ee/WFLY-1160-Provide_ability_to_configure_a_global_directory_which_puts_the_contents_in_the_classpath.adoc
```diff
@@ -0,0 +1,94 @@
...
+This RFE is opened to have an alternative new operation to address simpler cases, where it is not required any additional configuration beyond on the default module.xml configurations, and supported in production environments and domain mode.
+
+The simplest way to accomplished this is to use the EE subsystem to configure a path to the shared folder. An example of this configuration could be:
+
+  [standalone@localhost:9990 /] /subsystem=ee:list-add(name=global-folders,value={name=my-common-libs, path=/Users/me/libs})
+
+  [domain@localhost:9990 /] /profile=default/subsystem=ee:list-add(name=global-folders,value={name=my-common-libs, path=/Users/me/libs})
+
+The name and path attributes will be mandatory.
+Internally, this operation will create a module with the name `my-common-libs` that will have as resource-roots the folder `/Users/me/libs/` and all the jars files included in this folder. Subdirectories will not be scanned, but if it is required, users can create a new global-folder using the subdirectory path.
```
vmarekrh: This sentence suggests the functionality work only locally, then only in standalone mode.
It is not clear if functionality work in standalone mode or domain mode.
Can you explain it?

ee/WFLY-1160-Provide_ability_to_configure_a_global_directory_which_puts_the_contents_in_the_classpath.adoc
```diff
@@ -0,0 +1,94 @@
...
+
+=== QE Contacts
+
+* mailto:vmarek@redhat.com[Vratislav Marek]
+
+
+=== Testing By
+[ ] Engineering
+
+[X] QE
```
vmarekrh: Basic unit tests for the functionality must be prepared by Engineering.
The more complex integration testing will be covered by QE.
In case of need to be the complete testing cover only by QE, you must give a reason why.

ee/WFLY-1160-Provide_ability_to_configure_a_global_directory_which_puts_the_contents_in_the_classpath.adoc
yersan: My initial idea regarding domain mode is we will able to have a local folder path relative to the host controller file system. 

We will be configuring the EE subsystem with a path. The EE subsystem in domain mode is configured under a profile. Each profile configuration is transferred to each slave HC, so each HC will see just a path configuration which should be local to its filesystem.

That means we are not going to transfer all the available jar files from the DC to the slaves. Is the user responsibility to have a folder available on each host where an HC instance is running. If an HC is not managing servers in a server group affected by the profile which was configured, then this requirement should not be necessary, since there weren't deployments running with this configuration.

Does it make sense to you? I'm still testing if this is possible and what could be the best way to give support in domain mode for this feature.

ee/WFLY-1160-Provide_ability_to_configure_a_global_directory_which_puts_the_contents_in_the_classpath.adoc
yersan: ok, resolving this conversation then

ee/WFLY-1160-Provide_ability_to_configure_a_global_directory_which_puts_the_contents_in_the_classpath.adoc
vmarekrh: Does it work remotely in the standalone mode? Or do you plan it?

Do you expect the customer will be responsible for distribution jars in the shared library to the server and all slaves manually?

If it will be distributed only name and path. What about path collision between Linux and Windows?
Some nodes can be running on Windows and other nodes can be running on Linux.

yersan: > Does it work remotely in the standalone mode? Or do you plan it?

Yes, it should work remotely too, it is the server jvm responsibility to check the paths internally, not the client jvm. I could be giving a vague response, please, would you mind to be more specific about your concern? Why do you think it couldn't work if you are configuring the server remotely? What exactly do you think it couldn't work?

> Do you expect the customer will be responsible for distribution jars in the shared library to the server and all slaves manually?

For a standalone mode, they should have in the shared folder the jars they want to include as a dependency of all the deployed applications. In domain mode, the jar should be accesible from the host controller file system too. It is the same if a customer wants to add a new module, they would have to create a module.xml with a resource root pointing to a jar file accessible from the server (standalone mode) or host controller (domain mode) file system.

> If it will be distributed only name and path. What about path collision between Linux and Windows?

It will work like any other operation in domain mode that it is using path and relative-to attributes (we can have paths with a similar name but different values per HC, they can use expressions ...)

ee/WFLY-1160-Provide_ability_to_configure_a_global_directory_which_puts_the_contents_in_the_classpath.adoc
```diff
@@ -0,0 +1,94 @@
...
+=== Hard Requirements
+- The new module created from the global-folder configuration will get automatically a dependency on `javaee.api` module name.
+- Hot JBoss modules redeployments will require a server restart to make them available to all existing and new applications.
+- New modules created from the content of the shared folder will be added to the deployed applications as a system dependencies.
+
+=== Nice-to-Have Requirements
+- We could have a deployment pattern name to filter which deployment is affected by this global shared folder.
+- We could allow the possibility of adding existing module dependencies to the new module created from the global-folder configuration, and, if there is no dependencies configured, use by default `javaee.api`.
+
+=== Non-Requirements
```
bstansberry: Please add an item here stating that putting the module content on the filesystem of any machine is not a requirement; it's completely a user responsibility.

There's some question in the comments here that seem to be about whether we're managing the module content.

ee/WFLY-1160-Provide_ability_to_configure_a_global_directory_which_puts_the_contents_in_the_classpath.adoc
```diff
@@ -0,0 +1,94 @@
...
+
```
bstansberry: A possible non-requirement to state is doing any Stage.MODEL validation of the path configuration, beyond validating that any relative-to value references a valid path capability. (The latter definitely happens.)

Whether we state it here or not we shouldn't do that kind of validation in MODEL, as the location doesn't need to exist until RUNTIME. Sometimes we get demands to 'help' by validating stuff in MODEL though and we shouldn't do that.

ee/WFLY-1160-Provide_ability_to_configure_a_global_directory_which_puts_the_contents_in_the_classpath.adoc
yersan: Fixed, I changed the description to make use of path and relative-to.

ee/WFLY-1160-Provide_ability_to_configure_a_global_directory_which_puts_the_contents_in_the_classpath.adoc
yersan: Fixed. Added a non-requirement following your description.

ee/WFLY-1160-Provide_ability_to_configure_a_global_directory_which_puts_the_contents_in_the_classpath.adoc
yersan: Yes, paths will be know at Stage.RUNTIME. Added a non-requirement following your description.

ee/WFLY-1160-Provide_ability_to_configure_a_global_directory_which_puts_the_contents_in_the_classpath.adoc
vmarekrh: > > Does it work remotely in the standalone mode? Or do you plan it?
> 
> Yes, it should work remotely too, it is the server jvm responsibility to check the paths internally, not the client jvm. I could be giving a vague response, please, would you mind being more specific about your concern? Why do you think it couldn't work if you are configuring the server remotely? What exactly do you think it couldn't work?

My question was be if can be jar uploaded to the server via CLI.
You've already answered my question, It will don't upload jars, thank you.

yersan:
@vmarekrh @bstansberry 

I've updated a bit the analysis document with these two changes:

1. Use a subsystem EE child resource instead of an attribute type LIST to handle the directories. Although it loses a bit of consistency with the global-modules operation, the user will gain in simplicity to configure shared directories and it will be easier to expand the behavior with additional functionalities.
2. Rename global-folders attribute as to global-directory resource. The word 'folder' 
has a wider concept and related to Windows Explorer, directory sounds more appropriate for this feature.

yersan:
Ups, I closed it instead of adding a comment

bstansberry:
@yersan @vmarekrh +1 to the changes in your last comment, Yeray.

ee/WFLY-1160-Provide_ability_to_configure_a_global_directory_which_puts_the_contents_in_the_classpath.adoc
```diff
@@ -0,0 +1,130 @@
...
+e.g. Suppose that our global directory path is /lib and it contains two jar files, lib-1.jar and lib-2.jar. The module generated internally by the global-directory operation will be equivalent to declare a module.xml with the following resources:
+
+  <resources>
+    <resource-root path="/lib"/>
+    <resource-root path="/lib/lib-1.jar"/>
+    <resource-root path="/lib/lib-2.jar"/>
+  </resources>
+
+The module created from the shared directory will be loaded as soon as the first application is deployed in the server. That means, if the server is started/restarted and there are no applications deployed, then the shared directory is neither scanned nor the module loaded.
```
vmarekrh: Can you add the sentence about the scanned jars to the log with the jar loading order, please?
The log of the scanned and really loaded jars will be critical for debugging customer solution for the GSS.

ee/WFLY-1160-Provide_ability_to_configure_a_global_directory_which_puts_the_contents_in_the_classpath.adoc
```diff
@@ -0,0 +1,130 @@
...
+[domain@localhost:9990 /] /subsystem=ee/global-directory=my-common-libs:read-resource
+{
+    "outcome" => "success",
+    "result" => {
+        "path" => "lib",
+        "relative-to" => "jboss.server.data.dir"
+    }
+}
+
+This RFE will not cover the availability of hot JBoss modules redeployments, that means, modifications of existing jar files as well as adding a new jar or removal an existing one will require a complete server restart, however, changes in plain resources as property files will be available immediately to the applications.
```
vmarekrh: Can you add the sentence if is enough to reload of the server only or why the reload of the server is not enough, please?

ee/WFLY-1160-Provide_ability_to_configure_a_global_directory_which_puts_the_contents_in_the_classpath.adoc
```diff
@@ -0,0 +1,130 @@
...
+  [standalone@localhost:9990 /] /subsystem=ee/global-directory=my-common-libs:remove()
+
+  [domain@localhost:9990 /] /profile=default/subsystem=ee/global-directory=my-common-libs:remove()
+
+As any other resources, users can read the current configuration of a global-directory using resource-resource operation:
+
+[standalone@localhost:9990 /] /subsystem=ee/global-directory=my-common-libs:read-resource
+{
+    "outcome" => "success",
+    "result" => {
```
vmarekrh: Can be here more information about the shared library?
E.q.:
- If is the shared library used/loaded
- If loading is succeeded of thrown Exception?

ee/WFLY-1160-Provide_ability_to_configure_a_global_directory_which_puts_the_contents_in_the_classpath.adoc
vmarekrh: I think it will be good to write info log message or lowest to write into debug message.

ee/WFLY-1160-Provide_ability_to_configure_a_global_directory_which_puts_the_contents_in_the_classpath.adoc
yersan: Not really, this operation only returns the values of this resource. 

`If is the shared library used/loaded`

As soon as there is an application deployed, the library is loaded

`If loading is succeeded of thrown Exception?`

If the library cannot be loaded, then the application deployment cannot be done, and the deployment will fail.

Libraries in the shared directory will be loaded when the deployment is done. Usually, the users can define one global directory, copy the jar files in it, then restart the server to apply the changes to the deployed application. If a library cannot be load, deployments will fail.

ee/WFLY-1160-Provide_ability_to_configure_a_global_directory_which_puts_the_contents_in_the_classpath.adoc
yersan: Added into "hard requirements" section

ee/WFLY-1160-Provide_ability_to_configure_a_global_directory_which_puts_the_contents_in_the_classpath.adoc
yersan: Yes, it is possible, I've added some logs in the service which process the jars and path resources

ee/WFLY-1160-Provide_ability_to_configure_a_global_directory_which_puts_the_contents_in_the_classpath.adoc
```diff
@@ -0,0 +1,130 @@
...
+        "path" => "lib",
+        "relative-to" => "jboss.home.dir"
+    }
+}
+
+[domain@localhost:9990 /] /subsystem=ee/global-directory=my-common-libs:read-resource
```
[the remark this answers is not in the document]
yersan: Fixed

ee/WFLY-1160-Provide_ability_to_configure_a_global_directory_which_puts_the_contents_in_the_classpath.adoc
```diff
@@ -0,0 +1,130 @@
...
+
+Removal of a global-directory could be done using the following operation:
+
+  [standalone@localhost:9990 /] /subsystem=ee/global-directory=my-common-libs:remove()
+
+  [domain@localhost:9990 /] /profile=default/subsystem=ee/global-directory=my-common-libs:remove()
+
+As any other resources, users can read the current configuration of a global-directory using resource-resource operation:
+
+[standalone@localhost:9990 /] /subsystem=ee/global-directory=my-common-libs:read-resource
```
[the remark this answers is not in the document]
yersan: Fixed

ee/WFLY-1160-Provide_ability_to_configure_a_global_directory_which_puts_the_contents_in_the_classpath.adoc
vmarekrh: Ok

ee/WFLY-1160-Provide_ability_to_configure_a_global_directory_which_puts_the_contents_in_the_classpath.adoc
vmarekrh: Thank you

ee/WFLY-1160-Provide_ability_to_configure_a_global_directory_which_puts_the_contents_in_the_classpath.adoc
vmarekrh: Thank you

yersan:
@bstansberry After finished a workable version, we realized it is very simple to be able to scan subfolders and include any jar found there into the generated module. We would like to include this possibility. 

Advantages:

1. It will allow interdependencies between any jars included in the folder tree, for example:
/my-libs/jar1.jar
/my-libs/sub-folder/jar2.jar

jar1 will see classes included in jar2.jar. We are not covering this case with our current implementation.

2. Users don't have to add multiple global folders, which is more natural. One global folder will be enough for the purposes of this issue. We could reject the possibility of adding more than one global-folder, but since our workable version is already prepared to work with multiple global folders we decided to keep them

Jar files will be added in the same order as jar resource roots, independently of the subfolder where they are. Order will be the same name/size. 

@vmarekrh @bstansberry Do you want to include it?

vmarekrh:
I accepted this change of analysis document for include subdirectories because limit a customer only for a directory without subdirectories isn't user-friendly.

bstansberry:
@yersan Handling subfolders is fine with me. Did you update the doc to reflect that? If you did I didn't see it. If you were waiting for a +1 from me first, well, 👍

ee/WFLY-1160-Provide_ability_to_configure_a_global_directory_which_puts_the_contents_in_the_classpath.adoc
```diff
@@ -0,0 +1,137 @@
...
+
+* path: The path of the directory to scan. (Mandatory)
+* relative-to: The name of another previously named path, or of one of the standard paths provided by the system. (Optional)
+
+When a new `global-directory` is created, the server will create a module that will have a https://jboss-modules.github.io/jboss-modules/manual/#path-resource-loader[Path Resource Loader] using 'path' and 'relative-to' attributes and one https://jboss-modules.github.io/jboss-modules/manual/#jar-resource-loader[Jar Resource loader] for each jar file included in this directory and subdirectories.
+
+The 'Path Resource Loader' will make available any file as a resource to the applications.
+The 'Jar Resource loader' will make available any class inside of the jar file to the applications.
+
+The jar resources of this module will be created iterating over all jar files in alphabetical order and size descending. The resource root order declaration will govern the class loading order for the classes available from the jar resources of this module.
```
marekkopecky: Hi @yersan, what if I have a/c.jar and d/b.jar. What is the first resource and what is the second? b.jar < c.jar? Or a/c.jar < d/b.jar?

ee/WFLY-1160-Provide_ability_to_configure_a_global_directory_which_puts_the_contents_in_the_classpath.adoc
```diff
@@ -0,0 +1,137 @@
...
+
+* https://issues.jboss.org/browse/MODULES-99[MODULES-99]
+
+=== Dev Contacts
+
+* mailto:{email}[{author}]
+
+=== QE Contacts
+
+* mailto:vmarek@redhat.com[Vratislav Marek]
```
tterem: Please add me as QE contact.
`* mailto:tterem@redhat.com[Tomas Terem]`

yersan: Hello @tterem, I've just added you as additional QE

ee/WFLY-1160-Provide_ability_to_configure_a_global_directory_which_puts_the_contents_in_the_classpath.adoc
yersan: Hello @marekkopecky, Sorry, I missed out this notification.

Currently, we are only taking into account jar file names, not the full pathname. So in your example, b.jar will be before than c.jar

There is a similar question from Tomas Terem in the Wildfly PR, we can discuss it there:
https://github.com/wildfly/wildfly/pull/12489#issuecomment-521720014

tterem:
https://github.com/wildfly/wildfly/pull/12489#issuecomment-522928177

This should also be changed in the analysis.

yersan:
@tterem I've updated the documentation in relation to your last comment

yersan:
Hello @tterem , I expanded a bit the example trying to clear it up how we are loading the jar files, could you review and approve in case? thanks

yersan:
This proposal can be merged now since the feature PR https://github.com/wildfly/wildfly/pull/12489 was already merged

PR #194, opened 2019-04-05
Proposal for WFLY-11989 offliner file with everything needed allowing to provision server using Galleon

jfdenise:
Issue: https://issues.jboss.org/browse/WFGP-132

### DIFF
--- a/wf-galleon/WFLY-11989_Galleon_Maven_Repository.adoc
+++ b/wf-galleon/WFLY-11989_Galleon_Maven_Repository.adoc
@@ -1,0 +1,93 @@
+= WFLY-11989: wildfly-dist should generate and deploy a zipped maven repo for galleon
+:author:            Jean-Francois Denise
+:email:             jdenise@redhat.com
+:toc:               left
+:icons:             font
+:idprefix:
+:idseparator:       -
+
+== Overview
+
+Offer the ability to package all Galleon required artifacts to provision WildFly server.
+
+== Issue Metadata
+
+=== Issue
+
+* https://issues.jboss.org/browse/WFLY-11989[WFLY-11989]
+
+=== Related Issues
+
+* https://issues.jboss.org/browse/EAP7-891[EAP7-891]
+
+* https://issues.jboss.org/browse/EAP7-1218[EAP7-1218]
+
+* https://issues.jboss.org/browse/WFGP-132[WFGP-132] (Obsolete, not followed approach)
+
+* https://issues.jboss.org/browse/WFGP-135[WFGP-135]
+
+=== Dev Contacts
+
+* mailto:{email}[{author}]
+
+=== QE Contacts
+
+* mailto:mjurc@redhat.com[Michal Jurc]
+* mailto:nziakova@redhat.com[Nikoleta Hlavickova]
+
+=== Testing By
+// Put an x in the relevant field to indicate if testing will be done by Engineering or QE. 
+// Discuss with QE during the Kickoff state to decide this
+[ ] Engineering
+
+[X] QE
+
+=== Affected Projects or Components
+
+* OpenShift
+
+=== Other Interested Projects
+
+* WildFly Galleon
+
+* Offliner: https://github.com/release-engineering/offliner
+
+== Requirements
+
+In order to address this requirement, we are defining a new WildFly Maven plugin (_org.wildfly.galleon-plugins:wildfly-galleon-maven-plugin_)
+goal named _generate-all-artifacts-list_.
+This goal aggregates all offliner files that have been generated when each Wildfly galleon feature-pack is built.
+
+offliner file == list of artifacts == "sha-256 checksum, artifact path inside maven repo" per line.
+
+When building the wildfly-dist module with release profile (-Drelease=true), this plugin is called to 
+generate /target/wildfly-galleon-${project.version}-all-artifacts-list.txt directory.
+
+The file is deployed with other maven artifacts as wildfly-dist-${project.version}-artifact-list.txt
+
+Then the workflow to produce a maven repository to provision WildFly is:
+
+1) Retrieve the org.wildfly:wildfly-dist-${project.version}-artifact-list.txt from maven repository
+
+2) Use offliner runtime (https://github.com/release-engineering/offliner) to download all files to populate a local maven repository.
+
+=== Hard Requirements
+
+* Generate and deploy the file with valid sha-256 checksum and paths for all required artifacts.
+
+=== Nice-to-Have Requirements
+
+=== Non-Requirements
+
+== Implementation Plan
+
+* Add the goal to WildFly galleon-plugins project.
+* Evolve dist/pom.xml to call the plugin.
+
+== Test Plan
+
+* Use the generated file to populate a maven local repo.
+* Test that wildfly can be provisioned from the resulting repo. 
+
+== Community Documentation
+This file could be treated as an implementation detail and not be documented.


wf-galleon/EAP7-1218_Galleon_Maven_Repository.adoc
```diff
@@ -0,0 +1,74 @@
...
+
+== Implementation Plan
+
+* Add the goal to WildFly galleon-plugins project.
+* Evolve dist/pom.xml to call the plugin.
+
+== Test Plan
+
+== Community Documentation
+We would need to advertise that this zip file exists (release notes?).
```
bstansberry: The blog post for the first such release seems reasonable, i.e. https://wildfly.org/news/2019/02/27/WildFly16-Final-Released/

I think that's what you meant. Release notes could also mean something like https://issues.jboss.org/secure/ReleaseNote.jspa?projectId=12315422&version=12340978 but that's not sufficient.

Mentioning it once is fine; after that it's a standard item on the downloads page.

Adding this to wildfly.or/downloads will take some work in the current site (please file a WFLY assigned to me) and on the WIP new site, https://jbossorg.github.io/wildflysite/ (which has a link at the top for filing issues.)

bstansberry:
How do you see this repo being used? That is, it's out there on the internet, and in the end we want it incorporated into an image. How is that done?

Having the repo inside the builder image is the ultimate goal. Having the repo independently downloadable may be required to achieve that ultimate goal. Or perhaps not. And even if not it might be a good thing to provide anyway (e.g. it's usable for other use cases.)

I'm a bit embarrassed as I'm the person who said we need to think about having this repo, but I didn't really think it through. I still think we need it but now's the time to ask. :)

jfdenise:
The zipped repository is retrieved (curl) from its location and unzipped to $Home/.m2 during docker build of the builder image.
It seems to me the proper way to do it although we could provision a full server during docker build, side effect being that the HOME/m2 repository would be populated with the needed content. Retrieving all artifacts during provisioning takes long time, downloading the zip is faster.

bstansberry:
@jfdenise Thanks. So this zip is definitely something the WildFly project should start producing.

Please file a WFLY for this and convert the PR title, commit message and primary issue link in the doc to it. A link to EAP7-1218 in the related issues section is fine.

wf-galleon/EAP7-1218_Galleon_Maven_Repository.adoc
```diff
@@ -0,0 +1,74 @@
...
+== Requirements
+
+In order to address this requirement, we are defining a new WildFly Maven plugin (_org.wildfly.galleon-plugins:wildfly-galleon-maven-plugin_)
+goal named _generate-maven-repository_.
+This goal generates a zipped maven repository from a feature-pack.
+
+When building a WildFly release (-Drelease=true), this plugin will be called to 
+generate /target/wildfly-<version>-galleon-maven-repository.zip file.
+
+This zip file should then be made available from WildFly public releases web site (along with released WildFly).
```
bstansberry: Ah, so this doesn't end up being sent to the remote maven repo as part of the normal mvn deploy process?

That's a concern as the remote maven repo acts as a kind of "source of truth" for what goes into wildfly.org/downloads.  We populate the folder that gets served by pulling down the needed files from the maven repo. The person doing that task need not be the person who did the mvn deploy.

jfdenise: @bstansberry , I am not familiar with the deployment process. I suppose we could attach the zipped repository as a maven artifact to deploy. When building dist with -Drelease, I can see wildfly-dist-17.0.0.Beta1-SNAPSHOT.tar.gz being installed in local repo but not wildfly-17.0.0.Beta1-SNAPSHOT.tar.gz that is downloadable from https://download.jboss.org/wildfly/
Is it renamed when pulling it down to wildfly.org folder?

jfdenise: @bstansberry, I made the zipped repository a proper deployed artifact. The proper way to do it is:
1) Change the Galleon plugin to output a repository directory
2) assembly is zipping the directory
3) install plugin is installing it as :
wildfly-dist-17.0.0.Beta1-SNAPSHOT-galleon-maven-repository.zip

jfdenise:
@bstansberry , I have updated the proposal to focus on community issue.

wf-galleon/EAP7-1218_Galleon_Maven_Repository.adoc
bstansberry: Sounds good.

And yes, when putting the zips, tars etc in the dir from which https://download.jboss.org/wildfly/ serves, the file name gets tweaked. I think that's likely for historical reasons.

psakar:
@bstansberry All open items were discussed / solved from product point of view (@jfdenise thank you for refining / fine tuning the solution)

wf-galleon/WFLY-11989_Galleon_Maven_Repository.adoc
```diff
@@ -0,0 +1,90 @@
...
+When building the wildfly-dist module with release profile (-Drelease=true), this plugin is called to 
+generate /target/wildfly-galleon-${project.version}-all-artifacts-list.txt directory.
+
+The file is deployed with other maven artifacts as wildfly-dist-${project.version}-artifact-list.txt
+
+Then the workflow to produce a maven repository to provision WildFly is:
+
+1) Retrieve the org.wildfly:wildfly-dist-${project.version}-artifact-list.txt from maven repository
+
+2) Use offliner runtime (https://github.com/release-engineering/offliner) to download all files to populate a local maven repository.
```
bstansberry: OK, so the relationship of this RFE to our image work is that the image creation process will execute these two steps.

That sounds fine.

wf-galleon/WFLY-11989_Galleon_Maven_Repository.adoc
```diff
@@ -0,0 +1,90 @@
...
+
+* WildFly Galleon
+
+* Offliner: https://github.com/release-engineering/offliner
+
+== Requirements
+
+In order to address this requirement, we are defining a new WildFly Maven plugin (_org.wildfly.galleon-plugins:wildfly-galleon-maven-plugin_)
+goal named _generate-all-artifacts-list_.
+This goal aggregates all offliner files that have been generated when each Wildfly galleon feature-pack is built.
```
bstansberry: Are these already being generated? I quick check of galleon-pack/target in my current wildfly checkout doesn't show one.

wf-galleon/WFLY-11989_Galleon_Maven_Repository.adoc
bstansberry: AIUI it's not desirable to download all files; e.g. a maven repo in an image doesn't want src zips. Does offliner handle that or is this something the logic coordinating the use of offliner would handle (e.g. remove the unwanted files that were downloaded)?

jfdenise: @bstansberry , sources are not added only pom and jar.

psakar: artifact list generated per galleon pack does not contain source files or javadocs. It contains required artifacts, their poms and parent poms. Thus the same applies for the aggregated artifact list

wf-galleon/WFLY-11989_Galleon_Maven_Repository.adoc
jfdenise: @bstansberry , this is coming. I just opened https://github.com/wildfly/wildfly-core/pull/3751
to make core to generate the offliner that will be aggregated in wildfly-dist.

psakar: during each galleon pack build (or precisely in another maven lifecycle phase after galleon pack build) will be generated artifact list (=offline file) for given pack. Those will be aggregates as described above.
On the product level we will assert that EAP maven repository contains all artifacts mentioned in the aggregated artifact list (it would be subset of maven repository, which contains additional artifacts required eg. for develepment)

psakar: @jfdenise  BTW - those additional artifacts which are not part of any galleon pack you may need as well - eg. jboss-ejb-client-legacy - not sure whether it is part of GP or not

mjurc:
@psakar Are you okay with this proposal as it is? If you are, please kindly add an approval.

psakar:
@mjurc I'm ok with the proposal  - see my comment https://github.com/wildfly/wildfly-proposals/pull/194#issuecomment-484425267

wf-galleon/WFLY-11989_Galleon_Maven_Repository.adoc
```diff
@@ -0,0 +1,90 @@
...
+
+* OpenShift
+
+=== Other Interested Projects
+
+* WildFly Galleon
+
+* Offliner: https://github.com/release-engineering/offliner
+
+== Requirements
```
mjurc: @jfdenise What artefacts do we expect to be present in Maven repository provisioned in this plugin compared to the zipped Maven product repository?

wf-galleon/WFLY-11989_Galleon_Maven_Repository.adoc
```diff
@@ -0,0 +1,90 @@
...
+
+== Implementation Plan
+
+* Add the goal to WildFly galleon-plugins project.
+* Evolve dist/pom.xml to call the plugin.
+
+== Test Plan
+
+* Use the generated file to populate a maven local repo.
+* Test that wildfly can be provisioned from the resulting repo.
```
nziakova: This will be tested within EAP7-891. However, we need additional test to ensure the artifacts in the resulting repository are the same as in delivered Maven repository ZIP for the same release.

wf-galleon/WFLY-11989_Galleon_Maven_Repository.adoc
jfdenise: @mjurc we will have *.pom[.sha|.md5] and *.jar[.sha|.md5] No jar file containing sources. So dependencies needed to build application will be not present.

psakar: It would contain artifacts required for server runtime (=feature packs and their dependencies). Zipped product repository contains also artifacts required for development / BOMs / quickstarts

wf-galleon/WFLY-11989_Galleon_Maven_Repository.adoc
jfdenise: @nziakova, it will be a subset, it doesn't contain it all.

nziakova: Sure, not all. However, those that will be included must be the same - have the same checksums as those in ZIP

jfdenise: Yes, exact same artifacts.

PR #196, opened 2019-04-12
[ELY-1778] Enhance Audit Logging by Feature Parity with Legacy Security Auditing

justinmcook:
Jira: https://issues.jboss.org/browse/ELY-1778

PR #195, opened 2019-04-11
HAL-1578: Customise Browser Title

hpehl:
https://issues.jboss.org/browse/HAL-1578

PR #197, opened 2019-04-15
HAL-1578: Customise Browser Title

hpehl:
https://issues.jboss.org/browse/HAL-1578

### DIFF
--- a/console/HAL-1578_Customise_Browser_Title.adoc
+++ b/console/HAL-1578_Customise_Browser_Title.adoc
@@ -1,0 +1,65 @@
+= Customise Browser Title
+:author:            Harald Pehl
+:email:             hpehl@redhat.com
+:toc:               left
+:icons:             font
+:idprefix:
+:idseparator:       -
+:issue-base-url:    https://issues.jboss.org/browse
+
+== Overview
+
+Customise the title of the management console. A custom title can be specified in the user settings of the console. The title can contain the following placeholders:
+
+- `%n`: The `name` attribute of the the root resource
+- `%o`: The `organization` (standalone) resp. `domain-organization` attribute of the the root resource
+
+If the title is not specified, the built in title will be used. If a placeholder is used, but the attribute is undefined, the built in title will be used.
+The built in title defaults to "%n | Management Console".
+
+== Issue Metadata
+
+=== Issue
+
+* {issue-base-url}/HAL-1578[HAL-1578]
+
+=== Related Issues
+
+* {issue-base-url}/EAP7-837[EAP7-837]
+
+=== Dev Contacts
+
+* mailto:hpehl@redhat.com[Harald Pehl]
+
+=== QE Contacts
+
+* mailto:jkasik@redhat.com[Jan Kasik]
+
+=== Testing By
+    
+[ ] Engineering
+    
+[x] QE
+
+=== Affected Projects or Components
+
+* WildFly
+* HAL
+
+== Requirements
+
+- Add an entry in the user setting to customise the title
+- If no title is specified the built in title will be used
+- If a title is specified use this title as browser / tab title
+- Replace `%n` with the value of the `name` attribute of the root resource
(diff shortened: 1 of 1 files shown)


hpehl:
Reopened using upstream issues from #195

honza-kasik:
Continuing previous discussion: 

I am just thinking if migration of such settings into "console subsystem" won't it be a problem for users? Settings won't be "per-user" anymore but rather "per-instance". What are the obstacles in introducing that subsystem now?

hpehl:
The introduction of a console subsystem needs more analysis: What are "per-user" settings, what are "per-instance" settings. I think this won't happen for WF17.

honza-kasik:
Oh, sorry, I meant "per-user" as in that each browser has one user, and using cookies to save title makes Web Console title unique for each user. On the other hand, introducing some server side storage would make title available for all users on given instance.

console/HAL-1578_Customise_Browser_Title.adoc
```diff
@@ -0,0 +1,58 @@
...
+:issue-base-url:    https://issues.jboss.org/browse
+
+== Overview
+
+Customise the title of the management console. A custom title can be specified in the user settings of the console. The title can contain the following placeholders:
+
+- `%n`: The `name` attribute of the the root resource
+- `%o`: The `organization` (standalone) resp. `domain-organization` attribute of the the root resource
+
+If the title is not specified, the built in title will be used. If a placeholder is used, but the attribute is undefined, the built in title will be used.
```
honza-kasik: Wouldn't it be better if users could directly benefit from the enhancement by using the name placeholder in default title value?

hpehl: AFAIK the name attribute is undefined for domain mode.

honza-kasik: It seems like it has a name, though not a nice one:
```
[domain@localhost:9990 /] :read-attribute(name=name)
{
    "outcome" => "success",
    "result" => "Unnamed Domain"
}
```

honza-kasik:
@hpehl To conclude that discussion. You don't think, that in future, when console subsystem will be introduced and it will be used to store the title value, the behavior (per-user vs per-instance) will not be changed/will not make any problems to users?

honza-kasik:
@hpehl Can you please also add information about who does the testing (QA in this case)?

hpehl:
> @hpehl To conclude that discussion. You don't think, that in future, when console subsystem will be introduced and it will be used to store the title value, the behavior (per-user vs per-instance) will not be changed/will not make any problems to users?

I don't know whether or when a console subsystem will be introduced. If so I'm sure we'll find a smooth migration for the settings. 

> @hpehl Can you please also add information about who does the testing (QA in this case)?

You mean in the analysis document?

honza-kasik:
> You mean in the analysis document?

Yes, please - similar to this: https://github.com/wildfly/wildfly-proposals/pull/196/files#diff-09c496d673c56836551e84041b44a56cR32

hpehl:
> > You mean in the analysis document?
> 
> Yes, please - similar to this: https://github.com/wildfly/wildfly-proposals/pull/196/files#diff-09c496d673c56836551e84041b44a56cR32

Done.

console/HAL-1578_Customise_Browser_Title.adoc
honza-kasik: @hpehl I understand there will be no `%n` in title by default? IMHO it would make sense...

hpehl: Ok, I agree. I'll change the implementation: Default title will be "%n | Management Console"

honza-kasik: Thanks!

hpehl:
@honza-kasik squashed to one commit.

PR #200, opened 2019-04-23
[WFLY-6892] Access logging for EJBs

chengfang:
Jira: https://issues.jboss.org/browse/WFLY-6892

### DIFF
--- a/ejb/WFLY-6892-ejb-access-logging.adoc
+++ b/ejb/WFLY-6892-ejb-access-logging.adoc
@@ -1,0 +1,118 @@
+= EJB Access Logging
+:author:            Cheng Fang
+:email:             cfang@redhat.com
+:toc:               left
+:icons:             font
+:idprefix:
+:idseparator:       -
+
+== Overview
+
+== Issue Metadata
+
+=== Issue
+
+* https://issues.jboss.org/browse/WFLY-6892 (Access logging for EJBs)
+
+=== Related Issues
+
+* https://issues.jboss.org/browse/WFCORE-4355 (Add utility for event logging such as audit and access logging)
+* https://issues.jboss.org/browse/EAP7-523 (Access logging for EJBs)
+* https://issues.jboss.org/browse/EAP7-699 (Enhanced Audit Logging - RFC support and reliability/speed customization)
+
+=== Dev Contacts
+
+* mailto:{email}[{author}]
+
+=== QE Contacts
+
+* mailto:jkasik@redhat.com[Jan Kašík]
+
+=== Testing By
+// Put an x in the relevant field to indicate if testing will be done by Engineering or QE. 
+// Discuss with QE during the Kickoff state to decide this
+[x] Engineering
+
+[ ] QE
+
+=== Affected Projects or Components
+
+* ejb3
+
+=== Other Interested Projects
+
+== Requirements
+
+=== Hard Requirements
+
+* implement the access logging feature similar to WildFly web access logging.
+
+* add access-logging definition to wildfly-ejb3 schema
+
+* configure ejb access logging with CLI
+
+* enable and disable ejb access logging
+
+* users should be able to configure condition/filter to determine which requests to log
+
+* configure how to rotate ejb access logs if applicable
+
+** default should be OFF for performance consideration
+
+* configure output location and format
+
+** directory/file
+
+** stdout
+
+** JSON
+
+* choose from pre-defined logging formats
+
+** common (default):
+
+** combined:
+
+* for more customization, define which attributes to include in ejb access logging
+
+** consider adapting the following attributes that are currently supported by web access logging:
+
+    %a - Remote IP address
+    %A - Local IP address
+    %b - Bytes sent, excluding HTTP headers, or '-' if no bytes were sent
+    %B - Bytes sent, excluding HTTP headers
+    %h - Remote host name
+    %H - Request protocol
+    %l - Remote logical username from identd (always returns '-')
+    %m - Request method
+    %p - Local port
+    %q - Query string (excluding the '?' character)
+    %r - First line of the request
+    %s - HTTP status code of the response
+    %t - Date and time, in Common Log Format format
+    %u - Remote user that was authenticated
+    %U - Requested URL path
+    %v - Local server name
+    %D - Time taken to process the request, in millis
+    %T - Time taken to process the request, in seconds
+    %I - current Request thread name (can compare later with stacktraces)
+
+=== Nice-to-Have Requirements
+
+* configure ejb access logging from management console
+
+=== Non-Requirements
+
+//== Implementation Plan
+////
+Delete if not needed. The intent is if you have a complex feature which can 
+not be delivered all in one go to suggest the strategy. If your feature falls 
+into this category, please mention the Release Coordinators on the pull 
+request so they are aware.
+////
+== Test Plan
+
+== Community Documentation
+////
+Generally a feature should have documentation as part of the PR to wildfly master, or as a follow up PR if the feature is in wildfly-core. In some cases though the documentation belongs more in a component, or does not need any documentation. Indicate which of these will happen.
+////
--- a/ejb/WFLY-6892-ejb-access-logging.adoc
+++ b/ejb/WFLY-6892-ejb-access-logging.adoc
@@ -19,6 +19,7 @@
 * https://issues.jboss.org/browse/WFCORE-4355 (Add utility for event logging such as audit and access logging)
 * https://issues.jboss.org/browse/EAP7-523 (Access logging for EJBs)
 * https://issues.jboss.org/browse/EAP7-699 (Enhanced Audit Logging - RFC support and reliability/speed customization)
+* https://issues.jboss.org/browse/WFLY-11031 (Json encoding of Access Log Events)
 
 === Dev Contacts
 
@@ -26,14 +27,14 @@
 
 === QE Contacts
 
-* mailto:jkasik@redhat.com[Jan Kašík]
+* mailto:szhantem@redhat.com[Sultan Zhantemirov]
 
 === Testing By
 // Put an x in the relevant field to indicate if testing will be done by Engineering or QE. 
 // Discuss with QE during the Kickoff state to decide this
 [x] Engineering
 
-[ ] QE
+[x] QE
 
 === Affected Projects or Components
 
@@ -53,8 +54,6 @@
 
 * enable and disable ejb access logging
 
-* users should be able to configure condition/filter to determine which requests to log
-
 * configure how to rotate ejb access logs if applicable
 
 ** default should be OFF for performance consideration
@@ -65,13 +64,11 @@
 
 ** stdout
 
-** JSON
+* choose from pre-defined logging formats (need to adapt for ejb invocation):
 
-* choose from pre-defined logging formats
+** common (default): %h %l %u %t "%r" %s %b
 
-** common (default):
-
-** combined:
+** combined: %h %l %u %t "%r" %s %b "%{i,Referer}" "%{i,User-Agent}"
 
 * for more customization, define which attributes to include in ejb access logging
 
@@ -91,7 +88,7 @@
     %s - HTTP status code of the response
     %t - Date and time, in Common Log Format format
     %u - Remote user that was authenticated
-    %U - Requested URL path
+    %U - Requested URL path (ejb invocation equivalent)
     %v - Local server name
     %D - Time taken to process the request, in millis
     %T - Time taken to process the request, in seconds
@@ -99,9 +96,14 @@
 
 === Nice-to-Have Requirements
 
-* configure ejb access logging from management console
+* users should be able to configure condition/filter to determine which requests to log
+
+* users should be able to configure JSON as the output format for ejb access logging
 
 === Non-Requirements
+
+* configure ejb access logging from management console
+
 
 //== Implementation Plan
 ////
--- a/ejb/WFLY-6892-ejb-access-logging.adoc
+++ b/ejb/WFLY-6892-ejb-access-logging.adoc
@@ -64,59 +64,308 @@
 
 ** stdout
 
+** JSON
+
 * choose from pre-defined logging formats (need to adapt for ejb invocation):
 
-** common (default): %h %l %u %t "%r" %s %b
+** short: date time ip user ejb method
 
-** combined: %h %l %u %t "%r" %s %b "%{i,Referer}" "%{i,User-Agent}"
+*** For example: 2019-05-05 12:23:27,003 127.0.0.1 admin hello/helloBean hello
 
-* for more customization, define which attributes to include in ejb access logging
+** long : date time ip user ejb method invocation event
 
-** consider adapting the following attributes that are currently supported by web access logging:
+*** For example: 2019-05-05 12:23:27,003 127.0.0.1 user1 hello/helloBean hello 1234 received
+                 2019-05-05 12:23:30,003 127.0.0.1 user1 hello/helloBean hello 1234 finished
 
-    %a - Remote IP address
-    %A - Local IP address
-    %b - Bytes sent, excluding HTTP headers, or '-' if no bytes were sent
-    %B - Bytes sent, excluding HTTP headers
-    %h - Remote host name
-    %H - Request protocol
-    %l - Remote logical username from identd (always returns '-')
-    %m - Request method
-    %p - Local port
-    %q - Query string (excluding the '?' character)
-    %r - First line of the request
-    %s - HTTP status code of the response
-    %t - Date and time, in Common Log Format format
-    %u - Remote user that was authenticated
-    %U - Requested URL path (ejb invocation equivalent)
-    %v - Local server name
-    %D - Time taken to process the request, in millis
-    %T - Time taken to process the request, in seconds
-    %I - current Request thread name (can compare later with stacktraces)
+** default: short
+
+* for more customization, define which attributes to include in ejb access logging with the following attributes:
+
+** date:
+** time
+** ip: client ip address
+** user: authenticated caller user name
+** ejb: ejb name (may need to prefix with application name)
+** method: ejb method
+** invocation: invocation id
+** event: event type ("received", "failed", "completed"; subject to available events from ejb container)
+** host: client host name
+** port: client port number
+** protocol: protocol used for this invocation
+** thread: thread name for this invocation
+** server: local server name
+
+* For example, one may configure the following custom pattern, which can be applied to both pattern formatter or
+JSON formatter:
+
+** pattern="date time client ejb method server thread"
+
 
 === New elements in wildfly-ejb3_6_0.xsd schema
-ejb30 subsystem is extended with a new top-level element access-log, defined as follows:
+ejb30 subsystem is extended with a new top-level element access-logging, defined as follows:
 [source,xml]
 ----
-    <xs:complexType name="accessLogType">
-        <xs:attribute name="pattern" use="optional" type="xs:string" default="common"/>
-        <xs:attribute name="worker" use="optional" type="xs:string" default="default"/>
-        <xs:attribute name="directory" use="optional" type="xs:string" default="${jboss.server.log.dir}"/>
-        <xs:attribute name="relative-to" use="optional" type="xs:string" />
-        <xs:attribute name="prefix" use="optional" type="xs:string" default="access_log."/>
-        <xs:attribute name="suffix" use="optional" type="xs:string" default="log"/>
-        <xs:attribute name="rotate" use="optional" type="xs:string" default="true"/>
-        <xs:attribute name="use-server-log" use="optional" type="xs:string" default="false"/>
-        <xs:attribute name="extended" use="optional" type="xs:string" default="false" />
-        <xs:attribute name="predicate" use="optional" type="xs:string" />
-    </xs:complexType>
+    <xs:element name="access-logging" type="accessLoggingType" minOccurs="0" maxOccurs="1" />
+
+    <xs:complexType name="accessLoggingType">
+            <xs:choice minOccurs="0" maxOccurs="unbounded">
+                <xs:element name="console-access-log" type="consoleAccessLogType"/>
+                <xs:element name="file-access-log" type="fileAccessLogType"/>
+                <xs:element name="periodic-rotating-file-access-log" type="periodicFileAccessLogType"/>
+                <xs:element name="size-rotating-file-access-log" type="sizeFileAccessLogType"/>
+                <xs:element name="server-log-access-log" type="serverLogAccessLogType"/>
+                <xs:element name="formatter" type="formatterType"/>
+            </xs:choice>
+        </xs:complexType>
+
+        <xs:complexType name="consoleAccessLogType">
+            <xs:annotation>
+                <xs:documentation>
+                    Defines an access log which writes to the console.
+                </xs:documentation>
+            </xs:annotation>
+            <xs:all>
+                <xs:element name="encoding" type="valueType" minOccurs="0"/>
+                <xs:element name="formatter" type="accessLogFormatterType" minOccurs="0"/>
+                <xs:element name="target" minOccurs="0">
+                    <xs:complexType>
+                        <xs:attribute name="name" use="required">
+                            <xs:simpleType>
+                                <xs:restriction base="xs:token">
+                                    <xs:enumeration value="System.out"/>
+                                    <xs:enumeration value="System.err"/>
+                                    <xs:enumeration value="console"/>
+                                </xs:restriction>
+                            </xs:simpleType>
+                        </xs:attribute>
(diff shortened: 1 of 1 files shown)


ejb/WFLY-6892-ejb-access-logging.adoc
```diff
@@ -0,0 +1,138 @@
...
+
+** consider adapting the following attributes that are currently supported by web access logging:
+
+    %a - Remote IP address
+    %A - Local IP address
+    %b - Bytes sent, excluding HTTP headers, or '-' if no bytes were sent
+    %B - Bytes sent, excluding HTTP headers
+    %h - Remote host name
+    %H - Request protocol
+    %l - Remote logical username from identd (always returns '-')
```
jamezp: I'd suggest not supporting this if it's always going to return a `-` like Undertow does. This was done for compatibility only.

ejb/WFLY-6892-ejb-access-logging.adoc
```diff
@@ -0,0 +1,138 @@
...
+        <xs:attribute name="worker" use="optional" type="xs:string" default="default"/>
+        <xs:attribute name="directory" use="optional" type="xs:string" default="${jboss.server.log.dir}"/>
+        <xs:attribute name="relative-to" use="optional" type="xs:string" />
+        <xs:attribute name="prefix" use="optional" type="xs:string" default="access_log."/>
+        <xs:attribute name="suffix" use="optional" type="xs:string" default="log"/>
+        <xs:attribute name="rotate" use="optional" type="xs:string" default="true"/>
+        <xs:attribute name="use-server-log" use="optional" type="xs:string" default="false"/>
+        <xs:attribute name="extended" use="optional" type="xs:string" default="false" />
+        <xs:attribute name="predicate" use="optional" type="xs:string" />
+    </xs:complexType>
```
jamezp: I'd like to suggest possibly looking at a different model for this. When working on the JSON and console logging for web access logging this model did not work well for expansion. I'd suggest having separate resources and not making the `use-server-log` mutually exclusive.

ejb/WFLY-6892-ejb-access-logging.adoc
```diff
@@ -0,0 +1,138 @@
...
+----
+
+=== Nice-to-Have Requirements
+
+* users should be able to configure condition/filter to determine which requests to log
```
jamezp: If you end up supporting this I'd like to see some detailed requirements on how this would be done.

ejb/WFLY-6892-ejb-access-logging.adoc
```diff
@@ -0,0 +1,138 @@
...
+
+* choose from pre-defined logging formats (need to adapt for ejb invocation):
+
+** common (default): %h %l %u %t "%r" %s %b
+
+** combined: %h %l %u %t "%r" %s %b "%{i,Referer}" "%{i,User-Agent}"
+
+* for more customization, define which attributes to include in ejb access logging
+
+** consider adapting the following attributes that are currently supported by web access logging:
```
jamezp: IMO before a pattern like this is defined we need to know what users expect to see. The patterns below to me are not intuitive. These were used in Undertow for legacy reasons AIUI.

That said if we feel we should keep standard patterns that's okay too. We just don't want to use a legacy format because that's how it's been done :)

ejb/WFLY-6892-ejb-access-logging.adoc
chengfang: I was thinking using the predicate attribute for some basic filtering, either by reusing the undertow predicate impl, or writing one more tailored to ejb invocations.

ejb/WFLY-6892-ejb-access-logging.adoc
chengfang: Can you elaborate on the model that is more friendly to JSON and console logging?  What do you mean by "having separate resources"?  Does it mean an ejb access-log model that is different from web access-log?  Also can you explain more on "not making user-server-log mutually exclusive?"

ejb/WFLY-6892-ejb-access-logging.adoc
chengfang: The intent was to stick to convention already in use in undertow and apache web server, which should make it easier for users.  The logging pattern will not be intuitive anyway and so I figured existing convention will be a bit easier.  But I'm open to other better patterns.

chengfang:
> @chengfang
> In the context of https://issues.jboss.org/browse/EAP7-523 discussion maybe we would need some convenient way to configure JSON logging. I think also, it may be useful to add most important information about the implementation (integration with current logging, DMR format etc).

Noted.  I will add to the proposal once we have a better undestanding of the model to be used.

ejb/WFLY-6892-ejb-access-logging.adoc
jamezp: I think we should only do this if it makes sense. I'm not sure on what details are expected to be logged. If they match what a web container would log then it likely makes sense. If not then we should rethink it IMO.

ejb/WFLY-6892-ejb-access-logging.adoc
jamezp: That seems fine. I only mentioned this because if the output is used by a log aggregator I think it most cases it makes sense for the aggregator to make those determinations. However there could be performance penalties if we're logging heaps of data.

ejb/WFLY-6892-ejb-access-logging.adoc
jamezp: I ended up adding the console logging as a separate resource, `setting=console-access-log`. This was because using new attributes would require too many mutually exclusive attributes. Meaning you would have something like `use-console-logging` which means `use-server-log` can't be `true` as well and it also means that it would turn off the file logging. This doesn't seem very intuitive. I also think users should be allowed to log in as many places as they'd like. For example if they want to use the server logging and log to a file that seems like a valid use case.

I would suggest doing something like, please not I'm guessing on the address,: 
* `/subsystem=ejb3/service=access-logging/json-formatter=my-json-formatter`
* `/subsystem=ejb3/service=access-logging/file=myfile`
* `/subsystem=ejb3/service=access-logging/console=my-console`
* `/subsystem=ejb3/service=access-logging/pattern-formatter=mypattern`
* `/subsystem=ejb3/service=access-logging:write-attribute(name=use-server-log, value=true)`

Note these names can definitely change it's just a concept. The idea though is if you end up with the requirement for an additional output type, say a socket out, you could easily add another resource.

chengfang: Got it.  I'll update it based on the discussion so far.

jamezp: Sounds good. It doesn't have to be specifically those address, but just a way to allow for any output and format to be used on that output.

Note too if you don't use the event logging API some special handling for the console will need to be used as we wrap `System.out` and `System.err` in loggers.

ejb/WFLY-6892-ejb-access-logging.adoc
chengfang: I've updated the proposal to use more readable logging patterns, since we will need to support JSON, which will require a somewhat readable key anyway.  So for a pattern = "date time ip ejb method", a log entry with simple pattern will be like "2019-05-01 20:50:54,975 127.0.0.1 hello hello", a JSON entry will be like: {"date": "2019-05-01", "time": "20:50:54,975", "ip": "127.0.0.1", "ejb": "hello", "method": "hello"}

chengfang: I've updated the proposal to use more readable logging patterns, since we will need to support JSON, which will require a somewhat readable key anyway.  So for a pattern = "date time ip ejb method", a log entry with simple pattern will be like "2019-05-01 20:50:54,975 127.0.0.1 hello hello", a JSON entry will be like: {"date": "2019-05-01", "time": "20:50:54,975", "ip": "127.0.0.1", "ejb": "hello", "method": "hello"}

ejb/WFLY-6892-ejb-access-logging.adoc
chengfang: I updated the schema segment to model after jboss-as-logging-7_0.xsd, and it ends up quite long addition.  Not sure if there's a better way.  Some schema elements definition differ from jboss-as-logging ones to adapt to access logging.  With this changes, one should be able to specify multiple ejb access logs, each with its own formatter (simple or JSON).

I also defined use-server-log as its own element (instead of an attribute), in order to apply any named or inline formatter.

ejb/WFLY-6892-ejb-access-logging.adoc
```diff
@@ -0,0 +1,387 @@
...
+                <xs:element name="encoding" type="valueType" minOccurs="0"/>
+                <xs:element name="formatter" type="accessLogFormatterType" minOccurs="0"/>
+                <xs:element name="target" minOccurs="0">
+                    <xs:complexType>
+                        <xs:attribute name="name" use="required">
+                            <xs:simpleType>
+                                <xs:restriction base="xs:token">
+                                    <xs:enumeration value="System.out"/>
+                                    <xs:enumeration value="System.err"/>
+                                    <xs:enumeration value="console"/>
```
jamezp: If we use this option we will need some changes to the API provided for access logging [WFCORE-4355](https://issues.jboss.org/browse/WFCORE-4355). The API only writes to `System.out` for now.

ejb/WFLY-6892-ejb-access-logging.adoc
```diff
@@ -0,0 +1,387 @@
...
+                                </xs:restriction>
+                            </xs:simpleType>
+                        </xs:attribute>
+                    </xs:complexType>
+                </xs:element>
+            </xs:all>
+            <xs:attribute name="autoflush" type="xs:boolean" use="optional" default="true"/>
+            <xs:attribute name="name" type="xs:string" use="required"/>
+            <xs:attribute name="enabled" type="xs:boolean" use="optional" default="true"/>
```
jamezp: I'd not add this unless you feel it's necessary. If this is just a separate resource it could be removed instead of disabled.

ejb/WFLY-6892-ejb-access-logging.adoc
```diff
@@ -0,0 +1,387 @@
...
+            <xs:annotation>
+                <xs:documentation>
+                    Defines an access log which writes to the server log.
+                </xs:documentation>
+            </xs:annotation>
+            <xs:all>
+                <xs:element name="formatter" type="accessLogFormatterType" minOccurs="0"/>
+            </xs:all>
+            <xs:attribute name="name" type="xs:string" use="required"/>
+            <xs:attribute name="enabled" type="xs:boolean" use="optional" default="true"/>
```
jamezp: Same comment as above for this.

ejb/WFLY-6892-ejb-access-logging.adoc
```diff
@@ -0,0 +1,387 @@
...
+            </xs:annotation>
+            <xs:all>
+                <xs:element name="encoding" type="valueType" minOccurs="0"/>
+                <xs:element name="formatter" type="accessLogFormatterType" minOccurs="0"/>
+                <xs:element name="file" type="pathType" minOccurs="1"/>
+                <xs:element name="append" type="booleanValueType" minOccurs="0"/>
+            </xs:all>
+            <xs:attribute name="autoflush" type="xs:boolean" use="optional" default="true"/>
+            <xs:attribute name="name" type="xs:string" use="required"/>
+            <xs:attribute name="enabled" type="xs:boolean" use="optional" default="true"/>
```
jamezp: Same comment as above for this.

ejb/WFLY-6892-ejb-access-logging.adoc
```diff
@@ -0,0 +1,387 @@
...
+            <xs:all>
+                <xs:element name="encoding" type="valueType" minOccurs="0"/>
+                <xs:element name="formatter" type="accessLogFormatterType" minOccurs="0"/>
+                <xs:element name="file" type="pathType"/>
+                <xs:element name="suffix" type="valueType"/>
+                <xs:element name="append" type="booleanValueType" minOccurs="0"/>
+            </xs:all>
+            <xs:attribute name="autoflush" type="xs:boolean" use="optional" default="true"/>
+            <xs:attribute name="name" type="xs:string" use="required"/>
+            <xs:attribute name="enabled" type="xs:boolean" use="optional" default="true"/>
```
jamezp: Same comment as above for this.

ejb/WFLY-6892-ejb-access-logging.adoc
```diff
@@ -0,0 +1,387 @@
...
+                <xs:element name="formatter" type="accessLogFormatterType" minOccurs="0"/>
+                <xs:element name="file" type="pathType"/>
+                <xs:element name="rotate-size" type="sizeType" minOccurs="0"/>
+                <xs:element name="max-backup-index" type="positiveIntType" minOccurs="0"/>
+                <xs:element name="suffix" type="valueType" minOccurs="0"/>
+                <xs:element name="append" type="booleanValueType" minOccurs="0"/>
+            </xs:all>
+            <xs:attribute name="autoflush" type="xs:boolean" use="optional" default="true"/>
+            <xs:attribute name="name" type="xs:string" use="required"/>
+            <xs:attribute name="enabled" type="xs:boolean" use="optional" default="true"/>
```
jamezp: Same comment as above for this.

ejb/WFLY-6892-ejb-access-logging.adoc
```diff
@@ -0,0 +1,387 @@
...
+        </xs:complexType>
+
+        <xs:complexType name="accessLogFormatterType">
+            <xs:annotation>
+                <xs:documentation>
+                    Defines a formatter.
+                </xs:documentation>
+            </xs:annotation>
+            <xs:choice minOccurs="1" maxOccurs="1">
+                <xs:element name="pattern-formatter" type="simpleFormatterType" maxOccurs="1"/>
```
jamezp: I would suggest just using the `named-formatter` and not include this one.

chengfang:
> The XSD seems to look a lot like the logging subsystem. While I like the consistency, I'm not sure we need all of this for the EJB access logging. Part of the reason for the logging subsystem design is limitations with how the log manager itself works. I'm not sure what the plan is for how the logs will be written to a file, but we shouldn't use the limitations based on the log manager.
> 
> I also don't think we likely need all the file type of handlers. We should leave this a little more generic if possible and we can extend it later if required.

Among all file access log types, sizeFileAccessLog seems the least used option, and I think we can skip this one at least for now.  After that, users still have flexibility to choose from console-access-log, server-log-access-log, file-access-log, periodic-file-access-log.  I guess the last one will be the most used one.

ejb/WFLY-6892-ejb-access-logging.adoc
chengfang: I will remove this option and just directly use System.out for console-access-log.

ejb/WFLY-6892-ejb-access-logging.adoc
chengfang: Make sense, and I'll remove this attribute (enabled).

ejb/WFLY-6892-ejb-access-logging.adoc
chengfang: Makes sense, and I'll remove this attribute (enabled).

ejb/WFLY-6892-ejb-access-logging.adoc
chengfang: Makes sense, and I'll remove this attribute (enabled).

ejb/WFLY-6892-ejb-access-logging.adoc
chengfang: Makes sense, and I'll remove this attribute (enabled).

ejb/WFLY-6892-ejb-access-logging.adoc
chengfang: Makes sense, and I'll remove this attribute (enabled).

ejb/WFLY-6892-ejb-access-logging.adoc
chengfang: Okay, I'll just use named-formatter for applying a formatter to a access-log type.  The inline formatter only takes pattern-formatter (not json-formatter), so removing it will make it more consistent without much loss.

chengfang: Thanks for the detailed review.  Now that a `formatter` can only be a `named-formatter`, would it be easier to just change it to a simpel attribute `formatter`?

Also a more general question is, I noticed jboss-as-logging.xsd tends to use value-type elemennt instead of simple attribute, is that the convention, or part of the limitation imposed by logmanager you mentioned above?  Would it be more concise if they are xml attributes in ejb access logging schema?

```xml
<console-access-log name="access-log1" formatter="formatter1" >
  <encoding value="UTF-8" />
</console-access-log>
```

vs

```xml
<console-access-log name="access-log1" >
  <encoding value="UTF-8" />
  <formatter>
    <named-formatter name="format1" />
  </formatter>
</console-access-log>
```

chengfang: Or we could remove the container `<formatter>` element to be something like:

```xml
<console-access-log name="access-log1" >
  <encoding value="UTF-8" />
  <named-formatter name="format1" />
</console-access-log>
```

ejb/WFLY-6892-ejb-access-logging.adoc
jamezp: For now we could even not use an option if we it's a separate resource.

ejb/WFLY-6892-ejb-access-logging.adoc
jamezp: The main reason is backwards compatibility and the logging subsystem was the first subsystem written for JBoss AS 7 :) If I could I would completely re-write as we've learned a lot since then.

That said I would probably just go with `<formatter name="formatter/>`. One thing to consider too is if we want to use the logging terms. It likely makes sense, but we don't have to.

jamezp:
I'm really okay with having any of the "handlers" that seem logical. I was more wondering if we want to keep them at a minimum to hopefully make the feature for WildFly 17.

One question I have is the plan to write your own way of writing to files or are you going to use the handlers from the log manager?

chengfang:
I'd like to reuse any infrastructure support from logging subsystem, to avoid many pitfalls associated with implementing it my own.  Will that need any extra dependencies in ejb3 subsystem?  Right now I don't see wildfly-logging in the dependency list of ejb3 subsysem and module.  Also any pointer how to use logging handler will be very helpful.

ejb/WFLY-6892-ejb-access-logging.adoc
chengfang: I have no problem switching to logging terms like `file-handler` as opposed to `file-access-log`.

jamezp:
You won't be able to depend on the logging subsystem itself as it would make the ejb subsystem dependent on the logging subsystem. If you want to use the handlers and formatters from the log manager you'll need to depend on the `org.jboss.logmanager` module.

ejb/WFLY-6892-ejb-access-logging.adoc
```diff
@@ -0,0 +1,309 @@
...
+Delete if not needed. The intent is if you have a complex feature which can 
+not be delivered all in one go to suggest the strategy. If your feature falls 
+into this category, please mention the Release Coordinators on the pull 
+request so they are aware.
+////
+== Test Plan
+
+== Community Documentation
+////
+Generally a feature should have documentation as part of the PR to wildfly master, or as a follow up PR if the feature is in wildfly-core. In some cases though the documentation belongs more in a component, or does not need any documentation. Indicate which of these will happen.
```
tommaso-borgato: @chengfang this part is not yes complete

ejb/WFLY-6892-ejb-access-logging.adoc
```diff
@@ -0,0 +1,309 @@
...
+
+//== Implementation Plan
+////
+Delete if not needed. The intent is if you have a complex feature which can 
+not be delivered all in one go to suggest the strategy. If your feature falls 
+into this category, please mention the Release Coordinators on the pull 
+request so they are aware.
+////
+== Test Plan
+
```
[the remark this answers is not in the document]
chengfang: Just like turning on any logging, there will be some overhead.  To quantify the performance overhead, there need to be dedicated performance benchmarking and testing.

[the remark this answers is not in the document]
chengfang: yes, good points.

ejb/WFLY-6892-ejb-access-logging.adoc
chengfang: Will work on it once the feature is implemented and finalized.

ejb/WFLY-6892-ejb-access-logging.adoc
```diff
@@ -0,0 +1,309 @@
...
+*** For example: 2019-05-05 12:23:27,003 127.0.0.1 admin hello/helloBean hello
+
+** long : date time ip user ejb method invocation event
+
+*** For example: 2019-05-05 12:23:27,003 127.0.0.1 user1 hello/helloBean hello 1234 received
+                 2019-05-05 12:23:30,003 127.0.0.1 user1 hello/helloBean hello 1234 finished
+
+** default: short
+
+* for more customization, define which attributes to include in ejb access logging with the following attributes:
```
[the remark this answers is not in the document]
chengfang: not all will have counterparts.  Some are specific to web app or that are less useful will not be implemented in ejb access log.  On the other hand, there are some fields in ejb access log but are not in web access log.

ejb/WFLY-6892-ejb-access-logging.adoc
```diff
@@ -0,0 +1,309 @@
...
+** thread: thread name for this invocation
+** server: local server name
+
+* For example, one may configure the following custom pattern, which can be applied to both pattern formatter or
+JSON formatter:
+
+** pattern="date time client ejb method server thread"
+
+
+=== New elements in wildfly-ejb3_6_0.xsd schema
```
mchoma: @chengfang Is it necessary to introduce so complex model addition?

Seems to me a lot of types from logging subsystem are duplicated here. Couldn't be leveraged logging subsystem through capability reference?

Or when I compare with undertow `accessLogType`, there is just one level configuration. Isn't that enough for ejb?
```
    <xs:complexType name="accessLogType">
        <xs:attribute name="pattern" use="optional" type="xs:string" default="common"/>
        <xs:attribute name="worker" use="optional" type="xs:string" default="default"/>
        <xs:attribute name="directory" use="optional" type="xs:string" default="${jboss.server.log.dir}"/>
        <xs:attribute name="relative-to" use="optional" type="xs:string" />
        <xs:attribute name="prefix" use="optional" type="xs:string" default="access_log."/>
        <xs:attribute name="suffix" use="optional" type="xs:string" default="log"/>
        <xs:attribute name="rotate" use="optional" type="xs:string" default="true"/>
        <xs:attribute name="use-server-log" use="optional" type="xs:string" default="false"/>
        <xs:attribute name="extended" use="optional" type="xs:string" default="false" />
        <xs:attribute name="predicate" use="optional" type="xs:string" />
    </xs:complexType>
```

jamezp: Unfortunately there's not much we can leverage from the logging subsystem as the output would likely need to be in a different format.

I do share the concerns though that we're going to end up with 4 ways of using different file handlers, console handlers, syslog handlers, etc.

mchoma: If we are going to duplicate existing configuration and implementation, then that sounds to me as proper fit for using capabilities [1] Different output can be managed by exposing configuration of capability.

What do you think?

[1] https://docs.jboss.org/author/display/WFLY/Working+with+WildFly+Capabilities

jamezp: Unfortunately I don't think that will work in this case. The reason being is the logging subsystem would be likely providing the base capabilities. This means it would need to be present in order for access logging to work. While I would guess most people do not remove the logging subsystem, a change like this would prevent it from being possible if they want to use EJB access logging.

I'm definitely up for suggestions too as I do share the concerns with essentially duplicate resources.

The other catch is going through logging likely won't have the information needed. You'd need to use a `java.util.logging.Formatter` to use any log handlers and the `java.util.logging.LogRecord` likely won't have all the information an EJB access event would need.

We do have an [event logger](https://github.com/wildfly/wildfly-core/tree/master/event-logger) project within WildFly Core that we could however consider extending to get a more standard output from. I would not be opposed to this, but we'd need to open a discussion about that.

tommaso-borgato: @chengfang Still comparing against undertow/src/main/resources/schema/wildfly-undertow_10_0.xsd .... would it be possible to maintain, at least, the closest XSD possible in order to minimize differences from the user point of view?

PR #207, opened 2019-05-28
[WFCORE-4314] enchance keystore commands alias

baranowb:
https://issues.jboss.org/browse/WFCORE-4314
https://issues.jboss.org/browse/EAP7-1266

### DIFF
--- a/core/WFCORE-4314_enchance_keystore_commands_alias.adoc
+++ b/core/WFCORE-4314_enchance_keystore_commands_alias.adoc
@@ -1,0 +1,336 @@
+= Enchance read-alias and read-aliases keystore commands
+:author:            Bartosz Baranowski
+:email:             bbaranow@redhat.com
+:toc:               left
+:icons:             font
+:idprefix:
+:idseparator:       -
+
+== Overview
+
+ Elytron keystore expose `read` commands to access content:
+ * read-alias
+ * read-aliases
+
+ There are two problems with those commands. The `read-alias` command is too verbose by default. When alias is displayed after read command, output include hex format of public-key and encoded fields. This is usually not something required if one inspects key and it clouds output with dump of unreadable string of hex digits.
+ Second problem is that read-aliases is in fact "read-aliases-names". In order to read content of keystore one has to read alias names with `read-aliases` and issue `read-alias` for each name. To make it more user friendly, simple changes are required:
+ * VERBOSE - new parameter for both commands. Default value would be `true`. This parameter control inclusion of bulk hex fields of key(public-key and encoded)
+ * RECURSIVE - new parameter for `read-aliases`. Default value `false`. If set otherwise, `read-aliases` will read recurse into keys, displaying more than names.
+
+
+== Issue Metadata
+
+=== Issue
+
+* https://issues.jboss.org/browse/WFCORE-4314
+
+=== Related Issues
+
+=== Dev Contacts
+
+* mailto:mailto:{email}[{author}]
+* mailto:darran.lofthouse@redhat.com[Darran Lofthouse]
+* mailto:fjuma@redhat.com[Farah Juma]
+
+=== QE Contacts
+
+* mailto:vobratil@redhat.com[Vilem Obratil]
+
+=== Testing By
+// Put an x in the relevant field to indicate if testing will be done by Engineering or QE. 
+// Discuss with QE during the Kickoff state to decide this
+[ ] Engineering
+
+[X] QE
+
+=== Affected Projects or Components
+
+ * wildfly-core
+ 
+=== Other Interested Projects
+
+ n/a
+
+== Requirements
+=== Hard Requirements
+
+ Make read commands of elytron keystore tad more user friendly.
+
+=== Nice-to-Have Requirements
+
+
+
+=== Non-Requirements
+
+=== Design
+
+Preparation:
+[source]
+--
+/subsystem=elytron/key-store=serverKS:add(path=server.keystore.jsk,relative-to=jboss.server.config.dir,credential-reference={clear-text=secret}, type=JKS)
+cd subsystem=elytron/key-store=serverKS
+:generate-key-pair(alias=server1,distinguished-name="CN=www.example.com")
+[standalone@localhost:9990 key-store=serverKS] :generate-key-pair(alias=server2,distinguished-name="CN=www.example.com")
+[standalone@localhost:9990 key-store=serverKS] :generate-key-pair(alias=backup,distinguished-name="CN=www.example.com")
+--
+
+
+`read-aliases` - no change in default output.
+[source]
+--
+[standalone@localhost:9990 key-store=serverKS] :read-aliases
+{
+    "outcome" => "success",
+    "result" => [
+        "backup",
+        "server2",
+        "server1"
+    ]
+}
+--
+
+
+`read-aliases` - inclusion of single parameter does not change output.
+[source]
+--
+[standalone@localhost:9990 key-store=serverKS] :read-aliases(verbose=true)
+{
+    "outcome" => "success",
+    "result" => [
+        "backup",
+        "server2",
+        "server1"
+    ]
+}
+--
+
+
+`read-aliases` - recursive read, with hex field included.
+[source]
+--
+[standalone@localhost:9990 key-store=serverKS] :read-aliases(recursive=true,verbose=true)
+{
+    "outcome" => "success",
+    "result" => {
+        "backup" => {
+            "alias" => "backup",
+            "entry-type" => "PrivateKeyEntry",
+            "creation-date" => "2019-05-27T08:29:42.093+0200",
+            "certificate-chain" => [{
+                "type" => "X.509",
+                "algorithm" => "DSA",
+                "format" => "X.509",
+                "public-key" => "30:82:03:42:30:82:02:35:06:07:2a:86:48:ce:38:04:01:30:82:02:28:02:82:01:01:00:8f:79:35:d9:b9:aa:e9:bf:ab:ed:88:7a:cf:49:51:b6:f3:2e:c5:9e:3b:af:37:18:e8:ea:c4:96:1f:3e:fd:36:06:e7:43:51:a9:c4:18:33:39:b8:09:e7:c
+2:ae:1c:53:9b:a7:47:5b:85:d0:11:ad:b8:b4:79:87:75:49:84:69:5c:ac:0e:8f:14:b3:36:08:28:a2:2f:fa:27:11:0a:3d:62:a9:93:45:34:09:a0:fe:69:6c:46:58:f8:4b:dd:20:81:9c:37:09:a0:10:57:b1:95:ad:cd:00:23:3d:ba:54:84:b6:29:1f:9d:64:8e:f8:83:44:86:77:97:9c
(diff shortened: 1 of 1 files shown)


fjuma:
@baranowb Thanks for working on this! Please drop the EAP7 issue from the titles and filename, we should focus on the community issue here.

A link to the EAP7 issue can still be kept within the doc.

core/EAP7-1266_WFCORE-4314_enchance_keystore_commands_alias.adoc
```diff
@@ -0,0 +1,343 @@
...
+ Elytron keystore expose `read` commands to access content:
+ * read-alias
+ * read-aliases
+
+ There are two problems with those commands. The `read-alias` command is too verbose by default. When alias is displayed after read command, output include hex format of public-key and encoded fields. This is usually not something required if one inspects key and it clouds output with dump of unreadable string of hex digits.
+ Second problem is that read-aliases is in fact "read-aliases-names". In order to read content of keystore one has to read alias names with `read-aliases` and issue `read-alias` for each name. To make it more user friendly, simple changes are required:
+ * VERBOSE - new parameter for both commands. Default value would be `false`. This parameter control inclusion of bulk hex fields of key(public-key and encoded)
+ * RECURSIVE - new parameter for `read-aliases`. Default value `false`. If set otherwise, `read-aliases` will read recurse into keys, displaying more than names.
+
+CAUTION: default value of VERBOSE parameter, in current implementation is `true`. Explicity definition of parameter with default value of `false` is a change in behavior.
```
fjuma: Instead of having `verbose` get set to false by default, I think it should actually default to true. That way, the operations will continue to return the same values they currently return today by default, in case anyone is relying on that behaviour.

core/EAP7-1266_WFCORE-4314_enchance_keystore_commands_alias.adoc
```diff
@@ -0,0 +1,343 @@
...
+NOTICE: `read-alias(alias=...)` and `read-alias(alias=..., verbose=true)` difference is the same as for `read-aliases`.
+
+
+== Test Plan
+
+ Modify existing key store tests to verify features don't break anything. Ensure QE can test HAL.
+
+== Community Documentation
+////
+No documentation needed, command help should be enough.
```
fjuma: Documentation on the new parameters should be added to the "Using the Elytron Subsystem" section in the WildFly documentation.

core/EAP7-1266_WFCORE-4314_enchance_keystore_commands_alias.adoc
```diff
@@ -0,0 +1,343 @@
...
+        "server1"
+    ]
+}
+--
+
+
+`read-aliases` - inclusion of single parameter does not change output.
+[source]
+--
+[standalone@localhost:9990 key-store=serverKS] :read-aliases(verbose=true)
```
fjuma: Should this be `verbose=false`?

core/EAP7-1266_WFCORE-4314_enchance_keystore_commands_alias.adoc
```diff
@@ -0,0 +1,343 @@
...
+        "backup" => {                                                                                                                                                                                                                               
+            "alias" => "backup",                                                                                                                                                                                                                    
+            "entry-type" => "PrivateKeyEntry",                                                                                                                                                                                                      
+            "creation-date" => "2019-05-27T08:29:42.093+0200",                                                                                                                                                                                      
+            "certificate-chain" => [{
+                "type" => "X.509",
+                "algorithm" => "DSA",
+                "format" => "X.509",
+                "public-key" => undefined,
```
fjuma: Instead of having the `public-key` appear as `undefined` to make the output shorter, I think it would be better just to not include `public-key` in the output at all.

core/EAP7-1266_WFCORE-4314_enchance_keystore_commands_alias.adoc
```diff
@@ -0,0 +1,343 @@
...
+                "sha-1-digest" => "4d:a9:16:f9:6d:39:da:03:a1:f0:56:bd:58:46:6c:12:16:59:72:f0",
+                "sha-256-digest" => "73:ea:31:e9:9d:de:38:1a:66:06:e7:66:a6:7d:fe:5d:b6:95:02:58:4f:8e:e0:01:19:ee:90:96:e0:18:2d:a1",
+                "encoded" => undefined,
```
fjuma: Same for `encoded`.

core/EAP7-1266_WFCORE-4314_enchance_keystore_commands_alias.adoc
```diff
@@ -0,0 +1,343 @@
...
+// Put an x in the relevant field to indicate if testing will be done by Engineering or QE. 
+// Discuss with QE during the Kickoff state to decide this
+[ ] Engineering
+
+[ ] QE
+
+=== Affected Projects or Components
+
+ * wildfly-core
+ * HAL(possibly)
```
vobratil: This should be more clear - is HAL going to be affected or not? How?

core/EAP7-1266_WFCORE-4314_enchance_keystore_commands_alias.adoc
```diff
@@ -0,0 +1,343 @@
...
+}
+--
+
+
+NOTICE: `read-alias(alias=...)` and `read-alias(alias=..., verbose=true)` difference is the same as for `read-aliases`.
+
+
+== Test Plan
+
+ Modify existing key store tests to verify features don't break anything. Ensure QE can test HAL.
```
vobratil: I could use some clarification on the "can test HAL" bit. I assume you mean general regression testing?

core/EAP7-1266_WFCORE-4314_enchance_keystore_commands_alias.adoc
baranowb: This is just to show that it, alone, should not trigger change in output.

core/EAP7-1266_WFCORE-4314_enchance_keystore_commands_alias.adoc
baranowb: Indeed. This is why caution meta is there. I did think about backward compatibility, however in this case impact could be negligible? Im fine with inverting this.

fjuma: For backwards compatibility, please update this so that ```verbose``` is set to ```true``` by default.

baranowb: Did, Im just waiting for test everything before I push

baranowb:
Hmm, git is wonky, for some reason it sipped another commit, will fix it shortly.

core/WFCORE-4314_enchance_keystore_commands_alias.adoc
```diff
@@ -0,0 +1,343 @@
...
+        "server1"
+    ]
+}
+--
+
+
+`read-aliases` - inclusion of single parameter does not change output.
+[source]
+--
+[standalone@localhost:9990 key-store=serverKS] :read-aliases(verbose=true)
```
fjuma: The ```verbose``` parameter here only makes sense if it is specified together with ```recursive```, correct? If so, I don't think we should allow the ```verbose``` parameter to be specified on its own, it should only be allowed if ```recursive``` has also been specified.

core/EAP7-1266_WFCORE-4314_enchance_keystore_commands_alias.adoc
baranowb: It wont, at least I could not find anything that would fail( though I might not be on point)

core/WFCORE-4314_enchance_keystore_commands_alias.adoc
baranowb: There is no harm  in having it. But if so - is there an example of such code/practice in WFLY ?

fjuma: There's a ```setRequires``` method that can be set on the ```SimpleAttributeDefinitionBuilder```.

core/WFCORE-4314_enchance_keystore_commands_alias.adoc
```diff
@@ -0,0 +1,337 @@
...
+
+=== Dev Contacts
+
+* mailto:mailto:{email}[{author}]
+* mailto:darran.lofthouse@redhat.com[Darran Lofthouse]
+* mailto:fjuma@redhat.com[Farah Juma]
+* mailto:vobratil@redhat.com[Vilem Obratil]
+
+=== QE Contacts
+
```
fjuma: Vilém Obrátil can be added to the QE contacts section

core/WFCORE-4314_enchance_keystore_commands_alias.adoc
```diff
@@ -0,0 +1,337 @@
...
+=== Testing By
+// Put an x in the relevant field to indicate if testing will be done by Engineering or QE. 
+// Discuss with QE during the Kickoff state to decide this
+[ ] Engineering
```
fjuma: Have you checked with Vilem who will be handling the test plan for this?

vobratil: The RFE has meanwhile moved up in rank, so I suppose that this one is up to me.

core/WFCORE-4314_enchance_keystore_commands_alias.adoc
baranowb: It does not work, or it works a bit differently - what it seems to do is strip some parameters if required is not present( that are sent to controller). But yeah, Ive added this as well.

core/WFCORE-4314_enchance_keystore_commands_alias.adoc
```diff
@@ -0,0 +1,337 @@
...
+        }
+    }
+}
+--
+
+
+`read-aliases` - recursive and without hext fields.
+[source]
+--
+[standalone@localhost:9990 key-store=serverKS] :read-aliases(recursive=false)
```
baranowb: this should be 'true'

core/EAP7-1266_WFCORE-4314_enchance_keystore_commands_alias.adoc
baranowb: done

core/EAP7-1266_WFCORE-4314_enchance_keystore_commands_alias.adoc
baranowb: done

core/WFCORE-4314_enchance_keystore_commands_alias.adoc
```diff
@@ -0,0 +1,336 @@
...
+        }
+    }
+}
+--
+
+
+`read-aliases` - recursive and without hex fields.
+[source]
+--
+[standalone@localhost:9990 key-store=serverKS] :read-aliases(recursive=true)
```
fjuma, line 258: Looks like this should be `read-aliases(recursive=true,verbose=false)` since the output isn't verbose.

baranowb: verbose is false by default, but I can add it just the emphasise.

fjuma: Isn't `verbose` true by default?

fjuma (approved):

PR #212, opened 2019-06-18
WFLY-12192 Add support for injecting Optional<T> parameter types

liweinan:


liweinan:
@darranl okay.

> Please just use the WFLY issue in the file name, the EAP7 issues are internal planning issues only so should just be listed under 'Related Issues'.

okay! I have:

- renamed the PR title
- renamed the file
- moved EAP issue in doc into 'related issues' seciton

jaxrs/WFLY-12192_optional_param.adoc
```diff
@@ -0,0 +1,64 @@
...
+* WildFly
+* RESTEasy
+
+== Requirements
+
+The behavior described in the first section is expected.
+
+=== Test Plan
+
+The feature is already extensively tested in RESTEasy testsuite, so that should be run against relevant WildFly version.
```
vmarekrh: This proposal document writes what this issue include.
Please write that the tests for this feature will be stored in the RESTEasy testsuite.

Currently, we integrate these changes to Wildfly. If I good understand.
These changes are in RESTEasy version 4 but Wildfly currently uses RESTEasy 3.7 Final.

liweinan:
@vmarekrh thanks for reviewing! I'm working on this doc, once I finished editing it I'll ask you to review again.

jaxrs/WFLY-12192_optional_param.adoc
```diff
@@ -0,0 +1,64 @@
+= WFLY-12192 / EAP7-1248 Support for java.util.Optional on PathParam and QueryParam
```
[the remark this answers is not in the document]
liweinan: okay! In addition the current title is not accurate, because the other parameter types are also affected. So I changed title of relative JIRA issues, and created sub-task: [RESTEASY-2287 / Add complete tests for different optional paramter types](https://issues.jboss.org/browse/RESTEASY-2287).

jaxrs/WFLY-12192_optional_param.adoc
```diff
@@ -0,0 +1,64 @@
...
+:idseparator:       -
+:issue-base-url:    https://issues.jboss.org/browse
+
+== Overview
+
+Currently, `PathParam` and `QueryParam` support the following construct (from https://docs.jboss.org/resteasy/docs/3.0.17.Final/userguide/html_single/#_PathParam):
+
+> The parameter type you inject into can be any primitive type, a String, or any Java object that has a constructor that takes a String parameter, or a static `valueOf` method that takes a String as a parameter.
+
+This proposal is to support `java.util.Optional.ofNullable` as a wrapper object for Strings, which is always called. This will eliminate all null checks for optional parameters in resource classes with the different Optional followup methods, such as optional.ifPresent.
```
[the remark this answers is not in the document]
liweinan: I have modified the text.

jaxrs/WFLY-12192_optional_param.adoc
```diff
@@ -0,0 +1,64 @@
...
+[] QE
+
+=== Affected Projects or Components:
+
+* WildFly
+* RESTEasy
+
+== Requirements
+
+The behavior described in the first section is expected.
```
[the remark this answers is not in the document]
liweinan: requirement section updated.

liweinan:
I have used some time to root cause other issue this week and finished working that part. I'll start to write test plan for this feature request.

liweinan:
prepared tests to write test plan: https://github.com/wildfly/wildfly-proposals/pull/212

jaxrs/WFLY-12192_optional_param.adoc
liweinan: @vmarekrh Yeah, the changes will be backported from resteasy 4.x to resteasy 3.x.

asoldano: @vmarekrh (cc @liweinan), the changes are in RESTEasy master but will be backported to 3.9 branch as soon as there's an agreement here. As part of the backport, new tests might have to be added.
WildFly is currently using 3.8.1.Final and will most likely be updated to 3.9.0.Final

jaxrs/WFLY-12192_optional_param.adoc
```diff
@@ -0,0 +1,66 @@
+= WFLY-12192 Add support for injecting Optional<T> parameter types
+:author:            Wei Nan Li
+:email:             weli@redhat.com
+:toc:               left
+:icons:             font
+:keywords:          comma,separated,tags
```
marekkopecky: @liweinan Can you please updates these default keywords?

liweinan: ok!

liweinan: @marekkopecky Sorry for the late reply. I was moving to some other tasks and let me come back to focus on this one now. For this one I have question: I checked other documents: https://github.com/wildfly/wildfly-proposals/tree/master/jaxrs
And they seems to contain the same keywords as above. I also checked the pr like this:
https://github.com/wildfly/wildfly-proposals/pull/233/files
And it contains similar keywords. Could you please let me know what should I change about these keywords? Thanks!

marekkopecky: Well, so which component is this this RFE related with? Possible keywords could be `RESTEasy`, `JAX-RS`, `Injection`, etc.

liweinan: good advice! Let me update the tags.

liweinan:
The document is updated according to requirement, @manovotn @jkasik could you please help to review it? Thanks!

manovotn:
Browsing the issues that are loosely linked to this, do I see it correctly that this planning issue is for something that is actually already implemented and working in resteasy?
E.g. PR for RESTEASY-1419 has been merged already

asoldano:
@manovotn @liweinan the changes for RESTEASY-1419 have been merged on master back in May and has such got included in 4.1.0.Final release. So, basically, yes, they're working. RESTEasy 4 series is not consumed by WFLY / EAP, though, hence the need for this WFLY-12192 jira. As soon as this is approved, we'll merge the changes to RESTEasy 3.9 branch, which is what WFLY currently consumes.

asoldano:
@manovotn : @liweinan is from the RESTEasy team. I lead the RESTEasy team (with Ron) ;-)

asoldano:
@honza-kasik , @vmarekrh (cc @marekkopecky, ), this is waiting for your approval...

marekkopecky:
> honza-kasik , vmarekrh (cc marekkopecky, ), this is waiting for your approval...

@asoldano See my last coment on EAP7 jira.

jaxrs/WFLY-12192_optional_param.adoc
```diff
@@ -0,0 +1,66 @@
...
+
+* mailto:weli@redhat.com[Wei Nan Li]
+* mailto:rsigal@redhat.com[Ron Sigal]
+* mailto:asoldano@redhat.com[Alessio Soldano]
+* mailto:ema@redhat.com[Jim Ma]
+* mailto:rsearls@redhat.com[Rebecca Searls]
+
+=== QE Contacts:
+
+* mailto:vmarek@redhat.com[Vratislav Marek]
```
tterem: Add me as QE contact here please.

tterem:
@liweinan I am new QE contact for this. It needs to be written in the analysis. Apart from this minor issue, the analysis is fine and I will approve it after it's changed.

jaxrs/WFLY-12192_optional_param.adoc
liweinan: okay!

liweinan:
@tterem Thanks for reviewing! I wrote that `testPathParamNeverAbsentThrowsException` test just to cover all types of params. Actually `@PathParam` is different from other type of params because it can't be "optional". Because the path param must be exist in path, and when we don't have it in path, actually we have a different path and it will either be served by a method mapped to that different path, or that path doesn't exist. So the test is just to cover this absent situation, and it's described in our test-plan. IMHO it's a special situation that path param can't be `optional`. If you agree we just  cover it in test-plan, I'll keep it as is, wdyt?

tterem:
@liweinan 

> Actually @PathParam is different from other type of params because it can't be "optional".

Exactly. That's why I think it should be mentioned directly in the analysis. Title and analysis only talk about parameters in general. It should be mentioned that @PathParam is an exception and that using Optional with @PathParam doesn't allow us to omit that parameter from the path. Would you mind to add one sentence about it?

liweinan:
@tterem sure! I have updated the `Overview` section.  Please help to check.

tterem:
@liweinan Thank you, but I was thinking about it more and I need to ask: Does using Optional with @PathParam have any benefit at all? Wouldn't it be less confusing for the users if we disabled that completely?

liweinan:
@tterem Nice suggestion! With the thought I checked the implementation in RESTEasy side: 

- https://github.com/resteasy/Resteasy/pull/1898/files#diff-3fd561a4c754fed4427cd68f808456c6R91

From above we can see the design is universal for all types of parameters. And here is the `PathParam` part:

- https://github.com/resteasy/Resteasy/pull/1898/files#diff-3fd561a4c754fed4427cd68f808456c6R110

IMHO, the current design doesn't affect the existing `PathParam` function(In other word, `Optionals` are only enabled for other types of annotations).

wdyt?

tterem:
@liweinan 
> IMHO, the current design doesn't affect the existing PathParam function(In other word, Optionals are only enabled for other types of annotations).

Maybe I don't understand what do you mean. Optionals are now enabled for PathParam as tested here:

https://github.com/resteasy/Resteasy/pull/2084/files#diff-509272a2a9b4858a9ebcba9a77a807c6R93

My concern is whether it's a good idea if path parameter cannot be null anyway?

liweinan:
@tterem The current `Optional` design does not change the behavior of `PathParam`. Do you  mean to remove this test case?  This test case ensures the `PathParam` works as before.

tterem:
@liweinan 
> This test case ensures the PathParam works the same way as before.

No, it doesn't. This RFE is not implemented in WF17. Goal of the analysis is to describe new feature for WF18. Using Optional with PathParam is new feature introduced by this RFE and I believe it could lead to confusion. So that's why I am asking if it has any benefit for users. If not, I would suggest to drop it and enable Optionals only for other parameter annotations.

liweinan:
@tterem How about rewrite this section:

https://github.com/wildfly/wildfly-proposals/pull/212/files#diff-229b287fb806a847d33b42623b239f17R19

from:

> Please note that `@PathParam` is an exception and that using `Optional` with `@PathParam` doesn't allow us to omit that parameter from the path. The reason is that the path param must be exist in path, and when we don't have it in path, actually we have a different path and it will either be served by a method mapped to that different path, or that path doesn't exist.

to:

> Please note that `@PathParam` is an exception that does not support the `Optional` feature and it works as before.

tterem:
@liweinan Yes, I agree with suggested change.

ronsigal:
Hi @tterem, @liweinan,

I guess I have a slightly different attitude toward the @Optional in @PathParam issue. To say that @Optional is not supported for @PathParam means that using @Optional there would break an application, which wouldn't happen if we just say that using @Optional is a NO-OP in that case. But I don't have strong feelings on the subject, so I wouldn't object either way.

-Ron

liweinan:
@tterem @ronsigal Thanks for advices! With both of your advices, how about we write it like this:

> Please note that `@PathParam` is an exception that keeps its original behavior and is not affected by this `@Optional` feature addition. The reason is that using `@Optional` under the `@PathParam` use case is a NO-OP.

ronsigal:
Hi @tterem,

Well, I think your version is a little ambiguous. In particular, it's not clear to me if you're saying

1. "@Optional is not available for @PathParam because it would just be a NO-OP anyway.", or
2. "@Optional is available for @PathParam, but it just acts like a NO-OP".

Do you see what I mean?

tterem:
Hi @ronsigal,

You probably meant to tag @liweinan.

Anyway, I suggest to go with the first option. Optional is implying that something is optional (and can be null) and that is not the case with path parameter. That's why I think it shouldn't be available to use there.

ronsigal:
@tterem, @liweinan, ok, I'm not going to argue.  In that case, I would change the text slightly:

Please note that @PathParam is an exception for which @Optional is not available. The reason is that @Optional for the @PathParam use case would just be a NO-OP, since an element of the path cannot be omitted.

Or something like that.

liweinan:
@ronsigal Looks good to me @tterem If you agree with Ron's version I'll go ahead to update this PR.

tterem:
@liweinan
Yes, I agree. Just note that it's **Optional** not @optional.

liweinan:
@tterem okay! @ronsigal Thanks for help!

liweinan:
@tterem Text is updated. Anything else needs me to edit please let me know.

tterem:
@liweinan You've put it there with @. As I said in my previous comment, it's just Optional. It's not an annotation.

liweinan:
@tterem Sorry my mistake. `@Optional` is fixed to `Optional`.

tterem:
@liweinan Thanks. I don't have any more objections.

ronsigal:
Ah, using @Optional was my fault. :-P

Thanks, @tterem.

liweinan:
@tterem @ronsigal Thanks for review! With my understanding, we need @asoldano to approve, he's on PTO and will be back next week.

asoldano:
@liweinan you don't need me to review / formal approve this, the RFE is assigned to you and Ron also already checked this ;-) Anyway... approved.

PR #213, opened 2019-06-20
[WFCORE-4360] Support expression resolution backed by a CredentialStore.

darranl:
This is very much work in progress however these are my initial plans in relation to delivery of WFCORE-4360 and the addition of supporting expression resolution backed by a credential store.

https://issues.jboss.org/browse/EAP7-677
https://issues.jboss.org/browse/WFCORE-4360

### DIFF
--- a/elytron/WFCORE-4360-CredentialStore_Expression_Resolution.adoc
+++ b/elytron/WFCORE-4360-CredentialStore_Expression_Resolution.adoc
@@ -193,7 +193,7 @@ We will not support the decryption of expressions other than the support within 
 
 Base64 representations will be used for both the export and import of the generated keys and for the expression value to be passed to the expression resolver - generally these details should be opaque to the end user however we may include the description within the documentation allowing for others to make use of the format.  We should consider that in the future alternative representations may be required either to support alternative key types or to support alternative representations so the representations will also be versioned.
 
-The general structure of the first representation will be (illestrated as an array): -
+The general structure of the first representation will be (illustrated as an array): -
 
     { 'E', 'L', 'Y', (VERSION), (TYPE), VALUE... }
 
--- a/elytron/WFCORE-4360-CredentialStore_Expression_Resolution.adoc
+++ b/elytron/WFCORE-4360-CredentialStore_Expression_Resolution.adoc
@@ -109,6 +109,55 @@ As with the command line utility it is expected each of these will need to take 
 
 NOTE: The credential store also offers a programatic API, should the tooling and operations be insufficient for an end user manual population of the credential store may remain an option.
 
+==== Command Line Encryption
+
+Whilst the Elytron tool does contain a `credential-store` command the encryption of values is really a special operation so a new command `encryption` will be added to the tool.
+
+When using the new `encryption` command it will be necessary to reference an existing `alias` within an existing `credential-store` to obtain an instance of the `SecretKey` that will be used for encryption of a plain text String, this command will make use of the same attributes as used by the `credential-store` command to identify the store and alias to use to load the key.  This is important so that using different commands in the same command line utility is consistent.
+
+The encryption of String instances will make use of `javax.crypto.Cipher` APIs, although the only `SecretKey` type supported will be `AES` various transformations will be supported, a transformation can be specified in the format: -
+
+    Algorithm Name/Algorithm Mode/Padding
+
+The support algorithm names will be: -
+
+ * `AES`
+ * `AES_128`
+ * `AES_192`
+ * `AES_256`
+
+The final three names restrict the algorithm to the specified key size.
+
+The supported modes will be: -
+
+ * `CBC`
+ * `CTR`
+ * `GCM`
+ 
+The following padding will be supported: -
+
+ * `NoPadding`
+ * `PKCS5Padding`
+
+Different permutations of name, mode, and padding require different `ParameterSpec` instances to configure the `Cipher`, by defining the supported set in advance selection of the correct `ParameterSpec` can be more deterministic.
+
+If no cipher is specified the tool will default to `????`.
+
+The new `encryption` command will support two actions: -
+
+ * `--encrypt` - To encrypt a value and return the result Base64 encoded.
+ * `--create-expression` - To encrypt a value and return it as a complete expression.
+ 
+In both cases the value to be encrypted can be passed in with the action or if not specified the user will be prompted to enter the value.  In the case of `--create-expression` the user can also optionally specify the `resolver` that will be used to create the expression String.
+
+The following arguments can also be specified to both actions: -
+
+ * `--iv-length` - The size in bytes of the initialisation vector, valid range <12,16> with a default of 12.
+ * `--authentication-tag-length` - The length of the authentication tag in bits, only applicable for the `GCM` mode.  Default 128.
+
+The tool will not add any support for the decryption of expressions or their values.
+
+
 ==== Expression Resolver
 
 Presently WildFly Core supports an expression resolver that can delegate to a Vault configuration and if that is not available fall back to use either system properties or environment variables.  This will be updated to make use of the `CapabilityRegistry` and attempt to lookup a capability using a pre-defined constant (To be defined) which exposes an expression resolver runtime API.  The expression resolver looked up using a capability will be used first to attempt to resolve any expression, if this is successful the resolved value will be used - if not we will fallback to the existing behaviour of using the vault and if that fails using a system property or environment variable.
@@ -138,22 +187,7 @@ The return value of the operation will be the complete expression ready to be us
 
 We will not support the decryption of expressions other than the support within the management model to resolve existing expressions.
 
-==== Command Line Encryption
 
-Whilst the Elytron tool does contain a `credential-store` command the encryption of values is really a special operation so a new command `encryption` will be added to the tool.
-
-This command will duplicate the parameters used by the `credential-store` command to configure the location and initialisation of the underlying credential store, additionally it will take an argument to specify the alias from the credential store to load the `SecretKey` to use for encryption.
-
-Additional arguments to the tool will be used to specify the algorithm to use for encryption as well as any additional parameters requires such as a salt or iteration count.
-
-The new `encryption` command will support two actions: -
-
- * `--encrypt` - To encrypt a value and return the result Base64 encoded.
- * `--create-expression` - To encrypt a value and return it as a complete expression.
- 
-In both cases the value to be encrypted can be passed in with the action or if not specified the user will be prompted to enter the value.  In the case of `--create-expression` the user can also optionally specify the `resolver` that will be used to create the expression String. 
-
-The tool will not add any support for the decryption of expressions or their values.
 
 ==== Base64 Representations
 
@@ -173,7 +207,7 @@ As version `1` of this implementation only supports `AES` `SecretKey` instances 
 
 An encrypted expression will be encoded with a type of 2: -
 
-    { 'E', 'L', 'Y', (VERSION), 2, DATA... }
+    { 'E', 'L', 'Y', (VERSION), 2, IV Length, Initialisation Vector, Cipher Text }
 
 At the moment only a single version of the representation will be supported, in the future the commands and operations that generated the encoded values may take a parameter to specify which version to generate however that is not required within the first implementation.  Additionally later versions of the tooling and operations should always attempt to use the oldest version possible to encode a value allowing newer tools to encode values for use by older versions.
 
--- a/elytron/WFCORE-4360-CredentialStore_Expression_Resolution.adoc
+++ b/elytron/WFCORE-4360-CredentialStore_Expression_Resolution.adoc
@@ -87,6 +87,20 @@ This enhancement is only in relation to expression resolution within the applica
 
 Expression resolution will only be supported against attributes that already support expression resolution, this enhancement will not perform a review of which attributes support expression resolution and will not be changing any attributes to support expression resolution.
 
+== Implementation Plan
+
+A large portion of this enhancement is going to be in relation to the addition of the expression resolver within the management model and demonstrating the viability of the approach for this reason the implementation will be divided into two stages.
+
+=== Hard Requirements
+
+A first pass of the implementation will focus exclusively on the defined "Hard Requirements" this will result in an end to end delivery of the enhancement.
+
+=== Nice-to-Have Requirements
+
+The nice to have requirements are effectively additional enhancements that can be added to this work such as alternative key types etc..
+
+After the hard requirements have been implemented we will make a decision as to which of the nice to have requirements we will also include in this work, a part of this decision process will take into account any upcoming feature freeze deadlines.  This enhancement can be considered complete once the hard requirements have been implemented, if any of the nice to have requirements have not been included in the enhancement we will consider raising future feature requests to further enhance this in a later stage.
+
 == Test Plan
 
 Any utilities added to the Wildfly Elytron project will also be accompanied with their own unit tests.
--- a/elytron/WFCORE-4360-CredentialStore_Expression_Resolution.adoc
+++ b/elytron/WFCORE-4360-CredentialStore_Expression_Resolution.adoc
@@ -16,11 +16,18 @@ The present `CredentialStore` integration is implemented to provide a tight inte
 
 * https://issues.jboss.org/browse/WFCORE-4360[WFCORE-4360]
 * https://issues.jboss.org/browse/WFLY-12215[WFLY-12215]
+* https://issues.jboss.org/browse/ELY-1842[ELY-1842] - Command line support for `SecretKey` manipulation.
 
 === Related Issues
 
 * https://issues.jboss.org/browse/EAP7-677[EAP7-677]
 * https://issues.jboss.org/browse/WFLY-11101[WFLY-11101]
+
+=== Other Links
+
+The following is a link to a discussion on the mailto:wildfly-dev@lists.jboss.org[wildfly-dev] mailing list.
+
+https://lists.jboss.org/pipermail/wildfly-dev/2019-July/006932.html[Expression Resolver Discussion]
 
 === Dev Contacts
 
--- a/elytron/WFCORE-4360-CredentialStore_Expression_Resolution.adoc
+++ b/elytron/WFCORE-4360-CredentialStore_Expression_Resolution.adoc
@@ -1,0 +1,93 @@
+= WFCORE-4360 CredentialStore Backed Expression Resolution
+:author:            Darran Lofthouse
+:email:             darran.lofthouse@redhat.com
+:toc:               left
+:icons:             font
+:idprefix:
+:idseparator:       -
+
+== Overview
+
+The present `CredentialStore` integration is implemented to provide a tight integration between the locations in the model which require access to credentials, however their is also a desire to encrypt other attribute values stored in the management model.  This enhancement is specifically to focus on that latter requirement.
+
+== Issue Metadata
+
+=== Issue
+
+* https://issues.jboss.org/browse/WFCORE-4360[WFCORE-4360]
+* https://issues.jboss.org/browse/WFLY-12215[WFLY-12215]
+
+=== Related Issues
+
+* https://issues.jboss.org/browse/EAP7-677[EAP7-677]
+* https://issues.jboss.org/browse/WFLY-11101[WFLY-11101]
+
+=== Dev Contacts
+
+* mailto:{email}[{author}]
+
+=== QE Contacts
+
+* mailto:jkasik@redhat.com[Jan Kasik]
+
+=== Testing By
+
+[ ] Engineering
+[x] QE
+
+=== Affected Projects or Components
+
+Utility classes may be required within the WildFly Elytron project and will be developed as the need is identified, however the primary integration will be performed within the WildFly Core project as this enhancment is predominantly in relation to management model expression resolution.
+
+Additionally all documentation lives within the WildFly project so updates will be applied there accordingly.
+
+=== Other Interested Projects
+
+== Requirements
+
+=== Hard Requirements
+
+The general intent for this enhancement is to develop some form of configurable expression resolver.
+
+The expression resolver will reference a specific alias within a specific credential store, this credential will be used to decrypt the expression which was previously encrypted using this credential.
+
+Using the credential store it will be possible to either add or generate suitable credentials to be used for the encryption / decryption of values.  This will need to include: -
+
+ * The standalone elytron tool.
+ * The management operations exposed on the credential store.
+
+Functionality should be added to allow an administrator to encrypt a clear text String so that it can be stored as an expression, this will be added to: -
+
+ * The standalone elytron tool.
+ * New management operations, this may however be against the expression resolver not the credential store. 
+
+The domain mode relationships will need to considered further, where we used the PicketBox vault we defined the vault within the host.xml and used it's values in the domain.xml.  As the credential store can both be defined in the host.xml and the domain.xml it would make sense to make use of the one stored within that descriptor, however this could be impacted based on how the configurable expression resolver is defined.
+
+=== Nice-to-Have Requirements
+
+Support for multiple expression resolvers being defined, allowing the correct one to be identified for a specific expression value.  It may not be possible to implement this in the initial enhancement however this should still be considered to allow it's potential addition at a later point.
+
+=== Non-Requirements
+
+This enhancement will not support the retrieval of plain text strings from the credential store, this enhancement is specifically adding support for decrypting reversibly encrypted attribute expression values.
+
+Automatic encryption of attribute values will not be supported via this enhancement, as multiple steps are required that would be better performed within enhancements to the management tooling - each of which would require special consideration based on their own user interfaces.
+
+As with other CredentialStore use cases no automatic replication of the store or it's entries are supported with this enhancement.
+
+This enhancment will not add support for migrating expressions to a different credential, however if support for multiple expression resolvers is added at a later point there may be opportunities to support migration.
+
+== Test Plan
+
+Any utilities added to the Wildfly Elytron project will also be accompanied with their own unit tests.
+
+The major testing of this enhancement will happen within the WildFly Core project allowing us to test in the same location it is implemented.
+
+== Community Documentation
+
+The CredentialStore is presently missing community documentation so before development of this enhancement can proceed this needs to be added so further enhancements can be added for this feature, this will be added under the following issue: -
+
+* https://issues.jboss.org/browse/WFLY-11101[WFLY-11101]
+
+The documentation will then subsequently be enhanced to include details of the new feature.
+
--- a/elytron/WFCORE-4360-CredentialStore_Expression_Resolution.adoc
+++ b/elytron/WFCORE-4360-CredentialStore_Expression_Resolution.adoc
@@ -33,11 +33,14 @@ The present `CredentialStore` integration is implemented to provide a tight inte
 === Testing By
 
 [ ] Engineering
+
 [x] QE
 
 === Affected Projects or Components
 
-Utility classes may be required within the WildFly Elytron project and will be developed as the need is identified, however the primary integration will be performed within the WildFly Core project as this enhancement is predominantly in relation to management model expression resolution.
+Changes will be required within WildFly Elytron to both provide utilities needed for this enhancement and also updates will be required within the Elytron tool to support additional key management operations and potentially offline encryption of values.
+ 
+Beyond this the primary integration will be performed within the WildFly Core project as this enhancement is predominantly in relation to management model expression resolution.
 
 Additionally all documentation lives within the WildFly project so updates will be applied there accordingly.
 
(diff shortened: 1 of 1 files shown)


elytron/WFCORE-4360-CredentialStore_Expression_Resolution.adoc
```diff
@@ -0,0 +1,105 @@
...
+:author:            Darran Lofthouse
+:email:             darran.lofthouse@redhat.com
+:toc:               left
+:icons:             font
+:idprefix:
+:idseparator:       -
+
+== Overview
+
+The present `CredentialStore` integration is implemented to provide a tight integration between the locations in the model which require access to credentials, however their is also a desire to encrypt other attribute values stored in the management model.  This enhancement is specifically to focus on that latter requirement.
```
honza-kasik: > ...however their `<?missing?>`...

elytron/WFCORE-4360-CredentialStore_Expression_Resolution.adoc
```diff
@@ -0,0 +1,105 @@
...
+* mailto:jkasik@redhat.com[Jan Kasik]
+
+=== Testing By
+
+[ ] Engineering
+[x] QE
+
+=== Affected Projects or Components
+
+Utility classes may be required within the WildFly Elytron project and will be developed as the need is identified, however the primary integration will be performed within the WildFly Core project as this enhancment is predominantly in relation to management model expression resolution.
```
honza-kasik: *enhancement

elytron/WFCORE-4360-CredentialStore_Expression_Resolution.adoc
```diff
@@ -0,0 +1,105 @@
...
+
+=== Other Interested Projects
+
+== Requirements
+
+=== Hard Requirements
+
+The general intent for this enhancement is to develop some form of configurable expression resolver.
+
+The expression resolver will reference a specific alias within a specific credential store, this credential will be used to decrypt the expression which was previously encrypted using this credential.  The expression resolver will also contain configuration such as the algorithm to use.
```
honza-kasik: Will be this configuration stored in the model? Where it will be stored?

elytron/WFCORE-4360-CredentialStore_Expression_Resolution.adoc
```diff
@@ -0,0 +1,105 @@
...
+
+Additionally all documentation lives within the WildFly project so updates will be applied there accordingly.
+
+=== Other Interested Projects
+
+== Requirements
+
+=== Hard Requirements
+
+The general intent for this enhancement is to develop some form of configurable expression resolver.
```
honza-kasik: Will be this expression resolver presented by a management resource? If it will can you please present an example?

elytron/WFCORE-4360-CredentialStore_Expression_Resolution.adoc
```diff
@@ -0,0 +1,105 @@
...
+
+The expression resolver will reference a specific alias within a specific credential store, this credential will be used to decrypt the expression which was previously encrypted using this credential.  The expression resolver will also contain configuration such as the algorithm to use. 
+
+The expression resolver should support both generated keys and password based encryption.
+
+Using the credential store it will be possible to either add or generate suitable keys to be used for the encryption / decryption of values.  This will need to include: -
+
+ * The standalone elytron tool.
+ * The management operations exposed on the credential store.
```
honza-kasik: Which management operations? How will they look? Will there be any changes to existing ones?

elytron/WFCORE-4360-CredentialStore_Expression_Resolution.adoc
```diff
@@ -0,0 +1,105 @@
...
+
+Where generated keys are used the tools and management operations should support exporting and importing the keys to allow them to be 'moved' between credential stores.  Where possible this should be achieved using specification defined formats.
```
honza-kasik: > Where generated keys are used the tools and management operations...

Can you please give an example of such management operation and how it will change?

> ...specification defined formats.

Can you please amend which specification are you talking about and which formats are taken into account?

elytron/WFCORE-4360-CredentialStore_Expression_Resolution.adoc
```diff
@@ -0,0 +1,105 @@
...
+
+Functionality should be added to allow an administrator to encrypt a clear text String so that it can be stored as an expression, this will be added to: -
+
+ * The standalone elytron tool.
+ * New management operations, this may however be against the expression resolver not the credential store.
```
honza-kasik: How will such management operation look? What parameters will be there? What parameters will be optional/mandatory?

elytron/WFCORE-4360-CredentialStore_Expression_Resolution.adoc
```diff
@@ -0,0 +1,105 @@
...
+ * The management operations exposed on the credential store.
+
+Where generated keys are used the tools and management operations should support exporting and importing the keys to allow them to be 'moved' between credential stores.  Where possible this should be achieved using specification defined formats.
+
+Functionality should be added to allow an administrator to encrypt a clear text String so that it can be stored as an expression, this will be added to: -
+
+ * The standalone elytron tool.
+ * New management operations, this may however be against the expression resolver not the credential store. 
+
+The domain mode relationships will need to considered further, where we used the PicketBox vault we defined the vault within the host.xml and used it's values in the domain.xml.  As the credential store can both be defined in the host.xml and the domain.xml it would make sense to make use of the one stored within that descriptor, however this could be impacted based on how the configurable expression resolver is defined.
```
honza-kasik: Do I understand it correctly that the problem is whether allow expressions in both `domain.xml` and `host xml` or is the issue something else?

To be honest, I am not certain that I fully understand how definition of expression resolver will affect the location where credential store is defined. Can you please explain it to me?

elytron/WFCORE-4360-CredentialStore_Expression_Resolution.adoc
```diff
@@ -0,0 +1,105 @@
...
+
+=== Nice-to-Have Requirements
+
+Support for multiple expression resolvers being defined, allowing the correct one to be identified for a specific expression value.  It may not be possible to implement this in the initial enhancement however this should still be considered to allow it's potential addition at a later point.
+
+It would be nice to cross reference subsystem managed security providers for the expression resolver, however this component needs to be usable at the start of `Stage.RUNTIME` so there will be a limit as to how many subsystem managed resources can be depended upon.
```
honza-kasik: What would be benefits of this requirement?

elytron/WFCORE-4360-CredentialStore_Expression_Resolution.adoc
```diff
@@ -0,0 +1,105 @@
...
+
+=== Non-Requirements
+
+This enhancement will not support the retrieval of plain text strings from the credential store, this enhancement is specifically adding support for decrypting reversibly encrypted attribute expression values.
+
+Automatic encryption of attribute values will not be supported via this enhancement, as multiple steps are required that would be better performed within enhancements to the management tooling - each of which would require special consideration based on their own user interfaces.
+
+As with other CredentialStore use cases no automatic replication of the store or it's entries are supported with this enhancement.
```
honza-kasik: Can you please explain to me what replication means in this context?

elytron/WFCORE-4360-CredentialStore_Expression_Resolution.adoc
```diff
@@ -0,0 +1,105 @@
...
+
+This enhancment will not add support for migrating expressions to a different credential, however if support for multiple expression resolvers is added at a later point there may be opportunities to support migration.
```
honza-kasik: *enhancement

By credential you mean credential store?

elytron/WFCORE-4360-CredentialStore_Expression_Resolution.adoc
darranl: At the moment I need to verify my plans to incorporate an expression resolver in the way I am envisioning  - the majority of this enhancement is relatively straight forward but it all hinges on that component.

Once I have that verified I will add some configuration detail.

elytron/WFCORE-4360-CredentialStore_Expression_Resolution.adoc
darranl: As I say in my previous comment I just need to verify the feasibility of this resolver and then I can start to add some anticipated configuration.

elytron/WFCORE-4360-CredentialStore_Expression_Resolution.adoc
darranl: More details will be added soon.  Writing this analysis it has become apparent this will not be one large RFE, instead it will be small improvements to various components.  They will probably need to be visited for some in-depth work one at a time to clarify their exact details.

elytron/WFCORE-4360-CredentialStore_Expression_Resolution.adoc
darranl: At the moment this is really TBD and will evolve as the enhancement progresses.

elytron/WFCORE-4360-CredentialStore_Expression_Resolution.adoc
darranl: When using the PicketBox Vault the Vault definition always lived in the host.xml and was used to handle expressions in both the host.xml and the domain.xml of that node.

The credential store on the other hand is now defined within the Elytron subsystem and is always used to resolve credentials in the co-located configuration i.e. the one in the host.xml will be used to obtain credentials for host.xml and the one in domain.xml will be used for credentials in domain.xml.

I suspect the latter would be preferable for consistency with the new configuration making use of subsystems - however the interaction between the host.xml and domain.xml handling is quite complex so when we did into this further it may throw up some additional complexities that may force our direction one way or the other.

elytron/WFCORE-4360-CredentialStore_Expression_Resolution.adoc
darranl: The expression resolver is likely to be using APIs such as https://docs.oracle.com/javase/8/docs/api/javax/crypto/Cipher.html - the implementations of the algorithms coming from the installed java.security.Provider instances - being able to cross reference registered providers would allow for alternative implementations.

One example could be the administrator prefers to use the BouncyCastle provider for this functionality.

Another possibility which we would not directly cover in this RFE is both the Key and the encryption could be handled by a HSM - this integration again is generally configured via providers.

Having said all that custom providers can still be added to the java.security.policy file so there are different ways to achieve the same although they would not be manageable in the same way.

elytron/WFCORE-4360-CredentialStore_Expression_Resolution.adoc
darranl: In domain mode we will not automatically transfer the contents of one credential store to another node in the domain.

darranl:
Thanks for the comments, I will incorporate the typos in the next commit - the other questions will be expanded on in more detail as each component is visited.

honza-kasik:
Thanks for updates @darranl! IMHO it should be clear at least what minimal set of arguments will be required by new management operations (fine tuning will be possible during development process). It would be also great if you could present an example of usage of this RFE - e.g. how user will use these new credential store backed expression and expression resolvers.
I agree with you, that this needs to be split to several smaller RFE. Does it than make sense to cover this set of RFEs with this one large analysis?

elytron/WFCORE-4360-CredentialStore_Expression_Resolution.adoc
```diff
@@ -0,0 +1,210 @@
...
+ * Any additional parameters such as `salt`, `iteration count` etc.. as required to configure the algorithm.
+
+The `expression=encryption` resource will also contain an operation `create-expression` the purpose of this operation being to take a clear text value and using one of the defined `resolvers` output an expression that can be used elsewhere in the management model.  The operation will just require two parameters: -
+
+ * `resolver` - The name of the defined resolver to use to encrypt the data.
+ * `data` - The clear text value to be encrypted.
+
+The return value of the operation will be the complete expression ready to be used elsewhere in the mode.
+
+The format of the expressions resolved using the WildFly Elytron expression resolver will be `${ENC:Resolver:ENCRYPTED_DATA}` where `Resolver` is a reference to the specific resolver defined in the `expression=encryption` resource and `ENCRYPTED_DATA` is the data to be decrypted encoded using Base64.
```
honza-kasik: Is there a reason why it is not simply ${CredentialStore:Name:Alias} instead of ${ENC:Resolver:ENCRYPTED_DATA}?

darranl: The point here is we need to be cross referencing the new "expression=encryption" resource, when we reference the credential store directly we are only referencing a specific entry within that credential store but we still need to configure the Cipher used for the decryption.

Additionally within the credential store multiple credentials can be associated with a single alias so although today we are just looking to use a SecretKey in the future we may choose to use public / private keys etc..

elytron/WFCORE-4360-CredentialStore_Expression_Resolution.adoc
```diff
@@ -0,0 +1,250 @@
...
+
+The tool will not add any support for the decryption of expressions or their values.
+
+==== Base64 Representations
+
+Base64 representations will be used for both the export and import of the generated keys and for the expression value to be passed to the expression resolver - generally these details should be opaque to the end user however we may include the description within the documentation allowing for others to make use of the format.  We should consider that in the future alternative representations may be required either to support alternative key types or to support alternative representations so the representations will also be versioned.
+
+The general structure of the first representation will be: -
+
+    { 'E', 'L', 'Y', (VERSION), (TYPE), VALUE... }
```
honza-kasik: I have never seen this format (structure) before. Can you please explain to me (and maybe add a note too to the document) what it represents? An example would be best.

darranl: I am not sure what else to add for this, really it is a Java / pseudocode representation of the population of an array.

elytron/WFCORE-4360-CredentialStore_Expression_Resolution.adoc
```diff
@@ -0,0 +1,250 @@
...
+
+We will not support the decryption of expressions other than the support within the management model to resolve existing expressions.
+
+As expressions are already widely in use with the application server there is a small possibility that users may have already defined expressions with a prefix of `ENC:` expecting it to be resolved as a system property.  The `expression=encryption` resource will also have a `prefix` attribute which will allow an alternative prefix to `ENC` to be specified. 
+
+==== Command Line Encryption
+
+Whilst the Elytron tool does contain a `credential-store` command the encryption of values is really a special operation so a new command `encryption` will be added to the tool.
+
+This command will duplicate the parameters used by the `credential-store` command to configure the location and initialisation of the underlying credential store, additionally it will take an argument to specify the alias from the credential store to load the `SecretKey` to use for encryption.
```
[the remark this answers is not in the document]
darranl: Not the actions, just the parameters needed to load the actual store.

elytron/WFCORE-4360-CredentialStore_Expression_Resolution.adoc
honza-kasik: So this represents and array - that's what wasn't clear to me. Thanks!

darranl: I will change the wording above to say this is an array.

elytron/WFCORE-4360-CredentialStore_Expression_Resolution.adoc
darranl: In this case I do mean credential not credential store.

The expression=encryption resource contains one or more resolver definitions, each resolver definition will be associated with a single credential.  The purpose of this line is to clarify that we will not support automatically switching to a different credential i.e. decrypting using the original credential and encrypting using a new credential.

elytron/WFCORE-4360-CredentialStore_Expression_Resolution.adoc
```diff
@@ -0,0 +1,250 @@
...
+
+==== Credential Store
+
+The enhancement will exclusively make use of `javax.crypt.SecretKey` instances generated from a `javax.crypto.SecretKeyFactory`, these will be stored in a `CredentialStore`.
+
+For the purpose of this enhancement each of the secret key types specified in <<https://docs.oracle.com/en/java/javase/11/docs/specs/security/standard-names.html#secretkeyfactory-algorithms, Java Security Standard Algorithm Names>> for Java 11 will be supported, however a user will be restricted to using the subset supported by their own JVM / installed `java.security.Provider` instances.
+
+It will be possible to generate, export, and import the SecretKey instances using both the command line tool and management operations.
+
+The Elytron tool `credential-store` command already has an `--add` action, rather than overloading this action for different types of key and different operations which will become complex very quickly a new set of actions will be added: -
```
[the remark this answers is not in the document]
darranl: I may be able to add a link to the adoc within the WildFly repo once I merge an outstanding PR but the doc itself will not be published until WildFly 18 is released so won't have access to a true doc link,

elytron/WFCORE-4360-CredentialStore_Expression_Resolution.adoc
```diff
@@ -0,0 +1,250 @@
...
+
+ * `resolver` - The name of the defined resolver to use to encrypt the data.
+ * `data` - The clear text value to be encrypted.
+
+The return value of the operation will be the complete expression ready to be used elsewhere in the mode.
+
+The format of the expressions resolved using the WildFly Elytron expression resolver will be `${ENC:Resolver:ENCRYPTED_DATA}` where `Resolver` is a reference to the specific resolver defined in the `expression=encryption` resource and `ENCRYPTED_DATA` is the data to be decrypted encoded using Base64.
+
+We may also consider supporting a default resolver, in that case the `expression=encryption` resource would have an additional attribute `default-resolver` to specify which resolver is the default, the expression can then be simplified to `${ENC:ENCRYPTED_DATA}`.  In this case the `resolver` parameter on the `create-expression` resource could also become optional allowing the default to be used.
```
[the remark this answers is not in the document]
darranl: Just reworking this section, we may as well add support for the default - it does not change the implementation that much.

elytron/WFCORE-4360-CredentialStore_Expression_Resolution.adoc
honza-kasik: Hi @darranl, I'm looking to the RFE description and all linked issues. This RFE is about replacing vaults usage by CS. This is requested by users. AFAIK, vaults has this format: `${VAULT::VAULT_BLOCK_name::attribute_name::1}`. Originally, users have been used to use attribute name for vaults, not some base64 code. Readable attribute name is more user-friendly than some base-64 unreadable text.

> The point here is we need to be cross referencing the new "expression=encryption" resource, when we reference the credential store directly we are only referencing a specific entry within that credential store but we still need to configure the Cipher used for the decryption.

It sounds reasonable. So WDYT about this? Would this or similar approach be feasible in this situation? If not, why?
```
/subsystem=elytron/secret-key=custom-secret-key:add(generate-new=true)
/subsystem=elytron/credential-store=STORE_NAME:add(modifiable=true, implementation-properties={"cipher-type"="AES", cipher-parameters=[], "secret-key"=custom-secret-key}, create=true)
/subsystem=elytron/credential-store=STORE_NAME:add-alias(alias="Alias", secret-value="SENSITIVE_STRING")
"${CredentialStore:STORE_NAME:Alias}"
```

> Additionally within the credential store multiple credentials can be associated with a single alias so although today we are just looking to use a SecretKey in the future we may choose to use public / private keys etc..

These can be implemented by additional parameters to `/subsystem=elytron/secret-key=custom-secret-key:add(generate-new=true)` or can't it (why?)?

Thank you!

elytron/WFCORE-4360-CredentialStore_Expression_Resolution.adoc
```diff
@@ -0,0 +1,248 @@
...
+
+The `--export-secret-key` action will return a single Base64 value which is the key in it's encoded form.
+
+The `--import-secret-key` action will take the alias to store the key as an argument, it will also take the following additional option: -
+
+ * `--key` - The key to import encoded using Base64.
+
+If the `--key` option is not specified the user will be prompted to enter the key interactively, this will avoid the key being seen in the current running processes and avoid the key being cached in the users command line history.
+
+NOTE: For `generate-secret-key` and `import-secret-key` if an entry already exists for that alias / algorithm combination the behaviour will be the same as the `add` action in relation to replacing the entry or reporting an error.  (To be checked)
```
honza-kasik: @darranl Should this note still be here if we are going to support just the AES?

elytron/WFCORE-4360-CredentialStore_Expression_Resolution.adoc
honza-kasik: @darranl Do you already know which approach will be used? IMHO this analysis should contain final decision.

elytron/WFCORE-4360-CredentialStore_Expression_Resolution.adoc
darranl: Within this RFE we are not attempting to provide a direct 1:1 mapping with the Vault, the vault was used to handle the storage of the encrypted Strings that is not the change that is being made here.

One issue that we do have with RFEs is they are often expressed in terms of previously understood features and not in terms of the actual requirement to be solved.

Within this RFE the issue that we are specifically addressing is how to allow encryption of other attributes within the management model that are not credentials and the chosen approach is to support expressions containing inlined encrypted values which can be decrypted at runtime.

The Base64 part of this description is to specifically hide those details from an end user, those are an internal implementation detail.  The reason I am adding some additional internal details to those values is to give us options in the future if we want to change the representations / add additional functionality - in a future RFE we can add version 2 and make changes as needed.

The expression the end user will use will still be in a human readable format: -

  ${ENC:Resolver:ENCRYPTED_DATA}

Comparing this to the Vault solution the string 'ENC' is the equivalent of 'VAULT' from the prior expression. The value 'Resolver' allows the correct resolver to be identified and then the 'ENCRYPTED_DATA' is binary data so this needs to represented as a String so for this choosing to use Base64 but could equally use Hex.  So it is within this ENCRYPTED_DATA field only that some support for versioning is being added to add some future protection.

Regarding the 'expression=encryption' resource, one further option that could be considered is to make this a child resource of the credential-store resource.  Generally I am adamantly against child resources within the management model as they are often used to express a relationship where the parent depends on the child which is incorrect, however in this case the child would depend on the parent which would be correct.

If we did that however we have made extensive use of the capabilities and requirements APIs of the management tier of the server - this means other subsystems can advertise these capabilities and use them either with the elytron subsystem or as an alternative.  If we make the expression-encryption resource a child of the credential-store resource it will not be possible to use this with credential stores advertised by other subsystems, if however we make it an independent resource it can make use of credential stores advertised from elsewhere.

In both cases however we could also avoid using a singleton and instead use a named resource e.g.

expression-resolver=PREFIX

The expressions used in the management model could now be: -

  ${PREFIX:ENCRYPTED_DATA}

So now the PREFIX in the expression is used to cross reference the correct expressions-resolver.

Adding this level of configuration however does have the potential to make things a lot more complex, this RFE is bridging the gap between the core management of the server and subsystem provided capabilities - using a singleton drastically simplifies this.  If we switch to have multiple resources a lot more internal coordination of those resources becomes required.

elytron/WFCORE-4360-CredentialStore_Expression_Resolution.adoc
```diff
@@ -0,0 +1,248 @@
...
+
+The `expression=encryption` resource will contain three attributes, the first attribute being `resolvers` which will be used to define one or more resolvers to handle the encryption and decryption of values.  Each resolver will support the following configuration: -
+
+ * Reference to a credential store and alias to identify a specific previously created `SecretKey` to use.
+ * Name of the algorithm to use for the encryption / decryption.
+ * Any additional parameters such as `salt`, `iteration count` etc.. as required to configure the algorithm.
+
+The second attribute on the `expression=encryption` resource will be `default-resolver` which will be used to optionally specify which resolver is the default, the expression can then be simplified to `${ENC:ENCRYPTED_DATA}`.
+
+As expressions are already widely in use with the application server there is a small possibility that users may have already defined expressions with a prefix of `ENC:` expecting it to be resolved as a system property.  The `expression=encryption` resource will also have a third attribute `prefix` attribute which will allow an alternative prefix to `ENC` to be specified.
```
honza-kasik: What if users will already have defined `${ENC}` variable in their configuration. Won't this cause a backward compatibility issues when the `${ENC}` variable will start causing errors? Customer must be aware of fact, that this might happen.

This also leads me to following questions:

1. Will be the `expression=encryption` resource defined by default?
1. What will happen if I set the `prefix` to another value with expressions already defined? Do I have to as a user manually update all my expression prefixes?

elytron/WFCORE-4360-CredentialStore_Expression_Resolution.adoc
```diff
@@ -0,0 +1,248 @@
...
+
+The format of the expressions resolved using the WildFly Elytron expression resolver will be `${ENC:Resolver:ENCRYPTED_DATA}` where `Resolver` is a reference to the specific resolver defined in the `expression=encryption` resource and `ENCRYPTED_DATA` is the data to be decrypted encoded using Base64.
+
+The `expression=encryption` resource will register it's expression resolver capability with the `CapabilityRegistry` making it available for runtime resolution of expressions.
+
+The `expression=encryption` resource will contain three attributes, the first attribute being `resolvers` which will be used to define one or more resolvers to handle the encryption and decryption of values.  Each resolver will support the following configuration: -
+
+ * Reference to a credential store and alias to identify a specific previously created `SecretKey` to use.
+ * Name of the algorithm to use for the encryption / decryption.
```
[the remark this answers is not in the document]
darranl: No the algorithm configuration can be relaxed so will not be needed, one point I need to confirm is if we want to make the size of the initialisation vector configurable so that may be one that gets added.

elytron/WFCORE-4360-CredentialStore_Expression_Resolution.adoc
darranl: The reason we will make the prefix configurable is so they can select a prefix not already in use, it is unlikely they will have something that starts 'ENC' but not impossible so if they are using this new expression resolution they will be able to override that.

The expression=encryption resource will not be included in the default configuration as we do not ship a default credential store so there will be no compatibility issue just by moving to a new version.

honza-kasik: Thank you for response! Though you probably forgot to respond to question number 2.

elytron/WFCORE-4360-CredentialStore_Expression_Resolution.adoc
darranl: Yes, I think this can remain - the credential store already supports the different permutations so that is not new behaviour being added by this RFE.

elytron/WFCORE-4360-CredentialStore_Expression_Resolution.adoc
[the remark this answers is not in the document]
darranl: Actually this will need to be adjusted, the attribute will be `transformation` not algorithm, I have added details for the command line tool that will be duplicated in this section but there will be a set of algorithm name / algorithm mode / padding mode permutations that will need to be specified as the transformation.

elytron/WFCORE-4360-CredentialStore_Expression_Resolution.adoc
darranl: Regarding '2', they MUST use a prefix that is not already in use in the model - if they specify a prefix already used in their expressions then their expression will be incorrectly picked up by the expression=encryption expression resolver.  In that situation it is safer to fail with an error.

We could do something where we silently log failures and continue allowing a fallback but those can become a pain to debug where we have allowed a single expression through multiple resolvers that it matches.

elytron/WFCORE-4360-CredentialStore_Expression_Resolution.adoc
honza-kasik: Hi @darranl! Is there any update for this situation?

darranl: @honza-kasik No there is no further update at the moment, I do not expect much activity before March.

bstansberry: @darranl The host controller process is not allowed to resolve expressions based on any configuration in the profiles in domain.xml. Those profiles are only usable at runtime by servers.

I don't believe we have any cases of host controller subsystems (i.e. configured in host.xml) being used as part of the server configuration, and if we don't we should not make this the first such case.

elytron/WFCORE-4360-CredentialStore_Expression_Resolution.adoc
```diff
@@ -0,0 +1,284 @@
...
+
+These representations are independent of the overall representation of the expressions as it is intended a portion of those expressions is both human readable and modifiable to allow alternative resolver definitions to be used without requiring the Base64 representation to be recreated. 
+
+==== Domain Mode
+
+The domain mode relationships will need to considered further, where we used the PicketBox vault we defined the vault within the host.xml and used it's values in the domain.xml.  As the credential store can both be defined in the host.xml and the domain.xml it would make sense to make use of the one stored within that descriptor, however this could be impacted based on how the configurable expression resolver is defined.
+
+
```
bstansberry, line 222: @darranl This section should include discussion of how the various runtime-only management operations discussed earlier (i.e. manipulating the keystore) will be supported in domain mode; i.e. whether they will be available on /profile=foo/subsystem=elytron or whether the more typical behavior of only being available at /host=x/server=y/subsystem=elytron will be followed.

darranl:
Thank you for the additional comments @bstansberry - FYI we have had a discussion about making the scope of this smaller so we are going to target standalone mode only - the comments you make highlight some of the additional complexities that would need to be addressed for domain mode due to the multiple processes in use at once.

darranl:
Superseded by https://github.com/wildfly/wildfly-proposals/pull/373

PR #215, opened 2019-06-24
[WFCORE-4150] Support automatically adding and updating credentials in a credential store

fjuma:
https://issues.jboss.org/browse/WFCORE-4150

### DIFF
--- a/elytron/WFCORE-4150-credential-store-automatic-updates.adoc
+++ b/elytron/WFCORE-4150-credential-store-automatic-updates.adoc
@@ -1,0 +1,125 @@
+= WFCORE-4150 Support automatically adding and updating credentials in a credential store
+:author:            Farah Juma
+:email:             fjuma@redhat.com
+:toc:               left
+:icons:             font
+:idprefix:
+:idseparator:       -
+
+== Overview
+
+Many subsystems currently support a ```credential-reference``` which is used to specify either a reference to
+a credential that is stored in an Elytron credential store under a given alias or a clear text password. In
+particular, a ```credential-reference``` supports specifying either of the following two mutually exclusive
+things:
+
+1. ```store``` and ```alias``` attributes which specify the Elytron credential store and the alias within the
+store that hold a credential
+2. ```clear-text``` attribute which specifies a clear text password
+
+For the first case, a user needs to first create a credential store and add a credential to it before being able
+to reference it from a ```credential-reference```.
+
+This enhancement looks at relaxing the mutual exclusivity requirement to automate the process of adding or
+updating a credential in an existing credential store to simplify the number of steps that a user needs to
+perform.
+
+== Issue Metadata
+
+=== Issue
+
+* https://issues.jboss.org/browse/WFCORE-4150[WFCORE-4150]
+* https://issues.jboss.org/browse/WFLY-12218[WFLY-12218]
+
+=== Related Issues
+
+* https://issues.jboss.org/browse/EAP7-676[EAP7-676]
+
+=== Dev Contacts
+
+* mailto:{email}[{author}]
+
+=== QE Contacts
+
+* mailto:tterem@redhat.com[Tomas Terem]
+
+=== Testing By
+
+[ ] Engineering
+[x] QE
+
+=== Affected Projects or Components
+
+* WildFly Core
+* WildFly
+
+=== Other Interested Projects
+
+== Requirements
+
+=== Hard Requirements
+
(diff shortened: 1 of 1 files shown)


elytron/WFCORE-4150-credential-store-automatic-updates.adoc
```diff
@@ -0,0 +1,102 @@
...
+=== Other Interested Projects
+
+== Requirements
+
+=== Hard Requirements
+
+It should be possible for a ```credential-reference``` to support both a credential store (i.e., the ```store```
+attribute) and a clear text password (i.e., the ```clear-text``` attribute) to be specified at the same time.
+
+* If the ```alias``` attribute is also specified, one of the following will then occur:
```
honza-kasik: Will be users notified which of these two situations occurred? I can imagine a situation when they want to edit existing alias and they end up with new without them knowing. Or they would want to create new alias and they end up "rewriting" existing one.

elytron/WFCORE-4150-credential-store-automatic-updates.adoc
```diff
@@ -0,0 +1,102 @@
...
+
+* https://issues.jboss.org/browse/EAP7-676[EAP7-676]
+
+=== Dev Contacts
+
+* mailto:{email}[{author}]
+
+=== QE Contacts
+
+TBD
```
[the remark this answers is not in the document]
fjuma: Added

elytron/WFCORE-4150-credential-store-automatic-updates.adoc
fjuma: We could indicate this in the output of the management operation when adding / updating a ```credential-reference```. I've updated the "Hard Requirements" section to mention this.

honza-kasik: Thanks!

honza-kasik:
@fjuma Do I understand correctly, that Web Console changes are part of this enhancement? From description and components on product JIRA it seems so...

fjuma:
@honza-kasik Correct, this RFE involves management model changes so HAL will need corresponding updates. Thanks for adding the "affects-model" label to the JIRA issue.

fjuma:
@honza-kasik Sorry for the delayed response. Currently, for the corner case you mentioned, the result of the operation would indicate the credential store update that was made and then indicate the reason the operation failed. However, I can look at updating my implementation to ensure that any credential store updates get rolled back when an operation fails.

fjuma:
I've now updated this to specify that if an operation that includes a ```credential-reference``` parameter fails for any reason, then the credential store update will get rolled back.

elytron/WFCORE-4150-credential-store-automatic-updates.adoc
```diff
@@ -0,0 +1,125 @@
...
+
+* https://issues.jboss.org/browse/EAP7-676[EAP7-676]
+
+=== Dev Contacts
+
+* mailto:{email}[{author}]
+
+=== QE Contacts
+
+* mailto:jkasik@redhat.com[Jan Kasik]
```
tterem: I am new QE contact for this RFE:

* mailto:tterem@redhat.com[Tomas Terem]

elytron/WFCORE-4150-credential-store-automatic-updates.adoc
```diff
@@ -0,0 +1,125 @@
...
+
+In each of the above cases where adding or updating a ```credential-reference``` results in a new entry being added
+to the credential store or an existing entry being updated in the credential store, the output of the management
+operation will indicate this.
+
+==== Rollback upon operation failure
+
+If an operation that includes a ```credential-reference``` parameter fails for any reason, no automatic
+credential store update will take place, i.e., any credential store that was specified via the
+```credential-reference``` attribute will contain the same contents as it did before the operation was
```
tterem: Right now the document renders in a weird way - this line is missing and rest of the document is all in a grey code block.
It can be fixed by putting credential-reference into just one apostrophe instead of three.

elytron/WFCORE-4150-credential-store-automatic-updates.adoc
fjuma: Added.

elytron/WFCORE-4150-credential-store-automatic-updates.adoc
fjuma: Fixed, thanks!

PR #217, opened 2019-06-28
HAL-1600: Expose ability to configure the JSON encoded access logs

hpehl:
https://issues.jboss.org/browse/HAL-1600

### DIFF
--- a/console/HAL-1600_Configure_JSON_encoded_access_log.adoc
+++ b/console/HAL-1600_Configure_JSON_encoded_access_log.adoc
@@ -1,0 +1,93 @@
+= Expose ability to configure the JSON encoded access logs
+:author:            Harald Pehl
+:email:             hpehl@redhat.com
+:toc:               left
+:icons:             font
+:idprefix:
+:idseparator:       -
+:issue-base-url:    https://issues.jboss.org/browse
+
+== Overview
+
+Provide a way to configure the recently added resource `/subsystem=undertow/server=foo/host=bar/setting=console-access-log`.
+
+Currently undertow hosts are configured as part of the undertow server application. To open the application, the user has to select an undertow server and click on "View":
+
+Congiguration &rarr; Subsystem = Web &rarr; Settings = Server &rarr; Server = default-server
+
+In the server application screen, the user then selects "Hosts" in the left hand side navigation and finally selects the host in the table at the top. Host configuration follows the typical master-detail pattern with the attributes of the selected host showing up at the bottom (seperated across several tabs). This gives limited space for host configuration. With the addition of the `console-access-log` resource this becomes an issue.
+
+Therefore this RFE is also about an UI refactoring of the host configuration. Hosts will be moved to its own finder column. When selecting an undertow server in the finder, a new column will be appended which manages the hosts of this server. Adding hosts will be done using column actions. To configure the host, the user will select the host and press "View".
+
+The new host application screen will have its own left hand side navigation:
+
+- Attributes
+- Access Log
+- Console Access Log
+- HTTP Invoker
+- Single Sign On
+- Filters
+- Local Cache
+
+The 'Console Access Log' item will include the configuration for the `console-access-log` resource. It will have two tabs:
+
+1. Attributes: Manages the follwing attributes:
+    - `include-host-name`
+    - `metadata`
+    - `predicate`
+    - `worker`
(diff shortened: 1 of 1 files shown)


console/HAL-1600_Configure_JSON_encoded_access_log.adoc
```diff
@@ -0,0 +1,93 @@
...
+1. Attributes: Manages the follwing attributes:
+    - `include-host-name`
+    - `metadata`
+    - `predicate`
+    - `worker`
+2. Directives: Manages the attribute `attributes`
+
+The `attributes` attribute is a complex attribute containing nested objects with similar structure. It controls the output of the JSON log. Most of the nested objects only contain one attribute: `key`. This is the name / key in the JSON output. This attribute usualy has a default value and can be customized. Some nested objects have additional attributes to give more control about the JSON output.
+
+The UI for the directives will be a dynamic form where the user can add and remove directives. The form will follow the normal view-edit-save pattern used all across the console. In view mode the form will show the configured directives and their values. If the user switches to edit mode, he can modify existing directives, add new directives and remove existing directives.
```
MMarus: I think it is a good idea to move the hosts to its own finder column and have the 'Console Access Log' configuration like this in the 2 tabs. There might be some confusion between the 'Attributes' tab and 'attributes' attribute, however it would be reasonable to leave the 'Attributes' tab as @hpehl suggested since it is used this way in the other configuration pages through the console.

hpehl: I also thought about a better naming. But since the term "Attributes" is used all across the console for the main attributes of a resource, we need another term for the second tab. We could use "Keys" as an alternative. 

WDYT?

MMarus: Yes that sounds good.

bobjohnsrh: Since "key" is used in the code for the label of the attributes, I would recommend not to use that term as the label  for the interface.  How about "Log data"?

hpehl: I think "key" is a good choice, because it's also used in the management model. Users coming from the CLI will recognize it and know what it means.

tommaso-borgato:
@hpehl adding my comments here:

- this RFE is only about Web access logs, correct? a similar RFE will be created for EJB access logs EAP7-523, correct?
- is resource '`/subsystem=undertow/server=foo/host=bar/setting=console-access-log`' already configurable though cli, correct? where can I find examples?
- I could not find info about the test plan, how are test added usually?
- how a log aggregator like Prometheus can consume this logs? In my experience they are exposed at HTTP endpoint

hpehl:
> this RFE is only about Web access logs?

Right. For EAP7-523 the related console RFE is [HAL-1589](https://issues.jboss.org/browse/HAL-1589). This RFE is expected to be implemented for WildFly 19.

> is resource '/subsystem=undertow/server=foo/host=bar/setting=console-access-log' already configurable?

Yes it is already available in WildFly Full 18.0.0.Beta1-SNAPSHOT. You can partly configure it already in HAL and configure it completely using the CLI.

> I could not find info about the test plan, how are test added usually?

I plan to add tests to https://github.com/hal/testsuite.next.

> how a log aggregator like Prometheus can consume this logs?

The log is written to stdout (see also the analysis document for [WFLY-11031](https://github.com/wildfly/wildfly-proposals/blob/master/undertow/WFLY-11031_Formatted_Access_Logs.adoc#non-requirements))

tommaso-borgato:
@hpehl is it possible to send these JSON formatted access logs to logstash or prometheus using a custom-handler or a socket-handler?

console/HAL-1600_Configure_JSON_encoded_access_log.adoc
```diff
@@ -0,0 +1,132 @@
...
+* {issue-base-url}/EAP7-1183[EAP7-1183]
+* {issue-base-url}/WFLY-11031[WFLY-11031]
+
+=== Dev Contacts
+
+* mailto:hpehl@redhat.com[Harald Pehl]
+
+=== QE Contacts
+
+* mailto:jstourac@redhat.com[Jan Stourac]
```
simkam: please change QE contact to msimka@redhat.com[Martin Simka]

PR #225, opened 2019-07-05
[WFLY-5286]: Provide CLI operation which returns used journal type -NIOxASYNCIO.

ehsavoie:
* Expose the effectively used journal type

Jira: https://issues.jboss.org/browse/WFLY-5286

### DIFF
--- a/messaging/WFLY-5286_add_runtime_journal_type_attribute.adoc
+++ b/messaging/WFLY-5286_add_runtime_journal_type_attribute.adoc
@@ -1,0 +1,81 @@
+= Provide CLI operation which returns used journal type -NIOxASYNCIO.
+:author:            Emmanuel hugonnet
+:email:             ehugonne@redhat.com
+:toc:               left
+:icons:             font
+:idprefix:          messaging,jms
+:idseparator:       -
+
+== Overview
+
+When configuring the use of ASYNCIO for journal persistence, the Artemis server may fall back to NIO if the libaio binaries are not located. This state is not displayed to the user so he doesn't know the effective journal type that is used.
+Adding a new `runtime-journal-type` attribute to read the effective journal type.
+
+== Issue Metadata
+
+=== Issue
+
+* https://issues.jboss.org/browse/WFLY-5286[WFLY-5286]
+
+=== Related Issues
+
+* https://issues.jboss.org/browse/EAP7-850[EAP7-850]
+
+=== Dev Contacts
+
+* mailto:{email}[{author}]
+
+=== QE Contacts
+
+* mailto:tterem@redhat.com[Tomas Terem]
+
+=== Testing By
+// Put an x in the relevant field to indicate if testing will be done by Engineering or QE.
+// Discuss with QE during the Kickoff state to decide this
+[X] Engineering
+
+[ ] QE
+
+=== Affected Projects or Components
+
+* Wildfly
+* ActiveMQ Artemis (finished)
+
+=== Other Interested Projects
+
+== Requirements
+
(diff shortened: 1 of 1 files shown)


messaging/WFLY-5286_add_runtime_journal_type_attribute.adoc
```diff
@@ -0,0 +1,71 @@
+= <INSERT TITLE HERE>
```
tterem: The title is missing.

messaging/WFLY-5286_add_runtime_journal_type_attribute.adoc
```diff
@@ -0,0 +1,71 @@
...
+When configuring the use of AIO for journal persistence, the Artemis server may fall back to NIO if the libaio binaries are not located. This state is not displayed to the user so he doesn't know the effective journal type that is used.
+
+== Issue Metadata
+
+=== Issue
+
+* https://issues.jboss.org/browse/WFLY-5286
+
+=== Related Issues
+
```
tterem: Please add https://issues.jboss.org/browse/EAP7-850 here.

messaging/WFLY-5286_add_runtime_journal_type_attribute.adoc
```diff
@@ -0,0 +1,71 @@
...
+=== Dev Contacts
+
+* mailto:{email}[{author}]
+
+=== QE Contacts
+
```
tterem: I will be QE contact for this.

messaging/WFLY-5286_add_runtime_journal_type_attribute.adoc
```diff
@@ -0,0 +1,71 @@
...
+:email:             ehugonne@redhat.com
+:toc:               left
+:icons:             font
+:idprefix:          messaging,jms
+:idseparator:       -
+
+== Overview
+
+When configuring the use of AIO for journal persistence, the Artemis server may fall back to NIO if the libaio binaries are not located. This state is not displayed to the user so he doesn't know the effective journal type that is used.
+
```
tterem: I think new `runtime-journal-type` attribute should be also mentioned in an overview so it was clear that this is how the issue is going to be solved.

messaging/WFLY-5286_add_runtime_journal_type_attribute.adoc
ehsavoie: I'm not sure this is required as this is community documentation

tterem: We always mentioned related EAP7 issues here. For example:
https://github.com/wildfly/wildfly-proposals/blob/master/messaging/WFLY-10326_messaging_global_resource_usage.adoc
https://github.com/wildfly/wildfly-proposals/blob/master/messaging/WFLY-11829_artemis_clusters_with_http_load_balancer.adoc
We always mentioned related EAP7 issues here.

messaging/WFLY-5286_add_runtime_journal_type_attribute.adoc
```diff
@@ -0,0 +1,76 @@
...
+* Wildfly
+* ActiveMQ Artemis (finished)
+
+=== Other Interested Projects
+
+== Requirements
+
+* Add a new `runtime-journal-type` attribute to the https://wildscribe.github.io/WildFly/17.0/subsystem/messaging-activemq/server/[/subsystem=messaging-activemq/server resource] to obtain the effective journal type.
+The authorized values are *NIO*,*ASYNCIO* and *DATABASE*.
```
jmesnil: * just to be sure but this attribute is a `runtime read-only` attribute, right?
* What is the value if persistence is disabled for the Artemis server?

ehsavoie: yes you are right
We should add a NONE value, I thought it would not make sense to read this attribute without persistence but adding NONE is no big deal so I'll put it there

PR #226, opened 2019-07-05
WFLY-12271 Extend RxInvoker interfaces to return Response objects

ronsigal:
WFLY-12271 / EAP7-1313


darranl:
@ronsigal Both in the long and short term please only use the community issue in the file name and title of the proposal anyway as this is a community visible project - the EAP7 issue can then be listed in the related issues section of the analysis doc.

ronsigal:
Thanks, @darranl. Done.

PR #227, opened 2019-07-08
WFLY-10057 EJB subsystem configure max threads and core threads indep…

chengfang:
…endently

Proposal for RFE https://issues.jboss.org/browse/WFLY-10057

### DIFF
--- a/ejb/WFLY-10057-thread-pool.adoc
+++ b/ejb/WFLY-10057-thread-pool.adoc
@@ -1,0 +1,120 @@
+= EJB subsystem configure max threads and core threads independently
+:author:            Cheng Fang
+:email:             cfang@redhat.com
+:toc:               left
+:icons:             font
+:idprefix:
+:idseparator:       -
+
+== Overview
+
+== Issue Metadata
+
+=== Issue
+
+* https://issues.jboss.org/browse/WFLY-10057 (EJB subsystem configure max threads and core threads independently)
+
+=== Related Issues
+
+* https://issues.jboss.org/browse/EAP7-488 (Fine-grained control on ejb3 thread-pool and reuse idle threads in pool before creating new)
+
+
+=== Dev Contacts
+
+* mailto:{email}[{author}]
+
+=== QE Contacts
+
+* mailto:rjanik@redhat.com[Richard Janik]
+
+=== Testing By
+// Put an x in the relevant field to indicate if testing will be done by Engineering or QE. 
+// Discuss with QE during the Kickoff state to decide this
+[x] Engineering
+
+[x] QE
+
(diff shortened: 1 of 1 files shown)


ejb/WFLY-10057-thread-pool.adoc
```diff
@@ -0,0 +1,165 @@
...
+
+== Implementation Plan
+////
+Delete if not needed. The intent is if you have a complex feature which can 
+not be delivered all in one go to suggest the strategy. If your feature falls 
+into this category, please mention the Release Coordinators on the pull 
+request so they are aware.
+////
+
+* switch ejb3 subsystem thread-pool to the thread-pool backed by `org.jboss.threads.EnhancedQueueExecutor`
```
bstansberry: Will this be able to support all the metrics currently exposed by UnboundedQueueThreadPoolMetricsHandler?

```
[standalone@localhost:9990 /] /subsystem=ejb3/thread-pool=default:read-resource(include-runtime=true)
{
    "outcome" => "success",
    "result" => {
        "active-count" => 0,
        "completed-task-count" => 0L,
        "current-thread-count" => 0,
        "keepalive-time" => {
            "time" => 100L,
            "unit" => "MILLISECONDS"
        },
        "largest-thread-count" => 0,
        "max-threads" => 10,
        "name" => "default",
        "queue-size" => 0,
        "rejected-count" => 0,
        "task-count" => 0L,
        "thread-factory" => undefined
    }
}
```

chengfang: Yes, I've added a sample CLI output above (around line 121) that shows the same metrics.  It's taken after some custom config and test runs hence the non-default values.

ejb/WFLY-10057-thread-pool.adoc
```diff
@@ -0,0 +1,223 @@
...
+|New Requirement |Existing Behavior of EJB3 Subsystem thread-pool
+
+|Ability to configure core-threads and max-threads independently
+|max-threads is configurable, but core-threads is not configurable and always equals to max-threads
+
+|Ability to reuse available threads as much as possible, without unnecessary creation of new threads
+|Upon new request, new threads are created up to the limit of max-threads, even though idle threads are available
+
+|Ability to timeout idle non-core threads after keepalive-timeout
+|Idle threads are not timed out, and keepalive-timeout is ignored
```
mnovak1: @chengfang There is `100 ms` by default set for `keepalive-timeout` in current implementation. Do we have any recommendation to which value it should be set? I think that current default is very low as for example for Artemis thread pools there was keepalive timeout set to `60 seconds`.

PR #229, opened 2019-07-15
[WFLY-12228] Upgrade MicroProfile Health 2.0.1

jmesnil:
JIRA: https://issues.jboss.org/browse/WFLY-12228

### DIFF
--- a/microprofile/WFLY-12228_upgrade_microprofile-health_2.0.1.adoc
+++ b/microprofile/WFLY-12228_upgrade_microprofile-health_2.0.1.adoc
@@ -30,7 +30,7 @@ MicroProfile 3.0 has been released with a new major version for MicroProfile Hea
 
 === QE Contacts
 
-* TBD
+* mailto:jstourac@redhat.com [Jan Stourac]
 
 === Testing By
 
--- a/microprofile/WFLY-12228_upgrade_microprofile-health_2.0.1.adoc
+++ b/microprofile/WFLY-12228_upgrade_microprofile-health_2.0.1.adoc
@@ -1,0 +1,82 @@
+= [WFLY-12228] Upgrade MicroProfile Health 2.0.1.
+:author:            Jeff Mesnil
+:email:             jmesnil@redhat.com
+:toc:               left
+:icons:             font
+:idprefix:
+:idseparator:       -
+:keywords:          observability,microprofile,health,openshift
+
+== Overview
+
+WildFly is providing MicroProfile Health 1.0 for applications deployed in WildFly.
+
+MicroProfile 3.0 has been released with a new major version for MicroProfile Health 2.0.1 that contains new features and breaking changes as described in its https://github.com/eclipse/microprofile-health/releases/tag/2.0.1[release notes].
+
+== Issue Metadata
+
+=== Issue
+
+* https://issues.jboss.org/browse/WFLY-12228[WFLY-12228]
+
+=== Related Issues
+
+* https://issues.jboss.org/browse/EAP7-1206[EAP7-1206]
+* https://issues.jboss.org/browse/CLOUD-3326[CLOUD-3326] - Integrate MicroProfile Health 2.0.1 HTTP endpoints in EAP probes
+
+=== Dev Contacts
+
+* mailto:{email}[{author}]
+
+=== QE Contacts
+
+* TBD
+
+=== Testing By
+
+* [x] QE
+
+=== Affected Projects or Components
+
+The implementation of Eclipse MicroProfile Health is provided by https://github.com/smallrye/smallrye-health[smallrye-health].
+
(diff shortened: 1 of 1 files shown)


jmesnil:
implemented in WilFly in https://github.com/wildfly/wildfly/pull/12392

microprofile/WFLY-12228_upgrade_microprofile-health_2.0.1.adoc
```diff
@@ -0,0 +1,72 @@
...
+The microprofile-health-smallrye will provide its own "reporter" that aggreated healtch checks procecures when applications are deployed/undeployed.
+
+
+== Requirements
+
+* Support Eclipse MicroProfile Health 2.0.1 for applications deployed in WildFly.
+* Provide two additional HTTP endpoints `/health/live` and `/health/ready` to check exclusilvely liveness checks probes and readiness check probes.
+* Repurpose the existing HTTP endpoint `/health` to check *all* check probes (both liveness and readiness probes)
+* Provides a management attribute to specify the global status when no probes are defined. This allows applications to report `DOWN` until their probes verify that the application is ready/live. By default, this attribute will report `UP`.
+
```
bstansberry: Should there be some mention of wildfly/wildfly-s2i here?

The EAP images provide scripts that can be used for kubernetes liveness/readiness commands[1]. Those scripts invoke against the management endpoint and utilize this health data.

If WildFly s2i doesn't incorporate the equivalent, perhaps it should. That may be out of scope for this RFE though.

[1] https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-probes/#define-a-liveness-command

jmesnil: I've created https://issues.jboss.org/browse/CLOUD-3326 to track the integration of these liveness/readiness checks in EAP probes.
This can be further discussed in CLOUD-3326 but I agree that wildfly-s2i should integrate these probes.
However, I explicitly stated in the non-requirements that this is out of scope of this RFE.

Having said that, your comment made me realise that this RFE do not update the DMR operations that are used by the EAP probes. I've updated the dev analysis to stipulate that this RFE should also add `:check-live` and `:check-ready` operations suitable for use by the EAP probes.

jmesnil: https://github.com/wildfly/wildfly/pull/12392 has been update to provide management operations that can be used by the EAP liveness/readiness probes

microprofile/WFLY-12228_upgrade_microprofile-health_2.0.1.adoc
```diff
@@ -0,0 +1,82 @@
...
+The microprofile-health-smallrye will provide its own "reporter" that aggreated healtch checks procecures when applications are deployed/undeployed.
+
+
+== Requirements
+
+* Support Eclipse MicroProfile Health 2.0.1 for applications deployed in WildFly.
+* Provide two additional HTTP endpoints `/health/live` and `/health/ready` to check exclusilvely liveness checks probes and readiness check probes.
+* Provide two additional management operations `:check-live` and `:check-ready` to check exclusilvely liveness checks probes and readiness check probes.
+* Repurpose the existing HTTP endpoint `/health` and `:check` operation to check *all* check probes (both liveness and readiness probes)
+* Provides two management attributes to specify the global status when no readiness or liveness probes are defined. This allows applications to report `DOWN` until their probes verify that the application is ready/live. By default, these attributes will report `UP`.
```
bstansberry: I think the Galleon layer that configures the default variant of the mp health subsystem needs to make it easy for cloud users to do this using env vars. Something like 

empty-readiness-checks-status="${env.SOME_ENV_VAR_NAME_TBD:UP}"

We tend to think in terms of system props for expressions in our standard configs, but for cloud use cases env vars are more useful.

jmesnil: good point, I'll update the feature PR to use an expression resolved by an env var at https://github.com/wildfly/wildfly/pull/12392/files#diff-44104d2318acaca0ae7b10645fb14056

microprofile/WFLY-12228_upgrade_microprofile-health_2.0.1.adoc
```diff
@@ -0,0 +1,82 @@
...
+* https://issues.jboss.org/browse/EAP7-1206[EAP7-1206]
+* https://issues.jboss.org/browse/CLOUD-3326[CLOUD-3326] - Integrate MicroProfile Health 2.0.1 HTTP endpoints in EAP probes
+
+=== Dev Contacts
+
+* mailto:{email}[{author}]
+
+=== QE Contacts
+
+* TBD
```
jstourac: feel free to put me in

PR #231, opened 2019-07-15
WFLY-11898 Add ability to group distributed session attributes to avoid cost of SESSION granularity

pferraro:
https://issues.jboss.org/browse/WFLY-11898


clustering/web/WFLY-11898_Session_Attribute_Groups.adoc
```diff
@@ -0,0 +1,74 @@
...
+* Add @Group(name="...") annotation to public org.wildfly.clustering.web.api module that to be used by users to annotate session attribute classes.
+* Enhance logic within FineSessionAttributes (for both hotrod and infinispan session manager implementations) to transparently group attributes sharing the same group name.
+
+=== Nice-to-Have Requirements
+
+=== Non-Requirements
+
+== Test Plan
+
+* Add new failover test that validates that 2 session attributes sharing the same @Group name retain referential integrity following failover
```
tommaso-borgato: @pferraro you mean in the WildFly test-suite or in the QE Clustering test-suite? or both?

clustering/web/WFLY-11898_Session_Attribute_Groups.adoc
```diff
@@ -0,0 +1,74 @@
...
+:idseparator:       -
+
+== Overview
+
+Distributed web sessions support 2 levels of granularity, by which session attributes are mapped to a distributed/remote cache: SESSION and ATTRIBUTE.
+SESSION granularity maps all attributes to a single cache entry, whereas ATTRIBUTE granularity maps each session attribute to a separate cache entry.
+For most applications, ATTRIBUTE granularity is far more efficient than SESSION granularity, since the former only requires the persistence/replication of those attributes that were added/modified/removed within a given request.
+However, if any attributes share object references (e.g. parent/child relationships, foreign keys, etc.) they must use SESSION granularity to ensure that these shared object references are preserved following serialization/deserialization.
+
+This feature request adds a mechanism for users to annotate session attributes into groups, such that users can leverage the efficiency for ATTRIBUTE granularity, while allowing specific attributes to be grouped together into a single cache entry.
```
tommaso-borgato: @pferraro what is the behavior if granularity is set to SESSION and some classes are annotated? annotations get ignored?

clustering/web/WFLY-11898_Session_Attribute_Groups.adoc
```diff
@@ -0,0 +1,74 @@
...
+
+Clustering
+
+=== Other Interested Projects
+
+== Requirements
+
+=== Hard Requirements
+
+* Add @Group(name="...") annotation to public org.wildfly.clustering.web.api module that to be used by users to annotate session attribute classes.
```
tommaso-borgato: @pferraro it is a class level annotation, correct?

clustering/web/WFLY-11898_Session_Attribute_Groups.adoc
tommaso-borgato: @pferraro should it be tested with web cache and ejb cache separately?

pferraro: WF testsuite.

clustering/web/WFLY-11898_Session_Attribute_Groups.adoc
pferraro: When using SESSION granularity, *all* session attributes shared the same group - so this annotation is technically still respected, but is generally redundant.

clustering/web/WFLY-11898_Session_Attribute_Groups.adoc
pferraro: Correct.

clustering/web/WFLY-11898_Session_Attribute_Groups.adoc
pferraro: This feature only affects distributed web sessions.  Also it only affects top level classes, i.e. the class of the object stored as a session attribute.

clustering/web/WFLY-11898_Session_Attribute_Groups.adoc
tommaso-borgato: @pferraro one last clarification: the annotation is e.g. `@Group(name="MY_GROUP_ONE")` hence, when forcing SESSION granularity, the content of attribute name gets ignored, correct?

pferraro: Correct. You can think of session granularity as a generalization of attribute granularity, where all session attributes belong to the same logical group.

tommaso-borgato:
@pferraro tnks for all clarifications; I would suggest to add to the AD some of your responses (to make everything clear also to the un-experienced):
- "This feature only affects distributed web sessions. Also it only affects top level classes, i.e. the class of the object stored as a session attribute"
- When forcing SESSION granularity, and the annotation is e.g. @Group(name="MY_GROUP_ONE") the content of attribute name gets ignored since all attributes belong to the same group

pferraro:
Thanks @tommaso-borgato .  I'll update the analysis document.

tommaso-borgato:
@pferraro a couple of questions:

* what is the final name of the annotation? I see `@SessionAttributeGroup` in the Jira and `@Group` here in the AD
* this feature just works for objects stored in the "web" cache and not for objects stored in the "ejb" cache (i.e. SFSB): is this correct?

pferraro:
@tommaso-borgato 

> what is the final name of the annotation? I see `@SessionAttributeGroup` in the Jira and `@Group` here in the AD

SInce this document was writte, I've concluded that it is better to group session attributes by name, rather than by type.  This grants finer control over the logical grouping of attributes.  I'll update the analysis document once I resume working on this feature.
The logical grouping of session attribute names will be defined in distributable-web.xml.

> this feature just works for objects stored in the "web" cache and not for objects stored in the "ejb" cache (i.e. SFSB): is this correct?

`@Stateful` EJBs caching is an orthogonal concern.  If a user stores a SFSB instance in the HttpSession (e.g. `@SessionScoped`), they are merely storing a reference to the SFSB instance.  SFSB instance caching is a separate concern, and an implementation detail of the EJB container.

> if you annotate a couple of JPA Entities e.g. User and Address with e.g. `@Group(name="COUPLED_ENTITIES")` (or `@SessionAttributeGroup("COUPLED_ENTITIES")` ?) and then store them as web session attributes then the two entities get replicated together, right?

See above.

> hibernate/jpa second level cache has nothing to do with this feature since it uses an invalidation-cache, since 2lc isn't bound to the web session and for other reasons too, correct?

Correct.  JPA-related caching is a separate concern, and an implementation detail of the EntityManager.

PR #233, opened 2019-07-18
[WFLY-12298] Allow configuring RESTEasy through Wildfly management model

ronsigal:


jaxrs/WFLY-12298_Change_RESTEASY_settings.adoc
```diff
@@ -0,0 +1,65 @@
...
+
+* {issue-base-url}/RESTEASY-1510 [RESTEASY-1510]
+
+=== Dev Contacts:
+
+* mailto:asoldano@redhat.com[Alessio Soldano]
+* mailto:ema@redhat.com[Jim Ma]
+* mailto:rsigal@redhat.com[Ron Sigal]
+* mailto:rsearls@redhat.com[Rebecca Searls]
+* mailto:weli@redhat.com[Wei Nan Li]
```
asoldano: @ronsigal , you should be either the first or second contact in the list here ;-)

jaxrs/WFLY-12298_Change_RESTEASY_settings.adoc
```diff
@@ -0,0 +1,65 @@
...
+
+=== QE Contacts:
+
+* mailto:rsvoboda@redhat.com[Rostislav Svoboda]
+* mailto:mkopecky@redhat.com[Marek Kopecky]
```
asoldano: @ronsigal , you should get in touch with QE and provide the actual references here. I'm writing you an email about this.

jaxrs/WFLY-12298_Change_RESTEASY_settings.adoc
```diff
@@ -0,0 +1,65 @@
+= WFLY-12298 Support configuring RESTEasy with CLI
```
darranl: Regarding terminology it would probably be better to loose the reference to the CLI, the CLI is just one of many potential management clients.

As I understand it this RFE is specifically about exposing RESTEasy configuration using the application servers management model.

It is true that the admin console may need it's own additional RFE for any console specifics but I don't think that makes this specifically about the CLI.

jaxrs/WFLY-12298_Change_RESTEASY_settings.adoc
ronsigal: Done.

jaxrs/WFLY-12298_Change_RESTEASY_settings.adoc
ronsigal: Done. Thanks, @darranl.

fmarchioni:
@ronsigal one thing I'd like to discuss is which quota of the related JIRA description will be implemented at this stage. First things coming up in my head:
Does it include changes in both /subsytem and /deployment model ?
Adding other providers, interceptors, etc., or disable some in global level would be allowed for all deployments (JIRA description) or only for future deployments (as I understand from the analysis docs) ?

jaxrs/WFLY-12298_Change_RESTEASY_settings.adoc
ronsigal: Done.

ronsigal:
Hi @fmarchioni, 

I have added some additional information which (I hope) answers your questions. 

1. Just /subsystem
2. Just future deployments.

-Ron

jaxrs/WFLY-12298_Change_RESTEASY_settings.adoc
```diff
@@ -0,0 +1,79 @@
+= WFLY-12298 Support configuring RESTEasy through Wildfly management model
+:author:            Ron Sigal
+:email:             rsigal@redhat.com
+:toc:               left
+:icons:             font
+:keywords:          comma,separated,tags
```
OndrejKotek: Can you set a real value please?

ronsigal:
Replaced by https://github.com/wildfly/wildfly-proposals/pull/245.

PR #237, opened 2019-08-01
[WFLY-12342] Add server probes for readiness health check

jmesnil:
Dev analysis for WFLY-12342

JIRA: https://issues.jboss.org/browse/WFLY-12342

### DIFF
--- a/microprofile/WFLY-12342_server_readiness_probes.adoc
+++ b/microprofile/WFLY-12342_server_readiness_probes.adoc
@@ -1,0 +1,99 @@
+= [WFLY-12342] Integrate server probes in MP Health readiness check
+:author:            Jeff Mesnil
+:email:             jmesnil@redhat.com
+:toc:               left
+:icons:             font
+:idprefix:
+:idseparator:       -
+:keywords:          observability,microprofile,health,openshift
+
+== Overview
+
+MicroProfile Health 2.0 introduced a readiness health check.
+WFLY-12228 added support for MicroProfile Health 2.0 without enhancing the readiness health check.
+
+This means that the readiness probe always return UP by default (as long as the microprofile-health-smallrye is installed).
+
+This RFE enhances it with 3 probes (that are part of the https://github.com/jboss-container-images/jboss-eap-modules/blob/251f422c97e1bcd3625f57295bc79973193482a6/os-eap-probes/2.0/added/probes/probe/eap/dmr.py#L23[EAP readiness probe script]):
+
+* `server-status` - returns UP when the server-state is `running`
+* `boot-errors` - return UP if there are no boot-errors
+* `deployment-status` - returns UP if status for all deployments is OK
+
+Once this RFE is done, we could simplify the EAP readiness probe script by replacing its various DMR checks by a single check to `/subsystem=microprofile-health-smallrye:check-ready` (or a HTTP call to `http://:9990/health/ready`).
+
+== Issue Metadata
+
+=== Issue
+
+* https://issues.jboss.org/browse/WFLY-12342[WFLY-12342]
+
+=== Related Issues
+
+* https://issues.jboss.org/browse/EAP7-1322[EAP7-1322]
+
+=== Dev Contacts
+
+* mailto:{email}[{author}]
+
+=== QE Contacts
+
+* TBD
+
+=== Testing By
+
+[X] QE
+
+=== Affected Projects or Components
+
+* This RFE impacts the `microprofile-health-smallrye` extension provided by WildFly
+* Once done, it can be used by Kubernetes/Openshift to configure their readiness probe
+
+== Requirements
+
+* Add 3 readiness health checks to MicroProfile readiness probe:
+** `server-status` - returns UP when the server-state is `running`
+** `boot-errors` - return UP when there are no boot-errors
+** `deployment-status` - returns UP when status for all deployments is OK
+
+== Non-Requirements
+
+* The MicroProfile Health subsystem is not supported in domain mode.
+
+== Implementation Plan
+
+* Update `org.wildfly.extension.microprofile.health.HealthReporterService` to add these 3 `HealthCheck` probes to the HealthReporter's `readinessChecks`.
+* These 3 checks will execute a local DMR operation to determine their status.
+
+A call to `/subsystem=microprofile-health-smallrye:check-ready` would aggregates results from these 3 probes:
+
+[source]
+----
+[standalone@localhost:9990 /] /subsystem=microprofile-health-smallrye:check-ready
+{
+    "outcome" => "success",
+    "result" => {
+        "status" => "UP",
+        "checks" => [
+            {
+                "name" => "server-state",
+                "status" => "UP",
+                "data" => {"value" => "running"}
+            },
+            {
+                "name" => "boot-errors",
+                "status" => "UP"
+            },
+            {
+                "name" => "deployments-status",
+                "status" => "UP",
+                "data" => {"helloworld-rs.war" => "OK"}
+            }
+        ]
+    }
+}
+----
+
+== Community Documentation
+
+The feature will be documented in WildFly Admin Guide (in its MicroProfile Health section) to specify that server readiness incorporates these 3 checks (in addition to any readiness check contributed by the deployments)


jmesnil:
Dev PR is at https://github.com/wildfly/wildfly/pull/12465

microprofile/WFLY-12342_server_readiness_probes.adoc
```diff
@@ -0,0 +1,95 @@
...
+:keywords:          observability,microprofile,health,openshift
+
+== Overview
+
+MicroProfile Health 2.0 introduced a readiness health check.
+WFLY-12228 added support for MicroProfile Health 2.0 without enhancing the readiness health check.
+
+This means that the readiness probe always return UP by default (as long as the microprofile-health-smallrye is installed).
+
+This RFE enhances it with 3 probes (that are part of the https://github.com/jboss-container-images/jboss-eap-modules/blob/251f422c97e1bcd3625f57295bc79973193482a6/os-eap-probes/2.0/added/probes/probe/eap/dmr.py#L23[EAP readiness probe script]):
```
mchoma, line 17: @jmesnil I assume implementation of probes is different. Or is there chance they reuse shared code?

jmesnil: The probes that check server states delegates to internal DMR operations (see https://github.com/wildfly/wildfly/pull/13397/files#diff-ca09758a51bffeb0751512740f73f2ccR27). So they rely on the same implementation code.

mchoma: Good. But still differences could be in environment properties driving the logic e.g. [1] and output format?

[1] https://github.com/jboss-container-images/jboss-eap-modules/blob/251f422c97e1bcd3625f57295bc79973193482a6/os-eap-probes/2.0/added/probes/probe/eap/dmr.py#L79

microprofile/WFLY-12342_server_readiness_probes.adoc
```diff
@@ -0,0 +1,95 @@
...
+
+* `server-status` - returns UP when the server-state is `running`
+* `boot-errors` - return UP if there are no boot-errors
+* `deployment-status` - returns UP if status for all deployments is OK
+
+Once this RFE is done, we could simplify the EAP readiness probe script by replacing its various DMR checks by a single check to `/subsystem=microprofile-health-smallrye:check-ready` (or a HTTP call to `http://:9990/health/ready`).
```
mchoma, line 23: I can't see `/subsystem=microprofile-health-smallrye:check-ready` in wf20

jmesnil: that's weird, I confirm it's already in WildFly 20.0.0.Final: https://docs.wildfly.org/20/wildscribe/subsystem/microprofile-health-smallrye/index.html#op-check-ready

```
[standalone@localhost:9990 /] /subsystem=microprofile-health-smallrye:check-ready
{
    "outcome" => "success",
    "result" => {
        "status" => "UP",
        "checks" => []
    }
}
```

mchoma: I confirm I can see as well.

I was using embeded server. Unless this is expected I will fill a jira for that

```
./jboss-cli.sh 
You are disconnected at the moment. Type 'connect' to connect to the server or 'help' for the list of supported commands.
[disconnected /] embed-server 
[standalone@embedded /] /subsystem=microprofile-health-smallrye:check-ready
{
    "outcome" => "failed",
    "failure-description" => "WFLYCTL0031: No operation named 'check-ready' exists at address [(\"subsystem\" => \"microprofile-health-smallrye\")]",
    "rolled-back" => true
}
```

microprofile/WFLY-12342_server_readiness_probes.adoc
```diff
@@ -0,0 +1,95 @@
...
+* TBD
+
+=== Testing By
+
+[X] QE
+
+=== Affected Projects or Components
+
+* This RFE impacts the `microprofile-health-smallrye` extension provided by WildFly
+* Once done, it can be used by Kubernetes/Openshift to configure their readiness probe
```
mchoma, line 50: @jmesnil What about liveness probe on OpenShift. For example I expect bootable jar image should configure readiness probe as http call to http://:9990/health/ready. But what should be the liveness probe in this case?

jmesnil: it uses the existing http call to `/health/live`: https://github.com/eclipse/jkube/blob/97571e56bb72c3e4d7f5f0aa979b872897eab336/jkube-kit/jkube-kit-wildfly-jar/src/main/java/org/eclipse/jkube/wildfly/jar/enricher/WildflyJARHealthCheckEnricher.java#L50

mchoma: Ok so you are saying in case when user is not using jkube he should configure liveness probe as `check-live`. Does `check-live` has same semantic as s2i liveness-probe.sh?

microprofile/WFLY-12342_server_readiness_probes.adoc
mchoma: Thinking about significance of output similarities of both ways. In case of Kubernetes it is just return code of call important [1] In case of shell probe it is 0 for success and in case of http call it is 200 for success.

So I assume as far as it is true that /health `UP` => 200 and `DOWN` => 50x and it maps to pyton probe logic `UP` => 0 and `DOWN` => 1, we should be safe.

Regarding env variable `PROBE_DISABLE_BOOT_ERRORS_CHECK`. Not sure if we want to copy this to /health implementation.  This is not documented to users and seems to me not consistent to have only one disable check from 3 checks. But If I am reading this correctly it is true by default in our images, so seems it has its meaning? [2]

[1] https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/#define-a-liveness-command
[2] https://github.com/jboss-container-images/jboss-eap-modules/blob/master/os-eap-probes/common/module.yaml

microprofile/WFLY-12342_server_readiness_probes.adoc
jmesnil: As part of this RFE, we are not changing the liveness probes.
This RFE focuses on the readiness probe and makes sure that we have a way to check the readiness of the application server directly from HTTP (without any external scripts).

mchoma: I understand. I am thinking how bootable jar will do without python liveness probes? And eventually how wildfly can live without that python liveness probes in future. 

I understand one motivation for this RFE is provide support for bootable jar on OpenShift. But to have fully support on OpenShift we need readiness probes and liveness probes as well. Were liveness probes already discussed somewhere before? Does current capabilities  (e.g. check-live) provide enough support for bootable jar on OpenShift?

jmesnil: I'm not sure to follow you.

WildFly Bootable jar is already using the http `/health/live` endpoint for the liveness probes . There is nothing else to do here. We did not need to do anything in WildFly as the server does not provide additional liveness health check procedures.

However we needed additional readiness health check procedures that are now exposed with the `/health/ready` http endpoint.

Just to clarify, the Bootable jar will rely on the *HTTP endpoints* (`/health/live` & `/health/ready`) and not on the DMR operations in the `microprofile-health-smallrye` subsystem.

mchoma: `Bootable jar will rely on the HTTP endpoints (/health/live & /health/ready) ` I was looking for that statement. Thank you.

microprofile/WFLY-12342_server_readiness_probes.adoc
mchoma, line 23: @jmesnil What about domain mode? Is it supported? Or this is just meant for standalone mode?

When playing locally, I don't see `microprofile-health-smallrye` subsystem by default.

jmesnil: MicroProfile Health is not supported in domain mode, I'll update the dev analysis to make that explicit

PR #243, opened 2019-08-26
[WFLY-11255] Managed Executor Runtime Stats

emmartins:
Issue: https://issues.jboss.org/browse/WFLY-11255


bstansberry:
@emmartins This seems fine, but please add a domain mode discussion. Typically this would be a note in the Non Requirements sections stating that the new attributes will not be registered with the /profile=*/subsystem=ee resources.

emmartins:
@bstansberry proposal updated with Non-Requirements note for domain mode.

PR #246, opened 2019-09-02
[WFLY-12473] Allow instance-id obfuscation in Undertow session route

fl4via:
WFLY Jira: https://issues.jboss.org/browse/WFLY-12473
EAP7 Jira: https://issues.jboss.org/browse/EAP7-1187
WildFly PR: https://github.com/wildfly/wildfly/pull/12883

### DIFF
--- a/undertow/WFLY-12473_obfuscate-session-route.adoc
+++ b/undertow/WFLY-12473_obfuscate-session-route.adoc
@@ -15,7 +15,7 @@ is handled by HTTP Invoker. This data is sent across connections as part of the 
 cookie.
 EAP6 Web Server would similarly allow obfuscation of instance-id at routing by allowing a generated
 UUID to be used as the instance-id.
-This UUID was the default value used whenever instance-id configuration was absent, while Undertow default value
+This UUID was the default value used whenever instance-id configuration was absent. But in EAP7, Undertow default value
 for instance-id is ${jboss.node.name}.
 Some users would like to have a way of obfuscating the instance-id value in the route.
 
@@ -36,12 +36,12 @@ Some users would like to have a way of obfuscating the instance-id value in the 
 
 === QE Contacts
 
-=== Testing By
-// Put an x in the relevant field to indicate if testing will be done by Engineering or QE. 
-// Discuss with QE during the Kickoff state to decide this
-[X] Engineering
+* mailto:tborgato@redhat.com[Tommaso Borgato]
 
-[ ] QE
+=== Testing By
+* [x] Engineering
+
+* [ ] QE
 
 === Affected Projects or Components
 
@@ -60,7 +60,12 @@ when used at the server route.
 
 === Hard Requirements
 
-The route must not change across server restarts when `obfuscate-session-route` value is `true`.
+The route must not change across server restarts when `obfuscate-session-route` value is `true`, unless there is a change in the values
+ the route is derived from:
+
+* the route equals `instance-id` attribute value, defined at `subsystem=undertow` management model address
+* `instance-id` default value is `"${jboss.node.name}"`, meaning that, when the attribute is kept unchanged, changes to `${jboss.node.name}`
+accross restarts will result in changes to the route
 
 === Nice-to-Have Requirements
 It is ideal to make it hard to find out the instance-id value via reverse engineering of the value used in route. For that, we are going
@@ -73,6 +78,8 @@ to use the server name as a salt when generating the obfuscated instance-id byte
 
 A few test cases will be submitted to verifying that the instance-id of Undertow subsystem is obfuscated on routing when `obfuscate-session-route` is `true`, and to
 verify that affected functionalities (namely, remote invocation via HTTP invoker) work as expected when session route is obfuscated.
+
+Also, we also need tests verifying that the route does not changes across server restarts, as this is a hard requirement.
 
 == Community Documentation
 
--- a/undertow/WFLY-12473_obfuscate-session-route.adoc
+++ b/undertow/WFLY-12473_obfuscate-session-route.adoc
@@ -79,10 +79,18 @@ to use the server name as a salt when generating the obfuscated instance-id byte
 
 == Test Plan
 
-A few test cases will be submitted to verifying that the instance-id of Undertow subsystem is obfuscated on routing when `obfuscate-session-route` is `true`, and to
-verify that affected functionalities (namely, remote invocation via HTTP invoker) work as expected when session route is obfuscated.
+A few test cases will be submitted to verify the requirements. We need to verify that the instance-id of Undertow subsystem is obfuscated on
+routing when `obfuscate-session-route` is `true`, and that the route does not change across server restarts.
 
-Also, we need tests verifying that the route does not changes across server restarts, as this is a hard requirement.
+Also, every functionality that relies on the session route must work as expected when the session route is obfuscated.
+In this regard, tests for the following features will be incremented to cover for the `obfuscate-session-route="true"` scenario:
+
+* EJB remote invocations via HTTP invoker
+* session management CLI operations
+* mod cluster load balancing
+
+The last item, mod cluster load balancing, is currently tested in QE mod cluster tests and is not part of the WildFly testsuite. Those tests
+will be updated in their current repository and will not added to WildFly testsuite.
 
 == Community Documentation
 
--- a/undertow/WFLY-12473_obfuscate-session-route.adoc
+++ b/undertow/WFLY-12473_obfuscate-session-route.adoc
@@ -1,0 +1,79 @@
+= WFLY-12473 Obfuscate session-route
+:author:            Flavia Rainone
+:email:             frainone@redhat.com
+:toc:               left
+:icons:             font
+:keywords:          undertow,instance-id,obfuscate-session-route
+:idprefix:
+:idseparator:       -
+
+== Overview
+
+This feature is to be able to obfuscate the value of UndertowService.instanceId when routing. Routes are used as
+part of generated session ids when a remote invocation (such as EJB, naming or serving a wildfly-services address)
+is handled by HTTP Invoker. This data is sent across connections as part of the request and response JSESSION_ID
+cookie.
+EAP6 Web Server would similarly allow obfuscation of instance-id at routing by allowing a generated
+UUID to be used as the instance-id.
+This UUID was the default value used whenever instance-id configuration was absent, while Undertow default value
+for instance-id is ${jboss.node.name}.
+Some users would like to have a way of obfuscating the instance-id value in the route.
+
+== Issue Metadata
+
+=== Issue
+
+* https://issues.jboss.org/browse/WFLY-12473[WFLY-12473]
+
+=== Related Issues
+
+* https://issues.jboss.org/browse/EAP7-1187
+* https://issues.jboss.org/browse/PRODMGT-1934
+
+=== Dev Contacts
+
+* mailto:{email}[{author}]
+
+=== QE Contacts
+
+=== Testing By
+// Put an x in the relevant field to indicate if testing will be done by Engineering or QE. 
+// Discuss with QE during the Kickoff state to decide this
+[X] Engineering
+
+[ ] QE
+
+=== Affected Projects or Components
+
+WildFly, Undertow component
+
+=== Other Interested Projects
+
+WildFly clustering
+
+== Requirements
+
+This will be implemented as an additional attribute under the `subsystem=undertow` management model address.
+
+This will be a simple boolean attribute called `obfuscate-session-route` that indicates that instance-id value must be obfuscated
+when used at the server route.
+
+=== Hard Requirements
+
+The route must not change across server restarts when `obfuscate-session-route` value is `true`.
+
+=== Nice-to-Have Requirements
+It is ideal to make it hard to find out the instance-id value via reverse engineering of the value used in route. For that, we are going
+to use the server name as a salt when generating the obfuscated instance-id bytes.
+
+
+=== Non-Requirements
+
+== Test Plan
+
+A few test cases will be submitted to verifying that the instance-id of Undertow subsystem is obfuscated on routing when `obfuscate-session-route` is `true`, and to
+verify that affected functionalities (namely, remote invocation via HTTP invoker) work as expected when session route is obfuscated.
+
+== Community Documentation
+
+Update the WildFly documentation to document the new attribute.
--- a/undertow/WFLY-12473_obfuscate-session-route.adoc
+++ b/undertow/WFLY-12473_obfuscate-session-route.adoc
@@ -66,6 +66,9 @@ The route must not change across server restarts when `obfuscate-session-route` 
 * the route equals `instance-id` attribute value, defined at `subsystem=undertow` management model address
 * `instance-id` default value is `"${jboss.node.name}"`, meaning that, when the attribute is kept unchanged, changes to `${jboss.node.name}`
 accross restarts will result in changes to the route
+* the server name: the server that is serving the request and, hence, providing the obfuscated route, uses its `name` attribute value as
+a salt for the hash that obfuscates the route (see Nice-to-Have Requirements below). So, if the `name` is changed across restarts, this will
+result in a change in the obfuscated route value
 
 === Nice-to-Have Requirements
 It is ideal to make it hard to find out the instance-id value via reverse engineering of the value used in route. For that, we are going
@@ -79,7 +82,7 @@ to use the server name as a salt when generating the obfuscated instance-id byte
 A few test cases will be submitted to verifying that the instance-id of Undertow subsystem is obfuscated on routing when `obfuscate-session-route` is `true`, and to
 verify that affected functionalities (namely, remote invocation via HTTP invoker) work as expected when session route is obfuscated.
 
-Also, we also need tests verifying that the route does not changes across server restarts, as this is a hard requirement.
+Also, we need tests verifying that the route does not changes across server restarts, as this is a hard requirement.
 
 == Community Documentation
 


darranl:
Just a minor comment - it would be useful for the commit message and PR title to contain a few words about the actual RFE - when only Jira issue numbers are used they need to be looked up to find the context,

pferraro:
Is it just me, or does this seem like a lot of work for something that could easily be specified on the command line?
e.g.
`` ./bin/standalone.sh -Djboss.node.name=`eval uuidgen` ``

pferraro:
What is the use case for UUIDs for the instance-id?  It seems to me that it is only really usable with mod_cluster, as other load balancers use static configuration, where worker IDs need to be enumerated.  That is not possible when the instance-id is auto-generated.

fl4via:
@pferraro  the reason behind this is that users that worked with JBoss EAP6 used the uuid-instance-id and wouldn't like to have the node name undisclosed as the Undertow instance-id

pferraro:
@fl4via For that small subset of users, what is wrong with https://github.com/wildfly/wildfly-proposals/pull/246#issuecomment-572677443 ?

rhusar:
👎 I am still not convinced this is a good idea.

* What is the actual problem this is trying to solve? I don't think the statement "customers would like to have it default to a UUID instead" is exactly correct, they are probably just looking for a feature which was default in the past, but no longer, for which we have better solutions now.
* UUIDs are impractical because there are large - remember this needs to be then sent and parsed with every request.
* When server restarts a new UUID will be assigned so even though the cache topology will be the same, the load balancer will not be able to route the session to the original node.
* The implementation itself is not well designed, setting `uuid-instance-id` and `instance-id` at the same time results in unclear configuration, much better way to implement this would be along the lines of attribute `routing="STATIC|UUID|HOSTNAME"` to allow future evolution of route handling.

pferraro:
> users that worked with JBoss EAP6 used the uuid-instance-id

Also, the EAP6 feature to which you are referring was intended as a developer convenience, so that one could use mod_cluster with no additional configuration.  This was never intended for use in production.

> and wouldn't like to have the node name undisclosed as the Undertow instance-id

I think this is actually refers to the node name defaulting to the host name, i.e. the user doesn't want to expose the host name to the public.  That is exactly why the node name is independently configurable.  Thus, the requirement of the user is already fulfilled.

fl4via:
@pferraro  and @rhusar  I'll forward this comments to Bilge in the internal EAP7 so we can discuss it with him. TBH, I don't see a problem with the way it is either, all I know is that this was requested.

undertow/WFLY-12473_UUID-instance-id.adoc
```diff
@@ -0,0 +1,65 @@
...
+
+WildFly, Undertow component
+
+=== Other Interested Projects
+
+== Requirements
+
+This will be implemented as an additional attribute under the `subsystem=undertow` management model address.
+
+This will be a simple boolean attribute called `uuid-instance-id` that marks instance-id to be set equal to a uuid.
```
tommaso-borgato, line 51: @fl4via `instance-id` is set to UUID only if it was not specified? does this implies that `uuid-instance-id` works just when `instance-id` is set to its default value `${jboss.node.name}`?

the PR `|uuid-instance-id | set this to true to indicate that a UUID should be used as the instance id. Can be used only when instance-id is undefined` suggest so, I just wanted a confirm because it's not explicitly stated here

undertow/WFLY-12473_UUID-instance-id.adoc
```diff
@@ -0,0 +1,65 @@
...
+:keywords:          instance-id,uuid,uuid-instance-id
+:idprefix:
+:idseparator:       -
+
+== Overview
+
+This feature is to be able to set the value of UndertowService.instanceId to a UUID,
+if instance-id was not defined and user desires to have it so.
+EAP6 Web Server would set it as a default whenever instance-id configuration was absent, and Undertow defaults to ${jboss.node.name} instead.
+Some customers would like to have it default to a UUID instead.
```
tommaso-borgato, line 15: @fl4via why adding the new attribute `uuid-instance-id` instead of setting `instance-id` to an expression like `${UUID}` instead of `${jboss.node.name}` ? I am asking because on WFLY-12473 I read `Since we can't change it for compatibility reasons a new attribute such as 'uuid-instance-id' can be added; if it was true then it would force the instance id to a UUID.` and I can't see why we can't set it explicitly instead of introducing a new attribute given the final effect is `instance-id` isn't `undefined` any more

tommaso-borgato:
> @pferraro and @rhusar I'll forward this comments to Bilge in the internal EAP7 so we can discuss it with him. TBH, I don't see a problem with the way it is either, all I know is that this was requested.

Any update on this?

undertow/WFLY-12473_UUID-instance-id.adoc
fl4via: @tommaso-borgato, both attributes cannot be defined at the same time, so you're right.

rhusar:
@tommaso-borgato That's a question for @fl4via .

fl4via:
@tommaso-borgato  the request remains open besides all considerations taken into account. We will proceed with this RFE. Implementation PR is updated.

fl4via:
this PR is final and ready for review

undertow/WFLY-12473_UUID-instance-id.adoc
tommaso-borgato: @fl4via it took me a while to extract the rule:

```if [[ instance-id == undefined AND uuid-instance-id == true ]] --> UndertowService.instance-id := UUID (--> default route := UUID) ```

* can you confirm that the rule is actually correct?
* given the final customer's goal is to set the default route to UUID (correct me if I am wrong), are we sure we want use such a name as `uuid-instance-id`? are we doing this in order to discourage utilization in customers other than one who asked for this?
* what don't we use something like `instance-id=UUID_UNIVERSALLY_UNIQUE_IDENTIFIER` (or some other string that is unlikely to be used otherwise) to trigger this behavior?

**Review by tommaso-borgato — CHANGES_REQUESTED** [review]

File: undertow/WFLY-12473_UUID-instance-id.adoc [hunk 3]
```diff
@@ -0,0 +1,65 @@
...
+=== Hard Requirements
+
+=== Nice-to-Have Requirements
+
+=== Non-Requirements
+
+== Test Plan
+
+A new test will be added to verify that the instance-id of Undertow subsystem is set to a UUID instead of ${jboss.node.name}.
+
```
tommaso-borgato, line 62: @fl4via I am posting this comment here, because I can't find the TP doc, even though it actually belongs to the "manual experiments" section of the TP doc (or to the "Integration testing", in case we have such test already);

In the linked case https://access.redhat.com/support/cases/#/case/02167248 it says that the customer's ultimate goal is `... to configure JBPSS EAP 7 to match the behavior of JBOSS EAP 6 so that the Nodes registered with Apache using an generated UUID rather than the jboss.node.name. If so, how is this done?` (btw, are they referring to mod_cluster/mod_jk used as load balancer?)

Hence I propose to have such a check there in order to be 100% to fulfill the customer's expectations;

tommaso-borgato (changes requested):

fl4via:
> -1 I am still not convinced this is a good idea.
> 
>     * What is the actual problem this is trying to solve? I don't think the statement "customers would like to have it default to a UUID instead" is exactly correct, they are probably just looking for a feature which was default in the past, but no longer, for which we have better solutions now.
> 
>     * UUIDs are impractical because there are large - remember this needs to be then sent and parsed with every request.
> 
>     * When server restarts a new UUID will be assigned so even though the cache topology will be the same, the load balancer will not be able to route the session to the original node.
> 
>     * The implementation itself is not well designed, setting `uuid-instance-id` and `instance-id` at the same time results in unclear configuration, much better way to implement this would be along the lines of attribute `routing="STATIC|UUID|HOSTNAME"` to allow future evolution of route handling.

@rhusar I am reviewing all the feedback I received for this Jira in both this PR and in the WildFly PR. Can you ellaborate a little bit more what you suggested here? Taking into account that we are replacing UUID by obfuscated route, it would be rounting="STATIC|OBFUSCATED|HOSTNAME" What do you mean exactly with static routing and hostname routing in the light of how routing is currently implemented today: https://github.com/wildfly/wildfly/blob/master/undertow/src/main/java/org/wildfly/extension/undertow/Server.java#L174 ?

undertow/WFLY-12473_obfuscate-session-route.adoc
```diff
@@ -0,0 +1,79 @@
...
+
+== Overview
+
+This feature is to be able to obfuscate the value of UndertowService.instanceId when routing. Routes are used as
+part of generated session ids when a remote invocation (such as EJB, naming or serving a wildfly-services address)
+is handled by HTTP Invoker. This data is sent across connections as part of the request and response JSESSION_ID
+cookie.
+EAP6 Web Server would similarly allow obfuscation of instance-id at routing by allowing a generated
+UUID to be used as the instance-id.
+This UUID was the default value used whenever instance-id configuration was absent, while Undertow default value
```
tommaso-borgato, line 18: @fl4via do you mean `, while Undertow default value with EAP 7, for instance-id is ${jboss.node.name}.` ?

undertow/WFLY-12473_obfuscate-session-route.adoc
```diff
@@ -0,0 +1,79 @@
...
+
+* https://issues.jboss.org/browse/EAP7-1187
+* https://issues.jboss.org/browse/PRODMGT-1934
+
+=== Dev Contacts
+
+* mailto:{email}[{author}]
+
+=== QE Contacts
+
```
tommaso-borgato, line 38: @fl4via can you add tborgato@redhat.com here?

undertow/WFLY-12473_obfuscate-session-route.adoc
```diff
@@ -0,0 +1,79 @@
...
+=== Testing By
+// Put an x in the relevant field to indicate if testing will be done by Engineering or QE.
```
tommaso-borgato, line 40: @fl4via please remove these two commented lines

tommaso-borgato (changes requested):

undertow/WFLY-12473_obfuscate-session-route.adoc
```diff
@@ -0,0 +1,79 @@
...
+=== Dev Contacts
+
+* mailto:{email}[{author}]
+
+=== QE Contacts
+
+=== Testing By
+// Put an x in the relevant field to indicate if testing will be done by Engineering or QE. 
+// Discuss with QE during the Kickoff state to decide this
+[X] Engineering
```
rhusar, line 42: btw these don't render correctly - use * [X] instead.

* [X] this is what it looks like rendered correctly
[X] this doesn't rener

undertow/WFLY-12473_obfuscate-session-route.adoc
fl4via: Yes, I'll update

undertow/WFLY-12473_obfuscate-session-route.adoc
fl4via: Adding.

undertow/WFLY-12473_obfuscate-session-route.adoc
fl4via: thanks for the feedback!

undertow/WFLY-12473_obfuscate-session-route.adoc
```diff
@@ -0,0 +1,79 @@
...
+== Requirements
+
+This will be implemented as an additional attribute under the `subsystem=undertow` management model address.
+
+This will be a simple boolean attribute called `obfuscate-session-route` that indicates that instance-id value must be obfuscated
+when used at the server route.
+
+=== Hard Requirements
+
+The route must not change across server restarts when `obfuscate-session-route` value is `true`.
```
tommaso-borgato, line 63: @fl4via is this true regardless of the physical node where the server is restarted as long as the server parameters stay the same?

tommaso-borgato (changes requested):

undertow/WFLY-12473_obfuscate-session-route.adoc
```diff
@@ -0,0 +1,79 @@
...
+to use the server name as a salt when generating the obfuscated instance-id bytes.
+
+
+=== Non-Requirements
+
+== Test Plan
+
+A few test cases will be submitted to verifying that the instance-id of Undertow subsystem is obfuscated on routing when `obfuscate-session-route` is `true`, and to
+verify that affected functionalities (namely, remote invocation via HTTP invoker) work as expected when session route is obfuscated.
+
```
tommaso-borgato, line 76: @fl4via given the **Hard requirement**: _"The route must not change across server restarts..."_, can you also add a test to verify this?

tommaso-borgato (changes requested):

tommaso-borgato:
> > -1 I am still not convinced this is a good idea.
> > ```
> > * What is the actual problem this is trying to solve? I don't think the statement "customers would like to have it default to a UUID instead" is exactly correct, they are probably just looking for a feature which was default in the past, but no longer, for which we have better solutions now.
> > 
> > * UUIDs are impractical because there are large - remember this needs to be then sent and parsed with every request.
> > 
> > * When server restarts a new UUID will be assigned so even though the cache topology will be the same, the load balancer will not be able to route the session to the original node.
> > 
> > * The implementation itself is not well designed, setting `uuid-instance-id` and `instance-id` at the same time results in unclear configuration, much better way to implement this would be along the lines of attribute `routing="STATIC|UUID|HOSTNAME"` to allow future evolution of route handling.
> > ```
> 
> @rhusar I am reviewing all the feedback I received for this Jira in both this PR and in the WildFly PR. Can you ellaborate a little bit more what you suggested here? Taking into account that we are replacing UUID by obfuscated route, it would be rounting="STATIC|OBFUSCATED|HOSTNAME" What do you mean exactly with static routing and hostname routing in the light of how routing is currently implemented today: https://github.com/wildfly/wildfly/blob/master/undertow/src/main/java/org/wildfly/extension/undertow/Server.java#L174 ?

@fl4via, @rhusar is there any follow up to this discussion that's worth mentioning here?

undertow/WFLY-12473_obfuscate-session-route.adoc
fl4via: @tommaso-borgato right, updating

fl4via:
@tommaso-borgato  yes, there is. I discussed with @pferraro at the chat room a while ago what would be the best approach here, and basically in light of the changes that were done, he agreed that obfuscate-session-route argument was the best approach.
@pferraro  feel free to add any info here if you have more input to give, it is always welcome!

fl4via:
@tommaso-borgato I just added a new commit to incorporate the latest feedback. Once we reach a final agreement on this last points, I'll stash the two commits into a single, final one.

undertow/WFLY-12473_obfuscate-session-route.adoc
```diff
@@ -0,0 +1,86 @@
...
+Also, we also need tests verifying that the route does not changes across server restarts, as this is a hard requirement.
```
tommaso-borgato, line 82: @fl4via just a typo .... a repetition of `also` ....

tommaso-borgato (changes requested):

undertow/WFLY-12473_obfuscate-session-route.adoc
```diff
@@ -0,0 +1,86 @@
...
+It is ideal to make it hard to find out the instance-id value via reverse engineering of the value used in route. For that, we are going
+to use the server name as a salt when generating the obfuscated instance-id bytes.
+
+
+=== Non-Requirements
+
+== Test Plan
+
+A few test cases will be submitted to verifying that the instance-id of Undertow subsystem is obfuscated on routing when `obfuscate-session-route` is `true`, and to
+verify that affected functionalities (namely, remote invocation via HTTP invoker) work as expected when session route is obfuscated.
```
tommaso-borgato, line 80: @fl4via a clarification: do `remote invocation via HTTP invoker` includes scenarios like when we use a load balancer such as mod_cluster? or is this outside the scope of the RFE?

tommaso-borgato (changes requested):

undertow/WFLY-12473_obfuscate-session-route.adoc
tommaso-borgato: @fl4via I see the new `SessionManagementWithObfuscatedRouteTestCase` test (btw tnks for the route retrieval code) ... can you add a line for it? it's about checking correct HTTP session management when the route is obfuscated if I don't go wrong ...

undertow/WFLY-12473_obfuscate-session-route.adoc
```diff
@@ -0,0 +1,86 @@
...
+The route must not change across server restarts when `obfuscate-session-route` value is `true`, unless there is a change in the values
+ the route is derived from:
+
+* the route equals `instance-id` attribute value, defined at `subsystem=undertow` management model address
+* `instance-id` default value is `"${jboss.node.name}"`, meaning that, when the attribute is kept unchanged, changes to `${jboss.node.name}`
+accross restarts will result in changes to the route
+
+=== Nice-to-Have Requirements
+It is ideal to make it hard to find out the instance-id value via reverse engineering of the value used in route. For that, we are going
+to use the server name as a salt when generating the obfuscated instance-id bytes.
```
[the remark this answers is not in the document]
fl4via: @tommaso-borgato it is not the server.name, it is the jboss.node.name indeed that is used. I'll add an observation to hard requirements about changes in node name affecting the route.

undertow/WFLY-12473_obfuscate-session-route.adoc
fl4via: @tommaso-borgato sure I can add any missing checks to the test, but isn't the test already doing that? Please, explain better what is missing. I'll add a link to the updated test soon here, as I'm cleaning up and rebasing the code in the RFE PR.

undertow/WFLY-12473_obfuscate-session-route.adoc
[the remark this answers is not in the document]
fl4via: @tommaso-borgato now I see what you mean! At this point in the text, I mean the server name attribute in the model. Inside undertow subsystem we have one or more servers. The server service is the the one that provides the route via Server.getRoute(). In the implementation, we use the server name as a salt: https://github.com/fl4via/wildfly/blob/WFLY-12473_retrieve-route-from-servlet-test/undertow/src/main/java/org/wildfly/extension/undertow/Server.java#L176

undertow/WFLY-12473_obfuscate-session-route.adoc
tommaso-borgato: @fl4via this `Test Plan` paragraph in the AD is a high-level description of the tests that are detailed in the TP: I'm just asking to add a phrase here like `A test for checking correct HTTP session management when the route is obfuscated will be added` .... then in the TP we'll have `SessionManagementWithObfuscatedRouteTestCase` explicitly listed ....

undertow/WFLY-12473_obfuscate-session-route.adoc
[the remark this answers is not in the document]
tommaso-borgato: @fl4via tnks for adding the observation

PR #248, opened 2019-10-03
[WFLY-12609] MicroProfile JWT 1.1

darranl:
At the time of submission this PR is an initial draft, further updates will be added depending on discussion or any issues found during implementation.

https://issues.jboss.org/browse/WFLY-12609


microprofile/WFLY-12609_MicroProfile_JWT_1_1.adoc
```diff
@@ -0,0 +1,109 @@
...
+
+Once an identity has been established according to the steps described by the MicroProfile JWT specification the remaining security decisions required are covered by the relevant Jakarta EE specifications in use by the various subsystems. 
+
+=== Nice-to-Have Requirements
+
+It will not be a priority of this enhancement to add support for alternative authentication mechanisms to `MP-JWT` as this brings a lot of additional complexity.  For this initial integration how we provide support for the `MP-JWT` mechanism should be largely transparent to the deployment allowing us to evolve this aspect further.
+
+Some of the issues we need to consider may be worthwhile discussing at the specification level.
+
+Presently SmallRye JWT is making use of EE Security to provide an authentication mechanism.  A big benefit of making use of EE Security is that a deployment can also make use of this specification to provide an `@IdentityStore` implementation as described in the MP JWT specification.  Deployments also have the option of bringing in their own authentication mechanisms if they are not making use of `MP-JWT`.
```
sberyozkin: It is optional for the integrating containers to use, note the JAX-RS `DynamicFeature` which checks for the availability of the EE Security mechanism on the class path is not annotated with JAX-RS `@Provider` to avoid it being causing some side-effects so it is up to a given container whether it will be activated or not

MikeEdgar: +1 to using EE security. I've been using SmallRye JWT on WildFly for several months (bundled in the web archive) and it has worked nicely - currently running 17.0.1. The big benefit to using EE security is the availability of the principal in downstream components, i.e. EJBs.

darranl: I would suggest the availability of the principal in downstream components is not a benefit that will be gained by using EE Security - once this integration in in place the integration will be establishing a SecurityIdentity using the server's security SPIs so an identity would already be propagated as all of the sub-seployments are sharing a common SecurityDomain anyway.

This isn't going to need to be an immediate decision but the decisions around using EE Security vs Native Security vs ???? I suspect will be more focused on the mechanisms themselves with the following considerations: -
- MP-JWT
- Standard Mechanisms
- Vendor Mechansism
- Concurrent Mechanisms
- User supplied mechanisms (managed)
- User supplied mechanisms (deployed).
- Managed security realms
- Deployed IdentityStores

microprofile/WFLY-12609_MicroProfile_JWT_1_1.adoc
```diff
@@ -0,0 +1,109 @@
...
+
+=== Other Interested Projects
+
+The integration will also be making use of https://github.com/smallrye/smallrye-jwt[SmallRye JWT] as well as pulling in the https://github.com/eclipse/microprofile-jwt-auth[Eclipse MicroProfile JWT] API library.
+
+== Requirements
+
+=== Hard Requirements
+
+Primarily this enhancement will be delivered using a new subsystem `microprofile-jwt` which will be added to WildFly.  The subsystem will install a DeploymentUnitProcessor which will detect if MicroProfile JWT is required and activate a suitable security configuration as described within this proposal.
```
honza-kasik: 1. Will this subsystem have any operations available in management interface? If it will - which?
1. Will this subsystem have any attributes interface? If it will - which?

microprofile/WFLY-12609_MicroProfile_JWT_1_1.adoc
```diff
@@ -0,0 +1,109 @@
...
+* mailto:{email}[{author}]
+
+=== QE Contacts
+
+=== Testing By
+// Put an x in the relevant field to indicate if testing will be done by Engineering or QE. 
+// Discuss with QE during the Kickoff state to decide this
+[ ] Engineering
+
+[ ] QE
```
honza-kasik: Done by QE - me respectively

microprofile/WFLY-12609_MicroProfile_JWT_1_1.adoc
```diff
@@ -0,0 +1,109 @@
...
+
+This integration will not be supporting the use of pre-configured SecurityDomains, the general motivation to use MicroProfile JWT is to use an authentication mechanism where identities are self described using the tokens exchanged eliminating the need for configured security resources for the secured endpoint(s).  Use cases could be identified where pre-configured resources add an additional benefit but those would be reviewed under their own enhancements.
+
+It is assumed when working with the MicroProfile specifications that we will be supporting a single deployment, this deployment may contain sub-deployments.  The SecurityDomain of each top level deployment will be isolated from the SecurityDomain of any parallel top level deployment.  Advanced WildFly Elytron features such identity propagation across domains / top level deployments will not be supported.  At a later point we could consider an enhancement to offer more control of the SecurityDomains in use but that is out of scope for this enhancement.
```
rsvoboda: > At a later point we could consider an enhancement to offer more control of the SecurityDomains in use but that is out of scope for this enhancement.

Meaning WF 20+ ?

microprofile/WFLY-12609_MicroProfile_JWT_1_1.adoc
```diff
@@ -0,0 +1,109 @@
...
+
+== Implementation Plan
+
+The primary entry point for this enhancement will be a new subsystem `microprofile-jwt` which will be added to the WildFly project.  The WildFly project will also pull in SmallRye JWT as well as the MicroProfile JWT API.
```
rsvoboda: Will be some statistics available `microprofile-jwt` subsystem? 

E.g. which deployments are using MP JWT, what realm names are the defined, which authentication mechanism are defined and used, details about virtual security domains, valid / invalid / corrupted tokens

microprofile/WFLY-12609_MicroProfile_JWT_1_1.adoc
```diff
@@ -0,0 +1,109 @@
...
+
+Finally there will be an additional project `Elytron JWT` added to the WildFly Elytron project, this will also depend on SmallRye JWT, the purpose of this project will be to add any tighter integration required with the WildFly Elytron APIs.  The contents of the `Elytron JWT` project will be considered private API and will be subject to change between releases.
+
+If enhancements are required these will be submitted directly to the SmallRye JWT project, the Elytron JWT project will be a back up if we need a location to hold enhancements whilst the availability in SmallRye JWT is considered.
+
+== Test Plan
+
+Details are still to be discussed but in general testing will make use of the MicroProfile TCK, additionally integration tests will be added to the WildFly testsuite using deployments configured to use MicroProfile JWT.  Any code added to WildFly Elytron will also be tested using co-located unit tests.
```
rsvoboda: Equivalence of Thorntail project coverage must be ensured by RFE owner.
Following test scenarios must be covered in WildFly testsuite:

 - https://github.com/thorntail/thorntail/tree/master/testsuite/testsuite-keycloak-mpjwt
 - https://github.com/thorntail/thorntail/tree/master/testsuite/testsuite-microprofile-jwt

microprofile/WFLY-12609_MicroProfile_JWT_1_1.adoc
```diff
@@ -0,0 +1,109 @@
...
+
+One aspect we may consider is the generation of tokens as that is outside the specification, within the testsuites and possibly as a stand alone example we may want an endpoint which is capable of generating the JWT tokens - this would not be a part of our supported distribution however it would be useful to allow for deployments to use MicroProfile JWT without additional infrastructure requirements.  Something such as this should have the ability to take a WildFly Elytron SecurityIdentity and convert it into a JWT token.
```
rsvoboda: You mean something like http://jwtenizr.sh/ ?

microprofile/WFLY-12609_MicroProfile_JWT_1_1.adoc
```diff
@@ -0,0 +1,109 @@
...
+
+== Community Documentation
+
+Documentation will be added to the WildFly documentation showing how a MicroProfile JWT deployment can be deployed, we may want to restrict the level of detail as the behaviour will predominantly be covered by the specification.
+
+
```
rsvoboda: What are the plans for Quickstart ? Have you spoke with @xstefank and @emmartins ?

microprofile/WFLY-12609_MicroProfile_JWT_1_1.adoc
darranl: No the subsystem will not have any operations or attributes at this point in time, the presence of the subsystem will activate it's DUPs and the remainder of the configuration will be based on MP Config resolution rules.

In the future we could add some configuration for alternative Elytron integration options but those should not be required at the outset.

microprofile/WFLY-12609_MicroProfile_JWT_1_1.adoc
```diff
@@ -0,0 +1,109 @@
...
+== Requirements
+
+=== Hard Requirements
+
+Primarily this enhancement will be delivered using a new subsystem `microprofile-jwt` which will be added to WildFly.  The subsystem will install a DeploymentUnitProcessor which will detect if MicroProfile JWT is required and activate a suitable security configuration as described within this proposal.
+
+At the outset activation will only occur in the following circumstances: -
+
+ * A `@org.eclipse.microprofile.auth.LoginConfig` annotation is detected with an `auth-method` of `MP-JWT`.
+ * An auth method of `MP-JWT` is specified in any web.xml.
```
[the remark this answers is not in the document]
darranl: If we use the auth-method from the web.xml we will also use any realm name specified in the web.xml - overall web.xml contents are higher priority than the annotation.

microprofile/WFLY-12609_MicroProfile_JWT_1_1.adoc
```diff
@@ -0,0 +1,109 @@
...
+
+This activation scope is deliberately constrained to allow us to incrementally increase the scope at a later point. 
+
+At the point activation occurs the deployment will not make use of a configured `SecurityDomain` instead a "Virtual" `SecurityDomain` will be created to meet the requirements of the deployment and will be used across all sub-deployments.  The common `SecurityDomain` will allow for identity propagation across all sub-deployments whilst the virtual nature of the domain will make it unavailable for propagation across unrelated deployments allowing for deployment level isolation.
```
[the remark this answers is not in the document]
darranl: Sub deployments could be in relation to an ear which contains multiple deployments, but it could also be in relation to say a war which contains ejbs where we will also end up deploying across multiple containers.

microprofile/WFLY-12609_MicroProfile_JWT_1_1.adoc
```diff
@@ -0,0 +1,109 @@
...
+
+Activation will add module dependencies as required on SmallRye JWT, Elytron JWT, and the MicroProfile JWT API.
+
+Activation will also ensure activation of the SmallRye JWT CDI extension as well as any Elytron JWT CDI extension if deemed required and any Elytron specific overrides.
```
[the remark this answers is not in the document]
darranl: This subsystem is operating on Elytron APIs ONLY there will be no PicketBox variant.

microprofile/WFLY-12609_MicroProfile_JWT_1_1.adoc
darranl: WF 19+ - but regardless not this enhancement.

microprofile/WFLY-12609_MicroProfile_JWT_1_1.adoc
darranl: No, that is an example of something that should be covered in a future enhancement.

microprofile/WFLY-12609_MicroProfile_JWT_1_1.adoc
darranl: Will review options for Quickstarts later and update here as required.

microprofile/WFLY-12609_MicroProfile_JWT_1_1.adoc
darranl: We can review of porting those existing tests into the WildFly testsuite could make sense.

microprofile/WFLY-12609_MicroProfile_JWT_1_1.adoc
darranl: Yes potentially something similar to that to generate tokens that can be used to call endpoints.

microprofile/WFLY-12609_MicroProfile_JWT_1_1.adoc
```diff
@@ -0,0 +1,109 @@
...
+
+Finally activation will ensure the activation of the `MP-JWT` authentication mechanism for the deployment.
+
+The combination of SmallRye JWT with Elytron JWT will be the pair of projects responsible for providing the functionality described in the MicorProfile JWT specification combined with tying in the integration to the application server's security framework to ensure aspects such as identity association and establishment are correctly performed.
+
+Once an identity has been established according to the steps described by the MicroProfile JWT specification the remaining security decisions required are covered by the relevant Jakarta EE specifications in use by the various subsystems. 
+
```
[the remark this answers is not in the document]
darranl: Yes SmallRye JWT, Elytron JWT and the new MicroProfile JWT susbsytem will each be distinct modules.

microprofile/WFLY-12609_MicroProfile_JWT_1_1.adoc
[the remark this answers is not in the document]
rsvoboda: Thanks for clarification

microprofile/WFLY-12609_MicroProfile_JWT_1_1.adoc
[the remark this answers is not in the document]
rsvoboda: What about case when that ejb from deployment A is calling another ejb from different deployment B? Will be the identity propagated or not ? Deployment A is using MP JWT ...

My guess is that this case falls into "virtual nature of the domain will make it unavailable for propagation across unrelated deployments", but want to be sure.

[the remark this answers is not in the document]
darranl: +1 This is one of the main motivations for using a common "virtual security domain" across all sub-deployments - once the identity is established it will be available in all of those deployments.

As a motivation for these deployments is more along the lines of package / deploy / run overall this eliminates the need for complex subsystem configuration for things like identity propagation etc...

I should probably add a statement somewhere but my aim is a developer should be able to take a clean app server and provided the configuration contains the microprofile-jwt subsystem based on our other packaging discussions they should be able to deploy their deployment making use of MicroProfile JWT without being required to make any configuration changes to the server.

microprofile/WFLY-12609_MicroProfile_JWT_1_1.adoc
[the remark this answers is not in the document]
rsvoboda: That's a bit confusing atm. PicketBox is still available and can be the default. In such case JWT functionability wont be available ? 

I thought this will be transparent for JWT usage / backend impl agnostic. 
Only if users will want to use something "advanced" they will need to use Elytron.

rsvoboda:
Hi @darranl, thanks for your comments. Replied to few of the, others are fine with me.

You missed few of my comments, probably because GH collapses them 😱. Please take a look at them.

microprofile/WFLY-12609_MicroProfile_JWT_1_1.adoc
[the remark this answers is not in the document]
darranl: From the perspective of deployments there is actually no notion of a default - for all deployments appropriate security resources need to be defined before a deployment is deployed to the application server and the administrator can choose to use either Elytron or the legacy PicketBox resources.

This specific enhancement is adding support for the security configuration to be provided exclusively using the MicroProfile Config.

For this reason there is no association with managed security resources instead everything is created dynamically and internally to meet the needs of the deployment.

From the perspective of the developer the security framework in use is entirely internal to the application server - it would only be if they wanted to make use of vendor specific APIs that they would need to use Elytron APIs.

If it wasn't for the server exposing it's management interfaces the Elytron subsystem and Security subsystem almost become optional - it would not take too many steps to be able to remove these completely leaving a server distribution to just support JWT.

microprofile/WFLY-12609_MicroProfile_JWT_1_1.adoc
```diff
@@ -0,0 +1,109 @@
...
+=== Nice-to-Have Requirements
+
+It will not be a priority of this enhancement to add support for alternative authentication mechanisms to `MP-JWT` as this brings a lot of additional complexity.  For this initial integration how we provide support for the `MP-JWT` mechanism should be largely transparent to the deployment allowing us to evolve this aspect further.
```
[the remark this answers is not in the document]
darranl: For the alternative authentication mechanisms they still remain fully supported without the activation of the new MicroProfile JWT specification and at the moment the two do not come together cleanly.

As an example the motivation for using MicroProfile JWT is these tokens fully describe the resulting identity, this is why we can avoid relying on managed security configuration - the support for the alternative authentication mechanisms then triggers a dependency on managed configuration to specify things such as database and ldap identity stores etc...

Additionally once you move to mechanisms other than MP-JWT the remainder of the spec also becomes obsolete - items such as the Principal type to inject etc.. are no longer available as there was no longer an incoming token to make available / propagate.

There are a lot of options regarding alternative mechanism support but the primary focus here is adding support for MP-JWT - once this is added will open up discussions on how to better solve the alternative mechanism support - this may be enhancement to WildFly, enhancements to SmallRye JWT and hopefully even clarifications within the specification.

microprofile/WFLY-12609_MicroProfile_JWT_1_1.adoc
```diff
@@ -0,0 +1,109 @@
...
+
+However there are a couple of draw backs from the EE Security approach.  A first drawback is this brings in dependencies on JASPI and JACC both of which must be correctly enabled on the server, the `MP-JWT` mechanism is a very trivial light weight mechanism so a lot of overhead is pulled in.  A second draw back is that it should be possible for standard and vendor specific mechanisms to be specified, this brings us additional challenges as many of our mechanisms are native WildFly Elytron mechanisms.  Making use of native mechanisms has it's own complexities as we may need to bridge into using an `@IdentityStore` made available by CDI, alternatively this may require configuration of domains not covered by MicroProfile Config values.  We could consider porting some of our vendor specific mechanisms to EE security but one of the features of our mechanisms is the ability to support multiple mechanisms concurrently which is not completely covered by EE security.  The `MP-JWT` mechanism is also ideally suited to being used in parallel with other authentication mechanisms however we could end up in a situation with a hybrid of native mechanisms and an EE Security mechanism.
```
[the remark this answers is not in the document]
darranl: This is a single paragraph, the formatting is just how the tools we have chosen to use represent a paragraph.

The overall issue is this is a lot of layers to activate for a relatively trivial authentication mechanism whilst also making it harder to support some of the advanced features we would want to support such as multiple concurrent mechanisms.

For now this is just capturing some thoughts, as we are making use of SmallRye JWT the aim of the initial integration it to make use of the mechanism as an EE Security mechanism - this is an internal detail so can re-evaluate later.

microprofile/WFLY-12609_MicroProfile_JWT_1_1.adoc
```diff
@@ -0,0 +1,109 @@
...
+
+Overall these issues can be considered and solved, however I believe the primary motivation for activating MicroProfile JWT is to make use of the `MP-JWT` authentication mechanism so that will be the primary focus of this RFE, depending on time constraints consideration of alternative authentication mechanisms will either take place towards the end of this enhancement or in a follow up enhancement.
```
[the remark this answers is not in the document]
darranl: Yes after further review and commencing work on the integration it makes no sense to avoid merging this RFE whilst solving alternative authentication mechanisms, it also does not make sense to deffer the complete RFE to a later WildFly release to support alternative authentication mechanisms so that should be discussed on it's own.

We will track ongoing work and towards the end of this RFE start to decide where it the correct home for ongoing work,

microprofile/WFLY-12609_MicroProfile_JWT_1_1.adoc
```diff
@@ -0,0 +1,109 @@
...
+
+=== Non-Requirements
+
+This integration will be making use of WildFly Elytron exclusively, no support will be added for the deprecated PicketBox project.  Having said that the underlying security framework will be transparent for deployments developed according to the specification - the framework only becomes apparent if there is a desire to use vendor specific APIs.
```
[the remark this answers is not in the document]
darranl: As of WildFly 11 it has been agreed we will not adding any further PicketBox enhancements.

darranl:
The notion of dynamic MP Config is currently being discussed https://groups.google.com/forum/#!topic/smallrye/CA-FJNjCkME

No immediate action but may need to come back and check if we need to support any dynamic changes.

rsvoboda:
Dynamic MP Config integration sounds like new feature request.

I think it would be MP Config 2.0+ thing. But in MP world new majors can happen every second month :)

darranl:
Yeah I don't think Dynamic MP Config is going to be a reality in the short term for this enhancement, just worth keeping in mind for any new code if it could be adapted to support dynamic updates.

darranl:
I have just pushed a follow up commit following the first round of discussions, hopefully this has added some clarification rather than making it more ambiguous :smiley: 

I have moved alternative authentication mechanisms into the "non-requirements" section as that does make sense to cover under it's own enhancement.  As  that topic should be it's own enhancement I have not added any further discussion there other than to clarify what alternative mechanisms mean.

microprofile/WFLY-12609_MicroProfile_JWT_1_1.adoc
```diff
@@ -0,0 +1,153 @@
...
+[source,xml]
+----
+<subsystem xmlns="urn:wildfly:microprofile-jwt-smallrye:1.0"/>
+----
+
+Additionally the initial version of this subsystem will expose no operations and no attributes.
+
+The default location to specify the authentication requirements is within the `web.xml` deployment descriptor, the MicorProfile JWT specification add a new annotation `@org.eclipse.microprofile.auth.LoginConfig` which allows a JAX-RS `javax.ws.rs.core.Application` to be annotated to specify the desired `auth-method` and `realm-name`.
+
+During deployment, deployments will be scanned for a class extending `Application` annotated with the new `@LoginConfig` annotation - if the authentication configuration was not previously specified within the `web.xml` the configuration from this annotation will be used instead.  If the authentication configuration has already been specified in the `web.xml` annotation scanning will be skipped as the configuration in the `web.xml` takes precendence over the annotation.
```
honza-kasik: What configuration has to be set in `web.xml` to enable JWT auth?

darranl: If the auth-method is set to "MP_JWT" MicroProfile JWT will be activated for the deployment - no other configuration is required.

microprofile/WFLY-12609_MicroProfile_JWT_1_1.adoc
```diff
@@ -0,0 +1,153 @@
...
+
+This integration will not be supporting the use of pre-configured SecurityDomains, the general motivation to use MicroProfile JWT is to use an authentication mechanism where identities are self described using the tokens exchanged eliminating the need for configured security resources for the secured endpoint(s).  Use cases could be identified where pre-configured resources add an additional benefit but those would be reviewed under their own enhancements.
+
+It is assumed when working with the MicroProfile specifications that we will be supporting a single deployment, this deployment may contain sub-deployments.  The SecurityDomain of each top level deployment will be isolated from the SecurityDomain of any parallel top level deployment.  Advanced WildFly Elytron features such identity propagation across domains / top level deployments will not be supported.  At a later point we could consider an enhancement to offer more control of the SecurityDomains in use but that is out of scope for this enhancement.
```
honza-kasik: >  It is also possible that MicroProfile JWT is activated for a deployment that it itself a sub-deployment such as a `war` contained within an `ear` or in cases where the `war` contains items to be deployed in another container such as EJBs within the `war`.  In all of these cases the common "Virtual" `SecurityDomain` will be used across all containers handling the deployment - by using a common `SecurityDomain` any identity established in one container will automatically propagate with any calls to the other containers without the need for any manual configuration. 

This example describes successive levels, so it is not in conflict with this non-requirement, right? Or is it regarding other issue?

darranl: For now it may be worth adding a note to say mixed deployments will not be supported, i.e. an ear that contains a war using mp-jwt in parallel with a war that uses a different mech.

The motivation here is MicroService development not monolith development so we can always pick that up later with the support for the other mechanisms.

microprofile/WFLY-12609_MicroProfile_JWT_1_1.adoc
```diff
@@ -0,0 +1,153 @@
...
+
+For this initial implementation propagation of identities is outside the current scope.  The specification does make provision to access the token passed to the server so deployments can manually retrieve a token.  Under WFLY-11868 enhancements are being provided to make use of the WildFly Elytron authentication client in RESTEasy, once available under a new RFE the microprofile-jwt subsystem can dynamically configure an `AuthenticationContext` configuration to automate propagation.  This could also introduce a new feature to the client to allow a feature to convert it's current configuration to obtain an alternative token.
```
honza-kasik: How would the identity propagation look in JWT context?

darranl: Identify propagation would be along the lines of re-using the token that we received for an ongoing call to another endpoint on another server.

For now the MicroProfile JWT specification already provides access to the token so any deployments can do this manually, the point we may consider in the future is the automation of this.

microprofile/WFLY-12609_MicroProfile_JWT_1_1.adoc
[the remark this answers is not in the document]
honza-kasik: @darranl Do I understand that you want to decide during work on this RFE where will the enhancement regarding alternative authentication mechanisms end up? As @rsvoboda said - this RFE needs to have clear scope. This decision needs to be included in this document.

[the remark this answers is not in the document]
darranl: No that discussion is entirely outside the scope of this RFE, this RFE is purely about providing support for the MicroProfile JWT specification with the MP-JWT authentication mechanism only.

All further planning will be outside this RFE.

microprofile/WFLY-12609_MicroProfile_JWT_1_1.adoc
xstefank: The quickstart is now tracked separately.

darranl:
Will think later about how to add this to the proposal but in the meantime will need to keep the following in mind when completing the addition of the TCK: -

https://docs.jboss.org/resteasy/docs/4.4.0.Final/userguide/html/Securing_JAX-RS_and_RESTeasy.html

RestEasy does not automatically trigger authentication, also the annotations are only detected if support for them is activated.  Additionally it recommends against activation for EJBs.

The TCK does not make use of a web.xml but we could likely add our own for the test runs, however the TCK does also seem to bundle servlet, Rest Endpoint and EJB Rest Endpoint in a single deployment.

microprofile/WFLY-12609_MicroProfile_JWT_1_1.adoc
honza-kasik: Please include this into the document in next update. This may be important for documentation in the future.

darranl: @honza-kasik This is what it says on line 68 

"After the previous steps have completed if the resulting `auth-method` for the web application being deployed is `MP-JWT` the MicroProfile JWT integration will be activated"

honza-kasik: Thank you - it wasn't clear to me that this relates to `web.xml` too. I had just the `@LoginConfig` annotation on my mind and when it wasn't clear to me, other may come and have same experience.

darranl: Yeah that is why I have tried to redefine it as a sequence of events with activation happening at the end of that sequence.

bstansberry:
@darranl Is there any expectation of any container configuration being used for this beyond the couple MP Config properties discussed in https://download.eclipse.org/microprofile/microprofile-jwt-auth-1.1.1/microprofile-jwt-auth-spec.html#_configuration_parameters ?

darranl:
@bstansberry the SmallRye implementation also has some MP Config values in addition to the ones in the spec - but overall no I am looking for this first version to exclusively use the JWT and SmallRye configuration values and not add any specific configuration.

IMO requests for specific functionality will need careful evaluation as we will need to identify which of the following locations to address it in: -

1. WildFly specific behaviour and configuration.
2. SmallRye specific behaviour and configuration.
3. Evolution of MicroProfile specification (JWT or other)

microprofile/WFLY-12609_MicroProfile_JWT_1_1.adoc
honza-kasik: @darranl Do you plan to reflect current state regarding quickstarts in this paragraph?

darranl: No the Quickstart will be handled under it's own RFE at the moment so that will have it's own analysis.

honza-kasik: @darranl Sorry, seems like I checked older revision. This is OK.

honza-kasik: @darranl My bad again, but when I was writing previous comment, only top paragraph was visible for me for some reason. What I;m talking about is this:

> ...quickstarts should be considered to ensure a quickstart exists that can be used with WildFly.

As you said, they've already been considered and are going to be added as a separate feature.

honza-kasik: Please replace ^^^ by something like "quickstart will be processed as a separate proposal".

marekkopecky:
`Release Note Content` section from [the template](https://github.com/wildfly/wildfly-proposals/blob/master/design-doc-template.adoc) is missing here

marekkopecky:
`s/MicorProfile/MicroProfile/g` for the whole document

darranl:
@marekkopecky I will add a set of release notes once we get ready for the final review before merge, release notes are for @bstansberry to make it easier to put together a set of release notes for the application server release.

marekkopecky:
@darranl Release notes section is a part of template, so this section is mandatory part of this analysis as well. This analysis can't be marked as done, if the release notes section is not available.

darranl:
@marekkopecky That section will be added once we are closer to the end of the enhancement.

darranl:
@bstansberry I have just added a commit for the release notes - is there anything else you need in that section?

darranl:
I have resolved the intermediate discussions on this document, the implementation is effectively complete so I think ready to check if any additional items are required.

microprofile/WFLY-12609_MicroProfile_JWT_1_1.adoc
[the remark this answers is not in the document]
darranl: +1 will change that - that now is 100% in follow and and TBH I want to discuss that at the spec level first.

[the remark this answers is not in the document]
darranl: @honza-kasik I have removed that sentence entirely, this block of text is now in the non-requirements section but that sentence came down when it was still in the nice to have section.

PR #250, opened 2019-10-09
Proposal for WFLY-12590 Integrate MicroProfile Fault Tolerance 2.0

rhusar:
Jira
https://issues.jboss.org/browse/WFLY-12590
https://issues.jboss.org/browse/EAP7-1347

### DIFF
--- a/microprofile/WFLY-12590_Integrate_MicroProfile_Fault_Tolerance_2.0.adoc
+++ b/microprofile/WFLY-12590_Integrate_MicroProfile_Fault_Tolerance_2.0.adoc
@@ -63,7 +63,7 @@ Not interested in our integration per se, but projects that have already integra
 
 * Support Eclipse MicroProfile Fault Tolerance version 2.0 for all applications deployed in WildFly.
 * The extension will be provided as Galleon layer called `microprofile-fault-tolerance`.
-* [WILL BE CLARIFIED] The extension will be added to relevant default configurations provided by WildFly (standalone and domain mode).
+* The extension will *not* be part of default configurations provided by WildFly (standalone and domain mode) – instead a galleon layer `microprofile-fault-tolerance` will be provided.
 * Upstream Eclipse MicroProfile Fault Tolerance TCK test suite will be added to `microprofile-tck` WildFly upstream test suite.
 * Add integration tests above the scope of TCK (e.g. from https://github.com/thorntail/thorntail/tree/master/testsuite/testsuite-microprofile-fault-tolerance[Thorntail])
 * Support for multiple deployments. Only the first deployment processed will however configure the underlying Hystrix library, the subsequent deployment issues a warning message.
--- a/microprofile/WFLY-12590_Integrate_MicroProfile_Fault_Tolerance_2.0.adoc
+++ b/microprofile/WFLY-12590_Integrate_MicroProfile_Fault_Tolerance_2.0.adoc
@@ -1,0 +1,105 @@
+= WFLY-12590 Integrate MicroProfile Fault Tolerance 2.0
+:author:            Radoslav Husar
+:email:             rhusar@redhat.com
+:toc:               left
+:icons:             font
+:idprefix:
+:idseparator:       -
+:keywords:          microprofile,fault-tolerance,openshift
+
+== Overview
+
+Integrate support for Eclipse MicroProfile (MP) Fault Tolerance (FT) for applications deployed in WildFly.
+This proposal pertains to MP FT version 2.0 given the timing of this RFE.
+The integration is based on SmallRye project, specifically `smallrye-fault-tolerance` module.
+The `smallrye-fault-tolerance` module is already in use by projects https://thorntail.io/[Thorntail] (WildFly Swarm) and https://quarkus.io/[Quarkus].
+
+MicroProfile Fault Tolerance specification simplifies the challenges brought by the distributed nature of microservices.
+This is because the communication with external systems is inherently unreliable.
+This increases demand on resiliency of applications.
+
+The patterns brought by this specification are annotations such as `@Timeout`, `@Retry`, `@Fallback`, `@CircuitBreaker`, `@Bulkhead` and `@Asynchronous`.
+
+== Issue Metadata
+
+=== Issue
+
+* https://issues.jboss.org/browse/WFLY-12590[WFLY-12590 Integrate MicroProfile Fault Tolerance 2.0]
+* https://issues.jboss.org/browse/EAP7-1347[EAP7-1347 Integrate MP Fault Tolerance 2.0 into EAP]
+
+=== Related Issues
+
+* https://issues.jboss.org/browse/WFLY-10522[WFLY-10522 Support Eclipse MicroProfile Config] - dependency; already integrated
+
+=== Dev Contacts
+
+* mailto:{email}[{author}]
+
+=== QE Contacts
+
+* mailto:mnovak@redhat.com[Miroslav Novak]
+
+=== Testing By
+// Put an x in the relevant field to indicate if testing will be done by Engineering or QE.
+// Discuss with QE during the Kickoff state to decide this
+[ ] Engineering
+
+[x] QE
+
+=== Affected Projects or Components
+
+* The specification repository resides at official Eclipse organization on GitHub in https://github.com/eclipse/microprofile-fault-tolerance[eclipse/microprofile-fault-tolerance] repository.
+* The implementation of Eclipse MicroProfile Fault Tolerance will be provided by SmallRye project https://github.com/smallrye/smallrye-fault-tolerance[smallrye/smallrye-fault-tolerance].
+* The underlying library used by _current_ SmallRye implementation is Hystrix which resides in official Netflix GitHub organization https://github.com/Netflix/Hystrix[Netflix/Hystrix].
+
+=== Other Interested Projects
+
+Not interested in our integration per se, but projects that have already integrated MP FT:
+
+* https://thorntail.io/[Thorntail] has integrated MicroProfile Fault Tolerance as a https://github.com/thorntail/thorntail/tree/master/fractions/microprofile/microprofile-fault-tolerance[fraction].
+* https://quarkus.io/[Quarkus] has integrated MicroProfile Fault Tolerance as an https://github.com/quarkusio/quarkus/tree/master/extensions/smallrye-fault-tolerance[extension].
+
+== Requirements
+
+* Support Eclipse MicroProfile Fault Tolerance version 2.0 for all applications deployed in WildFly.
+* The extension will be provided as Galleon layer called `microprofile-fault-tolerance`.
+* [WILL BE CLARIFIED] The extension will be added to relevant default configurations provided by WildFly (standalone and domain mode).
+* Upstream Eclipse MicroProfile Fault Tolerance TCK test suite will be added to `microprofile-tck` WildFly upstream test suite.
+* Add integration tests above the scope of TCK (e.g. from https://github.com/thorntail/thorntail/tree/master/testsuite/testsuite-microprofile-fault-tolerance[Thorntail])
+* Support for multiple deployments. Only the first deployment processed will however configure the underlying Hystrix library, the subsequent deployment issues a warning message.
+
+== Nice-to-Have Requirements
+
+* Create a MicroProfile Fault Tolerance https://github.com/wildfly/quickstart[quickstart] – delivered via separate RFE above of scope of this one.
+
+Items that are currently only being taken into consideration with next version of this feature:
+
+* The underlying library used by SmallRye FT implementation is Hystrix which is currently in "maintenance" mode and a final
+  release of the project (version 1.5.18) has been released. The discussion/efforts about replacing the library are ongoing.
+* WildFly's standard management model will be configuration-less:
+** all configuration is done through MicroProfile Config specification
+** given the maintenance problem above it is not desirable to expose a management model which will change in near future as concepts are not cleanly mappable to a different underlying library
+
+== Non-Requirements
+
+* N/A
+
+== Implementation Plan
+
+* Add dependencies on `smallrye-fault-tolerance` and `microprofile-fault-tolerance-api` artifacts.
+* Add a new `microprofile-fault-tolerance-smallrye` extension to provide Fault Tolerance APIs support for deployments.
+* Add configuration-less MicroProfile Fault Tolerance subsystem.
+* Support configuration of Hystrix from MicroProfile Config specifications.
+  The configuration is queried for Hystrix properties which are then applied through Archaius configuration mechanism.
+  Since Hystrix configuration is singleton only globally scoped configuration is queried and applied.
+
+== Test Plan
+
+* The `smallrye-fault-tolerance` component is passing the MicroProfile Fault Tolerance TCK during its release process.
+* WildFly integration test suite will be enhanced with tests that deploys Java archives which uses the MicroProfile Fault Tolerance API.
+
+== Community Documentation
+
+The feature will be documented in WildFly Administration Guide in a new MicroProfile Fault Tolerance section.
+Since WildFly integration does not impose any special requirements and subsystem configuration, this section will just contain how to configure Hystrix and link to MP FT specification.
+This section will also describe migration path from Thorntail primarily regarding Hystrix configuration.


microprofile/WFLY-12590_Integrate_MicroProfile_Fault_Tolerance_2.0.adoc
```diff
@@ -0,0 +1,99 @@
...
+* The underlying library used by _current_ SmallRye implementation is Hystrix which resides in official Netflix GitHub organization https://github.com/Netflix/Hystrix[Netflix/Hystrix].
+
+=== Other Interested Projects
+
+Not interested in our integration per se, but projects that have already integrated MP FT:
+
+* https://thorntail.io/[Thorntail] has integrated MicroProfile Fault Tolerance as a https://github.com/thorntail/thorntail/tree/master/fractions/microprofile/microprofile-fault-tolerance[fraction].
+* https://quarkus.io/[Quarkus] has integrated MicroProfile Fault Tolerance as an https://github.com/quarkusio/quarkus/tree/master/extensions/smallrye-fault-tolerance[extension].
+
+== Requirements
```
mnovak1: @rhusar It would be great to provide new quickstart with fault tolerance into Wildfly quickstart - https://github.com/wildfly/quickstart - wdyt?

microprofile/WFLY-12590_Integrate_MicroProfile_Fault_Tolerance_2.0.adoc
```diff
@@ -0,0 +1,99 @@
...
+
+* Support Eclipse MicroProfile Fault Tolerance version 2.0 for all applications deployed in WildFly.
+* [TO BE CLARIFIED] The extension is either added to all configurations provided by WildFly (standalone and domain mode)
+  or provided as a layer in a separate build profile.
+* Upstream Eclipse MicroProfile Fault Tolerance TCK test suite will be added to `microprofile-tck` WildFly upstream test suite.
+
+== Nice-to-Have Requirements
```
mnovak1: @rhusar it would be great to provide some information about state of CircuitBreaker/Bulkhead using CLI operation on the deployment. For example it would list all fault tolerance annotations on methods and provide information about its status and configured values.

For example whether `@CircuitBreaker` is open/closed/semi-open. For `@Bulkhead` how many requests is in progress and how many in `waitingTaskQueue`.

CLI operation which would list all open CircuitBreakers might be handle as well.

Looking at https://download.eclipse.org/microprofile/microprofile-fault-tolerance-2.0/microprofile-fault-tolerance-spec.html#_integration_with_microprofile_metrics - this is for free when MP Metrics is added. Could it be enabled on the fly without having to redeploy application?

microprofile/WFLY-12590_Integrate_MicroProfile_Fault_Tolerance_2.0.adoc
```diff
@@ -0,0 +1,99 @@
...
+* Add a new `microprofile-fault-tolerance-smallrye` extension to provide Fault Tolerance APIs support for deployments.
+* Add configuration-less MicroProfile Fault Tolerance subsystem.
+* Support configuration of Hystrix from MicroProfile Config specifications. The configuration is queried for Hystrix properties which are then applied through Archaius configuration mechanism.
+
+== Test Plan
+
+* The `smallrye-fault-tolerance` component is passing the MicroProfile Fault Tolerance TCK during its release process.
+* WildFly integration test suite will be enhanced with tests that deploys Java archives which uses the MicroProfile Fault Tolerance API.
+
+== Community Documentation
```
mnovak1: @rhusar As we'd like Thorntail users to migrate to WF/EAP. Documentation should describe how to migrate Thorntail app to work on WF.

microprofile/WFLY-12590_Integrate_MicroProfile_Fault_Tolerance_2.0.adoc
```diff
@@ -0,0 +1,99 @@
...
+Integrate support for Eclipse MicroProfile (MP) Fault Tolerance (FT) for applications deployed in WildFly.
+This proposal pertains to MP FT version 2.0 given the timing of this RFE.
+The integration is based on SmallRye project, specifically `smallrye-fault-tolerance` module.
+The `smallrye-fault-tolerance` module is already in use by projects https://thorntail.io/[Thorntail] (WildFly Swarm) and https://quarkus.io/[Quarkus].
+
+MicroProfile Fault Tolerance specification simplifies the challenges brought by the distributed nature of microservices.
+This is because the communication with external systems is inherently unreliable.
+This increases demand on resiliency of applications.
+
+The patterns brought by this specification are annotations such as `@Timeout`, `@Fallback`, `@Retry`, `@CircuitBreaker` and `@Bulkhead`.
```
mnovak1: @rhusar Looking at MP Fault Tolerance 2.0: https://download.eclipse.org/microprofile/microprofile-fault-tolerance-2.0/microprofile-fault-tolerance-spec.html#asynchronous

there is new interceptor binding `@Asynchronous` which should be added.

microprofile/WFLY-12590_Integrate_MicroProfile_Fault_Tolerance_2.0.adoc
mnovak1: This would be great to have in Web Console as well - cc-ing @hpehl wdyt?

hpehl: Sounds like a good idea. If there are management operations available, bringing the results into the console makes sense to me.

AIUI the management operations are currently not part of this RFE, right?

mnovak1: @hpehl No RFE exist yet, I think we need to discuss with @rhusar whether this will be follow up RFE or it will part of this one.

microprofile/WFLY-12590_Integrate_MicroProfile_Fault_Tolerance_2.0.adoc
rhusar: Coincidentally I was thinking about it this morning given how great guides some other projects have. Again, this needs to be clarified across all MP specs, since there are currently no quickstarts for MP.

microprofile/WFLY-12590_Integrate_MicroProfile_Fault_Tolerance_2.0.adoc
rhusar: Added a mention about migration.

microprofile/WFLY-12590_Integrate_MicroProfile_Fault_Tolerance_2.0.adoc
rhusar: This wasn't really meant to be a complete enumeration ,but while we are at it, it's complete now.

microprofile/WFLY-12590_Integrate_MicroProfile_Fault_Tolerance_2.0.adoc
rhusar: Currently, I think it's unlikely that we will be exposing management resources or operations for any of this (or any other MP specs). The MP Metrics will be the only interface for this since it's the best fit.

mnovak1: @rhusar Looking at other MP specs in WF there are exposed some management resources, for example for Config, Health or Metrics. It sounds complicated to query metrics rest api for those. Could you raise this question on Monday's Microprofile meeting what is general strategy, please?

microprofile/WFLY-12590_Integrate_MicroProfile_Fault_Tolerance_2.0.adoc
```diff
@@ -0,0 +1,102 @@
...
+* [WILL BE CLARIFIED] Create a MicroProfile Fault Tolerance https://github.com/wildfly/quickstart[quickstart].
+
+Items that are currently only being taken into consideration with next version of this feature:
+
+* The underlying library used by SmallRye FT implementation is Hystrix which is currently in "maintenance" mode and a final
+  release of the project (version 1.5.18) has been released. The discussion about replacing the library is ongoing.
+* WildFly's standard management model will be configuration-less
+** all configuration is done through MicroProfile Config specification
+** given the maintenance problem above it is not desirable to expose a management model which will change in near future as concepts are not cleanly mappable to a different underlying library
+* Support for plurality of isolated deployments; current requirement is single deployment analogous to Thorntail model.
```
mnovak1: @rhusar Interesting, I was not aware of that. Does it mean that it's possible to deploy only one microprofile deployment into WF atm?

microprofile/WFLY-12590_Integrate_MicroProfile_Fault_Tolerance_2.0.adoc
```diff
@@ -0,0 +1,102 @@
...
+=== Other Interested Projects
+
+Not interested in our integration per se, but projects that have already integrated MP FT:
+
+* https://thorntail.io/[Thorntail] has integrated MicroProfile Fault Tolerance as a https://github.com/thorntail/thorntail/tree/master/fractions/microprofile/microprofile-fault-tolerance[fraction].
+* https://quarkus.io/[Quarkus] has integrated MicroProfile Fault Tolerance as an https://github.com/quarkusio/quarkus/tree/master/extensions/smallrye-fault-tolerance[extension].
+
+== Requirements
+
+* Support Eclipse MicroProfile Fault Tolerance version 2.0 for all applications deployed in WildFly.
```
mnovak1: @rhusar This includes not only JAX-RS applications but other EE types (EJBs, Servlets,...) as well, right?

microprofile/WFLY-12590_Integrate_MicroProfile_Fault_Tolerance_2.0.adoc
mnovak1: @rhusar any update from the meeting?

microprofile/WFLY-12590_Integrate_MicroProfile_Fault_Tolerance_2.0.adoc
rhusar: I don't think there is any limitation where you can use this API wherever you can use CDI (Weld)

microprofile/WFLY-12590_Integrate_MicroProfile_Fault_Tolerance_2.0.adoc
rhusar: Updated the wording a little, I don't think its clear yet how multiple deployments will be handled.

microprofile/WFLY-12590_Integrate_MicroProfile_Fault_Tolerance_2.0.adoc
rhusar: > @rhusar Looking at other MP specs in WF there are exposed some management resources, for example for Config, Health or Metrics.

Those are configuration resources, none of those is exposing any runtime resources/ops.

rhusar: > @rhusar any update from the meeting?

The gist is we want to avoid exposing anything extra deployment resources/ops wherever possible. In case we decide to expose anything that we deem necessary (needs to be investigated) this would probably be in the next iteration of the development. 

What we will be currently exposing is as I mentioned earlier in metrics: https://download.eclipse.org/microprofile/microprofile-fault-tolerance-2.0/microprofile-fault-tolerance-spec.html#_integration_with_microprofile_metrics

microprofile/WFLY-12590_Integrate_MicroProfile_Fault_Tolerance_2.0.adoc
```diff
@@ -0,0 +1,102 @@
...
+// Discuss with QE during the Kickoff state to decide this
+[ ] Engineering
+
+[x] QE
+
+=== Affected Projects or Components
+
+* The specification repository resides at official Eclipse organization on GitHub in https://github.com/eclipse/microprofile-fault-tolerance[eclipse/microprofile-fault-tolerance] repository.
+* The implementation of Eclipse MicroProfile Fault Tolerance will be provided by SmallRye project https://github.com/smallrye/smallrye-fault-tolerance[smallrye/smallrye-fault-tolerance].
+* The underlying library used by _current_ SmallRye implementation is Hystrix which resides in official Netflix GitHub organization https://github.com/Netflix/Hystrix[Netflix/Hystrix].
```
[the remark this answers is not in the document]
kenfinnigan: I see the comment further down, so ignore this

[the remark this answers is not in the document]
rhusar: Thanks @kenfinnigan, hopefully we will be able to help drive this. But for now priority #1 is to get whatever we have up and running in WF.

microprofile/WFLY-12590_Integrate_MicroProfile_Fault_Tolerance_2.0.adoc
mnovak1: @rhusar thanks, will be CDI required?

microprofile/WFLY-12590_Integrate_MicroProfile_Fault_Tolerance_2.0.adoc
mnovak1: @rhusar Could you provide more details to this, some example. I'm still not sure what to image under  `plurality of isolated deployments`?

microprofile/WFLY-12590_Integrate_MicroProfile_Fault_Tolerance_2.0.adoc
rhusar: @mnovak1 Most (if not all) MP specs are based on CDI.

mnovak1: @rhusar the idea is to have possibility to use advantages of MP FT annotations for example in EJB (without CDI). Will it be possible? 
It's probably more question what is the trigger point for enabling MP FT in deployment?

rhusar: @mnovak1 You can do that. You just need to have the MP FT extension and empty subsystem in your profile xml. The DUPs add necessary module dependencies and CDI extension. This is in line with what the existing MP subsystems are doing.

rhusar: Just to rephrase, WildFly can't technically limit where these CDI annotations can be used since that one of the limitation in CDI. The scope of where these are supposed to be used is defined by the MP spec.

mnovak1: @rhusar @kenfinnigan thanks guys for clarification. I see your point. This is important for completing test plan for this feature whether there should be mixed EE and MP test scenarios. I think this is more general question. Is there a place to bring this up and having decided for the whole MP for WF?

mnovak1: @rhusar I have another question to this topic. Does it makes sense to test MP FT which package inside EAR/JAR (not just WAR)?

microprofile/WFLY-12590_Integrate_MicroProfile_Fault_Tolerance_2.0.adoc
```diff
@@ -0,0 +1,102 @@
...
+
+Not interested in our integration per se, but projects that have already integrated MP FT:
+
+* https://thorntail.io/[Thorntail] has integrated MicroProfile Fault Tolerance as a https://github.com/thorntail/thorntail/tree/master/fractions/microprofile/microprofile-fault-tolerance[fraction].
+* https://quarkus.io/[Quarkus] has integrated MicroProfile Fault Tolerance as an https://github.com/quarkusio/quarkus/tree/master/extensions/smallrye-fault-tolerance[extension].
+
+== Requirements
+
+* Support Eclipse MicroProfile Fault Tolerance version 2.0 for all applications deployed in WildFly.
+* [WILL BE CLARIFIED] The extension is either added to all configurations provided by WildFly (standalone and domain mode)
```
rsvoboda: `WILL BE CLARIFIED` - any estimation when this will be clear ?

microprofile/WFLY-12590_Integrate_MicroProfile_Fault_Tolerance_2.0.adoc
```diff
@@ -0,0 +1,102 @@
...
+  or provided as a layer in a separate build profile.
+* Upstream Eclipse MicroProfile Fault Tolerance TCK test suite will be added to `microprofile-tck` WildFly upstream test suite.
+
```
rsvoboda: Please add item to ensure same/similar test coverage in WF test-suite as is in Thorntail test-suite.
Thorntail test-suite: https://github.com/thorntail/thorntail/tree/master/testsuite/

Relevant tests in Thorntail test-suite:
 * https://github.com/thorntail/thorntail/tree/master/testsuite/testsuite-microprofile-fault-tolerance
 * other modules are using some aspects - e.g. https://github.com/thorntail/thorntail/blob/master/testsuite/testsuite-microprofile-restclient/src/test/java/org/wildfly/swarm/microprofile/restclient/ft/HelloClientClassLevelCircuitBreaker.java

microprofile/WFLY-12590_Integrate_MicroProfile_Fault_Tolerance_2.0.adoc
```diff
@@ -0,0 +1,102 @@
...
+== Nice-to-Have Requirements
+
+* [WILL BE CLARIFIED] Create a MicroProfile Fault Tolerance https://github.com/wildfly/quickstart[quickstart].
```
rsvoboda: I think it was agreed that QS per MP spec will be delivered.
Thus this belongs to Requirements section.

microprofile/WFLY-12590_Integrate_MicroProfile_Fault_Tolerance_2.0.adoc
```diff
@@ -0,0 +1,102 @@
...
+
+Items that are currently only being taken into consideration with next version of this feature:
+
+* The underlying library used by SmallRye FT implementation is Hystrix which is currently in "maintenance" mode and a final
+  release of the project (version 1.5.18) has been released. The discussion about replacing the library is ongoing.
+* WildFly's standard management model will be configuration-less
+** all configuration is done through MicroProfile Config specification
+** given the maintenance problem above it is not desirable to expose a management model which will change in near future as concepts are not cleanly mappable to a different underlying library
+* Support for plurality of isolated deployments.
```
rsvoboda: What exactly is the problem with multiple deployments ?

From end user perspective I don't see any reason why I should be limited to just one deployment when using MP FT.

microprofile/WFLY-12590_Integrate_MicroProfile_Fault_Tolerance_2.0.adoc
```diff
@@ -0,0 +1,102 @@
...
+
+== Non-Requirements
+
+* N/A
+
+== Implementation Plan
+
+* Add dependencies on `smallrye-fault-tolerance` and `microprofile-fault-tolerance-api` artifacts.
```
rsvoboda: I assume 2 new modules. Am I correct ?

microprofile/WFLY-12590_Integrate_MicroProfile_Fault_Tolerance_2.0.adoc
```diff
@@ -0,0 +1,102 @@
...
+* Add a new `microprofile-fault-tolerance-smallrye` extension to provide Fault Tolerance APIs support for deployments.
+* Add configuration-less MicroProfile Fault Tolerance subsystem.
+* Support configuration of Hystrix from MicroProfile Config specifications. The configuration is queried for Hystrix properties which are then applied through Archaius configuration mechanism.
```
rsvoboda: All Hystrix configuration properties or just subset of properties is going to be enabled?

Motivation: Some of the properties may not be relevant for application server use-case. So we have smaller area to cover if we can limit enabled properties.

microprofile/WFLY-12590_Integrate_MicroProfile_Fault_Tolerance_2.0.adoc
```diff
@@ -0,0 +1,102 @@
...
+== Test Plan
+
+* The `smallrye-fault-tolerance` component is passing the MicroProfile Fault Tolerance TCK during its release process.
+* WildFly integration test suite will be enhanced with tests that deploys Java archives which uses the MicroProfile Fault Tolerance API.
+
+== Community Documentation
+
+The feature will be documented in WildFly Administration Guide in a new MicroProfile Fault Tolerance section.
+Since WildFly integration does not impose any special requirements and subsystem configuration, this section will just contain how to configure Hystrix and link to MP FT specification.
+This section will also describe migration path from Thorntail primarily regarding Hystrix configuration.
```
rsvoboda: Will be TT properties supported too ?

https://docs.thorntail.io/2.5.0.Final/#_hystrix

microprofile/WFLY-12590_Integrate_MicroProfile_Fault_Tolerance_2.0.adoc
```diff
@@ -0,0 +1,102 @@
...
+Integrate support for Eclipse MicroProfile (MP) Fault Tolerance (FT) for applications deployed in WildFly.
+This proposal pertains to MP FT version 2.0 given the timing of this RFE.
+The integration is based on SmallRye project, specifically `smallrye-fault-tolerance` module.
+The `smallrye-fault-tolerance` module is already in use by projects https://thorntail.io/[Thorntail] (WildFly Swarm) and https://quarkus.io/[Quarkus].
+
+MicroProfile Fault Tolerance specification simplifies the challenges brought by the distributed nature of microservices.
+This is because the communication with external systems is inherently unreliable.
+This increases demand on resiliency of applications.
+
+The patterns brought by this specification are annotations such as `@Timeout`, `@Retry`, `@Fallback`, `@CircuitBreaker`, `@Bulkhead` and `@Asynchronous`.
```
rsvoboda: MP FT is bound to CDI, any plans to have these annotations available for things like EJBs, remoting, persistence, anything_else_present_in_EE8_or_WF?

microprofile/WFLY-12590_Integrate_MicroProfile_Fault_Tolerance_2.0.adoc
```diff
@@ -0,0 +1,102 @@
...
+Not interested in our integration per se, but projects that have already integrated MP FT:
+
+* https://thorntail.io/[Thorntail] has integrated MicroProfile Fault Tolerance as a https://github.com/thorntail/thorntail/tree/master/fractions/microprofile/microprofile-fault-tolerance[fraction].
+* https://quarkus.io/[Quarkus] has integrated MicroProfile Fault Tolerance as an https://github.com/quarkusio/quarkus/tree/master/extensions/smallrye-fault-tolerance[extension].
+
+== Requirements
+
+* Support Eclipse MicroProfile Fault Tolerance version 2.0 for all applications deployed in WildFly.
+* [WILL BE CLARIFIED] The extension is either added to all configurations provided by WildFly (standalone and domain mode)
+  or provided as a layer in a separate build profile.
```
bstansberry: This is not an either/or thing.

We should have a layer for all of these MP specs so users can readily add/remove them from a server provisioned from a set of layers.

Whether this subsystem belongs in our various standalone[-xxx].xml variants or the analogous domain.xml profiles depends on whether it makes sense for each. Whether it's a fit for the basic use case of the profile (seems like it is) and whether there is a downside to having it be present OOTB. Fro OpenAPI there's a downside (multiple deployment issues; exposes an endpoint the user may not want.) For OpenTracing there was a reported perf downside, with no benefit if nothing is consuming the traces. Any downside for FT? I'd think it wouldn't have any impact if the deployment doesn't use the annotations.

microprofile/WFLY-12590_Integrate_MicroProfile_Fault_Tolerance_2.0.adoc
bstansberry: If it's just a missing implementation that's a perfectly valid reason.

But this restriction argues against having this subsystem in the existing standard profiles we ship.

microprofile/WFLY-12590_Integrate_MicroProfile_Fault_Tolerance_2.0.adoc
bstansberry: The management API of the subsystem (whether implemented via MP Config or otherwise) needs to be spelled out in this doc.

microprofile/WFLY-12590_Integrate_MicroProfile_Fault_Tolerance_2.0.adoc
rhusar: This will be figured out in the coming weeks. I think Bilge is back this week so this should move quicker now.

microprofile/WFLY-12590_Integrate_MicroProfile_Fault_Tolerance_2.0.adoc
rhusar: It looks as though this will be driven by separate async RFE.

microprofile/WFLY-12590_Integrate_MicroProfile_Fault_Tolerance_2.0.adoc
rhusar: I believe this is should not be a restriction from the spec perspective. However this is a restriction from the (1) smallrye implementation and (2) underlying Hystrix library. The respective crucial problems here are (1) on undeployment smallrye resets the hystrix runtime and (2) on deployment hystrix configuration and entire runtime for that matter is singleton so a new deployment cannot bring additional nor different configuration.

It appears as @bstansberry is saying, if we will be rigorous about these, we might elect not to include this in the standard profiles.

microprofile/WFLY-12590_Integrate_MicroProfile_Fault_Tolerance_2.0.adoc
rhusar: From the deployment dependencies perspective that's correct. From perspective of the application server there are many new modules (hystrix, commons-config, archies, rxjava v1, etc).

microprofile/WFLY-12590_Integrate_MicroProfile_Fault_Tolerance_2.0.adoc
rhusar: > All Hystrix configuration properties or just subset of properties is going to be enabled?

I don't see an added value in restricting properties that are applied moreover not sure if thats even viable. Neither of the projects integrating this impose any limits.

> The management API of the subsystem (whether implemented via MP Config or otherwise) needs to be spelled out in this doc. I'm not talking about the spec defined MP Config properties.

@bstansberry I am not sure what you mean here – could you rephrase your comment? There will be no configuration in the subsystem.

microprofile/WFLY-12590_Integrate_MicroProfile_Fault_Tolerance_2.0.adoc
rhusar: That's not a plan – we have never supported configuration of different application servers/runtimes in WildFly.

microprofile/WFLY-12590_Integrate_MicroProfile_Fault_Tolerance_2.0.adoc
rhusar: > @rhusar I have another question to this topic. Does it makes sense to test MP FT which package inside EAR/JAR (not just WAR)? Is this type of deployment to be supported?
Yes.

rhusar: > @rhusar @kenfinnigan thanks guys for clarification. I see your point. This is important for completing test plan for this feature whether there should be mixed EE and MP test scenarios. I think this is more general question. Is there a place to bring this up and having decided for the whole MP for WF?

To add to the above, we give users tools but if they elect to shoot themselves in the foot, we cannot do anything about it. This is case for any API really.

microprofile/WFLY-12590_Integrate_MicroProfile_Fault_Tolerance_2.0.adoc
rhusar: @rsvoboda  See this discussion https://github.com/wildfly/wildfly-proposals/pull/250/files/e5f39102374cecc3e30cc6f8a52e18ca90b8fcb7#r335472725

microprofile/WFLY-12590_Integrate_MicroProfile_Fault_Tolerance_2.0.adoc
rhusar: You are right, I will correct the wording now.

The only downside I can think of is just the added 2 modules to CL which is negligible and possible disallowment of multiple deployments (but this should only pertain to deployments which do make use of FT).

microprofile/WFLY-12590_Integrate_MicroProfile_Fault_Tolerance_2.0.adoc
rhusar: Thanks, will add.

microprofile/WFLY-12590_Integrate_MicroProfile_Fault_Tolerance_2.0.adoc
rhusar: Updated.

microprofile/WFLY-12590_Integrate_MicroProfile_Fault_Tolerance_2.0.adoc
rhusar: Added an item for integration tests.

microprofile/WFLY-12590_Integrate_MicroProfile_Fault_Tolerance_2.0.adoc
rhusar: Corrected the statement in the proposal.

microprofile/WFLY-12590_Integrate_MicroProfile_Fault_Tolerance_2.0.adoc
bstansberry: @rhusar WildFly provides a configuration contract. For these MP specs we are moving toward using MP Config to implement configuration but that doesn't mean it's all uncontrolled and we have random sets of config properties that change from release to release. At a minimum we need to document what we support and note that things may change.

By "I'm not talking about the spec defined MP Config properties" I mean that if the MP FT spec says you use config property mp.ft.foo to configure your level of fooness then there's no need to document that here; it's documented in the spec. It's the custom configuration stuff that needs to be documented.

Re: enabling properties, understood that it may not be possible to turn things off, but if something is not supported that can be documented.

microprofile/WFLY-12590_Integrate_MicroProfile_Fault_Tolerance_2.0.adoc
```diff
@@ -0,0 +1,103 @@
...
+* Create a MicroProfile Fault Tolerance https://github.com/wildfly/quickstart[quickstart] – delivered via separate RFE above of scope of this one.
+
+Items that are currently only being taken into consideration with next version of this feature:
+
+* The underlying library used by SmallRye FT implementation is Hystrix which is currently in "maintenance" mode and a final
+  release of the project (version 1.5.18) has been released. The discussion about replacing the library is ongoing.
+* WildFly's standard management model will be configuration-less
+** all configuration is done through MicroProfile Config specification
+** given the maintenance problem above it is not desirable to expose a management model which will change in near future as concepts are not cleanly mappable to a different underlying library
+* Support for plurality of isolated deployments – not viable with the current implementation, will be possible once smallrye is based on different library than Hystrix.
```
bstansberry: @rhusar Here or in the implementation plan section please describe what will happen in this first version of the feature if:
1) User deploys > 1 deployment.
2) User deploys an ear.

rhusar: Will do once I figure it out.

microprofile/WFLY-12590_Integrate_MicroProfile_Fault_Tolerance_2.0.adoc
rhusar: I have added a warning that these might change in the future in the documentation. Also, the properties that we expect people to tune are described in the docs.

microprofile/WFLY-12590_Integrate_MicroProfile_Fault_Tolerance_2.0.adoc
mnovak1: @rhusar could you update this section, please? We need to define th behaviour for the two scenarios @bstansberry has mentioned in [1].

[1] https://github.com/wildfly/wildfly-proposals/pull/250#discussion_r342770835

mnovak1: @maeste hi, as @rhusar is OOO, could you take a look at above point [1] and provide your pov, please? We need some decision to finish this analyses and test plan.

I had discussion on chat with @rhusar about this topic and major limitation was that hystrix is singleton. So if 2 MP FT deployments are present then they must use the same instance of hystrix. So 2nd deployment cannot apply different configuration for hystrix and must use instance configured by 1st deployment.

We've discussed following options:
1. * to have only single deployment and provide multideployment later (with new MP FT implementation)
2. * to have multi-deployment but any config from microprofile-config.properties would not be applied (just what is set system/env variables and MP config subsystem)
3. * to change hystrix not to be singleton (not sure if it's possible)

There might be other options which we did not discuss. 

I've nacked option 2) as configuration using microprofile-config.properties is must have. 

Which way would you recommend to proceed?

[1] https://github.com/wildfly/wildfly-proposals/pull/250#discussion_r342770835

rhusar: If you look at the current wildfly repo PR, its allowing first deployment to configure and issue the warning on subsequent deployments.

rhusar: 3) would require maintaining our own fork and fixes to SR. Thus its 1).

mnovak1: thanks @rhusar so in case there are 2 MP FT deployments and server is started then it's race, right?

tobiasstadler:
@rhusar Will this be part of EAP 7.3?

rhusar:
@tobiasstadler Please follow https://issues.jboss.org/browse/WFLY-12590 for the schedule.

microprofile/WFLY-12590_Integrate_MicroProfile_Fault_Tolerance_2.0.adoc
rhusar: Moved to requirements describing expected behavior.

microprofile/WFLY-12590_Integrate_MicroProfile_Fault_Tolerance_2.0.adoc
```diff
@@ -0,0 +1,105 @@
...
+Not interested in our integration per se, but projects that have already integrated MP FT:
+
+* https://thorntail.io/[Thorntail] has integrated MicroProfile Fault Tolerance as a https://github.com/thorntail/thorntail/tree/master/fractions/microprofile/microprofile-fault-tolerance[fraction].
+* https://quarkus.io/[Quarkus] has integrated MicroProfile Fault Tolerance as an https://github.com/quarkusio/quarkus/tree/master/extensions/smallrye-fault-tolerance[extension].
+
+== Requirements
+
+* Support Eclipse MicroProfile Fault Tolerance version 2.0 for all applications deployed in WildFly.
+* The extension will be provided as Galleon layer called `microprofile-fault-tolerance`.
+* [WILL BE CLARIFIED] The extension will be added to relevant default configurations provided by WildFly (standalone and domain mode).
```
mnovak1: @rhusar Here should be mentioned that MP FT extenstion will not be part of default standlaone/domain config. Could you update it, please?

rhusar: Updated.

PR #251, opened 2019-10-11
WFLY-12313 Integrate MicroProfile OpenAPI 1.1

pferraro:
https://issues.jboss.org/browse/WFLY-12313

### DIFF
--- a/microprofile/WFLY-12313_MicroProfile_OpenAPI_1.1.adoc
+++ b/microprofile/WFLY-12313_MicroProfile_OpenAPI_1.1.adoc
@@ -83,7 +83,7 @@ These integrations will heavily inform, though not dictate, the corresponding in
 *** Additionally, if the Undertow server/host associated with the application does not define an HTTP endpoint (as required by the spec), we should log a WARN message.
 ** The MicroProfile OpenAPI specification suggests, but does not require, the use of a `format=JSON` or `format=YAML` parameter for the `/openapi` endpoint, used to define content type of the resulting document, in lieu of an `Accept` header.
 *** This is very reasonable, and is simple enough to do.  See: https://download.eclipse.org/microprofile/microprofile-open-api-1.1.2/microprofile-openapi-spec.html#_query_parameters
-* Use a `mp.openapi.extensions.disable` property to disable OpenAPI documentation globally (via mp-config subsystem) or per-application (e.g. /META-INF/mp-config.properties).
+* Use a `mp.openapi.extensions.enabled` property to disable OpenAPI documentation globally (via mp-config subsystem) or per-application (e.g. /META-INF/mp-config.properties).
 ** Provides a parameterizable means of disabling OpenAPI documentation in specific environments (e.g. prod vs dev)
 ** Provides a mechanism to control which application associated with a given virtual host should generate an MicroProfile OpenAPI model.
 * Use a `mp.openapi.extensions.path` property to customize the path of the OpenAPI endpoint.
@@ -110,7 +110,7 @@ These integrations will heavily inform, though not dictate, the corresponding in
 * Add the microprofile-tck tests to the MicroProfile TCK testsuite in WildFly.
 * Integration tests should cover vendor specific behavior, not covered by the TCK tests, including:
 ** Ensuring that services deployed using a non-root context path are documented with the correct `server` elements.
-** Any multi-deployment or multi-module deployment behavior, e.g. `mp.openapi.extensions.disable`
+** Any multi-deployment or multi-module deployment behavior, e.g. `mp.openapi.extensions.enabled`
 ** Alternate OpenAPI endpoint registration according to `mp.openapi.extensions.path`
 ** Fetching JSON vs YAML documentation via `format` parameter
 ** Loading JSON vs YAML static OpenAPI documentation
--- a/microprofile/WFLY-12313_MicroProfile_OpenAPI_1.1.adoc
+++ b/microprofile/WFLY-12313_MicroProfile_OpenAPI_1.1.adoc
@@ -90,6 +90,9 @@ These integrations will heavily inform, though not dictate, the corresponding in
 ** Provides a means for generating OpenAPI documentation for multiple applications associated with the same virtual host.
 ** Potentially useful to support legacy OpenAPI tooling (e.g. `/swagger`)
 ** Analogous to the `quarkus.smallrye-openapi.path` property in Quarkus.
+* Use a `mp.openapi.extensions.relative` property to configure whether auto-generated Server records should be absolute or relative.
+** When relative, the generated record will contain the context path of the deployment
+** When absolute, generated records will include all hosts/ports at which the given deployment is accessible.
 * Allow optional authentication/authorization constraints for the `/openapi` endpoint.
 ** Consider for a future enhancement
 * Add an optional endpoint for browsing OpenAPI documentation via SwaggerUI (e.g. https://quarkus.io/guides/openapi-swaggerui-guide).
--- a/microprofile/WFLY-12313_MicroProfile_OpenAPI_1.1.adoc
+++ b/microprofile/WFLY-12313_MicroProfile_OpenAPI_1.1.adoc
@@ -57,7 +57,7 @@ These integrations will heavily inform, though not dictate, the corresponding in
 *** io.smallrye:smallrye-open-api
 *** org.eclipse.microprofile.openapi:microprofile-openapi-api
 *** com.fasterxml.jackson.dataformat:jackson-dataformat-yaml
-** Which feature packs will use this layer is still TBD.
+** Subsystem will be included in the default standalone-openshift.xml feature pack.
 * The microprofile-openapi spec module must be exposed to web deployments
 * The `/openapi` endpoint shall be registered relative to the root of the virtual host associated with the deployment.
 ** N.B. The `/openapi` endpoint is intended for a public interface, not a management interface, as with the `/health` and `/metrics` endpoints.
--- a/microprofile/WFLY-12313_MicroProfile_OpenAPI_1.1.adoc
+++ b/microprofile/WFLY-12313_MicroProfile_OpenAPI_1.1.adoc
@@ -38,9 +38,11 @@ An MicroProfile OpenAPI subsystem exposes an HTTP GET `/openapi` endpoint that r
 The SmallRye implementation of the MicroProfile OpenAPI specification is already integrated by the following projects:
 
 * https://thorntail.io/[Thorntail] via the https://github.com/thorntail/thorntail/tree/master/fractions/microprofile/microprofile-openapi[microprofile-openapi fraction].
+** General migration documentation from ThornTail is tracked by https://issues.jboss.org/browse/WFLY-12749[WFLY-12749]
 * https://quarkus.io/[Quarkus] via the https://github.com/quarkusio/quarkus/tree/master/extensions/smallrye-openapi[smallrye-openapi extension].
 
 These integrations will heavily inform, though not dictate, the corresponding integration into WildFly.
+
 
 === Other Interested Projects
 
@@ -73,7 +75,7 @@ These integrations will heavily inform, though not dictate, the corresponding in
 * Optional features of the specification:
 ** Multiple deployments:
 *** If a container allows multiple deployments, the specification expects the `/openapi` endpoint to return the logical union of the documented services for all applications in the runtime. See: https://download.eclipse.org/microprofile/microprofile-open-api-1.1.2/microprofile-openapi-spec.html#_multiple_applications
-**** Initially, we will limit multiple deployment support to a single deployment per host (for a given endpoint).
+**** Initially, we will limit multiple deployment support to a single deployment per host (for a given endpoint). i.e. each deployment, for which the OpenAPI endpoint is enabled, must either use a distinct host, or a distinct endpoint path (if a given virtual host is associated with multiple deployments).
 **** Merging (and more interestingly, unmerging on undeploy) OpenAPI models for all applications deployed to a given host.
 ***** Consider for a future enhancement
 ** HTTPS transport for the `/openapi` endpoint.
@@ -81,7 +83,7 @@ These integrations will heavily inform, though not dictate, the corresponding in
 *** Additionally, if the Undertow server/host associated with the application does not define an HTTP endpoint (as required by the spec), we should log a WARN message.
 ** The MicroProfile OpenAPI specification suggests, but does not require, the use of a `format=JSON` or `format=YAML` parameter for the `/openapi` endpoint, used to define content type of the resulting document, in lieu of an `Accept` header.
 *** This is very reasonable, and is simple enough to do.  See: https://download.eclipse.org/microprofile/microprofile-open-api-1.1.2/microprofile-openapi-spec.html#_query_parameters
-* Use a `mp.openapi.extensions.disable` property to disable OpenAPI documentation globally or per-application.
+* Use a `mp.openapi.extensions.disable` property to disable OpenAPI documentation globally (via mp-config subsystem) or per-application (e.g. /META-INF/mp-config.properties).
 ** Provides a parameterizable means of disabling OpenAPI documentation in specific environments (e.g. prod vs dev)
 ** Provides a mechanism to control which application associated with a given virtual host should generate an MicroProfile OpenAPI model.
 * Use a `mp.openapi.extensions.path` property to customize the path of the OpenAPI endpoint.
--- a/microprofile/WFLY-12313_MicroProfile_OpenAPI_1.1.adoc
+++ b/microprofile/WFLY-12313_MicroProfile_OpenAPI_1.1.adoc
@@ -90,9 +90,9 @@ These integrations will heavily inform, though not dictate, the corresponding in
 ** Provides a means for generating OpenAPI documentation for multiple applications associated with the same virtual host.
 ** Potentially useful to support legacy OpenAPI tooling (e.g. `/swagger`)
 ** Analogous to the `quarkus.smallrye-openapi.path` property in Quarkus.
-* Use a `mp.openapi.extensions.relative` property to configure whether auto-generated Server records should be absolute or relative.
-** When relative, the generated record will contain the context path of the deployment
-** When absolute, generated records will include all hosts/ports at which the given deployment is accessible.
+* Use a `mp.openapi.extensions.relative` property to configure whether auto-generated Server records should be absolute or relative.  At a minimum, Server records are necessary to ensure, in the presence of a non-root context path, that consumers of an OpenAPI can construct valid URLs to REST services relative to the host of the OpenAPI endpoint.  See: https://download.eclipse.org/microprofile/microprofile-open-api-1.1.2/microprofile-openapi-spec.html#_context_root_behavior
+** When relative (i.e. true), the generated record will contain the context path of the deployment
+** When absolute (i.e. false), generated records will include all hosts/ports at which the given deployment is accessible.
 * Allow optional authentication/authorization constraints for the `/openapi` endpoint.
 ** Consider for a future enhancement
 * Add an optional endpoint for browsing OpenAPI documentation via SwaggerUI (e.g. https://quarkus.io/guides/openapi-swaggerui-guide).
--- a/microprofile/WFLY-12313_MicroProfile_OpenAPI_1.1.adoc
+++ b/microprofile/WFLY-12313_MicroProfile_OpenAPI_1.1.adoc
@@ -1,0 +1,126 @@
+= Integrate MicroProfile OpenAPI 1.1
+:author:            Paul Ferraro
+:email:             paul.ferraro@redhat.com
+:toc:               left
+:icons:             font
+:idprefix:
+:idseparator:       -
+
+== Overview
+
+This enhancement proposes to add support for https://download.eclipse.org/microprofile/microprofile-open-api-1.1.2/microprofile-openapi-spec.html[Eclipse MicroProfile OpenAPI 1.1] to WildFly by integrating the implementation from https://github.com/smallrye/smallrye-open-api[SmallRye].
+This spec provides a unified Java API for producing documentation that complies with the https://github.com/OAI/OpenAPI-Specification/blob/3.0.2/versions/3.0.2.md[OpenAPI v3 specification].
+
+An MicroProfile OpenAPI subsystem exposes an HTTP GET `/openapi` endpoint that returns an OpenAPI v3 compliant document (in YAML or JSON format) containing documentation of the RESTful services of an application.
+
+== Issue Metadata
+
+=== Issue
+
+* https://issues.jboss.org/browse/WFLY-12313
+
+=== Dev Contacts
+
+* mailto:{email}[{author}]
+
+=== QE Contacts
+
+* mailto:fburzigo@redhat.com[Fabio Burzigotti]
+
+=== Testing By
+
+[ ] Engineering
+
+[X] QE
+
+=== Affected Projects or Components
+
+The SmallRye implementation of the MicroProfile OpenAPI specification is already integrated by the following projects:
+
+* https://thorntail.io/[Thorntail] via the https://github.com/thorntail/thorntail/tree/master/fractions/microprofile/microprofile-openapi[microprofile-openapi fraction].
+* https://quarkus.io/[Quarkus] via the https://github.com/quarkusio/quarkus/tree/master/extensions/smallrye-openapi[smallrye-openapi extension].
+
+These integrations will heavily inform, though not dictate, the corresponding integration into WildFly.
+
+=== Other Interested Projects
+
+== Requirements
+
+=== Hard Requirements
+
+* Create an MicroProfile OpenAPI galleon layer containing the MP Open API feature encapsulated by a new MicroProfile OpenAPI subsystem.
+** Should depend on MicroProfile Config and JAX-RS (and their dependencies)
+** N.B. New modules to be included in WF:
+*** MicroProfile OpenAPI subsystem module
+*** io.smallrye:smallrye-open-api
+*** org.eclipse.microprofile.openapi:microprofile-openapi-api
+*** com.fasterxml.jackson.dataformat:jackson-dataformat-yaml
+** Which feature packs will use this layer is still TBD.
+* The microprofile-openapi spec module must be exposed to web deployments
+* The `/openapi` endpoint shall be registered relative to the root of the virtual host associated with the deployment.
+** N.B. The `/openapi` endpoint is intended for a public interface, not a management interface, as with the `/health` and `/metrics` endpoints.
+** This places a superficial constraint of a single application per host (see subsequent section re: multiple deployments)
+** In the event a web deployment uses a given host for which another deployment has already registered an `/openapi` endpoint, we will:
+... Log a warning.
+... Skip OpenAPI model generation for that deployment.
+* All WildFly/SmallRye specific configuration will be obtained via MicroProfile Config using the properties prefixed with `mp.openapi.extensions`.
+** Consequently, the microprofile-openapi subsystem should have an empty management model.
+* Support YAML and JSON for both input (static/pre-generated OpenAPI document) and output formats.
+* MicroProfile OpenAPI TCK must pass
+
+=== Nice-to-Have Requirements
+
+* Optional features of the specification:
+** Multiple deployments:
+*** If a container allows multiple deployments, the specification expects the `/openapi` endpoint to return the logical union of the documented services for all applications in the runtime. See: https://download.eclipse.org/microprofile/microprofile-open-api-1.1.2/microprofile-openapi-spec.html#_multiple_applications
+**** Initially, we will limit multiple deployment support to a single deployment per host (for a given endpoint).
+**** Merging (and more interestingly, unmerging on undeploy) OpenAPI models for all applications deployed to a given host.
+***** Consider for a future enhancement
+** HTTPS transport for the `/openapi` endpoint.
+*** If the server/host of a given application defines an HTTPS listener, we will allow it, but not require it.
+*** Additionally, if the Undertow server/host associated with the application does not define an HTTP endpoint (as required by the spec), we should log a WARN message.
+** The MicroProfile OpenAPI specification suggests, but does not require, the use of a `format=JSON` or `format=YAML` parameter for the `/openapi` endpoint, used to define content type of the resulting document, in lieu of an `Accept` header.
+*** This is very reasonable, and is simple enough to do.  See: https://download.eclipse.org/microprofile/microprofile-open-api-1.1.2/microprofile-openapi-spec.html#_query_parameters
+* Use a `mp.openapi.extensions.disable` property to disable OpenAPI documentation globally or per-application.
+** Provides a parameterizable means of disabling OpenAPI documentation in specific environments (e.g. prod vs dev)
+** Provides a mechanism to control which application associated with a given virtual host should generate an MicroProfile OpenAPI model.
+* Use a `mp.openapi.extensions.path` property to customize the path of the OpenAPI endpoint.
+** Provides a means for generating OpenAPI documentation for multiple applications associated with the same virtual host.
+** Potentially useful to support legacy OpenAPI tooling (e.g. `/swagger`)
+** Analogous to the `quarkus.smallrye-openapi.path` property in Quarkus.
+* Allow optional authentication/authorization constraints for the `/openapi` endpoint.
+** Consider for a future enhancement
+* Add an optional endpoint for browsing OpenAPI documentation via SwaggerUI (e.g. https://quarkus.io/guides/openapi-swaggerui-guide).
+** Consider for a future enhancement
+
+=== Non-Requirements
+
+* The MicroProfile OpenAPI specification does not require support for i18n of the OpenAPI document (e.g. via an Accept-Language header), though this would be a reasonable subsequent enhancement.  See: https://download.eclipse.org/microprofile/microprofile-open-api-1.1.2/microprofile-openapi-spec.html#_internationalization
+* The MicroProfile OpenAPI specification does not require support for validating the generated OpenAPI model, though this is a reasonable subsequent enhancement. A number of third-party validator exist and could be leveraged for such a feature.  See: https://download.eclipse.org/microprofile/microprofile-open-api-1.1.2/microprofile-openapi-spec.html#_validation
+* The MicroProfile OpenAPI specification does not require support for https://www.w3.org/TR/cors/[Cross-Origin Resource Sharing (CORS)].  See: https://download.eclipse.org/microprofile/microprofile-open-api-1.1.2/microprofile-openapi-spec.html#_cross_origin_resource_sharing_cors
+
+== Test Plan
+
+* Basic subsystem unit tests
+* Add the microprofile-tck tests to the MicroProfile TCK testsuite in WildFly.
+* Integration tests should cover vendor specific behavior, not covered by the TCK tests, including:
+** Ensuring that services deployed using a non-root context path are documented with the correct `server` elements.
+** Any multi-deployment or multi-module deployment behavior, e.g. `mp.openapi.extensions.disable`
+** Alternate OpenAPI endpoint registration according to `mp.openapi.extensions.path`
+** Fetching JSON vs YAML documentation via `format` parameter
+** Loading JSON vs YAML static OpenAPI documentation
+
+== Community Documentation
+
+* Document how to add the microprofile-openapi subsystem to WildFly
+* Document any WildFly specific `mp.openapi.extensions.*` configuration.
+* Summarize how a user would:
+** Provide a custom bootstrap of the OpenApiDocument model
+** Contribute to the OpenApiDocument model via OpenAPI annotations.
+** Register an OASFilter.
+* Add quickstart application to WildFly quickstarts.
+
+== Release Note Content
+
+Add support for standardized documentation of RESTful web services via the MicroProfile OpenAPI 1.1 specification.
+


microprofile/WFLY-12313_MicroProfile_OpenAPI_1.1.adoc
```diff
@@ -0,0 +1,97 @@
...
+=== Non-Requirements
+
+* The MP OpenAPI specification does not require support for i18n of the OpenAPI document (e.g. via an Accept-Language header), though this would be a reasonable subsequent enhancement.  See: https://github.com/eclipse/microprofile-open-api/blob/mp-openapi-1.1.1/spec/src/main/asciidoc/microprofile-openapi-spec.adoc#71-internationalization
+* The MP OpenAPI specification does not require support for validating the generated OpenAPI model, though this is a reasonable subsequent enhancement. A number of third-party validator exist and could be leveraged for such a feature.  See: https://github.com/eclipse/microprofile-open-api/blob/mp-openapi-1.1.1/spec/src/main/asciidoc/microprofile-openapi-spec.adoc#72-validation
+* The MP OpenAPI specification does not require support for https://www.w3.org/TR/cors/[Cross-Origin Resource Sharing (CORS)].  See: https://github.com/eclipse/microprofile-open-api/blob/mp-openapi-1.1.1/spec/src/main/asciidoc/microprofile-openapi-spec.adoc#73-cross-origin-resource-sharing-cors
+
+== Test Plan
+
+* Basic subsystem unit tests
+* Add the microprofile-tck tests to the MicroProfile TCK testsuite in WildFly.
```
fabiobrz: Just checking here: are you meaning something similar to https://github.com/microservices-api/mp-openapi-tck-runner?

microprofile/WFLY-12313_MicroProfile_OpenAPI_1.1.adoc
```diff
@@ -0,0 +1,97 @@
...
+* Integration tests should cover vendor specific behavior, not covered by the TCK tests, including:
+** Ensuring that services deployed using a non-root context path are documented with the correct `server` elements.
+** Any multi-deployment or multi-module deployment behavior
```
fabiobrz: This would be subject to the Multiple deployments "Nice-to-have" requirement, right?

microprofile/WFLY-12313_MicroProfile_OpenAPI_1.1.adoc
```diff
@@ -0,0 +1,97 @@
...
+* The MP OpenAPI specification does not require support for validating the generated OpenAPI model, though this is a reasonable subsequent enhancement. A number of third-party validator exist and could be leveraged for such a feature.  See: https://github.com/eclipse/microprofile-open-api/blob/mp-openapi-1.1.1/spec/src/main/asciidoc/microprofile-openapi-spec.adoc#72-validation
+* The MP OpenAPI specification does not require support for https://www.w3.org/TR/cors/[Cross-Origin Resource Sharing (CORS)].  See: https://github.com/eclipse/microprofile-open-api/blob/mp-openapi-1.1.1/spec/src/main/asciidoc/microprofile-openapi-spec.adoc#73-cross-origin-resource-sharing-cors
+
+== Test Plan
+
+* Basic subsystem unit tests
+* Add the microprofile-tck tests to the MicroProfile TCK testsuite in WildFly.
+* Integration tests should cover vendor specific behavior, not covered by the TCK tests, including:
```
fabiobrz: Are these category of tests going to be provided by the Wildfly testsuite?

microprofile/WFLY-12313_MicroProfile_OpenAPI_1.1.adoc
```diff
@@ -0,0 +1,97 @@
...
+:icons:             font
+:idprefix:
+:idseparator:       -
+
+== Overview
+
+This enhancement proposes to add support for https://github.com/eclipse/microprofile-open-api/tree/1.1.2[Eclipse MicroProfile OpenAPI 1.1] to WildFly by integrating the implementation from https://github.com/smallrye/smallrye-open-api[SmallRye].
+This spec provides a unified Java API for producing documentation that complies with the https://github.com/OAI/OpenAPI-Specification/blob/3.0.2/versions/3.0.2.md[OpenAPI v3 specification].
+
+An MP OpenAPI subsystem exposes an HTTP GET `/openapi` endpoint that returns an OpenAPI v3 compliant document (in YAML or JSON format) containing documention of the RESTful services of an application.
```
fabiobrz: "documention" -> "documentation"?

microprofile/WFLY-12313_MicroProfile_OpenAPI_1.1.adoc
```diff
@@ -0,0 +1,97 @@
...
+These integrations will heavily inform, though not dictate, the corresponding integration into WildFly.
+
+=== Other Interested Projects
+
+== Requirements
+
+=== Hard Requirements
+
+* The microprofile-openapi spec module must be exposed to deployments
+* The `/openapi` endpoint shall be registered with the Undertow server/host associated with a deployment.
```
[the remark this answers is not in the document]
kenfinnigan: It needs to be one `/openapi` per deployment.

There is currently no mechanism in the MP OpenAPI spec for aggregating multiple documents to provide a single `/openapi` for multiple deployments

microprofile/WFLY-12313_MicroProfile_OpenAPI_1.1.adoc
```diff
@@ -0,0 +1,97 @@
...
+* All WildFly/SmallRye specific configuration will be obtained via MP Config using the properties prefixed with `mp.openapi.extensions`.
+** Consequently, the microprofile-openapi subsystem should have an empty management model.
+* Support YAML and JSON for both input (static/pre-generated OpenAPI document) and output formats.
+* MicroProfile OpenAPI TCK must pass
+
+=== Nice-to-Have Requirements
+
+* Optional features of the specification:
+** Multiple deployments:
+*** The specification expects the `/openapi` endpoint to return the logical union of the documented services for all applications in the runtime. See: https://github.com/eclipse/microprofile-open-api/blob/mp-openapi-1.1.1/spec/src/main/asciidoc/microprofile-openapi-spec.adoc#55-multiple-applications
```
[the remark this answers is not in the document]
kenfinnigan: That's simply the spec saying if you do it, it needs to aggregate.

I don't think it's a good idea to aggregate

microprofile/WFLY-12313_MicroProfile_OpenAPI_1.1.adoc
```diff
@@ -0,0 +1,97 @@
...
+[X] QE
+
+=== Affected Projects or Components
+
+The SmallRye implementation of the MicroProfile Open API specification is already integrated by the following projects:
+
+* https://thorntail.io/[Thorntail] via the https://github.com/thorntail/thorntail/tree/master/fractions/microprofile/microprofile-openapi[microprofile-openapi fraction].
+* https://quarkus.io/[Quarkus] via the https://github.com/quarkusio/quarkus/tree/master/extensions/smallrye-openapi[smallrye-openapi extension].
+
+These integrations will heavily inform, though not dictate, the corresponding integration into WildFly.
```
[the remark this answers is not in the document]
pferraro: > it would be quite useful to mention how the projects listed here would be affected from the end user point of view

I did not imagine they would be affected.

> and also the chance for users to migrate their existing apps from TT to Wildfly.

Migration from TT should be trivial, as TT equivalent configuration won't touch the more ambiguous spec behavior mentioned below.

> Probably the community docs could also cover a typical migration scenario with just some high-level detail

We don't typically discuss migration in the community docs - but it'll need to be mentioned somewhere.  I'll find out the correct place for that stuff.

microprofile/WFLY-12313_MicroProfile_OpenAPI_1.1.adoc
```diff
@@ -0,0 +1,97 @@
...
+== Requirements
+
+=== Hard Requirements
+
+* The microprofile-openapi spec module must be exposed to deployments
+* The `/openapi` endpoint shall be registered with the Undertow server/host associated with a deployment.
+* All WildFly/SmallRye specific configuration will be obtained via MP Config using the properties prefixed with `mp.openapi.extensions`.
+** Consequently, the microprofile-openapi subsystem should have an empty management model.
+* Support YAML and JSON for both input (static/pre-generated OpenAPI document) and output formats.
+* MicroProfile OpenAPI TCK must pass
```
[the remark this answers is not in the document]
pferraro: AIUI, which profiles will include microprofile subsystems is still an open question.  I'll keep this updated as that stuff gets sorted out.

microprofile/WFLY-12313_MicroProfile_OpenAPI_1.1.adoc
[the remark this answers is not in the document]
pferraro: We can still have multiple /openapi endpoints (i.e. 1 per host), but they would return the same document.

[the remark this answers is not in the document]
pferraro: @kenfinnigan Thanks for the feedback.  I was a bit surprised that the spec would recommend aggregation.  Producing separate documents does seem much simpler.  Do we know why the specs says this?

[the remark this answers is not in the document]
kenfinnigan: I think it's to define what should happen, as I think OpenLiberty might take this approach, but not entirely sure.

Might be able to determine from commit history

microprofile/WFLY-12313_MicroProfile_OpenAPI_1.1.adoc
```diff
@@ -0,0 +1,97 @@
...
+
+=== Nice-to-Have Requirements
+
+* Optional features of the specification:
+** Multiple deployments:
+*** The specification expects the `/openapi` endpoint to return the logical union of the documented services for all applications in the runtime. See: https://github.com/eclipse/microprofile-open-api/blob/mp-openapi-1.1.1/spec/src/main/asciidoc/microprofile-openapi-spec.adoc#55-multiple-applications
+*** Users might want the ability to disable OpenAPI documentation of a given deployment.
+** HTTPS transport for the `/openapi` endpoint.
+*** If the server/host of a given application defines an HTTPS listener, we will allow it, but not require it.
```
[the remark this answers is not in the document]
pferraro: Correct.

microprofile/WFLY-12313_MicroProfile_OpenAPI_1.1.adoc
```diff
@@ -0,0 +1,97 @@
...
+
+== Community Documentation
+
+* Document how to add the microprofile-openapi subsystem to WildFly
+* Document any WildFly specific `mp.openapi.extensions.*` configuration.
+* Summarize how a user would:
+** Provide a custom bootstrap of the OpenApiDocument model
+** Contribute to the OpenApiDocument model via OpenAPI annotations.
+** Register an OASFilter.
+* Add quickstart application to WildFly quickstarts.
```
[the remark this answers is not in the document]
pferraro: Everything in community documentation is meant to be treated as a requirement, i.e. not optional.

microprofile/WFLY-12313_MicroProfile_OpenAPI_1.1.adoc
pferraro: I mean migrating: https://github.com/thorntail/thorntail/tree/master/testsuite/microprofile-tcks/openapi
to: https://github.com/wildfly/wildfly/tree/master/testsuite/integration/microprofile-tck

microprofile/WFLY-12313_MicroProfile_OpenAPI_1.1.adoc
pferraro: Yes.

microprofile/WFLY-12313_MicroProfile_OpenAPI_1.1.adoc
pferraro: Yes.

microprofile/WFLY-12313_MicroProfile_OpenAPI_1.1.adoc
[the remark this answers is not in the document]
pferraro: > It needs to be one /openapi per deployment.

@kenfinnigan Thank you for clarifying.

In the case where a second application is deployed to the same virtual host as an existing /openapi producing application, we can either:

1. Fail the second deployment
2. Allow the second deployment, but skip document generation process and log a warning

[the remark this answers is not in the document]
kenfinnigan: I'm not a virtual host expert, but if you have two deployments on a single virtual host would that not mean you have endpoints of `vhost/dep1/openapi` and `vhost/dep2/openapi`?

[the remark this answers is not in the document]
pferraro: @kenfinnigan According to https://github.com/eclipse/microprofile-open-api/blob/mp-openapi-1.1.1/spec/src/main/asciidoc/microprofile-openapi-spec.adoc#51-overview

> A fully processed OpenAPI document must be available at the root URL `/openapi`.

This would seem to preclude:
* A deployment with a context path of "/openapi"
* Multiple deployments with distinct context paths sharing the same virtual host, as the locations of their `/openapi` documents would collide.

Did I misunderstand something?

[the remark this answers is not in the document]
kenfinnigan: I think we need to get the spec clarified, as MP Health talks about the endpoint being off the context root of the deployment (https://download.eclipse.org/microprofile/microprofile-health-2.1/microprofile-health-spec.html#_rationale)

I believe the typical understanding is that if you have an application context root of `myapp`, then the endpoint is available at `/myapp/openapi` and not `/openapi`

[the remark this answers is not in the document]
pferraro: @kenfinnigan The MP Open API TCK deploys multiple applications and verifies the OpenAPI document available at the root of the ${test.url}, _irrespective_ of the deployment's default context path.

Additionally, both Thorntail and Quarkus both register the /openapi endpoint at the root of the host, without regard for any configured thorntail.context.path or quarkus.servlet.context-path, respectively.

microprofile/WFLY-12313_MicroProfile_OpenAPI_1.1.adoc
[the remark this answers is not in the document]
fabiobrz: Ok, thanks for clarifying this.

microprofile/WFLY-12313_MicroProfile_OpenAPI_1.1.adoc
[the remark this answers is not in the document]
fabiobrz: Ok, thanks.

microprofile/WFLY-12313_MicroProfile_OpenAPI_1.1.adoc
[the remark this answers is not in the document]
fabiobrz: Shouldn't this optional requirement actually be discussed in the same scope as hard requirement #1? 
I mean that - although I understand that the specs are covering this topic [1] by considering the number of `applications` in the "runtime" - yet, IIUC, typical Wildfly scenarios do allow for multiple applications deployments.
The only case that I can see for this requirement to be optional - given the discussion about requirement #1 - would be to enforce come kind of constraints like the ones you mentioned, see [2].
IMHO all this calls for a more detailed definition of this/these requirements. 
WDYT?

[1]
https://download.eclipse.org/microprofile/microprofile-open-api-1.1.2/microprofile-openapi-spec.html#_multiple_applications

[2]
https://github.com/wildfly/wildfly-proposals/pull/251/files#r335027624

microprofile/WFLY-12313_MicroProfile_OpenAPI_1.1.adoc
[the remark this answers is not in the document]
kenfinnigan: Thorntail and Quarkus working that way are by virtue of them not having more than one application present in a deployment.

Which TCK test deploys two JAX-RS applications into a single deployment?

Might be worth getting @EricWittmann thoughts on this too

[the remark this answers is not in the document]
fabiobrz: > Thorntail and Quarkus working that way are by virtue of them not having more than one application present in a deployment.

That's the point. Wildfly is not assuming the same "one-application" scenario, if I don't go wrong. This is the reason why I've been asking for sharing more detail about this first requirement. Just to have an idea, after reading through the above discussion, it is not clear yet "how many" `/openapi` endpoints there should be, at which level and how would they behave with respect to producing the OpenAPI document in different use cases.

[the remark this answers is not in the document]
pferraro: @kenfinnigan 
> Thorntail and Quarkus working that way are by virtue of them not having more than one application present in a deployment.

I understand that they only support a single application/deployment.  However, whether the Open API endpoint should be relative to the host or to the context path is _still_ relevant to the spec compliance accuracy of both frameworks, regardless of how many applications/deployments the framework supports.

> I believe the typical understanding is that if you have an application context root of myapp, then the endpoint is available at /myapp/openapi and not /openapi

If the above was true, and Thorntail's implementation is correct, then an application deployed to ThornTail using `thorntail.context.path=/mycontext` should have OpenAPI documentation available at: `host:port/mycontext/openapi`  However, this path returns a 404.  Rather, the document is only available at `host:port/openapi`.

The same holds true for Quarkus when using `quarkus.servlet.context-path=/mycontext`

> Which TCK test deploys two JAX-RS applications into a single deployment?

Sorry if I was unclear.  Let me elaborate.

The TCK runs a set of tests against a number of applications, including `airlines.war`, `airlinesReader.war`, `petstore.war`, etc.

Thorntail and Quarkus will, by default, deploy these applications to the root context.  Thus when the TCK looks for the Open API documentation at `host:port/openapi` there are no problems.

However, most multi-deployment servlet containers (including WildFly, Tomcat, etc.) will, by default (i.e. in the absence of additional meta data), deploy these applications using context paths corresponding to the name of the archive.
In WildFly, this means deploying to: `host:port/airlines.war`, `host:port/airlinesReader.war`, `host:port/petstore.war`, etc.

However, the TCK looks for the Open API documentation at ${test.url}/openapi, which is the same for each test application.  While we can certainly add WildFly-specific hooks to the test execution that appends requisite meta data necessary to change the context path of the test applications to use the root context - to me, the wording of the spec, as well as the behavior described by the TCK, suggests that the `/openapi` endpoint should be relative to the root of the host, not the context path.  For optimal service discoverability, it also makes sense to me that the `/openapi` documentation resides at a predictable location, without requiring foreknowledge of context paths.

@EricWittmann Could you help clarify this?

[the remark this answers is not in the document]
EricWittmann: My understanding of the spec is that the generated openapi doc should always be at `/openapi` regardless of how many applications are deployed to the platform.  The expectation in the spec is that all of the generated openapi docs would be merged (union).  The relevant section of the spec is here:

https://github.com/eclipse/microprofile-open-api/blob/master/spec/src/main/asciidoc/microprofile-openapi-spec.adoc#55-multiple-applications

I was always under the impression that Microprofile wasn't intended for classic app servers like Wildfly, but there it is. :)

I'm not sure the smallrye implementation supports this merging of multiple app/context OpenAPI docs very well (or at all).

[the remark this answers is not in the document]
kenfinnigan: Ok, so always having worked with single deployment runtimes I've clearly not understood the spec!

Thanks for the clarification @EricWittmann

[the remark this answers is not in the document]
fabiobrz: > My understanding of the spec is that the generated openapi doc should always be at /openapi regardless of how many applications are deployed to the platform. The expectation in the spec is that all of the generated openapi docs would be merged (union). The relevant section of the spec is here:

Indeed, as stated previously, IIUC and in my opinion, the `hard` requirement no. 1 definition needs to include what at `nice-to-have` requirement no. 1 (multiple deployments), because Wildfly - as a multi deployment servlet container/application server etc. - typically allows for this to happen - also, as said previously, let's consider `ear` deployments, too...

> I was always under the impression that Microprofile wasn't intended for classic app servers like Wildfly, but there it is. :)

Probably that's the whole point: _microprofile_ vs (yet fast & beautiful) _application server_ . For this very difference - consisting in what is the expected (from the user perspective) deployment and runtime execution model between - let's say - something like Thorntail and Wildfly - main use cases and feature behavior should be described right in this proposal to clarify how Wildfly MP OpenAPI implementation will work, IMO.

@pferraro what do you think about the above suggestions? See also what you commented about usual naming strategy that Wildfly uses for app context. 
IMHO there's really need to make clear: 
* what the endpoint URL will be (consider virtual hosts definition/scope here)
* how this will work in case there are multiple deployments
* what the contents of the generated OpenAPI document will be with respect to the previously mentioned scenarios. 
Simple examples would definitely help here.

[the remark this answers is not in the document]
pferraro: > what the endpoint URL will be

This is mandated by the spec.  See Eric's comment above.

> how this will work in case there are multiple deployments

Until we have the ability to merge the documentation of all apps associated with a given host, I only see 2 options:

1. Only process the OpenAPI model for one web application per host, and fail to deploy others
2. Only process the OpenAPI model for one web application per host, and skip OpenAPI processing for the others.

> what the contents of the generated OpenAPI document will be with respect to the previously mentioned scenarios.

Without the ability to merged OpenAPI models, we are only able to process a single deployment into a single OpenAPI document.

microprofile/WFLY-12313_MicroProfile_OpenAPI_1.1.adoc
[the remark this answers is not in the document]
pferraro: > Shouldn't this optional requirement actually be discussed in the same scope as hard requirement no. 1?

I don't see how these are related.  Exporting OpenAPI module dependencies to deployments is independent of the actual processing of the OpenAPI document.

microprofile/WFLY-12313_MicroProfile_OpenAPI_1.1.adoc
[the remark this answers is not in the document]
fabiobrz: First of all, let me apologize for referencing hard requirement no.1 - while instead I meant to refer to this one:
```
* The `/openapi` endpoint shall be registered with the Undertow server/host associated with a deployment.
```
Which is no. 2, actually. So - again - sorry pointing you at the wrong item.

Now, about your last comment:

> This is mandated by the spec. See Eric's comment above.

Ok, so I am requesting for some more detail here - given previous examples involving different scenarios, like the ones you shared too (e.g.: virtual hosts, app root context etc.).

Sorry for repeating but the reader _here_ needs to be able and get all of the information in sufficient detail to determine how Wildfly would work with this feature enabled, and not going back to the spec in order to derive any missing information about the expected implementation.
This should be done by providing detailed description and [hopefully] examples in this very document, IMHO.
The above options about processing `a single deployment` and failing/skipping any OpenAPI generation for any deployment after the first one have to be documented, too.

microprofile/WFLY-12313_MicroProfile_OpenAPI_1.1.adoc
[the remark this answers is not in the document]
fabiobrz: Sorry again @pferraro, it was my mistake to point you at the wrong hard requirement (see my previous comment in this review). I actually meant to refer to the _second_ one:

```
* The `/openapi` endpoint shall be registered with the Undertow server/host associated with a deployment.
```

So, that said, I hope you get how I meant that those two - hard req. about the _unique_ `/openapi` endpoint registration and nice-to-have req. about multiple deployments - relate to one another.

pferraro:
@fabiobrz I've pushed a new draft, which should address your above concerns.  Please review.  Thanks!

microprofile/WFLY-12313_MicroProfile_OpenAPI_1.1.adoc
```diff
@@ -0,0 +1,111 @@
...
+** Consequently, the microprofile-openapi subsystem should have an empty management model.
+* Support YAML and JSON for both input (static/pre-generated OpenAPI document) and output formats.
+* MicroProfile OpenAPI TCK must pass
+
+=== Nice-to-Have Requirements
+
+* Optional features of the specification:
+** Multiple deployments:
+*** If a container allows multiple deployments, the specification expects the `/openapi` endpoint to return the logical union of the documented services for all applications in the runtime. See: https://download.eclipse.org/microprofile/microprofile-open-api-1.1.2/microprofile-openapi-spec.html#_multiple_applications
+**** Initially, we will limit multiple deployment support to a single deployment per host.
```
rsvoboda: Maybe it would be worth to mention that current SRye implementation assumes single deployment case.

microprofile/WFLY-12313_MicroProfile_OpenAPI_1.1.adoc
```diff
@@ -0,0 +1,111 @@
...
+**** Merging (and more interestingly, unmerging on undeploy) OpenAPI models for all applications deployed to a given host.
+***** Consider for a future enhancement
+**** Until full multiple deployment support is available, users might want/need the ability to disable (i.e. opt out of) OpenAPI documentation of a given web deployment, e.g. to support deployment of multiple WARs associated with the same virtual host, where only 1 is a JAX-RS application.
+***** Consider for a future enhancement
```
rsvoboda: I think this should be part of this effort, users need to have a way to specify preferred deployment for `/openapi` exposure.

Users can have *.ear with multiple *.war archives inside or just simply deploying multiple *.war archives.

microprofile/WFLY-12313_MicroProfile_OpenAPI_1.1.adoc
```diff
@@ -0,0 +1,111 @@
...
+** HTTPS transport for the `/openapi` endpoint.
+*** If the server/host of a given application defines an HTTPS listener, we will allow it, but not require it.
+*** Additionally, if the Undertow server/host associated with the application does not define an HTTP endpoint (as required by the spec), we should log a WARN message.
+** The MP OpenAPI specification suggests, but does not require, the use of a `format=JSON` or `format=YAML` parameter for the `/openapi` endpoint, used to define content type of the resulting document, in lieu of an `Accept` header.
+*** This is very reasonable, and is simple enough to do.  See: https://download.eclipse.org/microprofile/microprofile-open-api-1.1.2/microprofile-openapi-spec.html#_query_parameters
```
rsvoboda: Is SmallRye impl providing this functionality or it needs to be implemented there first ?

microprofile/WFLY-12313_MicroProfile_OpenAPI_1.1.adoc
```diff
@@ -0,0 +1,111 @@
...
+* Allow optional authentication/authorization requirements for the `/openapi` endpoint.
+** Consider for a future enhancement
```
rsvoboda: So right now to "secure" `/openapi` endpoint user removes openapi subsystem?

microprofile/WFLY-12313_MicroProfile_OpenAPI_1.1.adoc
```diff
@@ -0,0 +1,111 @@
...
+* Add an optional endpoint for browsing OpenAPI documentation via SwaggerUI (e.g. https://quarkus.io/guides/openapi-swaggerui-guide).
+** Consider for a future enhancement
```
rsvoboda: Thorntail provides SwaggerUI, WildFly should have it too so the users have the same experience.

If SwaggerUI is out-of-scope of this RFE there should be new RFE with high priority/ranking. @maeste fyi

microprofile/WFLY-12313_MicroProfile_OpenAPI_1.1.adoc
```diff
@@ -0,0 +1,111 @@
...
+
+=== Non-Requirements
+
+* The MP OpenAPI specification does not require support for i18n of the OpenAPI document (e.g. via an Accept-Language header), though this would be a reasonable subsequent enhancement.  See: https://download.eclipse.org/microprofile/microprofile-open-api-1.1.2/microprofile-openapi-spec.html#_internationalization
+* The MP OpenAPI specification does not require support for validating the generated OpenAPI model, though this is a reasonable subsequent enhancement. A number of third-party validator exist and could be leveraged for such a feature.  See: https://download.eclipse.org/microprofile/microprofile-open-api-1.1.2/microprofile-openapi-spec.html#_validation
+* The MP OpenAPI specification does not require support for https://www.w3.org/TR/cors/[Cross-Origin Resource Sharing (CORS)].  See: https://download.eclipse.org/microprofile/microprofile-open-api-1.1.2/microprofile-openapi-spec.html#_cross_origin_resource_sharing_cors
```
rsvoboda: Is this considered for future development on WF side or we are waiting for specification to evolve before implementing ?

microprofile/WFLY-12313_MicroProfile_OpenAPI_1.1.adoc
```diff
@@ -0,0 +1,111 @@
...
+:icons:             font
+:idprefix:
+:idseparator:       -
+
+== Overview
+
+This enhancement proposes to add support for https://download.eclipse.org/microprofile/microprofile-open-api-1.1.2/microprofile-openapi-spec.html[Eclipse MicroProfile OpenAPI 1.1] to WildFly by integrating the implementation from https://github.com/smallrye/smallrye-open-api[SmallRye].
+This spec provides a unified Java API for producing documentation that complies with the https://github.com/OAI/OpenAPI-Specification/blob/3.0.2/versions/3.0.2.md[OpenAPI v3 specification].
+
+An MP OpenAPI subsystem exposes an HTTP GET `/openapi` endpoint that returns an OpenAPI v3 compliant document (in YAML or JSON format) containing documentation of the RESTful services of an application.
```
rsvoboda: > RESTful services of an application

These services will be detected by the SRye code or WF functionality will be utilized to gather details about deployed RESTful services?

microprofile/WFLY-12313_MicroProfile_OpenAPI_1.1.adoc
```diff
@@ -0,0 +1,111 @@
...
+
+=== Other Interested Projects
+
+== Requirements
+
+=== Hard Requirements
+
+* Create an MP OpenAPI galleon layer
+** Should depend on MP Config and JAX-RS (and their dependencies)
+* The microprofile-openapi spec module must be exposed to deployments
```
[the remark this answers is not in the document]
pferraro: Also the module containing the WF subsystem.  smallrye-openapi also adds a dependency on com.fasterxml.jackson.dataformat:jackson-dataformat-yaml.

However, only the microprofile-openapi-api module will need to be exported to web deployments.

microprofile/WFLY-12313_MicroProfile_OpenAPI_1.1.adoc
```diff
@@ -0,0 +1,111 @@
...
+* The `/openapi` endpoint shall be registered relative to the root of the virtual host associated with the deployment.
```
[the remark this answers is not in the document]
pferraro: /openapi is intended for public consumption, thus it would use public interfaces.  It is not management oriented in the same way that health and metrics are.

microprofile/WFLY-12313_MicroProfile_OpenAPI_1.1.adoc
```diff
@@ -0,0 +1,111 @@
...
+** This places a superficial constraint of a single application per host (see subsequent section re: multiple deployments)
+** In the event a web deployment uses a given host for which another deployment has already registered an `/openapi` endpoint, we will:
+... Log a warning.
+... Skip OpenAPI model generation for that deployment.
+* All WildFly/SmallRye specific configuration will be obtained via MP Config using the properties prefixed with `mp.openapi.extensions`.
+** Consequently, the microprofile-openapi subsystem should have an empty management model.
```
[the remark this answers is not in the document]
pferraro: Using an MP config attribute to globally enable/disable `/openapi` registration (without requiring a different configuration for dev/qa/staging/prod (i.e. to remove the microprofile-openapi extension) sounds very reasonable and is trivial enough to do for the initial integration.

microprofile/WFLY-12313_MicroProfile_OpenAPI_1.1.adoc
```diff
@@ -0,0 +1,111 @@
...
+* Support YAML and JSON for both input (static/pre-generated OpenAPI document) and output formats.
+* MicroProfile OpenAPI TCK must pass
+
```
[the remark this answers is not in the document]
pferraro: Those thorntail tests are extremely minimal and are already mentioned in the test plan section.

microprofile/WFLY-12313_MicroProfile_OpenAPI_1.1.adoc
pferraro: It will be implemented in the Undertow HttpHandler that is part of the servlet container integration.

microprofile/WFLY-12313_MicroProfile_OpenAPI_1.1.adoc
pferraro: Or does not add it.  It is still to be decided whether this subsystem will be included in the default configuration profiles.

microprofile/WFLY-12313_MicroProfile_OpenAPI_1.1.adoc
```diff
@@ -0,0 +1,111 @@
...
+These integrations will heavily inform, though not dictate, the corresponding integration into WildFly.
+
+=== Other Interested Projects
+
+== Requirements
+
+=== Hard Requirements
+
+* Create an MP OpenAPI galleon layer
```
fabiobrz: IIUC this is a new requirement. Is there any plan on how to test it? I cannot see any reference to it in the "Test plan" section later.

microprofile/WFLY-12313_MicroProfile_OpenAPI_1.1.adoc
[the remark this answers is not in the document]
fabiobrz: I think that this level of detail that has been provided in comments about modules should be consolidated as document contents as well.

microprofile/WFLY-12313_MicroProfile_OpenAPI_1.1.adoc
pferraro: Perhaps, depending on interest/need.

microprofile/WFLY-12313_MicroProfile_OpenAPI_1.1.adoc
pferraro: If afraid I don't understand the question.  This is just a description of the intention of the OpenAPI initiative.

microprofile/WFLY-12313_MicroProfile_OpenAPI_1.1.adoc
pferraro: The testsuite will inherently validate this, as the distribution against which it will test will have been built using this new layer.

microprofile/WFLY-12313_MicroProfile_OpenAPI_1.1.adoc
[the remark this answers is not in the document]
pferraro: Sure, no problem.

microprofile/WFLY-12313_MicroProfile_OpenAPI_1.1.adoc
```diff
@@ -0,0 +1,111 @@
...
+** Should depend on MP Config and JAX-RS (and their dependencies)
+* The microprofile-openapi spec module must be exposed to deployments
+* The `/openapi` endpoint shall be registered relative to the root of the virtual host associated with the deployment.
+** This places a superficial constraint of a single application per host (see subsequent section re: multiple deployments)
+** In the event a web deployment uses a given host for which another deployment has already registered an `/openapi` endpoint, we will:
+... Log a warning.
+... Skip OpenAPI model generation for that deployment.
+* All WildFly/SmallRye specific configuration will be obtained via MP Config using the properties prefixed with `mp.openapi.extensions`.
```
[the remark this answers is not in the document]
pferraro: Technically, such an attribute is redundant with the /subsystem=microprofile-openapi:add()/remove() operations, but using attribute makes it easier to parameterize this behavior for a given environment.

microprofile/WFLY-12313_MicroProfile_OpenAPI_1.1.adoc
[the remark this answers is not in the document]
pferraro: Sure. No problem.

microprofile/WFLY-12313_MicroProfile_OpenAPI_1.1.adoc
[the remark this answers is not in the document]
pferraro: Correct.

microprofile/WFLY-12313_MicroProfile_OpenAPI_1.1.adoc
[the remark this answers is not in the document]
rsvoboda: My point is to highlight WF TS is ensuring at least the same coverage as TT TS.
Agree those tests are minimal.

microprofile/WFLY-12313_MicroProfile_OpenAPI_1.1.adoc
rsvoboda: My point is that this shouldn't be marked as "Consider for a future enhancement"

microprofile/WFLY-12313_MicroProfile_OpenAPI_1.1.adoc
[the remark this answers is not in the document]
fabiobrz: Ok, if you agree, let's leave the subsystem activation/deactivation via the CLI, which is not relevant here since we're actually discussing (extended) MP config examples.
That said, IIUC we can state there will be one only _extended_ configuration feature, which the user will be able to use, i.e. setting one or more properties, in order to 

> enable/disable /openapi registration

(see your comment [1])

Right? If the above is correct, then I think the properties/property name/s should be defined in the document and the above mentioned case could be used as the example we're discussing about. WDYT?  

One more question: when you state:

> ... parameterize this behavior for a given environment

Do you mean "_for a given application and thus not globally_"? In such case what you commented at [1] should be revised to make clear this property would not affect the `/openapi` registration globally, correct?

[1]
https://github.com/wildfly/wildfly-proposals/pull/251#discussion_r337532401

microprofile/WFLY-12313_MicroProfile_OpenAPI_1.1.adoc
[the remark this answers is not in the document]
fabiobrz: @pferraro so basically you're assuming to leave the subsystem management model empty and to use MP config extensions to do all the "_needed job_".
Given my previous comments [1] and if we agree that the _needed job_ would just consist in 

> Using an MP config attribute to globally enable/disable /openapi registration

my question here would be the following: is a global impact still meant to be the choice or the MP config extended properties would allow the user to do that on a _per application_ basis?

microprofile/WFLY-12313_MicroProfile_OpenAPI_1.1.adoc
fabiobrz: @rsvoboda I agree and given previous comments regarding this topic i think that this should be covered by a new hard requirement ( "enable/disable `/openapi` endpoint registration on a per-application basis via MP config). @pferraro WDYT?

microprofile/WFLY-12313_MicroProfile_OpenAPI_1.1.adoc
fabiobrz: Probably we should list things like this, i.e. which still are waiting for a decision. 
I remember that the same was ongoing about the subsystem being available or not in both `standalone` and `domain` modes but it seems that the discussion was marked as resolved.

microprofile/WFLY-12313_MicroProfile_OpenAPI_1.1.adoc
```diff
@@ -0,0 +1,111 @@
...
+[X] QE
+
+=== Affected Projects or Components
+
+The SmallRye implementation of the MicroProfile OpenAPI specification is already integrated by the following projects:
+
+* https://thorntail.io/[Thorntail] via the https://github.com/thorntail/thorntail/tree/master/fractions/microprofile/microprofile-openapi[microprofile-openapi fraction].
+* https://quarkus.io/[Quarkus] via the https://github.com/quarkusio/quarkus/tree/master/extensions/smallrye-openapi[smallrye-openapi extension].
+
+These integrations will heavily inform, though not dictate, the corresponding integration into WildFly.
```
fabiobrz: @pferraro  do you have any idea on whether/where/how to cover a vary basic migration from TT? That should eventually consist in changing just some dependencies.

microprofile/WFLY-12313_MicroProfile_OpenAPI_1.1.adoc
fabiobrz: Ok, thanks.

microprofile/WFLY-12313_MicroProfile_OpenAPI_1.1.adoc
pferraro: There are no differences between standalone vs domain with respect to the subsystems included in a given profile.

microprofile/WFLY-12313_MicroProfile_OpenAPI_1.1.adoc
pferraro: What dependencies would change?  An TT application using MP OpenAPI would only have a dependency on the spec API module.

pferraro: @rsvoboda To which TT specific features are you referring?

rsvoboda: Just general comment, not related to MP OpenAPI.

microprofile/WFLY-12313_MicroProfile_OpenAPI_1.1.adoc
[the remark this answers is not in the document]
pferraro: @fabiobrz After further deliberation, I suggest we add support for a `mp.openapi.extensions.disable` property.  This can be specified via a global config source (e.g. system property) to disable openapi globally, or via an application config source (e.g. /META-INF/microprofile-config.properties) to enable/disable on a per-application basis.

microprofile/WFLY-12313_MicroProfile_OpenAPI_1.1.adoc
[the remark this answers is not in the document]
pferraro: The MP Config spec requires a number of specific ConfigSources, some of which are global, and others classpath (i.e. application) specific.  Thus, any MP config property can be specified per application or globally.

pferraro:
@fabiobrz New revision incorporating feedback.

microprofile/WFLY-12313_MicroProfile_OpenAPI_1.1.adoc
```diff
@@ -0,0 +1,111 @@
...
+* The MP OpenAPI specification does not require support for validating the generated OpenAPI model, though this is a reasonable subsequent enhancement. A number of third-party validator exist and could be leveraged for such a feature.  See: https://download.eclipse.org/microprofile/microprofile-open-api-1.1.2/microprofile-openapi-spec.html#_validation
+* The MP OpenAPI specification does not require support for https://www.w3.org/TR/cors/[Cross-Origin Resource Sharing (CORS)].  See: https://download.eclipse.org/microprofile/microprofile-open-api-1.1.2/microprofile-openapi-spec.html#_cross_origin_resource_sharing_cors
+
+== Test Plan
+
+* Basic subsystem unit tests
+* Add the microprofile-tck tests to the MicroProfile TCK testsuite in WildFly.
+* Integration tests should cover vendor specific behavior, not covered by the TCK tests, including:
+** Ensuring that services deployed using a non-root context path are documented with the correct `server` elements.
+** Any multi-deployment or multi-module deployment behavior
```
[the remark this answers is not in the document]
pferraro: What new requirement?

microprofile/WFLY-12313_MicroProfile_OpenAPI_1.1.adoc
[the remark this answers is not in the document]
fabiobrz: Ok thanks. Then it would probably be useful to understand how the priority between global and per-application scopes would be managed (e.g.: disabled globally, what happens if the user enables at application level or the other way round?)

microprofile/WFLY-12313_MicroProfile_OpenAPI_1.1.adoc
```diff
@@ -0,0 +1,126 @@
...
+***** Consider for a future enhancement
+** HTTPS transport for the `/openapi` endpoint.
+*** If the server/host of a given application defines an HTTPS listener, we will allow it, but not require it.
+*** Additionally, if the Undertow server/host associated with the application does not define an HTTP endpoint (as required by the spec), we should log a WARN message.
+** The MicroProfile OpenAPI specification suggests, but does not require, the use of a `format=JSON` or `format=YAML` parameter for the `/openapi` endpoint, used to define content type of the resulting document, in lieu of an `Accept` header.
+*** This is very reasonable, and is simple enough to do.  See: https://download.eclipse.org/microprofile/microprofile-open-api-1.1.2/microprofile-openapi-spec.html#_query_parameters
+* Use a `mp.openapi.extensions.disable` property to disable OpenAPI documentation globally or per-application.
+** Provides a parameterizable means of disabling OpenAPI documentation in specific environments (e.g. prod vs dev)
+** Provides a mechanism to control which application associated with a given virtual host should generate an MicroProfile OpenAPI model.
```
fabiobrz: > Provides a mechanism...

From the user point of view, given the [supposed] logic behind the property name, I take that this one would be a per-exclusion and  derivative mechanism, correct?

microprofile/WFLY-12313_MicroProfile_OpenAPI_1.1.adoc
```diff
@@ -0,0 +1,126 @@
...
+=== Hard Requirements
+
+* Create an MicroProfile OpenAPI galleon layer containing the MP Open API feature encapsulated by a new MicroProfile OpenAPI subsystem.
+** Should depend on MicroProfile Config and JAX-RS (and their dependencies)
+** N.B. New modules to be included in WF:
+*** MicroProfile OpenAPI subsystem module
+*** io.smallrye:smallrye-open-api
+*** org.eclipse.microprofile.openapi:microprofile-openapi-api
+*** com.fasterxml.jackson.dataformat:jackson-dataformat-yaml
+** Which feature packs will use this layer is still TBD.
```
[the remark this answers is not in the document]
pferraro: We should have answers for these questions in the coming weeks.

microprofile/WFLY-12313_MicroProfile_OpenAPI_1.1.adoc
```diff
@@ -0,0 +1,126 @@
...
+** Consequently, the microprofile-openapi subsystem should have an empty management model.
+* Support YAML and JSON for both input (static/pre-generated OpenAPI document) and output formats.
+* MicroProfile OpenAPI TCK must pass
+
+=== Nice-to-Have Requirements
+
+* Optional features of the specification:
+** Multiple deployments:
+*** If a container allows multiple deployments, the specification expects the `/openapi` endpoint to return the logical union of the documented services for all applications in the runtime. See: https://download.eclipse.org/microprofile/microprofile-open-api-1.1.2/microprofile-openapi-spec.html#_multiple_applications
+**** Initially, we will limit multiple deployment support to a single deployment per host (for a given endpoint).
```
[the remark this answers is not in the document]
pferraro: > Are you referring to the endpoint specific path [...]?

Yes.

microprofile/WFLY-12313_MicroProfile_OpenAPI_1.1.adoc
pferraro: Yes - the property would be applied according to the ConfigSource scope and its ordinal.

microprofile/WFLY-12313_MicroProfile_OpenAPI_1.1.adoc
```diff
@@ -0,0 +1,126 @@
...
+
+=== Affected Projects or Components
+
+The SmallRye implementation of the MicroProfile OpenAPI specification is already integrated by the following projects:
+
+* https://thorntail.io/[Thorntail] via the https://github.com/thorntail/thorntail/tree/master/fractions/microprofile/microprofile-openapi[microprofile-openapi fraction].
+* https://quarkus.io/[Quarkus] via the https://github.com/quarkusio/quarkus/tree/master/extensions/smallrye-openapi[smallrye-openapi extension].
+
+These integrations will heavily inform, though not dictate, the corresponding integration into WildFly.
+
```
fabiobrz: @pferraro, given what in the above resolved discussions, can we consolidate the snippet about migrating sample Thorntail app to MP featured Wildfly by just showing what @rsvoboda suggested (dependency modification in POM)? This would add great value to the users experience by showing them how concretely Wildfly would replace Thorntail.

microprofile/WFLY-12313_MicroProfile_OpenAPI_1.1.adoc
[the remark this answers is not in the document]
fabiobrz: If so then I think this should be reworded as it isn't clear: e.g. 1 virtual host -> _n_ applications -> _m_ uniquely named (each registered through the extension property) OpenAPI endpoints. Of course this is just a short way to express my point, please use the form you prefer.

microprofile/WFLY-12313_MicroProfile_OpenAPI_1.1.adoc
fabiobrz: As said previously, from user point of view and in order to understand the logic in detail, IMO more information about how the implementation will prioritize global vs application with respect to MP OpenAPI module configuratio properties should be added here.

microprofile/WFLY-12313_MicroProfile_OpenAPI_1.1.adoc
pferraro: @fabiobrz What "dependency modification in POM" are you talking about?

microprofile/WFLY-12313_MicroProfile_OpenAPI_1.1.adoc
pferraro: > how the implementation will prioritize global vs application with respect to MP OpenAPI module configuratio properties

We don't control this.  The prioritization of properties is determined according to the MP Config spec.

microprofile/WFLY-12313_MicroProfile_OpenAPI_1.1.adoc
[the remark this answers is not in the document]
pferraro: OK - I will attempt to clarify this in the document.

microprofile/WFLY-12313_MicroProfile_OpenAPI_1.1.adoc
fabiobrz: Hi @pferraro, I was referring to the task of migrating from Thorntail topic. 
I've just created a basic JAX-RS Thorntail example [1] following the guide and then "migrated" it to Wildfly [2]. 
Of course, these are just samples and the goal is just to show that a [very simple] Thorntail app can be migrated easily to Wildfly by making some changes to the POM. I've been testing [2] against current feature branch, [3].

[1]
https://github.com/fabiobrz/openapi-demo

[2]
https://github.com/fabiobrz/openapi-demo-wfly

[3]
https://github.com/MikeEdgar/wildfly/tree/feature/WFLY-12313-openapi

microprofile/WFLY-12313_MicroProfile_OpenAPI_1.1.adoc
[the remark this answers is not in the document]
fabiobrz: Thanks, waiting for the change in the next draft.

microprofile/WFLY-12313_MicroProfile_OpenAPI_1.1.adoc
fabiobrz: Right I'll have a look to see whether the reader can derive the information from there and will get back.

microprofile/WFLY-12313_MicroProfile_OpenAPI_1.1.adoc
pferraro: @fabiobrz I'm still not sure what migration there is to do.  The only dependency such an app would have has not changed.

fabiobrz: What would the "only dependency" be? 
I see I had to remove all of Thorntail deps as the subsystem would provide module. 
Isn't that the migration change on the POM we're discussing about?

pferraro: That has to do with Thorntail provisioning, not the application itself.  Your application itself would have no references to any TT modules.  An application destined to run on WildFly has a completely different workflow.

fabiobrz: >  That has to do with Thorntail provisioning, not the application itself.

I don't think so. Probably - with _provisioning_ - you're referring to the plugin, which would execute the app by running the following command: 
```
mvn thorntail:run
```

But I wouldn't assume that's every application case. Quoting the project page:
```
Thorntail offers an innovative approach to _packaging_ and running Java EE applications by packaging them with just enough of the server runtime to "java -jar" your application. It's MicroProfile compatible, too...
```
So isn't TT one of the [currently] available tools that I actually need to have the feature? 
I think so. Otherwise I'd have to choose OpenLiberty or do something else. 
In such case, then, isn't adding the TT dependencies what would activate the needed fraction(s)? I think so, it is what I need to let a JVM execute my `uber jar` by running the commands:
```
mvn package
java -jar myapp-thorntail.war
```

> Your application itself would have no references to any TT modules.

My TT app would, indeed. My Wildfly app wouldn't, instead. And that's the whole point of migrating.
I am referring to what on TT project page, "Simple Project Setup" section: https://thorntail.io/
From that section I understand that my TT application itself should indeed have at least to depend on the TT bom (see TT examples) and also wouldn't have any MicroProfile OpenAPI feature if I wouldn't have the fraction declared, right? Quoting the snippet description there:
```
Import the BOM (Bill of Materials) dependency
...
Declare the fractions you need (or not, we’ll do our best to auto-detect them for you):
...
```
(I guess I would declare them in to be on the safe side a real world production scenario).

> An application destined to run on WildFly has a completely different workflow.

Indeed, that is quite clear but doesn't change the matter: a TT application would need dependencies to activate the fraction that support the feature, while an application that would run on Wildfly will have the feature thanks to a subsystem backed by a module and wouldn't need any MP dependency.

Please correct in case I am wrong on what above as I am no TT expert, as you know.
But if what I reported makes sense to you, too, then my point stands: 
"Do you have a TT app with MP fraction to support OpenAPI generation and you want to migrate that very app to Wildfly? Then remove the TT dependencies and let the Wildfly subsystem provide the same feature to you".

pferraro: I thinking I'm just tripping over terminology.  IMO, the uber jar isn't the application.  It's a jar containing both the application and the provisioned TT fractions.
General documentation about migrating TT provisioning to WF is a completely separate topic. I don't think it makes sense to have to repeat these instructions for every MP module.  However, to your point, general migration from TT is certainly something that needs documentation.   We should file a separate JIRA issue to track this migration documentation task.

fabiobrz: > General documentation about migrating TT provisioning to WF is a completely separate topic. I don't think it makes sense to have to repeat these instructions for every MP module. However, to your point, general migration from TT is certainly something that needs documentation. We should file a separate JIRA issue to track this migration documentation task

@pferraro that sounds reasonable to me as well. I think we could place a link for the JIRA once it gets created, so that it is at least traced here, WDYT?

microprofile/WFLY-12313_MicroProfile_OpenAPI_1.1.adoc
```diff
@@ -0,0 +1,128 @@
...
+*** com.fasterxml.jackson.dataformat:jackson-dataformat-yaml
+** Which feature packs will use this layer is still TBD.
+* The microprofile-openapi spec module must be exposed to web deployments
+* The `/openapi` endpoint shall be registered relative to the root of the virtual host associated with the deployment.
+** N.B. The `/openapi` endpoint is intended for a public interface, not a management interface, as with the `/health` and `/metrics` endpoints.
+** This places a superficial constraint of a single application per host (see subsequent section re: multiple deployments)
+** In the event a web deployment uses a given host for which another deployment has already registered an `/openapi` endpoint, we will:
+... Log a warning.
+... Skip OpenAPI model generation for that deployment.
+* All WildFly/SmallRye specific configuration will be obtained via MicroProfile Config using the properties prefixed with `mp.openapi.extensions`.
```
bstansberry: Any config properties we plan to support that are not defined in the MP OpenAPI spec should be enumerated in this document, a la the way we would describe planned management model changes.

pferraro: These are enumerated in the optional requirements section.

microprofile/WFLY-12313_MicroProfile_OpenAPI_1.1.adoc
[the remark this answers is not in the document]
pferraro: FYI - we will add microprofile-openapi to the default standalone-openshift.xml config.  I'll update the document.

microprofile/WFLY-12313_MicroProfile_OpenAPI_1.1.adoc
fabiobrz: I went back to MP Config specs and - apart from ordinal definition for config sources - no great detail is shared on how implementation should deal with configuration properties/sources priority.

pferraro: That is all there is to it.  The ordinal priority of a ConfigSource determines which configuration property values take precedence.

microprofile/WFLY-12313_MicroProfile_OpenAPI_1.1.adoc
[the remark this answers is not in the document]
pferraro: Only that we will provide a galleon layer for microprofile-openapi and that the subsystem will be included in the default standalone-openshift.xml.
As such, our general documentation will need to include adding the extension to the target server profile as a first step.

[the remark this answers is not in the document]
fabiobrz: > Only that we will provide a galleon layer for microprofile-openapi 

Ok, this one is covered in the proposal.

> ...the subsystem will be included in the default standalone-openshift.xml.

And this is probably where to add to the document. If so, will the feature be available by default _only_ in `standalone` operating mode and not `domain`, right? 
That said, then is there any limitation for the users to later enable the subsystem when operating Wildfly in both `standalone` and `domain` modes?

> As such, our general documentation will need to include adding the extension to the target server profile as a first step.

Ok. Will it be available for each profile then? Anyway, I suppose you're referring to community docs here and I agree in such case.
However the proposal should provide detailed answers to previous questions so that is clear from this very document how actually the feature will be available for end user cases.

microprofile/WFLY-12313_MicroProfile_OpenAPI_1.1.adoc
fabiobrz: Ok, thanks for confirmation.

microprofile/WFLY-12313_MicroProfile_OpenAPI_1.1.adoc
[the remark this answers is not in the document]
pferraro: > If so, will the feature be available by default only in standalone operating mode and not domain, right?

Yes. Domain mode is not relevant to OpenShift.

> That said, then is there any limitation for the users to later enable the subsystem when operating Wildfly in both standalone and domain modes?

No.  I am not aware of any subsystem with such a limitation.

> Will it be available for each profile then?

Any extension is available can be added to any profile.  There are no inherent limitations.

microprofile/WFLY-12313_MicroProfile_OpenAPI_1.1.adoc
```diff
@@ -0,0 +1,128 @@
...
+** Consequently, the microprofile-openapi subsystem should have an empty management model.
+* Support YAML and JSON for both input (static/pre-generated OpenAPI document) and output formats.
+* MicroProfile OpenAPI TCK must pass
+
+=== Nice-to-Have Requirements
+
+* Optional features of the specification:
+** Multiple deployments:
+*** If a container allows multiple deployments, the specification expects the `/openapi` endpoint to return the logical union of the documented services for all applications in the runtime. See: https://download.eclipse.org/microprofile/microprofile-open-api-1.1.2/microprofile-openapi-spec.html#_multiple_applications
+**** Initially, we will limit multiple deployment support to a single deployment per host (for a given endpoint). i.e. each deployment, for which the OpenAPI endpoint is enabled, must either use a distinct host, or a distinct endpoint path (if a given virtual host is associated with multiple deployments).
```
fabiobrz: As this is still listed among `Nice-to-have`s, it would be good to clarify here what will be the set of requirements - both _hard_ and _nice to have_- that will be delivered in first version of MP integration. This is fundamental information for the user to have a clear idea about if/how the multiple-deployment environment would work.

microprofile/WFLY-12313_MicroProfile_OpenAPI_1.1.adoc
```diff
@@ -0,0 +1,128 @@
...
+... Skip OpenAPI model generation for that deployment.
+* All WildFly/SmallRye specific configuration will be obtained via MicroProfile Config using the properties prefixed with `mp.openapi.extensions`.
+** Consequently, the microprofile-openapi subsystem should have an empty management model.
+* Support YAML and JSON for both input (static/pre-generated OpenAPI document) and output formats.
+* MicroProfile OpenAPI TCK must pass
+
+=== Nice-to-Have Requirements
+
+* Optional features of the specification:
+** Multiple deployments:
```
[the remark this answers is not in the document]
pferraro: > Is what follows here also covering sub-deployments topic as well?

It matters not to OpenAPI that a given WAR has a parent deployment.

> Would the behavior be the same from various end user perspectives (e.g. developer, administrator, etc.)?

Yes.

microprofile/WFLY-12313_MicroProfile_OpenAPI_1.1.adoc
```diff
@@ -0,0 +1,128 @@
...
+== Test Plan
+
+* Basic subsystem unit tests
+* Add the microprofile-tck tests to the MicroProfile TCK testsuite in WildFly.
+* Integration tests should cover vendor specific behavior, not covered by the TCK tests, including:
+** Ensuring that services deployed using a non-root context path are documented with the correct `server` elements.
+** Any multi-deployment or multi-module deployment behavior, e.g. `mp.openapi.extensions.disable`
+** Alternate OpenAPI endpoint registration according to `mp.openapi.extensions.path`
+** Fetching JSON vs YAML documentation via `format` parameter
+** Loading JSON vs YAML static OpenAPI documentation
```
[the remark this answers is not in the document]
pferraro: This is already covered under "Basic subsystem unit tests".  Every subsystem in WildFly uses the same unit test framework to validate its management model.

microprofile/WFLY-12313_MicroProfile_OpenAPI_1.1.adoc
pferraro: I don't understand what you're asking.  Multiple deployment support is explicitly an optional requirement of the specification, which is why it is listed in the "nice-to-have" requirement section.
I've already described how the initial implementation should handle multiple deployments, and options for future enhancements.  What exactly do you find unclear?

fabiobrz: Hi @pferraro, probably it's just about terminology: "hard" vs "nice-to-have" requirements let the reader think about this proposal development, not about the specification.
And yes - as a reader - I wanted to be sure that this document states as clearly as possible that initial implementation would have all of the "hard" + not-marked-for-future-enhancements "nice to have"s, and that implementation would work as detailed with respect to multiple deployments. 
Does this makes sense to you?

microprofile/WFLY-12313_MicroProfile_OpenAPI_1.1.adoc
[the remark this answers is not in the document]
fabiobrz: Ok, thanks.

microprofile/WFLY-12313_MicroProfile_OpenAPI_1.1.adoc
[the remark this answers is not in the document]
fabiobrz: Ok, thanks.

microprofile/WFLY-12313_MicroProfile_OpenAPI_1.1.adoc
pferraro: What do you still find unclear about the requirements surrounding multiple deployments?

fabiobrz: Probably I was just discussing about categories here, i.e. "hard" vs "nice to have" requirements. 
Anyway, I now assume the implementation will follow the hard requirements and those "nice to have"s that are not marked as "Consider for a future enhancement".

microprofile/WFLY-12313_MicroProfile_OpenAPI_1.1.adoc
```diff
@@ -0,0 +1,131 @@
...
+* Use a `mp.openapi.extensions.enabled` property to disable OpenAPI documentation globally (via mp-config subsystem) or per-application (e.g. /META-INF/mp-config.properties).
+** Provides a parameterizable means of disabling OpenAPI documentation in specific environments (e.g. prod vs dev)
+** Provides a mechanism to control which application associated with a given virtual host should generate an MicroProfile OpenAPI model.
+* Use a `mp.openapi.extensions.path` property to customize the path of the OpenAPI endpoint.
+** Provides a means for generating OpenAPI documentation for multiple applications associated with the same virtual host.
+** Potentially useful to support legacy OpenAPI tooling (e.g. `/swagger`)
+** Analogous to the `quarkus.smallrye-openapi.path` property in Quarkus.
+* Use a `mp.openapi.extensions.relative` property to configure whether auto-generated Server records should be absolute or relative.
+** When relative, the generated record will contain the context path of the deployment
+** When absolute, generated records will include all hosts/ports at which the given deployment is accessible.
```
fabiobrz: I think it would be good to have the allowed values for the property here, or just to mention that it is expecting a Boolean. At first sight it looks like an enumeration description.

microprofile/WFLY-12313_MicroProfile_OpenAPI_1.1.adoc
```diff
@@ -0,0 +1,131 @@
...
+*** This is very reasonable, and is simple enough to do.  See: https://download.eclipse.org/microprofile/microprofile-open-api-1.1.2/microprofile-openapi-spec.html#_query_parameters
+* Use a `mp.openapi.extensions.enabled` property to disable OpenAPI documentation globally (via mp-config subsystem) or per-application (e.g. /META-INF/mp-config.properties).
+** Provides a parameterizable means of disabling OpenAPI documentation in specific environments (e.g. prod vs dev)
+** Provides a mechanism to control which application associated with a given virtual host should generate an MicroProfile OpenAPI model.
+* Use a `mp.openapi.extensions.path` property to customize the path of the OpenAPI endpoint.
+** Provides a means for generating OpenAPI documentation for multiple applications associated with the same virtual host.
+** Potentially useful to support legacy OpenAPI tooling (e.g. `/swagger`)
+** Analogous to the `quarkus.smallrye-openapi.path` property in Quarkus.
+* Use a `mp.openapi.extensions.relative` property to configure whether auto-generated Server records should be absolute or relative.
```
fabiobrz: Can you please add some description about whether the "auto-generated server records" process is something peculiar to Wildfly implementation? Or at least point the readers to some section of the specs to which they can relate the feature and the rationale for implementing it this way?

microprofile/WFLY-12313_MicroProfile_OpenAPI_1.1.adoc
pferraro: Sure.

microprofile/WFLY-12313_MicroProfile_OpenAPI_1.1.adoc
pferraro: This is a spec requirement for any deployment with a context path.  Sure, I can add a link to the relevant section of the spec.

microprofile/WFLY-12313_MicroProfile_OpenAPI_1.1.adoc
```diff
@@ -0,0 +1,131 @@
...
+* The MicroProfile OpenAPI specification does not require support for validating the generated OpenAPI model, though this is a reasonable subsequent enhancement. A number of third-party validator exist and could be leveraged for such a feature.  See: https://download.eclipse.org/microprofile/microprofile-open-api-1.1.2/microprofile-openapi-spec.html#_validation
+* The MicroProfile OpenAPI specification does not require support for https://www.w3.org/TR/cors/[Cross-Origin Resource Sharing (CORS)].  See: https://download.eclipse.org/microprofile/microprofile-open-api-1.1.2/microprofile-openapi-spec.html#_cross_origin_resource_sharing_cors
+
+== Test Plan
+
+* Basic subsystem unit tests
+* Add the microprofile-tck tests to the MicroProfile TCK testsuite in WildFly.
+* Integration tests should cover vendor specific behavior, not covered by the TCK tests, including:
+** Ensuring that services deployed using a non-root context path are documented with the correct `server` elements.
+** Any multi-deployment or multi-module deployment behavior, e.g. `mp.openapi.extensions.disable`
```
fabiobrz: Typo: `mp.openapi.extensions.disable` -> `mp.openapi.extensions.enable`, sorry for missing this in previous review.

pferraro: Fixed.

microprofile/WFLY-12313_MicroProfile_OpenAPI_1.1.adoc
```diff
@@ -0,0 +1,131 @@
...
+** Provides a parameterizable means of disabling OpenAPI documentation in specific environments (e.g. prod vs dev)
+** Provides a mechanism to control which application associated with a given virtual host should generate an MicroProfile OpenAPI model.
+* Use a `mp.openapi.extensions.path` property to customize the path of the OpenAPI endpoint.
+** Provides a means for generating OpenAPI documentation for multiple applications associated with the same virtual host.
+** Potentially useful to support legacy OpenAPI tooling (e.g. `/swagger`)
+** Analogous to the `quarkus.smallrye-openapi.path` property in Quarkus.
+* Use a `mp.openapi.extensions.relative` property to configure whether auto-generated Server records should be absolute or relative.  At a minimum, Server records are necessary to ensure, in the presence of a non-root context path, that consumers of an OpenAPI can construct valid URLs to REST services relative to the host of the OpenAPI endpoint.  See: https://download.eclipse.org/microprofile/microprofile-open-api-1.1.2/microprofile-openapi-spec.html#_context_root_behavior
```
fabiobrz: Hi @pferraro, I noticed this typo while reviewing the actual sources: 
`mp.openapi.extensions.relative` -> `mp.openapi.extensions.servers.relative`, just as a side note.

pferraro: Fixed.

PR #236, opened 2019-07-25
[WFLY-12306]: Transition  OpenTracing to 1.3

ehsavoie:
Jira: https://issues.jboss.org/browse/WFLY-12306

PR #260, opened 2019-11-06
[WFLY-12758]: Integrate opentracing

ehsavoie:
Describing the management API changes that will be implemented in the
'microprofile-opentracing-smallrye' subsystem to cover the configuration
of tracers and their injections.

Jira: https://issues.jboss.org/browse/WFLY-12758

### DIFF
--- a/microprofile/WFLY-12758_integrate_opentracing.adoc
+++ b/microprofile/WFLY-12758_integrate_opentracing.adoc
@@ -1,0 +1,115 @@
+= Integrate Microprofile OpenTracing 1.3
+:author:            Emmanuel Hugonnet
+:email:             ehugonne@redhat.com
+:toc:               left
+:icons:             font
+:keywords:          microprofile,opentracing,observability,tracing,openshift
+:idprefix:
+:idseparator:       -
+
+== Overview
+
+Currently the integration of Microprofile Opentracing is minimal as you can't configure the tracer using the WildFly Management Model.
+So we will implement a way to define several tracer configurations (Jaeger) to be used in deployments.
+One of those configuration will become a default one so that deployments that don't define which configuration to use will use this one.
+In this proposal we will only support Jaeger as Tracer implementation but it should be able to be easily extendable to support other Tracers implementations:
+
+* Support https://github.com/eclipse/microprofile-opentracing/blob/1.3/spec/src/main/asciidoc/microprofile-opentracing.asciidoc[Eclipse MicroProfile OpenTracing 1.3] for applications deployed in WildFly, including JAX-RS and CDI integration. Passing the TCK is required.
+* Support a concrete tracer, such as Jaeger's.
+* Allow applications to supply their own tracer implementations via `TracerResolver`. The configuration is up to the user and the tracerResolver mechanism.
+* Each individual application deployed (WAR) should have its own `Tracer` instance.
+* Each WAR within an EAR should be treated as an individual WAR, having their own `Tracer` instance.
+
+== Issue Metadata
+
+=== Issue
+
+* https://issues.jboss.org/browse/WFLY-12758[WFLY-12758]
+
+=== Related Issues
+
+* https://issues.apache.org/jira/browse/EAP7-1351
+
+=== Dev Contacts
+
+* mailto:{email}[{author}]
+
+=== QE Contacts
+
+* mailto:szhantem@redhat.com[Sultan Zhantemirov]
+
+=== Testing By
+// Put an x in the relevant field to indicate if testing will be done by Engineering or QE. 
+// Discuss with QE during the Kickoff state to decide this
+[] Engineering
+
+[X] QE
+
+=== Affected Projects or Components
+The primary integration will be happening by the updating the subsystem `microprofile-opentracing-smallrye` within https://github.com/wildfly/wildfly[WildFly].
+
+=== Other Interested Projects
+
+== Requirements
+
+* Upgrade MicroProfile Opentracing 1.3.0
+* Run any OpenTracing test in WildFly integration test suite
+** tests may require changes as this major version has breaking changes.
+* Run the Eclipse MicroProfile OpenTracing TCK without failures
+
+=== Hard Requirements
+
+Adding a new opentracing subsystem management API to create Jaeger configuration so that Jaeger tracers can be injected into deployments.
+
+=== Nice-to-Have Requirements
+
+N/A
+
+=== Non-Requirements
+
+Other tracer implementations will not be configureable from the subsystem management API.
+
+== Implementation Plan
+
+We will only cover the creation of a Jaeger tracer configuration. Other tracer implementations may come afterwards but are not currently 
+
+* Create a `jaeger-tracer` resource that define the configuration of a Jaeger tracer. This resource will expose all Jaeger Tracer configuration attributes:
+    ** `propagation`: The supported trace context propagation formats:
+        *** JAEGER: The default Jaeger trace context propagation format.
+        *** B3: The Zipkin B3 trace context propagation format.
+    ** `sampler-type`: The type of sampler to use in the tracer. Optional. Valid values: remote (default),ratelimiting, probabilistic, const.
+    ** `sampler-param`: The floating point value that makes sense for the correct samplerType.
+    ** `sampler-manager-host-port`: the port of the sampling manager that can provide sampling strategy to this service.
+    ** `sender-binding`: The outbound socket binding to connecto the Agent.
+    ** `sender-endpoint`: The endpoint, like https://jaeger-collector:14268/api/traces
+    ** `sender-user`: The Basic Auth username to be added on Authorization headers for requests sent to the endpoint.
+    ** `sender-auth-password`: The Basic Auth password to be added on Authorization headers for requests sent to the endpoint.
+    ** `sender-auth-token`: The Auth Token to be added as "Bearer" on Authorization headers for requests sent to the endpoint.
+    ** `reporter-log-spans`: Boolean to indicates whether the reporter should log the spans.
+    ** `reporter-flush-interval`: The flush interval when reporting spans remotely in milliseconds.
+    ** `reporter-max-queue-size`: The reporter's maximum queue size.
+    ** `tracer_id_128bit`: Opt-in to use 128 bit traceIds. By default, uses 64 bits.
+    ** `tracer-tags`: A comma separated list of name = value tracer level tags, which get added to all reported spans. The value can also refer to an environment variable using the format ${envVarName:default}, where the :default is optional, and identifies a value to be used if the environment variable cannot be found.
+Those attributes will be 'reload-required' and the configuration will be only updated once the server is reloaded. A deployment would use the configuration that was set when it deployed.
+
+* Create a `default-tracer` attribute on the `microprofile-opentracing-smallrye` resource to define which tracer configuration is the default one to inject if none is specified in the deployment. Note that the default tracer configuration must exist, so you can't delete a tracer configuration that is referenced as default tracer configuration until that attribute is unset.
+
+* During the processing of the deployment, the module required for the tracer implementations will be added to the classloader as optional dependencies. Then the web.xml will be parsed and if an initial context parameter `smallrye.opentracing.tracer.configuration` is found then the configuration of the matching name will be used to create the tracer, otherwise the default configuration will be set as `smallrye.opentracing.tracer.configuration` value.
+
+* During the initialization of the deployment, the TracerResolver will be used to check for an internal defined tracer implementation at the deployment level. Otherwise the initial context parameter `smallrye.opentracing.tracer.configuration` will be used to obtain a Tracer that will be injected in the application. If this context parameter is not set by the user then it will use the default tracer if one exists, otherwise it should use the NoopTracer.
+If the value set is incorrect then the deployment will fail.
+
+Updating the configuration won't update the tracers already in use so it is required to redeploy the applications that will use the new tracer configuration.
+
+Since we need to be able to know what configuration is used by a deployment even if the model might have changed afterwards we will register a DeploymentModel for the `microprofile-opentracing-smallrye` subsystem with the `tracer-configuration-name` containing the name of the tracer configuration or the class found by the TracerResolver and a `tracer-configuration` (if one is found) corresponding to the configuration used at the time the tracer was instantiated.
+
+== Test Plan
+
+* Run the (updated) WildFly integration basic test suite and checks there are no failures related to opentracing tests.
+* Run the `microprofile-tck/opentracing` module from WildFly integration test suite and checks that there are no failures.
+* Add tests covering the injection from the subsystem.
+
+== Community Documentation
+
+* Community documentation for the `microprofile-opentracing-smallrye` subsystem in the admin guide needs to be updated to version 1.3.
+* Some community documentation is required to setup and run the MicroProfile OpenTracing TCK in WildFly.


rsvoboda:
QE contact is @zhantemirov

microprofile/WFLY-12758_integrate_opentracing.adoc
```diff
@@ -0,0 +1,107 @@
...
+=== Related Issues
+
+* https://issues.apache.org/jira/browse/EAP7-1351
+
+=== Dev Contacts
+
+* mailto:{email}[{author}]
+
+=== QE Contacts
+
```
zhantemirov: Should be mailto:szhantem@redhat.com[Sultan Zhantemirov]

microprofile/WFLY-12758_integrate_opentracing.adoc
```diff
@@ -0,0 +1,107 @@
...
+=== Testing By
+// Put an x in the relevant field to indicate if testing will be done by Engineering or QE. 
+// Discuss with QE during the Kickoff state to decide this
+[] Engineering
+
+[ ] QE
```
zhantemirov: As we know QE is responsible for testing => there should be a cross (x)

microprofile/WFLY-12758_integrate_opentracing.adoc
```diff
@@ -0,0 +1,107 @@
...
+    ** `reporter-max-queue-size`: The reporter's maximum queue size.
+    ** `tracer_id_128bit`: Opt-in to use 128 bit traceIds. By default, uses 64 bits.
+    ** `tracer-tags`: A comma separated list of name = value tracer level tags, which get added to all reported spans. The value can also refer to an environment variable using the format ${envVarName:default}, where the :default is optional, and identifies a value to be used if the environment variable cannot be found.
+
+* Create a `default-tracer` attribute on the `microprofile-opentracing-smallrye` resource to define which tracer configuration is the default one to inject if none is specified in the deployment.
+
+* During the processing of the deployment, the module required for the tracer implementations will be added to the classloader as optionnal dependencies. Then the web.xml will be parsed and if an initial context parameter `smallrye.opentracing.tracer.configuration` is found then the configuration of the matching name will be used to create the tracer, otherwise the default configuration will be set as `smallrye.opentracing.tracer.configuration` value.
```
zhantemirov: typo in _optional_ word

zhantemirov:
@ehsavoie despite of having a few formal inconsistencies in AD, it LGTM. 

Especially I'm satisfied with the _Test Plan_ section. There was an intention to add some lines about specific WAR archive testing but I think it could easily belong to the 1st point of this section. 

@rsvoboda another thing is https://github.com/rhoar-qe/thorntail-test-suite/tree/master/microprofile/microprofile-opentracing-1.3 tests. It has some additional test scenarios (http path regex) that are not covered in WildFly integration testsuite. Should it be mentioned in AD _Test Plan_ section too?   

Thanks.

rsvoboda:
We have to have it in our TP, I think presence in AD test plan section is optional.

microprofile/WFLY-12758_integrate_opentracing.adoc
```diff
@@ -0,0 +1,108 @@
...
+
+== Overview
+
+Currently the integration of Microprofile Opentracing is minimal as you can't configure the tracer using the WildFly Management Model.
+So we will implement a way to define several tracer configurations (Jaeger) to be used in deployments.
+One of those configuration will become a default one so that deployments that don't define which configuration to use will use this one.
+In this proposal we will only support Jaeger as Tracer implementation but it should be able to be easily extendable to support other Tracers implementations:
+* Support https://github.com/eclipse/microprofile-opentracing/blob/1.3/spec/src/main/asciidoc/microprofile-opentracing.asciidoc[Eclipse MicroProfile OpenTracing 1.3] for applications deployed in WildFly, including JAX-RS and CDI integration. Passing the TCK is required.
+* Support a concrete tracer, such as Jaeger's.
+* Allow applications to supply their own tracers via `TracerResolver`.
```
zhantemirov: Let me suggest removing this sentence. In my opinion, the wording "own tracers" could be a little bit confusing - other tracer implementations are beyond this RFE scope as we guarantee Jaeger tracer configuration ability only.

microprofile/WFLY-12758_integrate_opentracing.adoc
```diff
@@ -0,0 +1,108 @@
...
+:idprefix:
+:idseparator:       -
+
+== Overview
+
+Currently the integration of Microprofile Opentracing is minimal as you can't configure the tracer using the WildFly Management Model.
+So we will implement a way to define several tracer configurations (Jaeger) to be used in deployments.
+One of those configuration will become a default one so that deployments that don't define which configuration to use will use this one.
+In this proposal we will only support Jaeger as Tracer implementation but it should be able to be easily extendable to support other Tracers implementations:
+* Support https://github.com/eclipse/microprofile-opentracing/blob/1.3/spec/src/main/asciidoc/microprofile-opentracing.asciidoc[Eclipse MicroProfile OpenTracing 1.3] for applications deployed in WildFly, including JAX-RS and CDI integration. Passing the TCK is required.
```
zhantemirov: Looks like the bullet list is broken

microprofile/WFLY-12758_integrate_opentracing.adoc
```diff
@@ -0,0 +1,108 @@
...
+
+=== Non-Requirements
+
+Other tracer implementations will not be configureable from the subsystem management API.
+
+== Implementation Plan
+
+We will only cover the creation of a Jaeger tracer configuration. Other tracer implementations may come afterwards but are not currently 
+
+* Create a `jaeger-tracer` resource that define the configuration of a Jaeger tracer. This resource will expose all Jaeger Tracer configuration attributes:
```
zhantemirov: Is it necessary to have all the parameters settable (it would lead to the large number of tests)? Is there any stable set of parameters that won't be removed/modified in the future? Are all of them are mandatory or are there only few necessary parameters are need to be set?

microprofile/WFLY-12758_integrate_opentracing.adoc
```diff
@@ -0,0 +1,108 @@
...
+    ** `propagation`: The supported trace context propagation formats:
+        *** JAEGER: The default Jaeger trace context propagation format.
+        *** B3: The Zipkin B3 trace context propagation format.
```
zhantemirov: As I understood, we will be using existing jaeger-client with the Jaeger tracer. Do we need an option of setting Zipkin B3 trace context propagation format?

zhantemirov:
@ehsavoie 

1) Should I reload the application server when the Jaeger tracer configuration or default tracer implementation is changed? Or do the changes applied dynamically?    

2) Does WildFly have an option to define tracer properties via _microprofile-config.properties_ file? By defining I mean extracting properties from that file and setting them for the Tracer. I had no luck with a WAR archive (with _META-INF/microprofile-config.properties_) and _@Inject_ tracer instance - the tracer behaviour was not in compliance with desired properties.  

3) What if the jaeger-agent is down for a couple of minutes - are the spans from jaeger-client cached somehow or are they just thrown to nowhere? As soon as the jaeger-agent is live again - will these spans appear there?   

4) What if the customer wants to upgrade Jaeger agent version? Will there be an issue with the Jaeger client and agent versions mismatch?   

5) What if the customer wants to trace requests at certain time (e.g. a few hours per day) only? Is there an option to provide this functionality?

Thank you.

rsvoboda:
@zhantemirov please revoke your approval as you added new comments

microprofile/WFLY-12758_integrate_opentracing.adoc
ehsavoie: Hum, the support for TracerResolver was already in place, this would be a regression in terms of features over the previous releases. I'm talking tracer implementations here with their configuration being only supported by the user/TracerResolver mechanism.

microprofile/WFLY-12758_integrate_opentracing.adoc
ehsavoie: ok

microprofile/WFLY-12758_integrate_opentracing.adoc
ehsavoie: Well customer might wonder why they can't integrate WildFly with Zipkin anymore

rsvoboda: All the requirements I have seen are to integrate with Jaeger.
Because of that, there are no plans to test with Zipkin so it shouldn't be promoted.

microprofile/WFLY-12758_integrate_opentracing.adoc
rsvoboda: What are examples of own tracers users can supply ?

rsvoboda: @jstourac and @okotek you worked on opentracing stuff before [1], please comment on this topic

[1] https://github.com/wildfly/wildfly-proposals/pull/106 and https://github.com/wildfly/wildfly-proposals/pull/236

ehsavoie:
@zhantemirov 
1/ A reload is required as a new Tracer instance is to be created and injected from the updated configuration.
2/ No, I'm not aware of this being a requirement.
3/ I don't know, this is handled at the jaeger client level but to my understanding it should reconnect
4/ the user should use a jaeger agent compatible with our version: this depends on jaeger, nothing much we can do there.
5/ No

microprofile/WFLY-12758_integrate_opentracing.adoc
rsvoboda: @jstourac and @okotek you worked on opentracing stuff before [1], please comment on Zipkin topic

[1] https://github.com/wildfly/wildfly-proposals/pull/106 and https://github.com/wildfly/wildfly-proposals/pull/236

microprofile/WFLY-12758_integrate_opentracing.adoc
```diff
@@ -0,0 +1,111 @@
...
+:icons:             font
+:keywords:          microprofile,opentracing,observability,tracing,openshift
+:idprefix:
+:idseparator:       -
+
+== Overview
+
+Currently the integration of Microprofile Opentracing is minimal as you can't configure the tracer using the WildFly Management Model.
+So we will implement a way to define several tracer configurations (Jaeger) to be used in deployments.
+One of those configuration will become a default one so that deployments that don't define which configuration to use will use this one.
```
rsvoboda: What happens when deployments defines configuration to use but that configuration is not present on the server ?

not present can mean
 - configuration not yet present
 - configuration not present anymore (new one was introduced)
 - typo in configuration name - web.xml or server config

ehsavoie: Then it fails since we create a MSC service dependency on that configuration. If none is defined we fall back to the default one if one is defined otherwise no configuration is provided

microprofile/WFLY-12758_integrate_opentracing.adoc
```diff
@@ -0,0 +1,111 @@
...
+    ** `reporter-max-queue-size`: The reporter's maximum queue size.
+    ** `tracer_id_128bit`: Opt-in to use 128 bit traceIds. By default, uses 64 bits.
+    ** `tracer-tags`: A comma separated list of name = value tracer level tags, which get added to all reported spans. The value can also refer to an environment variable using the format ${envVarName:default}, where the :default is optional, and identifies a value to be used if the environment variable cannot be found.
+
+* Create a `default-tracer` attribute on the `microprofile-opentracing-smallrye` resource to define which tracer configuration is the default one to inject if none is specified in the deployment.
+
+* During the processing of the deployment, the module required for the tracer implementations will be added to the classloader as optional dependencies. Then the web.xml will be parsed and if an initial context parameter `smallrye.opentracing.tracer.configuration` is found then the configuration of the matching name will be used to create the tracer, otherwise the default configuration will be set as `smallrye.opentracing.tracer.configuration` value.
+
```
rsvoboda: Lets assume WF server defines 3 Jeager configurations, first of them is marked as default, server running in production.

case a) Admin comes and marks second one as a default. What happens with deployments which do not explicitly define tracer configuration ?

case b) Admin comes and marks second one as a default and removes first one. What happens with deployments which do not explicitly define tracer configuration? Can be first one removed ?

case c) Admin comes and removes first one. Can be first one removed ? Which config becomes the default one ? What happens with deployments which do not explicitly define tracer configuration?

ehsavoie: case a) nothing if those deployments are running, they still use old configuration until redeployment / reload
case b) nothing if those deployments are running, they still use old configuration until redeployment / reload.  We should be able to remove the configuration since the service would still be there.
case c) You need to undefine default when removing first one. No default is supported, they would get no configuration and thus rely on some external mean of configuration (env variables for example) or use a NoopTracer (as per specs).

microprofile/WFLY-12758_integrate_opentracing.adoc
ehsavoie: This is zipkin format, this can be supported at the jaeger agent level. i'm not talking about zipking agent there.

microprofile/WFLY-12758_integrate_opentracing.adoc
rsvoboda: > they still use old configuration until redeployment / reload

d) Is there a way to get information which configuration is in reality used by deployments?
Motivation is to detect such orphans which are using non-existing / non-relevant settings.

Based from your feedback I am thinking about new scenarios - reconfiguration.

e) When admin performs changes to the config (e.g. sampling rate change) are these changes propagated to all the relevant deployments or reload is needed ?

f) Will any change to configuration return "reload required" ?
This would ensure consistent state between server configuration for jaeger vs. what jaeger config is used in apps

microprofile/WFLY-12758_integrate_opentracing.adoc
rsvoboda: I understand that, one agent, but 2 possible servers (jaeger, zipkin) to consume traces

rsvoboda:
How will this subsystem relate to OpenTelemetry ? And what are the plans for OpenTracing vs. OpenTelemetry from server management perspective?

I would like to avoid jbossweb vs undertow situation when for end users it was just "web" layer and they hated that the need to deal with implementation (jbossweb, undertow) specific configuration.

We know OpenTelemetry will come, we should have some stuff designed having that in mind, for example configuration attributes.

ehsavoie:
To my understanding Microprofile OpenTracing should have a  bridge/shim to OpenTelemetry.
Those configuration attributes are Jaeger client configuration attributes, I'm not sure this is going to change a lot (like how to connect to the agent etc.). The changes would be at java code level mostly and if things go really bad we have only one resource to deprecate.

microprofile/WFLY-12758_integrate_opentracing.adoc
ehsavoie: d) Well it is either specified in the deployment via an init-parameter or it is injected. We could add a deployment subsystem for this
e) reload is required
f) yes I should add this to the doc

microprofile/WFLY-12758_integrate_opentracing.adoc
ehsavoie: https://www.jaegertracing.io/docs/1.15/features/#backwards-compatibility-with-zipkin

microprofile/WFLY-12758_integrate_opentracing.adoc
ehsavoie: Yes thjey all could be set but none is mandatory

microprofile/WFLY-12758_integrate_opentracing.adoc
ehsavoie: fixed

microprofile/WFLY-12758_integrate_opentracing.adoc
ehsavoie: some specific jaeger tracer implementation for example.

microprofile/WFLY-12758_integrate_opentracing.adoc
rsvoboda: d) is about deployments which do not explicitly define tracer configuration, follow up question for a) and b)

rsvoboda:
Will be Jaeger client able to communicate with OpenTelemetry server?
Or there will be a different client for it?

microprofile/WFLY-12758_integrate_opentracing.adoc
ehsavoie: Ok, well the trick is the tracer is created by the deployment 'itself' via a ServletContextListener. You would want to have some kind of way to know the source of the tracer configuration in the deployment model ?

rsvoboda: I want to know if there is / will be a way to know which configuration is used by deployments including all the parameters.
Asking because the configuration can change in the management model, but if "reload required" will be asked then config in management model and config used by applications will be in sync.

ehsavoie:
A new OpenTelemetry client will be required  according https://github.com/open-telemetry/opentelemetry-java as Jaeger will be supported by OpenTelemetry SDK. At this point I'm not sure about how this will be configurable and how much will be specific.

microprofile/WFLY-12758_integrate_opentracing.adoc
ehsavoie: d) ok we could add the name to the deployment subsystem so that the used tracer configuration is known.

ehsavoie: BUT there is the case of a tracer resolver. This is done via ServiceLoader by the deployment itself

ehsavoie: Would something like this suits you ? "microprofile-opentracing-smallrye" => {
                "tracer-configuration" => {
                    "sampler-type" => "const",
                    "sampler-param" => 1.0,
                    "propagation" => undefined,
                    "sampler-manager-host-port" => undefined,
                    "sender-binding" => undefined,
                    "sender-endpoint" => undefined,
                    "sender-auth-token" => undefined,
                    "sender-auth-user" => undefined,
                    "sender-auth-password" => undefined,
                    "reporter-log-spans" => undefined,
                    "reporter-flush-interval" => undefined,
                    "reporter-max-queue-size" => undefined,
                    "tags" => undefined,
                    "tracer_id_128bit" => undefined
                },
                "tracer-configuration-name" => "jaeger"
            }

rsvoboda: Looks good

PR #263, opened 2019-11-13
[WFCORE-1110] Add support for custom headers to the HTTP management endpoints.

darranl:
https://issues.jboss.org/browse/WFCORE-1110

### DIFF
--- a/management/WFCORE-1110_Custom_HTTP_Management_Headers.adoc
+++ b/management/WFCORE-1110_Custom_HTTP_Management_Headers.adoc
@@ -1,0 +1,84 @@
+= <INSERT TITLE HERE>
+:author:            Darran Lofthouse
+:email:             darran.lofthouse@redhat.com
+:toc:               left
+:icons:             font
+:idprefix:
+:idseparator:       -
+
+== Overview
+
+== Issue Metadata
+
+=== Issue
+
+* https://issues.jboss.org/browse/WFCORE-1110[WFCORE-1110]
+
+=== Related Issues
+
+* https://issues.jboss.org/browse/WFLY-12787[WFLY-12787]
+* https://issues.jboss.org/browse/EAP7-1335[EAP7-1335]
+=== Dev Contacts
+
+* mailto:{email}[{author}]
+
+=== QE Contacts
+
+=== Testing By
+// Put an x in the relevant field to indicate if testing will be done by Engineering or QE. 
+// Discuss with QE during the Kickoff state to decide this
+[ ] Engineering
+
+[X] QE
+
+=== Affected Projects or Components
+
+=== Other Interested Projects
+
+== Requirements
+
+=== Hard Requirements
+
+The HTTP management endpoint of the application server already returns a pre-defined set of HTTP headers in all of the responses that are sent to the clients, however it is sometimes desirable that additional headers can be returned to satisfy the requirements of an environment the application server is installed within.  This enhancement is to add support for an end user to specify a custom set of HTTP headers that should be returned within each response.
+
+Predominantly we have two endpoints made available from the HTTP management endpoint by default.
+
+ * `/management` - The endpoint for servicing management requests over HTTP.
+ * `/console` - The endpoint to provide the admin console initial page and complete JavaScript.
+
+As we have different endpoints it may be desirable for them to return different headers, as an example some headers are only applicable to the handling of pages or interpretation of the JavaScript so should only be applied to the `/console` endpoint whilst other headers may be applicable to both.
+
+Configuration will be added to the HTTP management endpoint to allow headers to be specified, an example configuration will be added later but in general the following will be apply: -
+
+ * The configuration will allow the end user to specify which context the headers apply to or if they apply to both contexts.
+ * This will be a new `custom-headers` attribute on the `/core-service=management/management-interface=http-interface` resource.
+ * Within the `custom-headers` attribute it will be possible to specify a set of headers represented as an ordered list of key / value pairs against a specific patch.
+ * If a header is already set by the corresponding endpoint and the configuration overrides the value the configuration will have higher precedence.
+ * If a header is specified to apply to both context and is redefined for a specific context the definition for the specific context will have higher precedence. 
+
+
+=== Nice-to-Have Requirements
+
+=== Non-Requirements
+
+This enhancement is only supporting the default endpoints, dynamic endpoints are excluded.  Typically the headers are most appropriate for and endpoint providing JavaScript.
+
+== Test Plan
+
+== Community Documentation
+
+Community documentation will need to be added describing how to configure custome HTTP headers that should be returned on every request.
+
+== Release Note Content
+
+TODO
+
+////
+Draft verbiage for up to a few sentences on the feature for inclusion in the
+Release Note blog article for the release that first includes this feature. 
+Example article: http://wildfly.org/news/2018/08/30/WildFly14-Final-Released/.
+This content will be edited, so there is no need to make it perfect or discuss
+what release it appears in.  "See Overview" is acceptable if the overview is
+suitable. For simple features best covered as an item in a bullet-point list 
+of features containing a few words on each, use "Bullet point: <The few words>" 
+////
--- a/management/WFCORE-1110_Custom_HTTP_Management_Headers.adoc
+++ b/management/WFCORE-1110_Custom_HTTP_Management_Headers.adoc
@@ -73,6 +73,7 @@ Some headers are integral to the correct operation of the server and so can not 
  * `Content-Length`
  * `Content-Type`
  * `Date`
+ * `Transfer-Encoding`
 
 If an administrator attempts to set these headers an error will be reported.
 
--- a/management/WFCORE-1110_Custom_HTTP_Management_Headers.adoc
+++ b/management/WFCORE-1110_Custom_HTTP_Management_Headers.adoc
@@ -1,4 +1,4 @@
-= <INSERT TITLE HERE>
+= WFCORE-1110 Custom HTTP Headers for the HTTP Management Interface
 :author:            Darran Lofthouse
 :email:             darran.lofthouse@redhat.com
 :toc:               left
@@ -52,7 +52,7 @@ Configuration will be added to the HTTP management endpoint to allow headers to 
 
  * The configuration will allow the end user to specify which context the headers apply to or if they apply to both contexts.
  * This will be a new `custom-headers` attribute on the `/core-service=management/management-interface=http-interface` resource.
- * Within the `custom-headers` attribute it will be possible to specify a set of headers represented as an ordered list of key / value pairs against a specific patch.
+ * Within the `custom-headers` attribute it will be possible to specify a set of headers represented as an ordered list of key / value pairs against a specific path.
  * If a header is already set by the corresponding endpoint and the configuration overrides the value the configuration will have higher precedence.
  * If a header is specified to apply to both context and is redefined for a specific context the definition for the specific context will have higher precedence. 
 
--- a/management/WFCORE-1110_Custom_HTTP_Management_Headers.adoc
+++ b/management/WFCORE-1110_Custom_HTTP_Management_Headers.adoc
@@ -43,27 +43,33 @@ The HTTP management endpoint of the application server already returns a pre-def
 
 === Hard Requirements
 
-Predominantly we have two endpoints made available from the HTTP management endpoint by default.
+The HTTP management interface provides access to a number of contexts, our primary contexts are accessed at the following paths: -
 
  * `/management` - The endpoint for servicing management requests over HTTP.
  * `/console` - The endpoint to provide the admin console initial page and complete JavaScript.
+ 
+Subsystems can also dynmically register additional contexts to dynamically provide additional functionality or content over the HTTP management interface. 
 
 As we have different endpoints it may be desirable for them to return different headers, as an example some headers are only applicable to the handling of pages or interpretation of the JavaScript so should only be applied to the `/console` endpoint whilst other headers may be applicable to both.
 
 Configuration will be added to the HTTP management endpoint to allow headers to be specified, an example configuration will be added later but in general the following will be apply: -
 
- * The configuration will allow the end user to specify which context the headers apply to or if they apply to both contexts.
+ * The configuration will allow an end user to specify headers to be added based on a specified path prefix.
  * This will be a new `custom-headers` attribute on the `/core-service=management/management-interface=http-interface` resource.
- * Within the `custom-headers` attribute it will be possible to specify a set of headers represented as an ordered list of key / value pairs against a specific path.
- * If a header is already set by the corresponding endpoint and the configuration overrides the value the configuration will have higher precedence.
- * If a header is specified to apply to both context and is redefined for a specific context the definition for the specific context will have higher precedence. 
+ * Within the `custom-headers` attribute it will be possible to specify a set of headers using key / value pairs against a specific path.
+ * For a specific path all keys must be unique i.e. a single header can not be added multiple times.
+ * All paths will be assumed to be prefixes starting with `/` - any paths missing a leading `/` will be handled as though the leading `/` has been specified.
+
+As the matching is path based it is possible that a single request can match against multiple path definitions e.g. `/` and `/console`.  All paths will be evaluated, if a header is defined against multiple paths the value for the most specific path will take precedence  i.e. if a header was defined both for `/` and `/console` if the request was to `/console` the header defined against the `/console` path would be used.  However in this example if the request was to `/management` then the header defined against `/` would be applied to the response.
+
+TODO can add a table here with some example configuration and the results of various requests.
 
 
 === Nice-to-Have Requirements
 
 === Non-Requirements
 
-This enhancement is only supporting the default endpoints, dynamic endpoints are excluded.  Typically the headers are most appropriate for an endpoint providing JavaScript.
+As the contexts supported by the HTTP management interface can be added dynamically by subsystems no validation will be performed to verify that the specified paths are genuinely reachable paths as this information is not available to us at the time we perform model validation.
 
 == Test Plan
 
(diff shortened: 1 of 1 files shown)


management/WFCORE-1110_Custom_HTTP_Management_Headers.adoc
```diff
@@ -0,0 +1,84 @@
...
+ * `/management` - The endpoint for servicing management requests over HTTP.
+ * `/console` - The endpoint to provide the admin console initial page and complete JavaScript.
+
+As we have different endpoints it may be desirable for them to return different headers, as an example some headers are only applicable to the handling of pages or interpretation of the JavaScript so should only be applied to the `/console` endpoint whilst other headers may be applicable to both.
+
+Configuration will be added to the HTTP management endpoint to allow headers to be specified, an example configuration will be added later but in general the following will be apply: -
+
+ * The configuration will allow the end user to specify which context the headers apply to or if they apply to both contexts.
+ * This will be a new `custom-headers` attribute on the `/core-service=management/management-interface=http-interface` resource.
+ * Within the `custom-headers` attribute it will be possible to specify a set of headers represented as an ordered list of key / value pairs against a specific patch.
```
bstansberry: s/patch/path

bstansberry: So each element in the list would consist of a node encapsulating

1) the relevant path
2) header name
3) header value

?

The path being any possible path, starting with "/" to cover all contexts, and then going beyond, e.g.  a specific context or even /management/subsystem/logging/console-handler.  But nothing fancy like regex matching.

(I'm not advocating for /management/subsystem/logging/console-handler, although doing it is fine if it's easier. I'm just clarifying.)

darranl: I will add a clarification for this I think we will restrict it to: -

- /
- /console
- /management

Keeping the options simple, at a later point deeper paths / redgex etc could be considered but TBH I think the root contexts will be sufficient - at this level we pick up that one handles JavaScript and once handles API calls with an option to apply headers to both equally.

bstansberry: Sounds good.

management/WFCORE-1110_Custom_HTTP_Management_Headers.adoc
```diff
@@ -0,0 +1,84 @@
+= WFCORE-1110 Custom HTTP Headers for the HTTP Management Interface
+:author:            Darran Lofthouse
+:email:             darran.lofthouse@redhat.com
+:toc:               left
+:icons:             font
+:idprefix:
+:idseparator:       -
+
+== Overview
+
```
tterem: Please add basic overview of this feature.

management/WFCORE-1110_Custom_HTTP_Management_Headers.adoc
```diff
@@ -0,0 +1,84 @@
...
+=== Issue
+
+* https://issues.jboss.org/browse/WFCORE-1110[WFCORE-1110]
+
+=== Related Issues
+
+* https://issues.jboss.org/browse/WFLY-12787[WFLY-12787]
+* https://issues.jboss.org/browse/EAP7-1335[EAP7-1335]
+=== Dev Contacts
```
tterem: Please put empty line above `=== Dev Contacts`. It isn't displaying correctly now.

management/WFCORE-1110_Custom_HTTP_Management_Headers.adoc
```diff
@@ -0,0 +1,84 @@
...
+
+* mailto:{email}[{author}]
+
+=== QE Contacts
+
```
tterem: I was assigned as QE contact for this.
`* mailto:tterem@redhat.com[Tomas Terem]`

management/WFCORE-1110_Custom_HTTP_Management_Headers.adoc
```diff
@@ -0,0 +1,84 @@
...
+ * Within the `custom-headers` attribute it will be possible to specify a set of headers represented as an ordered list of key / value pairs against a specific path.
+ * If a header is already set by the corresponding endpoint and the configuration overrides the value the configuration will have higher precedence.
+ * If a header is specified to apply to both context and is redefined for a specific context the definition for the specific context will have higher precedence. 
+
+
+=== Nice-to-Have Requirements
+
+=== Non-Requirements
+
+This enhancement is only supporting the default endpoints, dynamic endpoints are excluded.  Typically the headers are most appropriate for and endpoint providing JavaScript.
```
tterem: and endpoint -> an endpoint ?

management/WFCORE-1110_Custom_HTTP_Management_Headers.adoc
```diff
@@ -0,0 +1,133 @@
...
+----
+
+=== Nice-to-Have Requirements
+
+=== Non-Requirements
+
+As the contexts supported by the HTTP management interface can be added dynamically by subsystems no validation will be performed to verify that the specified paths are genuinely reachable paths as this information is not available to us at the time we perform model validation.
+
+== Test Plan
+
```
tterem: I am currently preparing the test plan, however some high-level description of the test plan should be added here as well.

management/WFCORE-1110_Custom_HTTP_Management_Headers.adoc
tterem: This comment is still valid.

management/WFCORE-1110_Custom_HTTP_Management_Headers.adoc
```diff
@@ -0,0 +1,133 @@
...
+                        "expressions-allowed" => true,
+                        "required" => true,
+                    },
+                    "headers" => {
+                        "type" => LIST,
+                        "description" => "The headers to set for the matched path.",
+                        "expressions-allowed" => false,
+                        "required" => false,
+                        "value-type" => {
+                            "header" => {
```
tterem: "header" -> "name"

(based on discussion/commit on WF-CORE PR)

management/WFCORE-1110_Custom_HTTP_Management_Headers.adoc
```diff
@@ -0,0 +1,133 @@
...
+                            }
+                        }
+                    }
+                }
+----
+
+The following example hows how the interface can be configured to add a custom header for the `/management` endpoint: -
+
+----
+/] /core-service=management/management-interface=http-interface:write-attribute(name=constant-headers, value=[{path=/management, headers=[{header=X-Header, value=HeaderValue}]}])
```
tterem: "header" -> "name"

(based on discussion/commit on WF-CORE PR)

management/WFCORE-1110_Custom_HTTP_Management_Headers.adoc
```diff
@@ -0,0 +1,133 @@
...
+
+== Test Plan
+
+== Community Documentation
+
+Community documentation will need to be added describing how to configure custome HTTP headers that should be returned on every request.
+
+== Release Note Content
+
+TODO
```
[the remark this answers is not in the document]
darranl: I know that is why it says TODO - this section will be completed at the very end once I am happy no further changes are being applied in advance of the merge.

management/WFCORE-1110_Custom_HTTP_Management_Headers.adoc
```diff
@@ -0,0 +1,133 @@
...
+=== Nice-to-Have Requirements
+
+=== Non-Requirements
+
+As the contexts supported by the HTTP management interface can be added dynamically by subsystems no validation will be performed to verify that the specified paths are genuinely reachable paths as this information is not available to us at the time we perform model validation.
+
+== Test Plan
+
+== Community Documentation
+
```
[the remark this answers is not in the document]
darranl: I will update this section once I have started working on the docs.

management/WFCORE-1110_Custom_HTTP_Management_Headers.adoc
darranl: I will add a little bit for the tests I am adding to WildFly Core but feel free to send a PR against this branch if you would like to add some content to this block https://github.com/darranl/wildfly-proposals/tree/WFCORE-1110

management/WFCORE-1110_Custom_HTTP_Management_Headers.adoc
```diff
@@ -0,0 +1,133 @@
...
+ * The configuration will allow an end user to specify headers to be added based on a specified path prefix.
+ * This will be a new `custom-headers` attribute on the `/core-service=management/management-interface=http-interface` resource.
+ * Within the `custom-headers` attribute it will be possible to specify a set of headers using key / value pairs against a specific path.
+ * All paths will be assumed to be prefixes starting with `/` - any paths missing a leading `/` will be handled as though the leading `/` has been specified.
+ * If a specific header is defined multiple times it will be added to the response multiple times.
+ * Configured headers will not override any headers added by the corresponding endpoint.
+
+
+The new attribute to be added to the `/core-service=management/management-interface=http-interface` resource will be called `constant-headers` and will be defined as: -
```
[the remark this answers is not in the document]
darranl: Just updated the PR to use 'constant-headers' for the attribute name - TBH as there is no dynamic nature to these it felt more appropriate.

darranl:
Just sent another update, some of the final pieces such as community docs and release notes will clarify those at the end.

management/WFCORE-1110_Custom_HTTP_Management_Headers.adoc
jmesnil: Should the `/health` and `/metrics` also be covered?

Looking at the implementation, that's actually the case

management/WFCORE-1110_Custom_HTTP_Management_Headers.adoc
```diff
@@ -0,0 +1,139 @@
...
+As we have different endpoints it may be desirable for them to return different headers, as an example some headers are only applicable to the handling of pages or interpretation of the JavaScript so should only be applied to the `/console` endpoint whilst other headers may be applicable to both.
+
+Configuration will be added to the HTTP management endpoint to allow headers to be specified, the following will be apply when applying the headers: -
+
+ * The configuration will allow an end user to specify headers to be added based on a specified path prefix.
+ * This will be a new `constant-headers` attribute on the `/core-service=management/management-interface=http-interface` resource.
+ * Within the `constant-headers` attribute it will be possible to specify a set of headers using key / value pairs against a specific path.
+ * All paths will be assumed to be prefixes starting with `/` - any paths missing a leading `/` will be handled as though the leading `/` has been specified.
+ * If a specific header is defined multiple times it will be added to the response multiple times.
+ * Configured headers will not override any headers added by the corresponding endpoint.
```
tterem: I think we need to clarify what these two rules means for existing headers:
Based on first rule, if we add headers {A: abc, A: 123} then both will be included in a response.
Based on second rule, if /management endpoint already adds {Content-Length: 188} and we configure constant-headers {Content-Length: abc, Content-Length: 123} then the original value won't be overriden. 

I think the analysis should clearly say what happens in this case. "will not override" could mean that only existing value would be used, but also that all three would be used.

management/WFCORE-1110_Custom_HTTP_Management_Headers.adoc
```diff
@@ -0,0 +1,139 @@
...
+----
+/] /core-service=management/management-interface=http-interface:write-attribute(name=constant-headers, value=[{path=/management, headers=[{name=X-Header, value=HeaderValue}]}])
+----
+
+The `write-attribute` operation will trigger a `reload-required` state as the management interface will need to be re-initialised.
+
+=== Nice-to-Have Requirements
+
+=== Non-Requirements
+
```
tterem: I think we should mention here that the server won't give any special meaning to added headers and that this feature isn't meant to configure response content. I mean for example a situation when user adds constant header {Content-Type: application/yaml}, they can't expect management port to return content in yaml.

management/WFCORE-1110_Custom_HTTP_Management_Headers.adoc
```diff
@@ -0,0 +1,139 @@
...
+ * `/management` - The endpoint for servicing management requests over HTTP.
+ * `/console` - The endpoint to provide the admin console initial page and complete JavaScript.
+ 
+Subsystems can also dynmically register additional contexts to dynamically provide additional functionality or content over the HTTP management interface. 
+
+As we have different endpoints it may be desirable for them to return different headers, as an example some headers are only applicable to the handling of pages or interpretation of the JavaScript so should only be applied to the `/console` endpoint whilst other headers may be applicable to both.
+
+Configuration will be added to the HTTP management endpoint to allow headers to be specified, the following will be apply when applying the headers: -
+
+ * The configuration will allow an end user to specify headers to be added based on a specified path prefix.
```
tterem: We should add [this clarification ](https://github.com/wildfly/wildfly/pull/12809/files#diff-f724c56d2b4f9096cfe9b9739d0c889cR98)here as well.

darranl: For the analysis we need to focus on communicating the initial intent, the whole point of these documents was to share our plans - if we are now cross referencing the community documentation as a recommendation for a change then everyone on the feature change already has access to that information as it is in the documentation.

management/WFCORE-1110_Custom_HTTP_Management_Headers.adoc
darranl: Going to work through this one today with one of the WIPs in relation to the header handling.

A number of headers that we already set are known to be problematic, we may just need to end up identifying a set of headers we ban from being set,

Overall all our implementation can do is pro-actively set a set of headers as the request is inbound, how the internal Undertow code, management endpoints and extenstions respond to these headers is not under our direct control.

management/WFCORE-1110_Custom_HTTP_Management_Headers.adoc
darranl: Yes all paths below `/` are covered.

management/WFCORE-1110_Custom_HTTP_Management_Headers.adoc
tterem: It's good that it's in the documentation, but I believe that the analysis should describe new feature in a way that is clear so there was no misunderstandings later. This behaviour will need to be tested and it isn't obvious from the analysis now so that's why I wanted it included.

management/WFCORE-1110_Custom_HTTP_Management_Headers.adoc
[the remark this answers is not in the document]
bstansberry: FWIW that's fine with me. I added this section to the analysis template to help myself with writing the WildFly release announcement and I'm not wanting it to block progress in any way. I've written lots of release announcements without this.

management/WFCORE-1110_Custom_HTTP_Management_Headers.adoc
```diff
@@ -65,9 +65,18 @@ Configuration will be added to the HTTP management endpoint to allow headers to
...
+Any headers specified within the configuration will be applied just before the response is sent to the client, this means any headers specified will override any of the headers set by the endpoint which handled the request.
 
-The headers `Connection` and `Date` define the behaviour of the connection management of the endpoint and are integral to correctly caching resources so it will not be possible for these to be set, if an administrator attempts to set these headers an error will be reported.  An error will also be reported if the name of a header specified includes invalid characters as defined by the HTTP specification RFCs.
+Some headers are integral to the correct operation of the server and so can not be overridden, these headers are: -
+
+ * `Connection`
+ * `Content-Length`
+ * `Content-Type`
+ * `Date`
+
```
tterem: What about adding 'Transfer-Encoding' to the list? Also is this list complete or can you think of some other headers to add? I believe that if we want to be backward-compatible, we won't be able to change the list in the future, right?

darranl: Create a WFWIP please for that one and will get it added.

tterem: I've created WFWIP issue.

management/WFCORE-1110_Custom_HTTP_Management_Headers.adoc
```diff
@@ -0,0 +1,150 @@
...
+                                "description" => "The value to set for the HTTP header.",
+                                "expressions-allowed" => true,
+                                "required" => true,
+                            }
+                        }
+                    }
+                }
+----
+
+The following example hows how the interface can be configured to add a constant header for the `/management` endpoint: -
```
tterem: hows -> shows

management/WFCORE-1110_Custom_HTTP_Management_Headers.adoc
```diff
@@ -0,0 +1,150 @@
...
+=== Hard Requirements
+
+This enhancement is to allow an administrator to define a constant set of headers that can be returned from requests handled via the management HTTP interface of the application server.
+ 
+The HTTP management interface provides access to a number of contexts, our primary contexts are accessed at the following paths: -
+
+ * `/management` - The endpoint for servicing management requests over HTTP.
+ * `/console` - The endpoint to provide the admin console initial page and complete JavaScript.
+ 
+Subsystems can also dynmically register additional contexts to dynamically provide additional functionality or content over the HTTP management interface.
```
tterem: Subsystems can also dynmically -> dynamically

management/WFCORE-1110_Custom_HTTP_Management_Headers.adoc
```diff
@@ -0,0 +1,150 @@
...
+The HTTP management interface provides access to a number of contexts, our primary contexts are accessed at the following paths: -
+
+ * `/management` - The endpoint for servicing management requests over HTTP.
+ * `/console` - The endpoint to provide the admin console initial page and complete JavaScript.
+ 
+Subsystems can also dynmically register additional contexts to dynamically provide additional functionality or content over the HTTP management interface. 
+
+As we have different endpoints it may be desirable for them to return different headers, as an example some headers are only applicable to the handling of pages or interpretation of the JavaScript so should only be applied to the `/console` endpoint whilst other headers may be applicable to both.
+
+Configuration will be added to the HTTP management endpoint to allow headers to be specified, the following will be apply when applying the headers: -
```
tterem: apply -> applied

management/WFCORE-1110_Custom_HTTP_Management_Headers.adoc
```diff
@@ -0,0 +1,150 @@
...
+
+ * The configuration will allow an end user to specify headers to be added based on a specified path prefix.
+ * This will be a new `constant-headers` attribute on the `/core-service=management/management-interface=http-interface` resource.
```
tterem: This -> There

PR #268, opened 2019-11-26
[WFLY-12825]: Artemis network health check feature not configurable from WildFly

ehsavoie:
Jira: https://issues.jboss.org/browse/WFLY-12825

### DIFF
--- a/messaging/WFLY-12825_Artemis_network_health_check_feature.adoc
+++ b/messaging/WFLY-12825_Artemis_network_health_check_feature.adoc
@@ -1,0 +1,93 @@
+= Artemis network health check feature not configurable from WildFly
+:author:            Emmanuel Hugonnet
+:email:             ehugonne@redhat.com
+:toc:               left
+:icons:             font
+:idprefix:          messaging,jms,network,health
+:idseparator:       -
+
+== Overview
+
+== Issue Metadata
+
+=== Issue
+
+* https://issues.jboss.org/browse/WFLY-12825[WFLY-12825]
+
+=== Related Issues
+
+=== Dev Contacts
+
+* mailto:{email}[{author}]
+
+=== QE Contacts
+
+=== Testing By
+// Put an x in the relevant field to indicate if testing will be done by Engineering or QE. 
+// Discuss with QE during the Kickoff state to decide this
+[X] Engineering
+
+[ ] QE
+
+=== Affected Projects or Components
+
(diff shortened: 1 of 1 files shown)


messaging/WFLY-12825_Artemis_network_health_check_feature.adoc
```diff
@@ -0,0 +1,93 @@
...
+* Add a LIST of STRING `network-check-list` attribute.Make sure you understand your network topology as this is meant to check if your network is up. Using IPs that could eventually disappear or be partially visible may defeat the purpose. You can use a list of multiple IPs, any successful ping will make the server OK
+* Add a LIST of STRING `network-check-URL-list` attribute if you want to use an HTTP server to validate the network
+* Add a STRING `network-check-ping-command` attribute for testing the network in IPv4, with a default value set to `ping -c 1 -t %d %s`.
+* Add a STRING `network-check-ping6-command` attribute for testing the network in IPv6, with a default value set to `ping6 -c 1 %2$s`.
+
+Those attributes will be grouped in the `network-health` attribute group for the https://wildscribe.github.io/WildFly/18.0/subsystem/messaging-activemq/server/[/subsystem=messaging-activemq/server resource].
+
+== Test Plan
+
+The tests will check that setting the attribute values and reloading will return the new values and that the server will start properly.
```
honza-kasik, line 76: IMHO it is also important to check if the command supplied to `network-check-ping6?-command` is working properly on various platforms. @ehsavoie WDYT?

ehsavoie: Those are the one defined as default by artemis

honza-kasik: @ehsavoie So WildFly will load these values from Artemis or these values will be defined in WildFly management model? I see users might have issues with this if the platform's ping command doesn't support default options.

ehsavoie: WildFly defines its values in the model, even if those came initially from Artemis.

honza-kasik: Thanks, I guess this can be solved in separate test plan document for the product.

honza-kasik (approved):
Thank you - this looks good from QE POV.

PR #269, opened 2019-11-26
[WFLY-9213]: Implement the Pause method for a Topic via Management APIs

ehsavoie:
Jira: https://issues.jboss.org/browse/WFLY-9213

### DIFF
--- a/messaging/WFLY-9213_Implement_the_Pause_method_for_a_Topic.adoc
+++ b/messaging/WFLY-9213_Implement_the_Pause_method_for_a_Topic.adoc
@@ -1,0 +1,144 @@
+= Implement the Pause method for a Topic via Management APIs
+:author:            Emmanuel Hugonnet
+:email:             ehugonne@redhat.com
+:toc:               left
+:icons:             font
+:idprefix:          messaging,jms
+:idseparator:       -
+
+== Overview
+
+
+An Artemis Queue can be paused but there is nothing for a Topic. The goal of this feature is to be able to pause a Topic by pausing all its consumers . New subscriptions being registered while the topic is in pause will be paused also.
+This is implemented in Apache Artemis and we need to expose this feature in the WildFly management API.
+
+== Issue Metadata
+
+=== Issue
+
+* https://issues.jboss.org/browse/WFLY-9213[WFLY-9213]
+
+=== Related Issues
+
+* https://issues.jboss.org/browse/ENTMQBR-1002[ENTMQBR-1002]
+* https://issues.apache.org/jira/browse/ARTEMIS-2401[ARTEMIS-2401]
+
+=== Dev Contacts
+
+* mailto:{email}[{author}]
+
+=== QE Contacts
+
+*  mailto:jkasik@redhat.com[Jan Kašík]
+
+=== Testing By
+// Put an x in the relevant field to indicate if testing will be done by Engineering or QE. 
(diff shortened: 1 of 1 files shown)


messaging/WFLY-9213_Implement_the_Pause_method_for_a_Topic.adoc
```diff
@@ -0,0 +1,102 @@
...
+* mailto:{email}[{author}]
+
+=== QE Contacts
+
+=== Testing By
+// Put an x in the relevant field to indicate if testing will be done by Engineering or QE. 
+// Discuss with QE during the Kickoff state to decide this
+[ ] Engineering
+
+[ ] QE
```
bstansberry: Which of these should be checked?

messaging/WFLY-9213_Implement_the_Pause_method_for_a_Topic.adoc
```diff
@@ -0,0 +1,102 @@
...
+:email:             ehugonne@redhat.com
+:toc:               left
+:icons:             font
+:idprefix:          messaging,jms
+:idseparator:       -
+
+== Overview
+
+
+An Artemis Queue can be paused but there is nothing for a Topic. The goal of this feature is to be able to pause a Topic by pausing all its consumers . New subscriptions being registered while the topic is in pause will be paused also.
```
honza-kasik, line 12: @ehsavoie What does it mean to "pause" a topic? Does it mean no messages will be received by subscribers? If it does what happen to these messages?

messaging/WFLY-9213_Implement_the_Pause_method_for_a_Topic.adoc
```diff
@@ -0,0 +1,102 @@
...
+    "result" => false
+}
+----
+
+
+== Test Plan
+
+== Community Documentation
+
+* Covered by operation description in management model.
```
honza-kasik, line 91: If it will cover the effect of having topic "paused" I see no problem here.

messaging/WFLY-9213_Implement_the_Pause_method_for_a_Topic.adoc
ehsavoie: Quoting Artemis documentation:
The AddressControl can pause and resume an address and all the queues that are bound to it. Newly added queue will be paused too until the address is resumed. Thus all messages sent to the address will be received but not delivered. When it is resumed, delivering will occur again.

messaging/WFLY-9213_Implement_the_Pause_method_for_a_Topic.adoc
ehsavoie: Maybe we can also put a link towards Artemis documentation on this

honza-kasik: +1, though this would mean we would have to maintain the link to target correct destination...

messaging/WFLY-9213_Implement_the_Pause_method_for_a_Topic.adoc
honza-kasik: Thanks, I failed to find this in docs.

honza-kasik:
This proposal looks good from QE point of view. Thank you!

bstansberry (approved):

PR #272, opened 2019-12-17
[WFLY-12896] EE Concurrency Hung Tasks Termination

emmartins:
Issue: https://issues.jboss.org/browse/WFLY-12896

### DIFF
--- a/concurrency/WFLY-12896_EE_Concurrency_Hung_Tasks_Termination.adoc
+++ b/concurrency/WFLY-12896_EE_Concurrency_Hung_Tasks_Termination.adoc
@@ -1,0 +1,99 @@
+= EE Concurrency Hung Tasks Termination
+:author:            Eduardo Martins
+:email:             emartins@redhat.com
+:toc:               left
+:icons:             font
+:idprefix:
+:idseparator:       -
+
+== Overview
+
+The managed executors in the EE subsystem are capable of identifying tasks which are running for a long unexpected time, also know as hung tasks, and this proposal specifies new functionality that will allows users to manually and automatically terminate such tasks.
+
+== Issue Metadata
+
+=== Issue
+
+* https://issues.jboss.org/browse/WFLY-12896[WFLY-12896]
+
+=== Related Issues
+
+* https://issues.jboss.org/browse/EAP7-1397[EAP7-1397]
+
+=== Dev Contacts
+
+* mailto:{email}[{author}]
+
+=== QE Contacts
+
+* mailto:tborgato@redhat.com[Tommaso Borgato]
+
+=== Testing By
+// Put an x in the relevant field to indicate if testing will be done by Engineering or QE. 
+// Discuss with QE during the Kickoff state to decide this
+[x] Engineering
+
+[ ] QE
+
+=== Affected Projects or Components
+
+* WildFly
+
+=== Other Interested Projects
+
+* NONE
+
+== Requirements
+
+=== Hard Requirements
+
+* Add the following attribute to the https://wildscribe.github.io/WildFly/18.0/subsystem/ee/managed-executor-service/[/subsystem=ee/managed-executor-service resource]:
+** *hung-task-termination-period*: the period, in milliseconds, for hung tasks automatic termination. A value of 0, which is the default, deactivates the feature.
+
+CLI to read the attribute from the Managed Executor Service resource named *default*:
+----
+/subsystem=ee/managed-executor-service=default:read-attribute(name=hung-task-termination-period)
+----
+
+* Add the following operation to the https://wildscribe.github.io/WildFly/18.0/subsystem/ee/managed-executor-service/[/subsystem=ee/managed-executor-service resource]:
+** *terminate-hung-tasks*: manually terminates all executor tasks that are hung.
+
+CLI to execute the operation from the Managed Executor Service resource named *default*:
+----
+/subsystem=ee/managed-executor-service=default:terminate-hung-tasks()
+----
+
+* Add the following attribute to the https://wildscribe.github.io/WildFly/18.0/subsystem/ee/managed-scheduled-executor-service/[/subsystem=ee/managed-scheduled-executor-service resource]:
+** *hung-task-termination-period*: the period, in milliseconds, for hung tasks automatic termination. A value of 0, which is the default, deactivates the feature.
+
+CLI to read the attribute from the Managed Scheduled Executor Service resource named *default*:
+----
+/subsystem=ee/managed-scheduled-executor-service=default:read-attribute(name=hung-task-termination-period)
+----
+
+* Add the following operation to the https://wildscribe.github.io/WildFly/18.0/subsystem/ee/managed-scheduled-executor-service/[/subsystem=ee/managed-scheduled-executor-service resource]:
+** *terminate-hung-tasks()*: manually terminates all executor tasks that are hung.
+
+CLI to execute the operation from the Managed Scheduled Executor Service resource named *default*:
+----
+/subsystem=ee/managed-scheduled-executor-service=default:terminate-hung-tasks()
+----
+
+=== Nice-to-Have Requirements
+
+* NONE
+
+=== Non-Requirements
+
+* Domain mode:
+** the new attribute discussed in this proposal will be provided on the servers in a managed domain, and registered with the /profile=*/subsystem=ee resources, being automatically discarded for servers which do not support it.
+** the new operation discussed in this proposal will be provided on the servers in a managed domain, and registered only with the /host=*/server=*/subsystem=ee resources.
+
+== Test Plan
+
+* Changes in Wildfly management model will be covered in WFLY testsuite.
+* Checking the new featured functionality work as expected will be covered in WFLY testsuite.
+
+== Community Documentation
+
+* Covered by local descriptions in management model.


concurrency/WFLY-12896_EE_Concurrency_Hung_Tasks_Termination.adoc
```diff
@@ -0,0 +1,97 @@
...
+/subsystem=ee/managed-scheduled-executor-service=default:read-attribute(name=hung-task-termination-period)
+----
+
+* Add the following operation to the https://wildscribe.github.io/WildFly/18.0/subsystem/ee/managed-scheduled-executor-service/[/subsystem=ee/managed-scheduled-executor-service resource]:
+** *terminate-hung-tasks()*: manually terminates all executor tasks that are hung.
+
+CLI to execute the operation from the Managed Scheduled Executor Service resource named *default*:
+----
+/subsystem=ee/managed-scheduled-executor-service=default:terminate-hung-tasks()
+----
```
rhusar: I think this should be moved below to the "Hard Requirements" section.

rhusar: I think this should be moved below to the "Hard Requirements" section.

emmartins: fixed

concurrency/WFLY-12896_EE_Concurrency_Hung_Tasks_Termination.adoc
```diff
@@ -0,0 +1,95 @@
...
+/subsystem=ee/managed-executor-service=default:read-attribute(name=hung-task-termination-period)
+----
+
+* Add the following operation to the https://wildscribe.github.io/WildFly/18.0/subsystem/ee/managed-executor-service/[/subsystem=ee/managed-executor-service resource]:
+** *terminate-hung-tasks*: manually terminates all executor tasks that are hung.
+
+CLI to execute the operation from the Managed Executor Service resource named *default*:
+----
+/subsystem=ee/managed-executor-service=default:terminate-hung-tasks()
```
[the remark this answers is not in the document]
emmartins: IMHO I don't see much value in leaving tasks that are hung alive so I would prefer to keep the simplicity of the feature, having a match of what the new management op and the automatic periodic process does (terminate all hung tasks).

concurrency/WFLY-12896_EE_Concurrency_Hung_Tasks_Termination.adoc
```diff
@@ -0,0 +1,95 @@
...
+----
+/subsystem=ee/managed-scheduled-executor-service=default:read-attribute(name=hung-task-termination-period)
+----
+
+* Add the following operation to the https://wildscribe.github.io/WildFly/18.0/subsystem/ee/managed-scheduled-executor-service/[/subsystem=ee/managed-scheduled-executor-service resource]:
+** *terminate-hung-tasks()*: manually terminates all executor tasks that are hung.
+
+CLI to execute the operation from the Managed Scheduled Executor Service resource named *default*:
+----
+/subsystem=ee/managed-scheduled-executor-service=default:terminate-hung-tasks()
```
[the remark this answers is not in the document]
emmartins: IMHO I don't see much value in leaving tasks that are hung alive so I would prefer to keep the simplicity of the feature, having a match of what the new management op and the automatic periodic process does (terminate all hung tasks).

concurrency/WFLY-12896_EE_Concurrency_Hung_Tasks_Termination.adoc
```diff
@@ -0,0 +1,95 @@
...
+=== Related Issues
+
+* https://issues.jboss.org/browse/EAP7-1397[EAP7-1397]
+
+=== Dev Contacts
+
+* mailto:{email}[{author}]
+
+=== QE Contacts
+
```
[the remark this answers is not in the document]
emmartins: done

concurrency/WFLY-12896_EE_Concurrency_Hung_Tasks_Termination.adoc
```diff
@@ -0,0 +1,97 @@
...
+== Requirements
+
+=== Hard Requirements
+
+* Add the following attribute to the https://wildscribe.github.io/WildFly/18.0/subsystem/ee/managed-executor-service/[/subsystem=ee/managed-executor-service resource]:
+** *hung-task-termination-period*: the period, in milliseconds, for hung tasks automatic termination.
+
+CLI to read the attribute from the Managed Executor Service resource named *default*:
+----
+/subsystem=ee/managed-executor-service=default:read-attribute(name=hung-task-termination-period)
```
marekkopecky, line 55: What is the default value/behaviour of this property? The answer needs to be provided in this document.

emmartins: Updated to indicate the default value and related feature behaviour

concurrency/WFLY-12896_EE_Concurrency_Hung_Tasks_Termination.adoc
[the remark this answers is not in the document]
tommaso-borgato: @emmartins actually I was asking because of a scenario I experienced myself; let me depict it: we had something like 10 long running tasks marked as "hung" and we knew they were hung because they generated high concurrency in the the database connected to the application server; this pile of tasks was generated by manual retries; we wanted to get rid of such concurrency killing 9 of them to enable the one left to complete it's work: it would have been too expensive to restart it; our criteria back then was to kill the most recent 9, but that was just a simplistic criteria we used because we lacked further info on the tasks; the other solution would have been to engage a db administrator to work at the db level and we had none available back then; the vendor was Weblogic, and if remember correctly we could kill hung tasks selectively; anyway, if you think this is an over specific scenario I won't insist further

concurrency/WFLY-12896_EE_Concurrency_Hung_Tasks_Termination.adoc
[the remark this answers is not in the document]
tommaso-borgato: same as above

concurrency/WFLY-12896_EE_Concurrency_Hung_Tasks_Termination.adoc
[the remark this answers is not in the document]
emmartins: The managed tasks can have a listener attached with logic to deal with termination on their own, we also have a runtime stat providing the number of hung tasks, which would need to be reworked since having the stat and a retrieve tasks attr/op would be confusing IMHO. Let's leave this extended functionality for later, if customers ask for it.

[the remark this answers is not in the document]
tommaso-borgato: agree

concurrency/WFLY-12896_EE_Concurrency_Hung_Tasks_Termination.adoc
```diff
@@ -0,0 +1,97 @@
...
+/subsystem=ee/managed-scheduled-executor-service=default:terminate-hung-tasks()
+----
+
+=== Nice-to-Have Requirements
+
+* NONE
+
+=== Non-Requirements
+
+* Domain mode: the new attributes discussed in this proposal will be provided on the servers in a managed domain, but will not be registered with the /profile=*/subsystem=ee resources.
```
tommaso-borgato, line 88: @emmartins I might have overlooked this statement ... the meaning it's not clear to me: what does it mean that the new attributes will be provided but not registered? what are the implications? which is the reason why domain mode is treated differently?

emmartins: This was content from another AD which removal I missed, there is nothing particular/different wrt domain mode on this RFE

concurrency/WFLY-12896_EE_Concurrency_Hung_Tasks_Termination.adoc
```diff
@@ -0,0 +1,99 @@
...
+----
+
+=== Nice-to-Have Requirements
+
+* NONE
+
+=== Non-Requirements
+
+* Domain mode:
+** the new attribute discussed in this proposal will be provided on the servers in a managed domain, and registered with the /profile=*/subsystem=ee resources, being automatically discarded for servers which do not support it.
```
tommaso-borgato, line 89: @emmartins this means that from both the master and slave nodes you will be able to perform e.g.
```
/profile=full/subsystem=ee/managed-executor-service=default:read-attribute(name=hung-task-termination-period)
```
and 
```
/profile=full/subsystem=ee/managed-executor-service=default:write-attribute(name=hung-task-termination-period, value=1234)
```
correct?

concurrency/WFLY-12896_EE_Concurrency_Hung_Tasks_Termination.adoc
```diff
@@ -0,0 +1,99 @@
...
+** the new operation discussed in this proposal will be provided on the servers in a managed domain, and registered only with the /host=*/server=*/subsystem=ee resources.
```
tommaso-borgato, line 90: @emmartins this means that from both the master and slave nodes you will be able to perform e.g.
```
/host=slavenodeN/subsystem=ee/managed-executor-service=default:terminate-hung-tasks()
```
correct?

concurrency/WFLY-12896_EE_Concurrency_Hung_Tasks_Termination.adoc
emmartins: You do writes on master, so it updates slave's configs.

emmartins: I would prefer to not call master and slaves tho, it's read/write on the domain profile as in the text.

concurrency/WFLY-12896_EE_Concurrency_Hung_Tasks_Termination.adoc
emmartins: I would prefer to not call master and slaves, it's invoked on a host's server as in the text.

emmartins: e.g /host=h1/server=s1/subsystem=ee/managed-executor-service=default:terminate-hung-tasks()

concurrency/WFLY-12896_EE_Concurrency_Hung_Tasks_Termination.adoc
tommaso-borgato: @emmartins so, if my understanding is correct, this new property will work in domain mode just like any other existing property in the ee subsystem, right?

emmartins: yes

tommaso-borgato: @emmartins still not clear to me why you specified this under "Non Requirements"

emmartins: because those are implementation details which are not requirements, can we now move forward with the AD approval or there is still something else?

tommaso-borgato: yes

emmartins: thank you

tommaso-borgato (approved):

PR #275, opened 2020-01-07
[WFLY-12459] Initial analysis for active request tracking feature.

fl4via:
Analysis for Active request tracking

Jira: https://issues.redhat.com/browse/WFLY-12459

### DIFF
--- a/undertow/WFLY-12459_Active_Request_Tracking.adoc
+++ b/undertow/WFLY-12459_Active_Request_Tracking.adoc
@@ -1,0 +1,143 @@
+= WFLY-12473 Active Request Tracking
+:author:            Flavia Rainone
+:email:             frainone@redhat.com
+:toc:               left
+:icons:             font
+:idprefix:
+:idseparator:       -
+
+== Overview
+
+== Issue Metadata
+
+=== Issue
+
+* https://issues.redhat.com/browse/WFLY-12459
+
+=== Related Issues
+
+* https://issues.redhat.com/browse/UNDERTOW-1593
+
+=== Dev Contacts
+
+* mailto:{email}[{author}]
+
+=== QE Contacts
+
+=== Testing By
+// Put an x in the relevant field to indicate if testing will be done by Engineering or QE.
+// Discuss with QE during the Kickoff state to decide this
+[ ] Engineering
+
+[ ] QE
+
+=== Affected Projects or Components
+Undertow, WildFly Undertow Subsystem
+
+=== Other Interested Projects
+
+== Requirements
+The new feature consists of adding a way for users to retrieve active requests, ie.
+requests that haven't finished and are active on the server.
+There was a similar feature in EAP6 and the plan is to be able to provide something similar
+in EAP7 with Undertow.
+
+Since there is an added impact for tracking active requests, the feature has to be
+enabled to be used. To enable the feature, we are adding an attribute to Undertow
+subsystem, called `"active-request-statistics-enabled"`.
+
+That attribute can only be set to `true` if `"statistics-enabled"` is `true`.
+
+Once that feature is enabled, there is an added resource to undertow server
+called "active-request-tracker=enabled"
+As an example, look at the cli script below:
+
+....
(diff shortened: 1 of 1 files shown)


undertow/WFLY-12459_Active_Request_Tracking.adoc
```diff
@@ -0,0 +1,143 @@
...
+active-request-info(id="id")
+....
+See the example bellow:
+----
+subsystem=undertow/active-request-tracker=requestTracker:list-active-requests
+----
+
+The operation below will show just the requested attributes from the request:
+----
+subsystem=undertow/active-request-tracker=requestTracker:list-active-requests(attributes="bytes-sent,bytes-received")
```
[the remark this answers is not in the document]
jstourac: Also, is there specified a complete list of attributes that will be collected for each active request somewhere?

undertow/WFLY-12459_Active_Request_Tracking.adoc
```diff
@@ -0,0 +1,143 @@
...
+----
+
+Notice that, in order for the track to start tracking actual requests, it must be referenced as a filter inside the server as in:
+----
+/subsystem=undertow/server=default-server/host=default-host/filter-ref=requestTracker:add
+----
```
jstourac, line 86: If I understand this, then this filter reference will be kind of a reference to a `virtual` filter, right? How is this reflected in the configuration XML? What if user defines its own filter named `requestTracker`? Shall we deny this name?

undertow/WFLY-12459_Active_Request_Tracking.adoc
```diff
@@ -0,0 +1,143 @@
...
+
+[ ] QE
+
+=== Affected Projects or Components
+Undertow, WildFly Undertow Subsystem
+
+=== Other Interested Projects
+
+== Requirements
+The new feature consists of adding a way for users to retrieve active requests, ie.
```
mchoma, line 40: Can't be this use case covered by standard subsystem `/metrics` we already have in Wildfly? There are several undertow related values already? Or is this somehow different? 

@jmesnil wdyt?

undertow/WFLY-12459_Active_Request_Tracking.adoc
```diff
@@ -0,0 +1,143 @@
...
+subsystem, called `"active-request-statistics-enabled"`.
+
+That attribute can only be set to `true` if `"statistics-enabled"` is `true`.
+
+Once that feature is enabled, there is an added resource to undertow server
+called "active-request-tracker=enabled"
+As an example, look at the cli script below:
+
+....
+subsystem=undertow/set-attribute(name="active-request-statistics-enabled", "true")
```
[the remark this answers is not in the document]
bstansberry: Correct; there is no 'set-attribute' operation.

undertow/WFLY-12459_Active_Request_Tracking.adoc
```diff
@@ -0,0 +1,143 @@
...
+....
+
+As a result, we will have a new resource in the Undertow Subsystem with the name:
+
+.....
+subsystem=undertow/active-request-tracker=enabled
```
bstansberry, line 62: Isn't this /subsystem=undertow/active-request-tracker=request-tracker?

Also AIUI this is a runtime-only resource (e.g. it does not appear in the xml config shown below). That should be stated explicitly.

AIUI this is a fixed-name resource (aka a 'singleton'). Some name like /subsystem=undertow/service=active-request-tracker is better.  The general pattern of address elements is "general type"="specific name". It's better to make up a general type than to not follow the pattern. The only thing to be careful about is not using some 'general type' that later you will want to use for a resource type where the user can provide the specific name.

undertow/WFLY-12459_Active_Request_Tracking.adoc
bstansberry: AIUI the intent in the above bit is to describe the automatic internal behavior as being the same as what would happen if a filter-ref was actually in the config. Not that there would actually be such a thing in the config, and thus no "add" operation to set it up.

PR #277, opened 2020-01-08
[WFLY-12190] Http server to server invocation analysis

tadamski:


ejb/WFLY-12190-http-server-to-server-config.adoc
```diff
@@ -0,0 +1,153 @@
...
+[ ] Engineering
+
+[x] QE
+
+=== Affected Projects or Components
+EJBs
+
+== Introduction
+
+PRODMGT-268 (https://issues.redhat.com/browse/PRODMGT-268) added the ability to configure EJB invocations over HTTP protocol. Customers would like to use this feature in server-to-server configurations.
```
darranl: As this is a community project can we please try and cross reference community issues?

darranl:
I suspect a large aspect of this feature will be the consideration of security and the configuration of this for the connections.

Any of these connections could require an SSL configuration which may vary depending on the destination, the same would apply for any authentication requirements.  Additionally in the future it may be required for EJB over HTTP to support additional authentication mechanisms such as MP-JWT so any plans should consider this.

Could it make sense for maybe a member of the Elytron team to pair up on this feature to collaborate on the security aspects with you?

ejb/WFLY-12190-http-server-to-server-config.adoc
```diff
@@ -0,0 +1,153 @@
...
+[source]
+----
+<remote-outbound-connection name="http-connection" outbound-socket-binding-ref="binding-remote-ejb-connection" protocol="http"/>
+----
+
+Such connection can be then used from within jboss-ejb-client.xml configuration file. The following issues have to be taken into consideration regarding such configuration:
+
+=== Http config in remoting subystem
+
+Remoting subsystem is responsible for the configuration of connection based on jboss-remoting protocol. The attributes and properties of remote-outbound-connections are used to configure remoting connections and make no sense in for http protocol. Simply adding another protocol to remote-outbound-connection may seem like a simplest solution from the customer point of view, but it introduces incoherent subsystem configuration and in my opinion it is unnaceptable.
```
ochaloup: I see the incoherency in subsystem configuration just I can imagine the headache for users configuring the different outbound connection on different places.
Should not be then deprecated the usage of the `remote-outbound-connections` in `remoting` subsystem and document that the correct place for configuration of any protocol is the `ejb3` subsystem from now?

ejb/WFLY-12190-http-server-to-server-config.adoc
```diff
@@ -0,0 +1,153 @@
...
+<subsystem xmlns="urn:jboss:domain:ejb3:6.0">
+        (...)
+        <remote connector-ref="http-remoting-connector" thread-pool-name="default">
+            <channel-creation-options>
+                <option name="MAX_OUTBOUND_MESSAGES" value="1234" type="remoting"/>
+            </channel-creation-options>
+            <profiles>
+                <profile name="test-profile">
+                    <static-ejb-discovery>
+                        <module uri="http://${server.ip.address}:8180/wildfly-services" module-name="foo"/>
```
ochaloup: I'm unfamiliar with the http remoting so what code represents the `module-name` `foo` here? That's the deployment or what? (I'm sorry that I didn't catch it).

zhantemirov:
@tadamski could you please provide your feedback to the latest commentaries from ochaloupka and darranl so we can consider this AD as ready one. Thank you!

ejb/WFLY-12190-http-server-to-server-config.adoc
tadamski: fixed

ejb/WFLY-12190-http-server-to-server-config.adoc
tadamski: @ochaloup I think that such configuration is correct. Remoting connections have their own configuration and IMO it's place is in remoting subsystem. This is a bit elaborate config but it was always the case.

ejb/WFLY-12190-http-server-to-server-config.adoc
tadamski: It's module part of EJB identifier (as used f.e. in JDNI lookup). BTW this is just an example as static discovery implementation/upgrade is not part of this RFE.

ejb/WFLY-12190-http-server-to-server-config.adoc
ochaloup: ok, I see

ejb/WFLY-12190-http-server-to-server-config.adoc
ochaloup: I see.
Just I have still this doubt - from the design document it seems  that the `static ejb discovery` will be part of the solution. There is section which explains what is it, then this code snippet is used in both code pieces of the document.

The static ejb discovery won't be then configurable at all?

tadamski: static-ejb-discovery is already implemented and works - it was put into the analysis document to show the difference between the new solution (dynamic discovery) and the currently available one

ochaloup: oh, ok. my wrong. thanks for clarification.

tadamski:
security discussion:
https://wildfly.zulipchat.com/#narrow/stream/174184-wildfly-developers/topic/WFLY-12190.20Reintroduce.20EJB.20over.20HTTP.2FHTTPS.20capability.20with.20H

zhantemirov:
Could this analysis be merged now as all stakeholders do not have any objections/suggestions? Thanks.

darranl:
@zhantemirov FYI analysis is merged after the pull requests to WildFly Core and WildFly.

PR #281, opened 2020-01-22
[WFCORE-4807] Add option to route log messages based on the current thread's class loader

jamezp:
…n for this to be disabled.

https://issues.redhat.com/browse/WFCORE-4807

### DIFF
--- a/logging/WFCORE-4807-log-routing.adoc
+++ b/logging/WFCORE-4807-log-routing.adoc
@@ -1,0 +1,118 @@
+= Log Routing
+:author:            James R. Perkins
+:email:             jperkins@redhat.com
+:toc:               left
+:icons:             font
+:idprefix:
+:idseparator:       -
+
+== Overview
+
+Currently, any modules that log generally log onto the system log context. In some cases, for example a deployment using
+a logging profile, the loggers may end up getting created on the wrong `LogContext`. As an example if `a.war` and `b.war`
+both depend on module `org.jboss.custom.module` and at least one deployment uses a logging profile whichever deployment
+initializes the custom module first will have the loggers created on that `LogContext`. If a logging profile is being
+used it would be that profiles log context. This means if `a.war` uses a logging profile and the module is accessed first
+by `a.war`, then static loggers will be created on the logging profiles `LogContext`. Once `b.war` invokes something to
+be logged in the custom module, then `b.war` 's log messages end up in the logging profiles `LogContext` instead of the
+systems `LogContext`.
+
+https://issues.jboss.org/browse/LOGMGR-264[LOGMGR-264] will introduce a way to route log messages based on the log context.
+This would allow messages from modules to be routed to deployment specific log contexts. It would also introduce a
+change in behavior which is the reason for an RFE. A new attribute, `allow-log-routing` will be introduced to disable
+the new routing and allow the old behavior. What this will do is route log messages using the current threads context
+class loader (TCCL). In the above example if the module `org.jboss.custom.module` is accessed from `a.war` the TCCL will
+be set to `a.war` and the logged messages will be routed to the `LogContext` associated with `a.war` 's class loader. In
+turn `b.war` will route to the system log context since its class loader is not associated with a specific `LogContext`.
+
+== Issue Metadata
+
+=== Issue
+
+* https://issues.jboss.org/browse/WFCORE-4897[WFCORE-4807]
+
+=== Related Issues
+
+* https://issues.jboss.org/browse/EAP7-1420[EAP7-1420]
+* https://issues.jboss.org/browse/LOGMGR-264[LOGMGR-264]
+* https://issues.redhat.com/browse/WFCORE-3376[WFCORE-3376]
+* https://issues.redhat.com/browse/LOGTOOL-146[LOGTOOL-146]
+
+=== Dev Contacts
+
+* mailto:{email}[{author}]
+
+=== QE Contacts
+
+* mailto:mkopecky@redhat.com[Marek Kopeck&#253;]
+
+=== Testing By
+
+* [x] Engineering
+
+* [ ] QE
+
+=== Affected Projects or Components
+
+* WildFly Core
+* WildFly
+* JBoss Log Manager
+
+=== Other Interested Projects
+
+* Any consumer of WildFly Core that uses the logging subsystem.
+
+== Requirements
+
+=== Hard Requirements
+
+* The attribute, `allow-log-routing`, is a simple boolean attribute on the `/subsystem=logging` resource which
+  determines if the new or old behavior will be used.
+* When the new behavior is disabled the behavior should be the same as it currently works.
+    ** This means that when disabled the bug in https://issues.redhat.com/browse/WFCORE-3376[WFCORE-3376] would
+       still be present.
+* The `LogContext` should be associated with a deployments class loader.
+    ** This means that if the TCCL is set to the deployment class loader log messages will be routed to the deployments
+       log context.
+    ** That may include messages that previously went to the `server.log` could end up in a handler defined on the
+       deployments log context. This is desired behavior IMO.
+* Loggers created on any log context should be routable to another log context based on the TCCL, if enabled.
+* The attribute will be set to `false` by default to ensure there are no behavioral changes out of the box.
+
+=== Nice-to-Have Requirements
+
+* Allow the enabling/disabling of the behavior to not require a restart.
+
+=== Non-Requirements
+
+* This will not guarantee that all logs from modules will be routed to the deployments log context. If the TCCL is not
+  set to a deployments class loader logs will not end up being routed to the deployments log context.
+
+== Implementation Plan
+
+When enabled the default `LogContextSelector` will be `org.jboss.logmanager.ContextClassLoaderLogContextSelector`. This
+allows modules to log to a deployments log context if the TCCL is set to the deployments log context. Otherwise the
+modules should default to the default log context. An implementation of the new `org.jboss.logmanager.LoggerRouter`
+interface will be implemented in the logging subsystem.
+
+When disabled the currently used `LogContextSelector` should be used. No change should happen with how log messages are
+routed.
+
+== Test Plan
+
+If possible there will be some unit tests for the behavior. There will also be tests required in WildFly full as it
+will easily allow for testing with multiple deployments.
+
+The test should verify that:
+
+* When enabled log messages are correctly routed to the deployments log context.
+* When disabled log messages are correctly, in some cases incorrectly, routed to the deployment or default log context.
+  This will depend on the scenario.
+* Tests should be done on all deployment specific logging options:
+    ** Logging profiles
+    ** Log configuration files, e.g. log4j.properties, logging.properties, etc.
+
+== Community Documentation
+
+The logging subsystem documentation will be updated to show the new attribute, `allow-log-routing`. Detailed examples
+will be given on how logging is routed when the new behavior is enabled.


logging/WFCORE-4807-log-routing.adoc
```diff
@@ -0,0 +1,106 @@
...
+=== Hard Requirements
+
+* The attribute will be a simple boolean attribute which determines if the new or old behavior will be used.
+* When the new behavior is disabled the behavior should be the same as it currently works.
+    ** This means that when disabled the bug in https://issues.redhat.com/browse/WFCORE-3376[WFCORE-3376] would
+       still be present.
+* The `LogContext` should be associated with a deployments class loader.
+    ** This means that if the TCCL is set to the deployment class loader log messages will be routed to the deployments
+       log context.
+    ** That may include messages that previously went to the `server.log` could end up in a handler defined on the
```
bstansberry: What is the impact on messages related to deploying and undeploying the deployment?

jamezp: If the TCCL is set to the deployments class loader then messages get routed to the deployments log context. For example:
```
09:54:42,716 INFO  [org.wildfly.extension.undertow] (ServerService Thread Pool -- 85) WFLYUT0021: Registered web context: '/test-logging-1' for server 'default-server'
```
gets logged to the deployments log context.

But the following is logged in the servers log configuration:
```
2020-01-24 09:54:42,301 INFO  [org.jboss.as.server.deployment] (MSC service thread 1-6) WFLYSRV0027: Starting deployment of "test-logging-1.war" (runtime-name: "test-logging-1.war")
2020-01-24 09:54:42,437 WARN  [org.jboss.as.dependency.private] (MSC service thread 1-6) WFLYSRV0018: Deployment "deployment.test-logging-1.war" is using a private module ("org.wildfly.security.manager") which may be changed or removed in future versions without notice.
2020-01-24 09:54:42,611 INFO  [org.jboss.as.arquillian] (MSC service thread 1-6) Arquillian deployment detected: ArquillianConfig[service=jboss.arquillian.config."test-logging-1.war",unit=test-logging-1.war,tests=[org.jboss.as.test.integration.logging.LoggingProfileModuleTestCase]]
2020-01-24 09:54:42,748 INFO  [org.jboss.as.server] (management-handler-thread - 1) WFLYSRV0010: Deployed "test-logging-1.war" (runtime-name : "test-logging-1.war")
2020-01-24 09:54:43,362 INFO  [org.jboss.as.server.deployment] (MSC service thread 1-2) WFLYSRV0028: Stopped deployment test-logging-1.war (runtime-name: test-logging-1.war) in 42ms
2020-01-24 09:54:43,379 INFO  [org.jboss.as.repository] (management-handler-thread - 1) WFLYDR0002: Content removed from location /home/jperkins/projects/wildfly/wildfly/testsuite/integration/basic/target/wildfly/standalone/data/content/56/6daf8be2c6588d53cf5a1acda67e2745821633/content
2020-01-24 09:54:43,380 INFO  [org.jboss.as.server] (management-handler-thread - 1) WFLYSRV0009: Undeployed "test-logging-1.war" (runtime-name: "test-logging-1.war")
```

jamezp: Clicked comment too soon :)

Basically the idea is, and maybe I should add this somewhere in the doc, is that stuff coming from say Weld or Hibernate that is associated with the deployment should end up in the deployments logs. Basically any messages that come from server dependencies where the TCCL is set to the deployments class loader should end up in the applications logs.

jamezp:
Maybe this should be opt-in rather than opt-out by default. I don't have a strong opinion either way. I went with opt-out as it seems like users have been wanting some kind of filter or way to mimic this behavior for some time. This should act like the old TCCL filter as far as I understand how that filter worked.

logging/WFCORE-4807-log-routing.adoc
bstansberry: That might be a bit odd. I don't know if anything can be done about it. But I doubt that much thought is given to what TCCL is in place when logging calls get added. That "Registered web context" message is more of a server.log message than the "Deployment "deployment.test-logging-1.war" is using a private module" one is. (Both are valid server.log messages; the 'private module' one is just more a detail about a particular deployment than the 'Registered web context' is.

jamezp: I can definitely see your point. The problem we're attempting to solve is that modules used by deployments may create loggers on the wrong log context. For example given two deployments, `a.war` and `b.war`, that each have their own log context and both have a dependency on `org.jboss.example.module` have a race to see which log context creates the static loggers associated with the module.

In other words; if `b.war` first initializes the loggers in the `org.jboss.example.module` then all log messages, even if they are initiated by `a.war`, from the module end up in the `b.war` log context. Note this assumes that `b.war` does use either a logger profile or have a logging configuration file in it that is configured by a DUP in the logging subsystem.

The two solutions would be:

1. All messages from modules get logged on the server level log context
2. Messages from modules that seemingly belong to the deployment get logged with the deployment. Seemingly here means the TCCL is set to the deployments class loader.

The problem with 1 is there is no real way to know that the logger is being created on a module and not a deployment. Short of inspecting the class loader and assuming that anything who's module name that does *not* begin with "deployment" is global module.

The problem with 2, as you point out, is it's a guess too at which log messages are said to be "deployment" log messages. Checking the TCCL is the deployments class loader is how the assumption is made and seems somewhat safe. Though I do agree and can see this could be considered wrong. This was part of the reason for the RFE to turn this off.

It seems the options are:

1. Reject the RFE and live with the bug
2. Have the flag like this document proposes which we could change to opt-in instead of opt-out.
3. We change the RFE to be a flag that would inspect the `ClassLoader` of the caller class for a module name of "deployment.*". Possibly even use a user defined pattern, but I'm not too sure I love that idea :)

I'm definitely open for input here. I can say since JBoss AS 7 users have wanted something akin to the `TCCLFilter` though admittedly in recent years this has not come up as much. Also I'm making some assumptions on how the `TCCLFilter` actually worked too. I've only heard of it before :)

bstansberry: @jamezp 

What does this look like if something like kitchensink.war gets deployed? In the discussion above there were two "odd" (i.e. worth discussing) messages. With complex deployments are there a lot more?

Is it correct to say that in a case where we really want something to go to server.log but it ends up not, that we could wrap the log call in a TCCL set/restore block? So we have a fix for those. If something still ended up in the server.log, well maybe we could also do a TCCL set/restore block, if the relevant code had access to the deployment's CL.

jamezp: @bstansberry Here's the logs from the deploy operation. The `[app-log]` is from the `kitchensink.war`.
```
11:06:34,392 INFO  [org.jboss.as.repository] (management-handler-thread - 1) WFLYDR0001: Content added at location /home/jperkins/projects/wildfly/wildfly/build/target/wildfly-22.0.0.Beta1-SNAPSHOT/standalone/data/content/a8/58ea34cd6065235347abf41b90a6e2f0c07065/content
11:06:34,408 INFO  [org.jboss.as.server.deployment] (MSC service thread 1-7) WFLYSRV0027: Starting deployment of "kitchensink.war" (runtime-name: "kitchensink.war")
11:06:34,737 INFO  [org.jboss.as.jpa] (MSC service thread 1-4) WFLYJPA0002: Read persistence.xml for primary
11:06:34,745 WARN  [org.jboss.as.connector] (MSC service thread 1-4) WFLYJCA0091: -ds.xml file deployments are deprecated. Support may be removed in a future version.
11:06:34,864 INFO  [org.jboss.as.connector.subsystems.datasources] (MSC service thread 1-1) WFLYJCA0001: Bound data source [java:jboss/datasources/KitchensinkQuickstartDS]
11:06:34,892 INFO  [org.jboss.weld.deployer] (MSC service thread 1-8) WFLYWELD0003: Processing weld deployment kitchensink.war
11:06:34,988 INFO  [org.hibernate.validator.internal.util.Version] (MSC service thread 1-8) HV000001: Hibernate Validator 6.0.21.Final
11:06:35,107 INFO  [org.jboss.as.ejb3.deployment] (MSC service thread 1-8) WFLYEJB0473: JNDI bindings for session bean named 'MemberRegistration' in deployment unit 'deployment "kitchensink.war"' are as follows:

	java:global/kitchensink/MemberRegistration!org.jboss.as.quickstarts.kitchensink.service.MemberRegistration
	java:app/kitchensink/MemberRegistration!org.jboss.as.quickstarts.kitchensink.service.MemberRegistration
	java:module/MemberRegistration!org.jboss.as.quickstarts.kitchensink.service.MemberRegistration
	java:global/kitchensink/MemberRegistration
	java:app/kitchensink/MemberRegistration
	java:module/MemberRegistration

11:06:35,108 INFO  [org.infinispan.CONTAINER] (ServerService Thread Pool -- 78) ISPN000128: Infinispan version: Infinispan 'Corona Extra' 11.0.4.Final
11:06:35,136 INFO  [org.jboss.as.jpa] (ServerService Thread Pool -- 78) WFLYJPA0010: Starting Persistence Unit (phase 1 of 2) Service 'kitchensink.war#primary'
11:06:35,149 INFO  [org.hibernate.jpa.internal.util.LogHelper] (ServerService Thread Pool -- 78) HHH000204: Processing PersistenceUnitInfo [
	name: primary
	...]
11:06:35,168 INFO  [org.infinispan.CONFIG] (MSC service thread 1-3) ISPN000152: Passivation configured without an eviction policy being selected. Only manually evicted entities will be passivated.
11:06:35,169 INFO  [org.infinispan.CONFIG] (MSC service thread 1-3) ISPN000152: Passivation configured without an eviction policy being selected. Only manually evicted entities will be passivated.
11:06:35,214 INFO  [io.jaegertracing.internal.JaegerTracer] (MSC service thread 1-8) No shutdown hook registered: Please call close() manually on application shutdown.
11:06:35,215 INFO  [org.hibernate.Version] (ServerService Thread Pool -- 78) HHH000412: Hibernate Core {5.3.20.Final}
11:06:35,216 INFO  [org.hibernate.cfg.Environment] (ServerService Thread Pool -- 78) HHH000206: hibernate.properties not found
11:06:35,241 INFO  [org.infinispan.PERSISTENCE] (ServerService Thread Pool -- 16) ISPN000556: Starting user marshaller 'org.wildfly.clustering.infinispan.spi.marshalling.InfinispanProtoStreamMarshaller'
11:06:35,320 INFO  [org.hibernate.annotations.common.Version] (ServerService Thread Pool -- 78) HCANN000001: Hibernate Commons Annotations {5.0.5.Final}
11:06:35,329 INFO  [org.jboss.weld.Version] (MSC service thread 1-3) WELD-000900: 3.1.5 (Final)
11:06:35,378 INFO  [org.jboss.as.clustering.infinispan] (ServerService Thread Pool -- 16) WFLYCLINF0002: Started http-remoting-connector cache from ejb container
[app-log] 11:06:35,476 INFO  [org.jboss.as.jpa] (ServerService Thread Pool -- 16) WFLYJPA0010: Starting Persistence Unit (phase 2 of 2) Service 'kitchensink.war#primary'
[app-log] 11:06:35,597 INFO  [org.hibernate.dialect.Dialect] (ServerService Thread Pool -- 16) HHH000400: Using dialect: org.hibernate.dialect.H2Dialect
[app-log] 11:06:35,601 WARN  [org.hibernate.dialect.H2Dialect] (ServerService Thread Pool -- 16) HHH000431: Unable to determine H2 database version, certain features may not work
[app-log] 11:06:35,645 INFO  [org.hibernate.envers.boot.internal.EnversServiceImpl] (ServerService Thread Pool -- 16) Envers integration enabled? : true
[app-log] 11:06:36,030 INFO  [org.hibernate.tool.schema.internal.SchemaCreatorImpl] (ServerService Thread Pool -- 16) HHH000476: Executing import script 'ScriptSourceInputFromUrl(vfs:/content/kitchensink.war/WEB-INF/classes/import.sql)'
[app-log] 11:06:36,104 INFO  [io.smallrye.metrics] (MSC service thread 1-7) MicroProfile: Metrics activated (SmallRye Metrics version: 2.4.4)
[app-log] 11:06:36,141 WARN  [org.jboss.weld.Bootstrap] (MSC service thread 1-7) WELD-000146: BeforeBeanDiscovery.addAnnotatedType(AnnotatedType<?>) used for class com.sun.faces.flow.FlowDiscoveryCDIHelper is deprecated from CDI 1.1!
[app-log] 11:06:36,503 INFO  [javax.enterprise.resource.webcontainer.jsf.config] (ServerService Thread Pool -- 16) jsf.config.listener.version
[app-log] 11:06:37,067 INFO  [org.jboss.resteasy.resteasy_jaxrs.i18n] (ServerService Thread Pool -- 16) RESTEASY002225: Deploying javax.ws.rs.core.Application: class org.jboss.as.quickstarts.kitchensink.rest.JaxRsActivator$Proxy$_$$_WeldClientProxy
[app-log] 11:06:37,116 INFO  [org.wildfly.extension.undertow] (ServerService Thread Pool -- 16) WFLYUT0021: Registered web context: '/kitchensink' for server 'default-server'
11:06:37,178 INFO  [org.jboss.as.server] (management-handler-thread - 1) WFLYSRV0010: Deployed "kitchensink.war" (runtime-name : "kitchensink.war")
```

> Is it correct to say that in a case where we really want something to go to server.log but it ends up not, that we could wrap the log call in a TCCL set/restore block? So we have a fix for those. If something still ended up in the server.log, well maybe we could also do a TCCL set/restore block, if the relevant code had access to the deployment's CL.

Yes that is correct. We could wrap the TCCL to ensure the log ends up being logged to the correct context.

jamezp: Coming back to this. @kabir gave me a good idea of adding an annotation to automate the wrapping. I created a `@ThreadContext`, maybe I'll use a better name, annotation which sets and resets the TCCL based on the type passed in or `getClass().getClassLoader()`. I've updated all the branches and that seems to work quite well.

logging/WFCORE-4807-log-routing.adoc
```diff
@@ -0,0 +1,109 @@
...
+:email:             jperkins@redhat.com
+:toc:               left
+:icons:             font
+:idprefix:
+:idseparator:       -
+
+== Overview
+
+Currently any modules that log generally log onto the system log context. In some cases, for example a deployment using
+a logging profile, the loggers may end up getting created on the wrong `LogContext`. For example if `a.war` and `b.war`
```
marekkopecky, line 12: What do you mean by `logging profile`? I know logging subsystem, logging modules (eg `./modules/system/layers/base/org/jboss/logging/main/module.xml`) and standalone profile, but I'm not sure what `logging profile` is.

logging/WFCORE-4807-log-routing.adoc
```diff
@@ -0,0 +1,109 @@
...
+both depend on module `org.jboss.custom.module` and at least one deployment uses a logging profile whichever deployment
```
marekkopecky, line 13: Can you please elaborate on this example a more in analysis document? This is my current understanding of this example, is this correct?

1. `a.war` deployment uses custom logging profile and this deployment initializes `org.jboss.custom.module` module. Loggers in this custom module is configured to use custom logging profile and its `LogContext` from `a.war`
2. `b.war` deployment doesn't use any custom logging profile, so it uses default system `LogContext`. It also uses some classes from custom module, but this classes are already configured to use `LogContext` from custom profile, which is wrong.

Is this my understanding correct?

logging/WFCORE-4807-log-routing.adoc
```diff
@@ -0,0 +1,109 @@
...
+* https://issues.jboss.org/browse/LOGMR-264[LOGMR-264]
+* https://issues.redhat.com/browse/WFCORE-3376[WFCORE-3376]
+
+=== Dev Contacts
+
+* mailto:{email}[{author}]
+
+=== QE Contacts
+
+* TBD
```
marekkopecky, line 40: It's me, my mail is mkopecky@redhat.com

logging/WFCORE-4807-log-routing.adoc
```diff
@@ -0,0 +1,109 @@
...
+
+=== Other Interested Projects
+
+* Any consumer of WildFly Core that uses the logging subsystem.
+
+== Requirements
+
+=== Hard Requirements
+
+* The attribute will be a simple boolean attribute which determines if the new or old behavior will be used.
```
marekkopecky, line 62: Please explicitly specify the name of attribute, it location in management model and default value.

marekkopecky:
Hello James, I'm looking to the description of the RFE jira, linked jiras, RFE analysis and EAP documentation and I'm still not sure about a few technical information which prevents me to fully understand this analysis:

* What is LogContext, what is its purpose? Is there some relationship between LogContext and Logging profile? This class is not described in product doc and javadoc seems to be poor.
* Can you please describe some details (in the analysis) about routing of messages to LogContext objects? Or provide a link to the documentation where this is described?
* LOGMGR-264 description contains some information about switching LogContext for a logger. Can you please elaborate on this topic in the analysis?
* LOGMGR-264 has MR against 2.2 branch. Why not to use 2.3 or master branch for EAP8 RFE?

logging/WFCORE-4807-log-routing.adoc
marekkopecky: > location in management model and default value.

These are still missing in the analysis. Although I assume that default value would be true, it is not mentioned explicitly in the analysis.

logging/WFCORE-4807-log-routing.adoc
marekkopecky: Can you please describe on previous example, how logs would be routed with after this RFE will be implemented? I hope that I understand how logs are routed now, but I'm still not sure about the goal of this RFE.

logging/WFCORE-4807-log-routing.adoc
```diff
@@ -0,0 +1,112 @@
...
+    ** This means that when disabled the bug in https://issues.redhat.com/browse/WFCORE-3376[WFCORE-3376] would
+       still be present.
+* The `LogContext` should be associated with a deployments class loader.
+    ** This means that if the TCCL is set to the deployment class loader log messages will be routed to the deployments
+       log context.
+    ** That may include messages that previously went to the `server.log` could end up in a handler defined on the
+       deployments log context. This is desired behavior IMO.
+* Loggers created on any log context should be routable to another log context based on the TCCL, if enabled.
+
+=== Nice-to-Have Requirements
```
marekkopecky, line 76: What about the migration? This RFE changes the behaviour somehow, so we should have a plan about the migration between EAP7 -> EAP 8. Migration will be implemented as a part of the following 2 RFEs, but I believe that this analysis should cover at least some recommended way for a migration of logging configuration

* https://issues.redhat.com/browse/EAP7-1704
* https://issues.redhat.com/browse/EAP7-1778

jamezp: I do wonder if we should default to the current behavior instead of defaulting to the new behavior like I suggested. That would make migration not really an issue I think.

marekkopecky: New EAP major version is the best version where we can change wrong behaviour to correct behaviour. Do you have some specific reasons for keeping the wrong behaviour as default?

jamezp: I'm not sure I'd call it the "wrong behavior" really, but the reason I think we might want to keep the current default is just that it won't create any migration issues. However, I've got no real string opinion either way.

marekkopecky: > I'm not sure I'd call it the "wrong behavior" really

Hm, ok, so maybe that we should finish the clarification of this analysis, there are still some open threads. Btw:

https://github.com/wildfly/wildfly-proposals/pull/281#discussion_r695546316:

> This is my current understanding of this example: ... this classes are already configured to use LogContext from custom profile, which is wrong

https://github.com/wildfly/wildfly-proposals/pull/281#discussion_r696001988:

> Yes. Your understanding is correct.

logging/WFCORE-4807-log-routing.adoc
jamezp: Sorry, I'm really not sure how I missed this. With this RFE logs would be routed to any specific log context associated with the deployment. In the above example log messages from `org.jboss.custom.module` would end up in `a.war`'s log context for anything coming from `a.war` and to the `server.log` for anything in `b.war`.

marekkopecky: Thank you. Can you please clarify this in a hard requirements section of the analysis directly?

Maybe (if I understand the situation correctly):
* `This means that if the TCCL is set to the deployment class loader log messages will be routed to the deployments log context.` -> `This means that if the TCCL is set to the deployment class loader log messages (from any modules called within that deployment) will be routed to the deployments log context.` + When the TCCL is set to the deployment class loader? When allow-log-routing == true? If yes, please mention this in the analysis.

logging/WFCORE-4807-log-routing.adoc
marekkopecky: This is still not fixed in the analysis.

logging/WFCORE-4807-log-routing.adoc
```diff
@@ -0,0 +1,112 @@
...
+
+=== Other Interested Projects
+
+* Any consumer of WildFly Core that uses the logging subsystem.
+
+== Requirements
+
+=== Hard Requirements
+
+* The attribute, `allow-log-routing`, is a simple boolean attribute which determines if the new or old behavior will be used.
```
marekkopecky, line 65: Please explicitly specify location of `allow-log-routing`attribute in management model and default value.

logging/WFCORE-4807-log-routing.adoc
marekkopecky: Analysis still talks about https://issues.redhat.com/browse/WFCORE-3376 "bug". https://issues.redhat.com/browse/WFCORE-3376 itself is a jira with bug type. And the behaviour where logs from deployment A uses LogContext from deployment B really seems like a bug. On the other hand, "it won't create any migration issues" this doesn't looks like strong reasoning in my PoV. If "no migration issues" would be a requirement, we can't do any improvements.

jamezp: Well the bug cannot be fixed without this. However, this is more than just a bug fix because it will change the behavior of other logs which may be unexpected.

I guess what I meant by "won't create any migration issues" was just if the attribute is set to `false` there is no change in behavior.

logging/WFCORE-4807-log-routing.adoc
jamezp: @marekkopecky I've attempted to clarify this in the second paragraph of the overview. I hope this helps. I'm trying to think of a good way to explain this and I realize it's kind of complicated and requires some knowledge of the log manager. :) It might be a bit confusing for end-users so I'll try to think of some good wording for documentation.

logging/WFCORE-4807-log-routing.adoc
marekkopecky: > if the attribute is set to false there is no change in behavior.

+1. So the migration plan is just to set the attribute to false. What about the way how to achieve this? Manually? Or by transformers? Or by some another way? cc @emmartins

logging/WFCORE-4807-log-routing.adoc
```diff
@@ -0,0 +1,116 @@
...
+Currently, any modules that log generally log onto the system log context. In some cases, for example a deployment using
+a logging profile, the loggers may end up getting created on the wrong `LogContext`. As an example if `a.war` and `b.war`
+both depend on module `org.jboss.custom.module` and at least one deployment uses a logging profile whichever deployment
+initializes the custom module first will have the loggers created on that `LogContext`. If a logging profile is being
+used it would be that profiles log context. This means if `a.war` uses a logging profile and the module is accessed first
+by `a.war`, then static loggers will be created on the logging profiles `LogContext`. Once `b.war` invokes something to
+be logged in the custom module, then `b.war` 's log messages end up in the logging profiles `LogContext` instead of the
+systems `LogContext`.
+
+https://issues.jboss.org/browse/LOGMR-264[LOGMR-264] will introduce a way to route log messages based on the log context.
```
marekkopecky, line 20: Typo here, LOGMR-264 -> LOGMGR-264, same for "Related Issues" section

marekkopecky:
Can you please explicitly mention in the analysis what will be the default value of `allow-log-routing` attribute? I see two resolved threads about this topic, but  I still don't see this info mentioned in the analysis:

* https://github.com/wildfly/wildfly-proposals/pull/281#discussion_r695561754
* https://github.com/wildfly/wildfly-proposals/pull/281#discussion_r745565981

logging/WFCORE-4807-log-routing.adoc
jamezp: @marekkopecky If we default the attribute to `false` then there should be no migration issues. Only users that manually set the attribute to `true` will see the affects of this RFE.

jamezp: > Can you please explicitly mention in the analysis what will be the default value of `allow-log-routing` attribute? I see two resolved threads about this topic, but I still don't see this info mentioned in the analysis:
> 
> * [[WFCORE-4807] Add the analysis document for log routing with an optio… #281 (comment)](https://github.com/wildfly/wildfly-proposals/pull/281#discussion_r695561754)
> * [[WFCORE-4807] Add the analysis document for log routing with an optio… #281 (comment)](https://github.com/wildfly/wildfly-proposals/pull/281#discussion_r745565981)

Sorry about that @marekkopecky. I meant to do that when I resolved those comments and got side-tracked. It's fixed now.

jamezp:
> Can you please explicitly mention in the analysis what will be the default value of `allow-log-routing` attribute? I see two resolved threads about this topic, but I still don't see this info mentioned in the analysis:
> 
> * [[WFCORE-4807] Add the analysis document for log routing with an optio… #281 (comment)](https://github.com/wildfly/wildfly-proposals/pull/281#discussion_r695561754)
> * [[WFCORE-4807] Add the analysis document for log routing with an optio… #281 (comment)](https://github.com/wildfly/wildfly-proposals/pull/281#discussion_r745565981)

Sorry about that @marekkopecky. I meant to do that when I resolved those comments and got side-tracked. It's fixed now.

logging/WFCORE-4807-log-routing.adoc
marekkopecky: > If we default the attribute to false then there should be no migration issues

Ah, from what I understand from your last comment, default value would be true and migration path would be "set false". Anyway, as we discussed earlier, there is some bug, that is fixed by this change. And default behaviour should be without a bug in my PoV.

>  jamezp 10 days ago
> Well the bug cannot be fixed without this

jamezp: It's kind of a weird one I think. While it does fix a bug, it's not a common bug. It changes behavior more than just fixes a bug which is I thought opting for `false` seemed correct. I've got no strong opinion either way, but defaulting to `true` may be a surprise to users using per-deployment logging.

marekkopecky: Is there some option how to fix that bug without changing the behaviour in other cases? AFAIU new behaviour seems to be better, right? New EAP8 major release is the best opportunity when we can fix/change wrong/misleading behaviour

jamezp: @marekkopecky There is no way to fix the bug without this change. I'm personally fine if we default to the new behavior. If everyone agrees I'm happy to use `true` for the default. And you're right that EAP 8 is the release to do this in.

jamezp:
I'm closing this as we are likely going to use a different approach and a new analysis will be needed.

PR #283, opened 2020-01-27
WFLY-12733 Proposal for EJB Timers: Forcing refresh timers in a datab…

chengfang:
…ase-data-store in a clustered environment.

JIRA:
https://issues.redhat.com/browse/WFLY-12733

Related Issue:
https://issues.redhat.com/browse/EAP7-1377

### DIFF
--- a/ejb/WFLY-12733-refresh-timer-programatically.adoc
+++ b/ejb/WFLY-12733-refresh-timer-programatically.adoc
@@ -1,5 +1,5 @@
 = EJB Timers: Forcing refresh timers in a database-data-store in a clustered environment
-:author:            Cheng Fanbg
+:author:            Cheng Fang
 :email:             cfang@redhat.com
 :toc:               left
 :icons:             font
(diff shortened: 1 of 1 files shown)


ejb/WFLY-12733-refresh-timer-programatically.adoc
```diff
@@ -0,0 +1,120 @@
+= EJB Timers: Forcing refresh timers in a database-data-store in a clustered environment
+:author:            Cheng Fanbg
```
rhusar: Who is `Fanbg`? :-)

chengfang: typo fixed, thanks!

chengfang:
I've just updated the proposal document to change to use EE interceptor (instead of WildFly custom interceptor) for configuring programmatic timer refresh. It should be a bit easier for users.

chengfang:
@tadamski @fl4via I've updated the proposal a little bit, replacing the use of WildFly interceptor to EE interceptor.  Please review.  Thanks!

chengfang:
A predefined interceptor will then be a public api with support promise, and there is not a good place (libary jar) to contain this class. On the other hand, if applications already have their interceptors, they can just add this simple logic there; otherwise, it should be simple for them to implement it.

PR #286, opened 2020-02-05
Proposal for CLOUD-3496, S2i to handle CLI scripts generated by SSO c…

jfdenise:
…onsole


msve:
I'm a bit concerned on what is going to be accepted inside the deployment in those CLI scripts, but the documentation draft you prepared explicitly mentions CLI scripts need to be exported from SSO client, so I think we will be covered on that part.

So I think this is ok from QE side and can be merged

jfdenise:
@msve , Thank-you, would you mind approve the Analysis doc?

PR #288, opened 2020-02-12
[WFCORE-4842] Add support for TLS 1.3 using the OpenSSL TLS provider with Elytron

fjuma:
https://issues.redhat.com/browse/WFCORE-4842

### DIFF
--- a/elytron/WFCORE-4842-tls-1.3-with-openssl.adoc
+++ b/elytron/WFCORE-4842-tls-1.3-with-openssl.adoc
@@ -1,0 +1,122 @@
+= WFCORE-4842 Add support for TLS 1.3 using the OpenSSL TLS provider with Elytron
+:author:            Farah Juma
+:email:             fjuma@redhat.com
+:toc:               left
+:icons:             font
+:idprefix:
+:idseparator:       -
+
+== Overview
+
+https://issues.jboss.org/browse/WFCORE-4172[WFCORE-4172] added support for TLS 1.3 when using the
+JSSE TLS provider with Elytron. This task is to add support for TLS 1.3 when using the OpenSSL TLS
+provider with Elytron.
+
+== Issue Metadata
+
+=== Issue
+
+* https://issues.redhat.com/browse/WFCORE-4842[WFCORE-4842]
+
+=== Related Issues
+
+* https://issues.jboss.org/browse/WFSSL-16[WFSSL-16]
+* https://issues.redhat.com/browse/SSLNTV-3[SSLNTV-3]
+* https://issues.redhat.com/browse/WFLY-13085[WFLY-13085]
+* https://issues.jboss.org/browse/EAP7-1414[EAP7-1414]
+
+=== Dev Contacts
+
+* mailto:{email}[{author}]
+
+=== QE Contacts
+
+* mailto:jstourac@redhat.com[Jan Stourac]
+
+=== Testing By
+// Put an x in the relevant field to indicate if testing will be done by Engineering or QE.
+// Discuss with QE during the Kickoff state to decide this
+
+[ ] Engineering
+
+[x] QE
+
+* TBD
+
+=== Affected Projects or Components
+
+* WildFly OpenSSL
+* WildFly Core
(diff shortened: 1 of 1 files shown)


elytron/WFCORE-4842-tls-1.3-with-openssl.adoc
```diff
@@ -0,0 +1,121 @@
...
+
+=== Non-Requirements
+
+== Test Plan
+
+All testing should be done using the OpenSSL TLS provider. The TLS 1.3 cipher suites that are
+actually available will depend on this provider.
+
+One-way and two-way SSL tests will be added to the WildFly OpenSSL testsuite and to the WildFly
+Core testsuite that make use of the TLS 1.3 protocol and TLS 1.3 cipher suites. The `java.specification.version`
```
jstourac, line 105: Just checking whether this part is still true since we agreed that we'll cover this feature in one of our internal testsuites?

elytron/WFCORE-4842-tls-1.3-with-openssl.adoc
```diff
@@ -0,0 +1,121 @@
...
+system property will be used to make sure that these new tests only get run when JDK 11 or higher
+is in use since TLS 1.3 was only introduced in JDK 11.
```
jstourac, line 107: Is new system property necessary? Cannot we use combination of `org.junit.Assume` and `System.getProperty('java.version')`?

elytron/WFCORE-4842-tls-1.3-with-openssl.adoc
darranl: Overall I don't believe we should ever take a decision to not include tests in the open source test suites if we have the option to perform the tests here regardless of any internal testing.

Ideally test cases should be as close to the engineer making changes as is practically possible so that any regressions can be picked up early - test cases within WildFly Core give us an opportunity to detect regressions definitely before WildFly Core is tagged and potentially before WildFly Open SSL is tagged.

Tests within the internal testsuite on the other hand only detect regressions late in the process potentially meaning any fix could require WildFly OpenSSL, WildFly Core, and WildFly all to be tagged again for something we had the potential to detect earlier.

elytron/WFCORE-4842-tls-1.3-with-openssl.adoc
fjuma: Mentioned `System.getProperty('java.specification.version')` here just to be consistent with what is currently used in the existing TLS 1.3 tests in the Elytron and WildFly Core testsuites, e.g., see below:

https://github.com/wildfly/wildfly-core/blob/9dd533251b272b6da025086e794d392d5b750d3a/elytron/src/test/java/org/wildfly/extension/elytron/TlsTestCase.java#L348-L349

elytron/WFCORE-4842-tls-1.3-with-openssl.adoc
fjuma: @jstourac Tests will still be added in the WildFly OpenSSL testsuite (I am currently working through these). Ideally, it would be good to have some test cases in the WildFly Core testsuite as well but I'll need to check with Ken first to see if we can perform the required setup on the Linux and Windows CI machines used for these runs.

jstourac: @darranl I completely agree. This case is a little bit more complicated with the setup necessary on platforms where OpenSSL libraries are not available at the moment in WildFly CI. That is why we talked with Farah that there will be tests for this in `WildFly OpenSSL upstream TS` but we're gonna use internal TS instead of `WildFly upstream TS` for integraiton tests. I wasn't quite sure what was the decision regarding to `WildFly Core upstream TS`.

@fjuma okay then, let me know in case you need anything. Thanks!

elytron/WFCORE-4842-tls-1.3-with-openssl.adoc
jstourac: Oh, sorry, of course. :)

PR #290, opened 2020-02-14
WFLY-13024 Integrate MP Fault Tolerance 2.1

rhusar:
https://issues.redhat.com/browse/WFLY-13024
https://issues.redhat.com/browse/EAP7-1428

### DIFF
--- a/microprofile/WFLY-13024_Integrate_MP_Fault_Tolerance_2_1.adoc
+++ b/microprofile/WFLY-13024_Integrate_MP_Fault_Tolerance_2_1.adoc
@@ -1,0 +1,87 @@
+= WFLY-12590 Integrate MicroProfile Fault Tolerance 2.1
+:author:            Radoslav Husar
+:email:             rhusar@redhat.com
+:toc:               left
+:icons:             font
+:idprefix:
+:idseparator:       -
+:keywords:          microprofile,fault-tolerance,openshift
+
+== Overview
+
+The support for Eclipse MicroProfile (MP) Fault Tolerance (FT) spec version 2.0 was already integrated for all applications
+deployed in WildFly with https://issues.jboss.org/browse/WFLY-12590[WFLY-12590].
+This describes the continuation of this effort updating the spec to version 2.1 following the Eclipse MicroProfile 3.3 release.
+
+This task incorporates upgrading SmallRye Fault Tolerance implementation to 4.1.0 (or newer) and the spec to 2.1 jars and TCK,
+removing legacy libraries and dependencies such as hystrix, commons-config, hdrhistogram, etc.
+
+For changes in the 2.1 spec see https://download.eclipse.org/microprofile/microprofile-fault-tolerance-2.1/microprofile-fault-tolerance-spec.html#release_notes_21[the spec].
+
+NOTE: For convenience of interested parties some of the following parts are copied over from previous proposal and updated.
+
+== Issue Metadata
+
+=== Issue
+
+* https://issues.redhat.com/browse/WFLY-13024[WFLY-13024 Integrate MP Fault Tolerance 2.1]
+* https://issues.redhat.com/browse/EAP7-1428[EAP7-1428 Integrate MP Fault Tolerance 2.1 into EAP]
+
+=== Related Issues
+
+* https://issues.redhat.com/browse/WFLY-13122[WFLY-13122 Include MicroProfile Fault Tolerance subsystem in the default server profiles]
+
+=== Dev Contacts
+
+* mailto:{email}[{author}]
+
+=== QE Contacts
+
+* mailto:tterem@redhat.com[Tomas Terem] (just to note, the 2.0 spec was QE-ed by Miroslav Novak)
+
+=== Testing By
+// Put an x in the relevant field to indicate if testing will be done by Engineering or QE.
+// Discuss with QE during the Kickoff state to decide this
+[ ] Engineering
+
+[x] QE
+
+=== Affected Projects or Components
+
+* The specification repository resides at official Eclipse organization on GitHub in https://github.com/eclipse/microprofile-fault-tolerance[eclipse/microprofile-fault-tolerance] repository.
+* The implementation of Eclipse MicroProfile Fault Tolerance will be provided by SmallRye project https://github.com/smallrye/smallrye-fault-tolerance[smallrye/smallrye-fault-tolerance].
+
+Note this no longer requires external dependency.
+
+=== Other Interested Projects
+
+Not interested in our integration per se, but projects that have already integrated MP FT:
+
+* https://thorntail.io/[Thorntail] has integrated MicroProfile Fault Tolerance as a https://github.com/thorntail/thorntail/tree/master/fractions/microprofile/microprofile-fault-tolerance[fraction].
+* https://quarkus.io/[Quarkus] has integrated MicroProfile Fault Tolerance as an https://github.com/quarkusio/quarkus/tree/master/extensions/smallrye-fault-tolerance[extension].
+
+== Requirements
+
+* Update support to Eclipse MicroProfile Fault Tolerance version 2.1 for all applications deployed in WildFly.
+* Remove "restriction" imposed by hystrix configuration being singleton - which was essentially just a warning log message - when multiple deployments are deployed and configuring hystrix.
+
+== Nice-to-Have Requirements
+
+N/A
+
+== Non-Requirements
+
+N/A
+
+== Implementation Plan
+
+* Update `smallrye-fault-tolerance` and `microprofile-fault-tolerance-api` dependencies.
+* Remove no longer needed dependencies/licences/modules/etc.
+
+== Test Plan
+
+* The `smallrye-fault-tolerance` component is passing the MicroProfile Fault Tolerance 2.1 TCK during its release process.
+
+== Community Documentation
+
+The feature documentation in WildFly Administration Guide in a new MicroProfile Fault Tolerance section will be updated.


rhusar:
@tterem Any update on this?

rhusar:
There is really no standard, but I agree its redundant to say its proposal since the repo is only for proposals; changed now.

rhusar:
And regarding using [ around issue number ] is a legacy bugzilla standard where there was just a number a not a project identifier - some people still do that but they should stop. Its redundant.

microprofile/WFLY-13024_Integrate_MP_Fault_Tolerance_2_1.adoc
```diff
@@ -0,0 +1,88 @@
...
+
+Not interested in our integration per se, but projects that have already integrated MP FT:
+
+* https://thorntail.io/[Thorntail] has integrated MicroProfile Fault Tolerance as a https://github.com/thorntail/thorntail/tree/master/fractions/microprofile/microprofile-fault-tolerance[fraction].
+* https://quarkus.io/[Quarkus] has integrated MicroProfile Fault Tolerance as an https://github.com/quarkusio/quarkus/tree/master/extensions/smallrye-fault-tolerance[extension].
+
+== Requirements
+
+* Update support to Eclipse MicroProfile Fault Tolerance version 2.1 for all applications deployed in WildFly.
+* The extension will be made part of default configurations provided by WildFly.
```
tterem: Based on discussion on https://github.com/wildfly/wildfly/pull/13003, this probably won't be part of this RFE so please remove it.

rhusar: Looks like I only removed one reference and not this one - removed now.

microprofile/WFLY-13024_Integrate_MP_Fault_Tolerance_2_1.adoc
```diff
@@ -0,0 +1,87 @@
...
+:keywords:          microprofile,fault-tolerance,openshift
+
+== Overview
+
+The support for Eclipse MicroProfile (MP) Fault Tolerance (FT) spec version 2.0 was already integrated for all applications
+deployed in WildFly with https://issues.jboss.org/browse/WFLY-12590[WFLY-12590].
+This describes the continuation of this effort updating the spec to version 2.1 following the Eclipse MicroProfile 3.3 release.
+
+This task incorporates upgrading SmallRye Fault Tolerance implementation to 4.1.0 (or newer) and the spec to 2.1 jars and TCK,
+removing legacy libraries and dependencies such as hystrix, commons-config, hdrhistogram, etc.
```
tterem: After discussion with @mnovak1, I conclude that removing hystrix is much bigger change than just removing some legacy dependency. This analysis needs to contain more information about this change and how hystrix is replaced. For example how hystrix specific configuration properties are replaced in new implementation?

microprofile/WFLY-13024_Integrate_MP_Fault_Tolerance_2_1.adoc
```diff
@@ -0,0 +1,87 @@
...
+
+Not interested in our integration per se, but projects that have already integrated MP FT:
+
+* https://thorntail.io/[Thorntail] has integrated MicroProfile Fault Tolerance as a https://github.com/thorntail/thorntail/tree/master/fractions/microprofile/microprofile-fault-tolerance[fraction].
+* https://quarkus.io/[Quarkus] has integrated MicroProfile Fault Tolerance as an https://github.com/quarkusio/quarkus/tree/master/extensions/smallrye-fault-tolerance[extension].
+
+== Requirements
+
+* Update support to Eclipse MicroProfile Fault Tolerance version 2.1 for all applications deployed in WildFly.
+* Changes to supporting multiple deployments, since Hystrix is no longer used, there will be no warning on multiple deployments.
```
tterem: Please clarify what will be the exact behaviour of multiple deployments now?

rhusar: There is really nothing to be clarified, there is no special behavior.

microprofile/WFLY-13024_Integrate_MP_Fault_Tolerance_2_1.adoc
rhusar: > After discussion with @mnovak1, I conclude that removing hystrix is much bigger change than just removing some legacy dependency. 

The SR FT 4.x is a complete rewrite of the FT implementation , so yes, its a bigger change.

> This analysis needs to contain more information about this change and how hystrix is replaced.

That's an implementation detail, hystrix usage was removed completely from the project. If you need more details than that you would need to talk to the SR team or look at the code.

> For example how hystrix specific configuration properties are replaced in new implementation?

See the documentation in the PR, but simply put, there is no replacement since those properties are not relevant any more. The only configurable option are the thread pool sizes.

mnovak1: > The SR FT 4.x is a complete rewrite of the FT implementation , so yes, its a bigger change.

@rhusar It's risky as it's new implementation. Hystrix was battle-hardened for years. Does the new implementation have the same test coverage as hystrix? If not what's is missing. 

> That's an implementation detail, hystrix usage was removed completely from the project. If you need more details than that you would need to talk to the SR team or look at the code.

It's not as it made limitations for multiple deployment and tests written for previous MP FT 2.0 RFE were written with this in mind and started to fail.

microprofile/WFLY-13024_Integrate_MP_Fault_Tolerance_2_1.adoc
mnovak1: @rhusar could you describe the changes against "Hystrix" implementation, please?

microprofile/WFLY-13024_Integrate_MP_Fault_Tolerance_2_1.adoc
rhusar: > @rhusar It's risky as it's new implementation. 

Its not like we can choose which implementation we want to use, this is the only implementation we have that implements FT 2.1, so the question is moot. 

> Hystrix was battle-hardened for years.

On the contrary, its abandonware, outside our control and the integration has been cumbersome.

> Does the new implementation have the same test coverage as hystrix?

I believe same or better than the previous implementation. 

> It's not as it made limitations for multiple deployment and tests written for previous MP FT 2.0 RFE were written with this in mind and started to fail.

That's expected - those tests should be simply removed now.

mnovak1: > Its not like we can choose which implementation we want to use, this is the only implementation we have that implements FT 2.1, so the question is moot.

I do not suggest to return back. However this RFE should be about simple update incorporating diffs between MP FT 2.0 and 2.1. Replacing Hystrix which is the core of MP FT implementation is way above of this scope.
 
> > Does the new implementation have the same test coverage as hystrix?
> 
> I believe same or better than the previous implementation.

Could you confirm it, please? (note tier 0/1 tests are handled by DEV)
If it's really ok then only new tests for integration should be added. 

> > It's not as it made limitations for multiple deployment and tests written for previous MP FT 2.0 RFE were written with this in mind and started to fail.
> 
> That's expected - those tests should be simply removed now.

Those tests should be removed or changed based on new behaviour for multiple deployment. New tests for covering it should be added. However analyses does not cover this topic well. Could you elaborate on it, please?

rhusar: Hopefully the questions were confirmed on Monday presentation. Just to summarize, removing hystrix is not above the scope, coverage is good or better as confirmed by SR team and test coverage tests, there is no special behavior when it comes to multiple deployments so nothing to cover in the proposal.

microprofile/WFLY-13024_Integrate_MP_Fault_Tolerance_2_1.adoc
rhusar: I have changed the wording here so it should be less confusing.

mnovak1: @rhusar  So the only change is that Hystrix was singleton and now it's not? 

What about EARs and multiple modules? Each module can have its own `microprofile-config.properties` configuring MP FT?

rhusar: Hystrix configuration was singleton, the new impl doesn't impose those limitations. Supporting EAR and how the configuration is handled is a general MP question outside of the spec a definitely outside of the scope of this RFE.

tterem:
I suggest to add a note about the fact that Hystrix is replaced with Smallrye own implementation and that multideployment and EAR deployment are not supported in Wildfly at the moment. It would be good to have complete information in the analysis for future reference. However, these are clear for us now so i approve the analysis.

jamezp:
For the note of only single deployments are supported that should be a higher level note. As I understand it all of the MP projects run under the presumption of single deployments and it's been noted that will be need to be documented.

rhusar:
@jamezp Correct – ideally that needs to be addressed at an umbrella RFE for "MP 3.3 Support" which then applies to all specs. I don't want to add a note here that its not supported, given that there is nothing known here why this would not work, so thus this needs to be at a different level.

tterem:
Well, [MultipleFaultToleranceModuleEarTest](https://github.com/tterem/eap-microprofile-test-suite/blob/1428/microprofile-fault-tolerance/src/test/java/org/jboss/eap/qe/microprofile/fault/tolerance/MultipleFaultToleranceModuleEarTest.java#L39) is ignored and still doesn't pass. So while configuring two deployments differently work, configuring two modules of same EAR doesn't.

jamezp:
@tterem I'm not sure what the final decision will be, but it was discussed at a meeting that MP will only work with one deployment. If fault tolerance happens to work with two applications deployed it doesn't mean that other MP specs will unfortunately. As I understand it we're definitely not going to support multiple deployments in an EAR. The MP implementations were not created with multi-tenancy in mind.

rhusar:
@tterem so there are two test cases. One is out of date (hystrix) and the other one really tests MP Metrics integration the same class name. So that’s a deficiency in that spec and not this one or its implementation. Moreover the case is unlikely a real world use case - so basically a docs item.

rhusar:
Again that should be in a global MP 3.3 scope as its not specific to MP FT.

PR #294, opened 2020-02-19
WFLY-13129, Example CLI script for microprofile

jfdenise:


microprofile/WFLY_13129_Microprofile_CLI_Script_Example.adoc
```diff
@@ -0,0 +1,95 @@
...
+* By default it applies to standalone.xml, use _-Dconfig=<standalone-full.xml|standalone-ha.xml|standalone-full-ha.xml>_
+to apply to other standalone configurations.
+* The script informs the user with high level information on the modifications it applies to the configuration it is updating.
+* The script is expected to adapt to the differences that exist between various standalone configurations.
+* The script detects that a microprofile subsystem is already present in the config and skips it. 
+* The script secures the configuration using elytron security domain ApplicationDomain, http-authentication-factory and sasl-authentication-factory for items present in default configurations:
+** core-service/management
+*** access/identity security domain
+*** managementItf/http-upgrade sasl factory
+*** managementItf/ http auth factory
```
darranl: This one needs to be referencing the management-http-authentication resource.

microprofile/WFLY_13129_Microprofile_CLI_Script_Example.adoc
```diff
@@ -0,0 +1,95 @@
...
+* To launch the script: _bin/jboss-cli.sh --file=docs/examples/enable-microprofile.cli_
+* By default it applies to standalone.xml, use _-Dconfig=<standalone-full.xml|standalone-ha.xml|standalone-full-ha.xml>_
+to apply to other standalone configurations.
+* The script informs the user with high level information on the modifications it applies to the configuration it is updating.
+* The script is expected to adapt to the differences that exist between various standalone configurations.
+* The script detects that a microprofile subsystem is already present in the config and skips it. 
+* The script secures the configuration using elytron security domain ApplicationDomain, http-authentication-factory and sasl-authentication-factory for items present in default configurations:
+** core-service/management
+*** access/identity security domain
+*** managementItf/http-upgrade sasl factory
```
darranl: This one needs to be referencing the management-sasl-authentication resource.

microprofile/WFLY_13129_Microprofile_CLI_Script_Example.adoc
```diff
@@ -0,0 +1,95 @@
...
+*** managementItf/ http auth factory
+** remoting/http-connector sasl factory
```
darranl: This one needs to reference the application-sasl-authentication resource.

microprofile/WFLY_13129_Microprofile_CLI_Script_Example.adoc
```diff
@@ -0,0 +1,95 @@
...
+** undertow 
+*** other application-domain
+*** http-invoker UNTOUCHED, references legacy security realm, no http-auth factory to set.
```
darranl: If the only thing missing is the http-auth-factory we can add that within a script.

microprofile/WFLY_13129_Microprofile_CLI_Script_Example.adoc
```diff
@@ -0,0 +1,95 @@
...
+* The script is expected to adapt to the differences that exist between various standalone configurations.
+* The script detects that a microprofile subsystem is already present in the config and skips it. 
+* The script secures the configuration using elytron security domain ApplicationDomain, http-authentication-factory and sasl-authentication-factory for items present in default configurations:
+** core-service/management
+*** access/identity security domain
+*** managementItf/http-upgrade sasl factory
+*** managementItf/ http auth factory
+** remoting/http-connector sasl factory
+** undertow 
+*** other application-domain
```
darranl: Should reference ApplicationDomain

microprofile/WFLY_13129_Microprofile_CLI_Script_Example.adoc
```diff
@@ -0,0 +1,95 @@
...
+*** http-invoker UNTOUCHED, references legacy security realm, no http-auth factory to set.
+** ejb3 other application-domain
```
darranl: Should reference ApplicationDomain

microprofile/WFLY_13129_Microprofile_CLI_Script_Example.adoc
```diff
@@ -0,0 +1,95 @@
...
+** jberet security domain
```
darranl: Should reference ApplicationDomain

microprofile/WFLY_13129_Microprofile_CLI_Script_Example.adoc
```diff
@@ -0,0 +1,95 @@
...
+** messaging-activemq server security-domain(full only)
+* The script remove the security subsystem
```
darranl: On removal we should double check if any JASPIC or JACC updates are needed in the Elytron subsystem, maybe not but the present configuration I think may be lenient to allow it to co-exist with PicketBox - that leniency would no longer be required.

microprofile/WFLY_13129_Microprofile_CLI_Script_Example.adoc
```diff
@@ -0,0 +1,95 @@
...
+* The script remove the management ManagementRealm. NB: the script doesn’t touch the ApplicationRealm (used by http-invoker).
```
darranl: If we can define the missing http auth factory I think we can get away with slimming the ApplicationRealm so it is just about default SSL.

microprofile/WFLY_13129_Microprofile_CLI_Script_Example.adoc
```diff
@@ -0,0 +1,95 @@
...
+* Add Microprofile subsystems and extensions
```
darranl: For individual extension adds it may be worth checking if they are already defined individually.

microprofile/WFLY_13129_Microprofile_CLI_Script_Example.adoc
```diff
@@ -0,0 +1,95 @@
...
+=== Other Interested Projects
+
+== Requirements
+
+=== Hard Requirements
+
+* The script is packaged in the galleon-pack maven module. It is a default package installed when WildFly is installed.
+* Installation path is _<JBOSS_HOME>/docs/examples/enable-microprofile.cli_.
+* To launch the script: _bin/jboss-cli.sh --file=docs/examples/enable-microprofile.cli_
+* By default it applies to standalone.xml, use _-Dconfig=<standalone-full.xml|standalone-ha.xml|standalone-full-ha.xml>_
```
darranl: Personally I am not a fan of this approach where a script updates multiple configurations, instead it would feel safer if the user had already started their chosen config and then apply the update to their config but this is a pattern we have had elsewhere.

microprofile/WFLY_13129_Microprofile_CLI_Script_Example.adoc
```diff
@@ -0,0 +1,95 @@
...
+to apply to other standalone configurations.
+* The script informs the user with high level information on the modifications it applies to the configuration it is updating.
+* The script is expected to adapt to the differences that exist between various standalone configurations.
+* The script detects that a microprofile subsystem is already present in the config and skips it. 
+* The script secures the configuration using elytron security domain ApplicationDomain, http-authentication-factory and sasl-authentication-factory for items present in default configurations:
+** core-service/management
+*** access/identity security domain
```
darranl: This one should reference ManagementDomain

microprofile/WFLY_13129_Microprofile_CLI_Script_Example.adoc
jfdenise: @darranl , just to clarify the script doesn't update multiple configurations during its execution. It updates a single configuration. By default it is standalone.xml but a user can choose to update another one (-Dconfig=<config to update>).

darranl: Ah that is much nicer

jfdenise: Thank-you, will update the description.

microprofile/WFLY_13129_Microprofile_CLI_Script_Example.adoc
jfdenise: @darranl , the question I have is what should be the content of this http-auth-factory? Should we add back the http-auth-factory that was existing at some point?

darranl: Yes I think we should.

A bit off topic for this feature but I think maybe we should also revisit at some point the resources that reference a factory and see if we can do direct security domain references maybe with some additional attributes to select the available authentication mechanisms.

microprofile/WFLY_13129_Microprofile_CLI_Script_Example.adoc
jfdenise: @darranl , for each subsystem, I was thinking to check if the subsystem is present. if not present add both extension and subsystem.

microprofile/WFLY_13129_Microprofile_CLI_Script_Example.adoc
jfdenise: @darranl , is there something required on the script front to allow for this removal?

darranl: I think it is just a point to double check once we have the script, if there is anything it will only be small.

microprofile/WFLY_13129_Microprofile_CLI_Script_Example.adoc
jfdenise: @darranl , the factory that we could add and use for http-invoker:
```
<http-authentication-factory name="application-http-authentication" http-server-mechanism-factory="global" security-domain="ApplicationDomain">
  <mechanism-configuration>
    <mechanism mechanism-name="BASIC">
      <mechanism-realm realm-name="Application Realm"/>
    </mechanism>
    <mechanism mechanism-name="FORM"/>
 </mechanism-configuration>
</http-authentication-factory>
```

darranl: As this is now only being used for the invoker I would say drop FORM form the list of mechanisms so it is BASIC only

jfdenise: @darranl , I will mark this possible evolution as "Future work", it seems to go beyond the scope of this example script that is not expected to optimize fully the dependency on Realms.

microprofile/WFLY_13129_Microprofile_CLI_Script_Example.adoc
```diff
@@ -0,0 +1,108 @@
...
+* The script secures the configuration using elytron security domain ApplicationDomain, http-authentication-factory and sasl-authentication-factory for items present in default configurations:
+** core-service/management
+*** access=identity security-domain=ManagementDomain
+*** management-interface=http-interface http-upgrade.sasl-authentication-factory=management-sasl-authentication
+*** management-interface=http-interface http-authentication-factory=management-http-authentication
+** http-remoting-connector sasl-authentication-factory=application-sasl-authentication
+** undertow 
+*** application-security-domain=other security-domain=ApplicationDomain
+** ejb3/application-security-domain=other security-domain=ApplicationDomain
+** batch-jberet security-domain=ApplicationDomain
```
jfdenise: @darranl , I am now wandering if we should secure jberet. In default configs jberet is unsecure, we go beyond replacing legacy security there it seems.

jfdenise: @darranl after our discussion, this is required once the legacy security subsystem is removed.

darranl: +1 looks good.

microprofile/WFLY_13129_Microprofile_CLI_Script_Example.adoc
bstansberry: My high-level expectation for this script when it came to security is it would basically do what docs/examples/enable-elytron.cli does. And it seems that script trims details from the application realm.

I think I may have said otherwise in a different context where the discuss was about the Galleon layers we provide.

microprofile/WFLY_13129_Microprofile_CLI_Script_Example.adoc
```diff
@@ -0,0 +1,108 @@
...
+=== Other Interested Projects
+
+== Requirements
+
+=== Hard Requirements
+
+* The script is packaged in the galleon-pack maven module. It is a default package installed when WildFly is installed.
+* Installation path is _<JBOSS_HOME>/docs/examples/enable-microprofile.cli_.
+* To launch the script: _bin/jboss-cli.sh --file=docs/examples/enable-microprofile.cli_
+* When run, the script updates a given standalone configuration. By default it applies to standalone.xml. Use _-Dconfig=<standalone-full.xml|standalone-ha.xml|standalone-full-ha.xml>_
```
OndrejKotek: Do we expect that after applying the script on standalone(-ha).xml the updated config will be the same like standalone-microprofile(-ha).xml? (WFLY-13099)

jfdenise: The resulting config is different, it contains the standalone(-xxx).xml content + added microprofile subsystems/extensions minus security subsystem and management realm.

microprofile/WFLY_13129_Microprofile_CLI_Script_Example.adoc
OndrejKotek: @darranl Can you share a hint how to easily check this please?

microprofile/WFLY_13129_Microprofile_CLI_Script_Example.adoc
```diff
@@ -0,0 +1,108 @@
...
+NONE
+
+=== Non-Requirements
+
+* The script is not supposed to be applied on top of custom standalone configurations.
+* The script doesn't support domain/host configurations.
+
+== Future Works
+
+* Introduce an http-authentication-factory to be used by the http-invoker.
```
OndrejKotek, line 82: Is this captured in a subsequent RFE?

jfdenise: No, no issue logged to track it.

PR #296, opened 2020-02-24
[WFLY-13003] Analysis document for SameSite cookie attribute changes

fl4via:
Jira: https://issues.redhat.com/browse/WFLY-13003
EAP7: https://issues.redhat.com/browse/EAP7-1422

### DIFF
--- a/undertow/WFLY-13003_Support_SameSite_Cookie_Attribute.adoc
+++ b/undertow/WFLY-13003_Support_SameSite_Cookie_Attribute.adoc
@@ -1,0 +1,140 @@
+= WFLY-13003 Support SameSite Cookie Attribute
+:author:            Flavia Rainone
+:email:             frainone@redhat.com
+:toc:               left
+:icons:             font
+:idprefix:
+:idseparator:       -
+
+== Overview
+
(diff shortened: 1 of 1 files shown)


jmesnil:
looks good to me.

As I understand this does not impact any of the HTTP endpoints we have in WildFly management interface as we don't rely on cookies for them. @fl4via is that correct?

fl4via:
@jmesnil  Yes, that's correct. It is just an added feature to make applications compliant with the new rules for cookie attribute `SameSite`. The users that wish to use this handler will do so via expression-filter, such as:
/subsystem=undertow/configuration=filter/expression-filter=addSameSiteLax:add(expression="path-prefix('/mypathprefix') -> samesite-cookie(Lax)")
And then apply the filter to the host via:
/subsystem=undertow/server=default-server/host=default-host/filter-ref=addSsameSiteLax:add
Or the user could use WEB-INF/undertow-handlers.conf as an alternative.

fl4via:
@jamezp I added the information on the parameter that adds secure for none. If you could re-review it, it would be great!

jamezp:
Looks good to me @fl4via

PR #299, opened 2020-03-06
Cleanup mail proposal; move to a corresponding directory mail/ from root

rhusar:


rhusar:
@kabir you mentioned something in January about bringing up un-merged PRs for proposals, so this might be a cleanup one that just won't get attention from a Jira workflow.

PR #309, opened 2020-05-04
[WFLY-13353] Galleon layer for the Pojo subsystem

yersan:
Jira issue: https://issues.redhat.com/browse/WFLY-13353

### DIFF
--- a/pojo/WFLY-13353_galleon_layer_for_pojo_subsystem.adoc
+++ b/pojo/WFLY-13353_galleon_layer_for_pojo_subsystem.adoc
@@ -1,0 +1,100 @@
+= Galleon layer for the Pojo subsystem
+:author:            Yeray Borges
+:email:             yborgess@redhat.com
+:toc:               left
+:icons:             font
+:idprefix:
+:idseparator:       -
+
+== Overview
+
+The Pojo subsystem enables the deployment of applications containing JBoss Microcontainer services, as supported by previous versions of JBoss Application Server.
+
+The goal of this feature is to supply a https://docs.wildfly.org/galleon/#_layers[Galleon layer] which will be responsible for providing the pojo subsystem.
+
+The following is a configuration example of the WildFly bootable jar maven plugin to provision a server with the `pojo` layer:
+
+[source,xml]
+----
+<plugin>
+    <groupId>org.wildfly.plugins</groupId>
+    <artifactId>wildfly-jar-maven-plugin</artifactId>
+    <configuration>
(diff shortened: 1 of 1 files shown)


darranl:

pojo/WFLY-13353_galleon_layer_for_pojo_subsystem.adoc
```diff
@@ -0,0 +1,100 @@
...
+
+=== Testing By
+
+* [X] Engineering
+
+* [ ] QE
+
+=== Affected Projects or Components
+
+WildFly
```
fabiobrz, line 63: Isn't WildFly core affected as well? Can you add a link for each affected project?

fabiobrz (commented):
Hi @yersan - thanks for this proposal. I've left just one comment, otherwise the AD LGTM. Let me know what you think

pojo/WFLY-13353_galleon_layer_for_pojo_subsystem.adoc
```diff
@@ -0,0 +1,100 @@
...
+. Testing the Galleon layer provisioning. This testing is done by https://github.com/wildfly/wildfly/blob/master/testsuite/layers/src/test/java/org/jboss/as/test/layers/LayersTestCase.java[LayersTestCase]. The testsuite will be modified to add a new server provisioned with this layer in isolation and with this layer combined with all the layers. For each kind of provisioning, this test does the following:
+
+.. Verifies the provisioned modules are the expected ones.
+.. Verifies the provisioned server starts successfully.
+
+. Execution of WildFly tests related to the feature provisioned by this layer. Reuse the existing tests available on the WildFly test suite, which are directly testing this layer functionalities, and execute them on a server installation provisioned with this layer.
+
+== Community Documentation
+
+Community documentation plan is adding the layer to https://docs.wildfly.org/19/Admin_Guide.html#wildfly-galleon-layers[WildFly Galleon layers] in the section it belongs to.
```
fabiobrz, line 96: s/19/20/

fabiobrz (changes requested):
Hi @yersan - noticed one minor typo too.

pojo/WFLY-13353_galleon_layer_for_pojo_subsystem.adoc
yersan: The `pojo` subsystem is on WildFly not in WildFly core, so only WildFly is affected by this RFE.

fabiobrz (approved):

PR #316, opened 2020-06-16
Proposal for WFCORE-5009, Enhance JBoss CLI generic 'command' with su…

jfdenise:
…pport for existing resource

### DIFF
--- a/cli/WFCORE-5009-Evolve-generic-command.adoc
+++ b/cli/WFCORE-5009-Evolve-generic-command.adoc
@@ -1,0 +1,77 @@
+= WFCORE-5009 Enhance JBoss CLI generic "command" with support for existing resource
+:author:            Jean-Francois Denise
+:email:             jdenise@redhat.com
+:toc:               left
+:icons:             font
+:keywords:          comma,separated,tags
+:idprefix:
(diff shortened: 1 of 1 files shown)


tommaso-borgato:
@jfdenise I read that it's not possible to add a command like the following at present: `command add --node-type=/core-service=management/access --command-name=access` but I wasn't able to figure out how to leverage this new attribute `--node-child` to overcome this limitation; could you please add such an example?

jfdenise:
@tommaso-borgato , with this new option you can add a command for an existing resource only. So the path must contain type=<instanceName> otherwise the resource instance is not identified.
For example: command add --node-child=/core-service=management/access=authorization --command-name=authorization

I am also adding the WIP branch that contains new tests and help doc: https://github.com/wildfly/wildfly-core/compare/master...jfdenise:generic-command-node-child

tommaso-borgato:
@jfdenise LGTM, I set the AD "approved" flag on https://issues.redhat.com/browse/EAP7-1501 but I don't have the permissions to merge this PR

jfdenise:
@tommaso-borgato , the proposal will get merged at some point.

PR #317, opened 2020-06-17
Proposal for WFLY-13354 Add Galleon layers for the EJB subsystem

chengfang:
JIRA: https://issues.redhat.com/browse/WFLY-13354

### DIFF
--- a/ejb/WFLY-13354-galleon-for-ejb.adoc
+++ b/ejb/WFLY-13354-galleon-for-ejb.adoc
@@ -59,26 +59,19 @@ If a user is trying to slim their server I think of four key things they want to
 
 Based on the above factors, the following layers are to be supported:
 
-1. ejb3-lite. This is like what we provide in standalone.xml, but without the MDB instance pool or the remote connector resource.
+1. ejb-lite. This is like what we provide in standalone.xml, but without the MDB instance pool or the remote connector resource.
 (A webservices layer would depend on this.)
 
-2. ejb3. Builds (i.e. depends) on ejb3-lite and adds the discovery subsystem, the MDB pool and the remote connector resource.
+2. ejb. Builds (i.e. depends) on ejb-lite and adds the discovery subsystem, the MDB pool and the remote connector resource.
 This looks what we provide in standalone-full.xml, except no IIOP integration.
-
-3. ejb3-iiop. Builds (i.e. depends) on *ejb3-lite* and adds the iiop subsystem plus the integration resource in the EJB3 subsystem.
-This *could* depend on *ejb3* instead, but that means EJB is unnecessarily exposed over the HTTP interface and a lot of
-dependencies are brought in for messaging integration.
 
 All ejb layers should be preferring elytron as security provider, with an option to select legacy security.
 We should also make sure not to accidentally leave anything unsecured by default.
 
-Provisioning ejb3 + ejb3-iiop is similar to what's in standalone-full.xml, except that ejb3 subsystem in standard distribution
-uses legacy security domain.
-
 There'd be a couple ancillary layers as well:
 
 1. ejb-local-cache. This provides the infinispan subsystem resources related to local ejb caching
-(i.e. the ones in standalone.xml and standalone-full.xml.) The ejb3-lite layer *optionally* depends on this.
+(i.e. the ones in standalone.xml and standalone-full.xml.) The ejb-lite layer *optionally* depends on this.
 It's an optional dependency because the user when provisioning can *exclude* this layer and instead provision...
 
 2. ejb-dist-cache. This provides the infinispan subsystem resources related to distributed ejb caching
@@ -90,10 +83,10 @@ and jpa caching are handled in the layers for those features.
 TBH I don't see big FS footprint improvements coming out of these permutations. Big EJB dependencies include (in no particular order.)
 
 * IIOP. But the TM uses IIOP libs so they will be there regardless.
-* Remoting. It would be nice to be able to eliminate the need for remoting, e.g. for ejb3-lite if the user app isn't a remote ejb client. But that doesn't seem trivial and if a server is manageable remoting will be there anyway to support the CLI.
+* Remoting. It would be nice to be able to eliminate the need for remoting, e.g. for ejb-lite if the user app isn't a remote ejb client. But that doesn't seem trivial and if a server is manageable remoting will be there anyway to support the CLI.
 * JCA stuff. But if there's a datasource it's there anyway.
 * Infinispan. A benefit of a future SLSB-only layer would be this dep could be eliminated. (But it might be used for web caching etc anyway.)
-* Remote artemis integration.  This we do save if ejb3-lite is used.
+* Remote artemis integration.  This we do save if ejb-lite is used.
 
 
 == Test Plan
--- a/ejb/WFLY-13354-galleon-for-ejb.adoc
+++ b/ejb/WFLY-13354-galleon-for-ejb.adoc
@@ -29,6 +29,7 @@ for the EJB subsystem to support common predefined EJB feature sets.
 === QE Contacts
 
 * mailto:fburzigo@redhat.com[Fabio Burzigotti]
+* mailto:tborgato@redhat.com[Tommaso Borgato]
 
 === Testing By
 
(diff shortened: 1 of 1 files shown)


chengfang:
@yersan please review.

darranl:

ejb/WFLY-13354-galleon-for-ejb.adoc
```diff
@@ -0,0 +1,110 @@
...
+2. ejb-dist-cache. This provides the infinispan subsystem resources related to distributed ejb caching
+(i.e. the ones in standalone0ha.xml and standalone-full-ha.xml.)
+
+This pattern of "exclude an optional local caching layer and add a distributed variant" is how web session
+and jpa caching are handled in the layers for those features.
+
+TBH I don't see big FS footprint improvements coming out of these permutations. Big EJB dependencies include (in no particular order.)
+
+* IIOP. But the TM uses IIOP libs so they will be there regardless.
+* Remoting. It would be nice to be able to eliminate the need for remoting, e.g. for ejb3-lite if the user app isn't a remote ejb client. But that doesn't seem trivial and if a server is manageable remoting will be there anyway to support the CLI.
```
darranl, line 88: FY the remoting layer is going to be needing some work, I am seeing it is defaulting to a dependency on legacy security realms without setting up the correct dependencies but also needs cleaning up to use elytron.

ejb/WFLY-13354-galleon-for-ejb.adoc
```diff
@@ -0,0 +1,110 @@
...
+
+2. Eliminate unnecessary exposure.
+
+3. Eliminate jars, both to save filesystem footprint and to reduce any theoretical attack surface.
+
+4. Reduce configuration clutter and unnecessary services.
+
+Based on the above factors, the following layers are to be supported:
+
+1. ejb3-lite. This is like what we provide in standalone.xml, but without the MDB instance pool or the remote connector resource.
```
darranl, line 61: Do we need the 3 in all the layer names?  This is the major version number of the specification but we don't currently plan to support any other ejb version in parallel do we?

The reason I ask is it came up late in the development of JBoss AS7 that really the subsystem should not have contained the 3 but it was too late at that point.

darranl:
I have just submitted the following ready for review to WildFly Core to adjust the current remoting layer definition:
  https://github.com/wildfly/wildfly-core/pull/4291

ejb/WFLY-13354-galleon-for-ejb.adoc
```diff
@@ -0,0 +1,110 @@
...
+(A webservices layer would depend on this.)
+
+2. ejb3. Builds (i.e. depends) on ejb3-lite and adds the discovery subsystem, the MDB pool and the remote connector resource.
+This looks what we provide in standalone-full.xml, except no IIOP integration.
+
+3. ejb3-iiop. Builds (i.e. depends) on *ejb3-lite* and adds the iiop subsystem plus the integration resource in the EJB3 subsystem.
```
darranl, line 67: I had to pause the work but the IIOP side may need this issue resolved as the legacy security subsystem will not be available as a layer https://issues.redhat.com/browse/WFLY-13679

ejb/WFLY-13354-galleon-for-ejb.adoc
```diff
@@ -0,0 +1,110 @@
...
+This *could* depend on *ejb3* instead, but that means EJB is unnecessarily exposed over the HTTP interface and a lot of
+dependencies are brought in for messaging integration.
+
+A user could provision ejb3 + ejb3-iiop and get the equivalent to what's in standalone-full.xml.
```
darranl, line 71: We won't be supporting the legacy security subsystem in bootable jar so this equivalence may not be possible.

Instead some consideration to the default security domain may be needed.

ejb/WFLY-13354-galleon-for-ejb.adoc
tommaso-borgato: @darranl shall we mention WFLY-13679 as a blocker for the current RFE?

ejb/WFLY-13354-galleon-for-ejb.adoc
```diff
@@ -0,0 +1,110 @@
...
+
+. Test the Galleon layer provision.
+https://github.com/wildfly/wildfly/blob/master/testsuite/layers/src/test/java/org/jboss/as/test/layers/LayersTestCase.java[LayersTestCase]
+will be expanded to test the provision of the added layers when it is used isolated and with other layers.
+For each sort of provision, this test does the following:
+.. Verifies the provisioned modules are the expected ones.
+.. Verifies the provisioned server starts successfully.
+. Execution of existing tests on a minimum set of layers. Reuse the existing tests available on the WildFly test suite,
+which are directly testing features added by these layers, and execute them on a server installation provisioned with this layer.
+
```
tommaso-borgato, line 106: @chengfang can we have some more details here? for example security and clustering test are topics of interest

tommaso-borgato (changes requested):

chengfang:
Will take a closer look at the comments and update this analysis and test plan soon.

ejb/WFLY-13354-galleon-for-ejb.adoc
darranl: For this piece I will try and look again when I am back, but we do need to make sure any supported layers don't inadvertently pull in the legacy security subsystem - I would say if they do we should raise that as a Blocker until we can triage.

Having said that some layers will still pull in the PicketBox modules but we can slowly work to clean those up.

ejb/WFLY-13354-galleon-for-ejb.adoc
chengfang: Good point. On the other hand, I think ejb3 distinghish itself from old-style, heavy-weight ejb 2.x, so it conveys some positive connotation. I'm neutral on this naming, but if there is strong perference to drop the 3, I'm fine with renaming it.

ejb/WFLY-13354-galleon-for-ejb.adoc
chengfang: @tadamski can you take a look at the iiop-openjdk's dependency on picketbox (https://issues.redhat.com/browse/WFLY-13679)? Is its dependency on legacy security really required?

chengfang:
> For all new layers I think we need to make sure security is considered in the requirements,
> 
> We don't want to accidentally leave anything unsecured by default but ideally if layers can avoid a dependency on any security layers that would be preferred.
> 
> Failing that for the new layers if security dependencies are required we should be preferring elytron with an option to select legacy security. At some point in the future all legacy security related layers will be removed.
> 
> Can we please add a check for these questions in the requirements?

incorporated the above comment into the analysis doc.

ejb/WFLY-13354-galleon-for-ejb.adoc
```diff
@@ -0,0 +1,115 @@
...
+== Community Documentation
+
+Community documentation plan is adding the layer to https://docs.wildfly.org/19/Admin_Guide.html#wildfly-galleon-layers[WildFly Galleon layers]
```
tommaso-borgato, line 114: @chengfang the link should be to https://docs.wildfly.org/20/Admin_Guide.html#wildfly-galleon-layers

tommaso-borgato (changes requested):

ejb/WFLY-13354-galleon-for-ejb.adoc
tommaso-borgato: @chengfang, @darranl, @tadamski does this conversation prevents the AD from being approved? do you think we can approve the AD and move this conversation somewhere else?

chengfang: Let's investigate it ([WFLY-13679](https://issues.redhat.com/browse/WFLY-13679) Make legacy security optional for "org.wildfly.iiop-openjdk") separately. I think it should not be blocking this proposal.

chengfang: We'll drop ejb-iiop and iiop-openjdk layers from this feature. Once WFLY-13679 is resolved, we can then add them back.

PR #318, opened 2020-06-22
[WFLY-13585] Add Galleon layer for JSON-P

yersan:
Jira issue: https://issues.redhat.com/browse/WFLY-13585

### DIFF
--- a/jaxrs/WFLY-13585_Galleon_layer_jsonp_support.adoc
+++ b/jaxrs/WFLY-13585_Galleon_layer_jsonp_support.adoc
@@ -1,0 +1,113 @@
+= Galleon layer for JSON-P support
+:author:            Yeray Borges
+:email:             yborgess@redhat.com
+:toc:               left
+:icons:             font
+:idprefix:
+:idseparator:       -
+
+== Overview
+
+Jakarta JSON Processing (JSON-P) is a Java API to process (e.g. parse, generate, transform and query) JSON documents. It produces and consumes JSON in a streaming fashion (similar to StAX API for XML) and allows to build a Java object model for JSON using API classes (similar to DOM API for XML).
+
+The goal of this feature is to supply a https://docs.wildfly.org/galleon/#_layers[Galleon layer] which will be responsible for providing the JSON-P API and Implementation.
+
+The Galleon packages that provision JSON-P are:
+
+ * `javax.json.api`: It provisions the JBoss Modules module for JSON-P API
+ * `org.glassfish.jakarta.json`: It provisions the JBoss Modules module for JSON-P implementation
+
+Since the JSON-P API Jboss Modules module depends on JSON-P Impl module, the implementation is always provisioned as soon as the API is present. The JSON-P API module is provisioned always by the `base-server` Galleon layer; there are modules provisioned by the `base-server` that have a hard dependency on the JSON-P modules, e.g. `org.jboss.logmanager` and `org.wildfly.security.elytron-private`.
+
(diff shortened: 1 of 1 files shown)


jaxrs/WFLY-13585_Galleon_layer_jsonp_support.adoc
```diff
@@ -0,0 +1,113 @@
...
+
+[source,xml]
+----
+<plugin>
+    <groupId>org.wildfly.plugins</groupId>
+    <artifactId>wildfly-jar-maven-plugin</artifactId>
+    <configuration>
+        <feature-pack-location>wildfly@maven(org.jboss.universe:community-universe)#${version.wildfly}</feature-pack-location>
+        <layers>
+            <layer>jaxrs-server</layer>
```
fabiobrz, line 36: Sorry because I might have got it wrong. Why do we use an existing layer (I am aware it provides the JSON-P AI and impl.) instead of just using the new module? BTW what would it be the new module name?

jaxrs/WFLY-13585_Galleon_layer_jsonp_support.adoc
```diff
@@ -0,0 +1,113 @@
...
+. Testing the Galleon layer provisioning. This testing is done by https://github.com/wildfly/wildfly/blob/master/testsuite/layers/src/test/java/org/jboss/as/test/layers/LayersTestCase.java[LayersTestCase]. The testsuite will be modified to add a new server provisioned with this layer in isolation and with this layer combined with all the layers. For each kind of provisioning, this test does the following:
+
+.. Verifies the provisioned modules are the expected ones.
+.. Verifies the provisioned server starts successfully.
+
+. Execution of WildFly tests related to the feature provisioned by this layer. Reuse the existing tests available on the WildFly test suite, which are directly testing this layer functionalities, and execute them on a server installation provisioned with this layer.
+
+== Community Documentation
+
+Community documentation plan is adding the layer to https://docs.wildfly.org/19/Admin_Guide.html#wildfly-galleon-layers[WildFly Galleon layers] in the section it belongs to.
```
fabiobrz, line 109: s/19/20/

fabiobrz (changes requested):
Hi @yersan and thanks for this. I've dropped two minor comments.

jaxrs/WFLY-13585_Galleon_layer_jsonp_support.adoc
yersan: I'm not sure what do you mean here with "module", could you elaborate it a bit more?

We could replace in the example the `jaxrs-server` Galleon layer by `jsonp`, however, it does not look as useful case, for this reason I used `jaxrs-server` and an example. Since `jsonp` is always provisioned,` base-server` Galleon layer, which is the smaller Galleon layer, would be enough to show an example, however as I said, `jaxrs-server` sounds more relevant.

jaxrs/WFLY-13585_Galleon_layer_jsonp_support.adoc
yersan: @fabiobrz Replaced

jaxrs/WFLY-13585_Galleon_layer_jsonp_support.adoc
fabiobrz: > I'm not sure what do you mean here with "module", could you elaborate it a bit more?

Sorry, I used the wrong word for "layer". 

> We could replace in the example the jaxrs-server Galleon layer by jsonp, however, it does not look as useful case, for this reason I used jaxrs-server and an example. Since jsonp is always provisioned, base-server Galleon layer, which is the smaller Galleon layer, would be enough to show an example, however as I said, jaxrs-server sounds more relevant.

Agree, thanks.

fabiobrz (approved):

PR #319, opened 2020-06-22
[WFLY-13586] Add a Galleon layer for JSON Binding support

yersan:
Jira issue: https://issues.redhat.com/browse/WFLY-13586

### DIFF
--- a/jaxrs/WFLY-13586_Galleon_layer_jsonb.adoc
+++ b/jaxrs/WFLY-13586_Galleon_layer_jsonb.adoc
@@ -1,0 +1,122 @@
+= Galleon layer for JSON Binding support
+:author:            Yeray Borges
+:email:             yborgess@redhat.com
+:toc:               left
+:icons:             font
+:idprefix:
+:idseparator:       -
+
+== Overview
+
+Jakarta JSON Binding (JSON-B) is a standard binding layer for converting Java objects to/from JSON messages. It defines a default mapping algorithm for converting existing Java classes to JSON, while enabling developers to customize the mapping process through the use of Java annotations.
+
+The goal of this feature is to supply a https://docs.wildfly.org/galleon/#_layers[Galleon layer] which will be responsible for providing the JSON-B API and Implementation.
+
+The Galleon packages that provision JSON-B are:
+
+* `javax.json.bind.api`: It provisions the JBoss Modules module for JSON-B API.
+* `org.eclipse.yasson`: It provisions the JBoss Modules module for JSON-B implementation.
+
+Currently, there are two subsystems that declare runtime package dependencies that make possible the provision of the JSON-B modules:
+
+* The EE subsystem, which declares JSON-B API and Impl. as optional runtime package dependencies.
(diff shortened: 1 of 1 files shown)


jaxrs/WFLY-13586_Galleon_layer_jsonb.adoc
```diff
@@ -0,0 +1,79 @@
...
+
+* `javax.json.bind.api`: It provisions the JBoss Modules module for JSON-B API.
+* `org.eclipse.yasson`: It provisions the JBoss Modules module for JSON-B implementation.
+
+Currently, there are two subsystems that declare runtime package dependencies that make possible the provision of the JSON-B modules:
+
+* The EE subsystem, which declares JSON-B API and Impl. as optional runtime package dependencies.
+* The JAXRS subsystem, in this case via the registration of `org.jboss.resteasy.resteasy-json-binding-provider` package as an optional runtime package dependency, which depends on JSON-B API and Impl. modules.
+
+We want to make JSON-B an optional layer for JAXRS so users can exclude it if they want. To make it possible, we need to remove the runtime package registration done by the EE subsystem and make EE layer depend optionally on JSON-B. For the JAXRS subsystem counterpart, we will make the `org.jboss.resteasy.resteasy-json-binding-provider` package registration done by JAXRS subsystem passive, so it will be registered only if the JSON-B package dependencies are satisfied by the presence of JSON-B layer. JAXRS layer will depend optionally on JSON-B via EE layer.
```
marekkopecky, line 25: @yersan (cc @fabiobrz @honza-kasik )AFAIU, there will be two options:

* JSON-B layer will be included, resteasy uses resteasy-json-binding-provider by default for json
* JSON-B layer won't be included, resteasy uses jackson2 for json

Is this intended? Basically, server handles json in http requests in both cases.

yersan: Hi @marekkopecky

Yes, server would be able o handle json on http requests. The JSON-B layer brings in the JSR 367(JSON-B) API and Implementation as well as the RestEasy provider for jsonb. If you exclude jsonb layer from your provisioning you won't be able to use the JSR 367 to serialize an object into a Json format. It is unrelated with the fact whether you are able to handle json in http requests.

Maybe I could have misunderstood your concern about it, could you expand it if you see a problem with this exclusion?

marekkopecky: > It is unrelated with the fact whether you are able to handle json in http requests.

Ok, so this feature is about JSON-B only.

> Maybe I could have misunderstood your concern about it, could you expand it if you see a problem with this exclusion?

I'm just thinking whether "json" layer wouldn't be better for users than JSON-B layer. I guess that JSON-B layer will be extra small and can do the same as "-Dresteasy.preferJacksonOverJsonB" system property. Anyway, I'm not the one who define requirements.

yersan: Thanks @marekkopecky any feedback is very valuable here, so thanks for sharing. The reason of this layer is a try to align with a similar thorntail fraction. That's the reason of this one.

darranl:

fabiobrz (approved):

PR #320, opened 2020-07-02
[WFLY-13640] MP Reactive Messaging with Focus on Apache Kafka

kabir:
Proposal for inclusion of Reactive Messaging (which in turn needs Reactive Streams Operators and Context Propagation)

https://issues.redhat.com/browse/WFLY-13640

### DIFF
--- a/microprofile/WFLY-13640_MicroProfile_Reactive_Messaging.adoc
+++ b/microprofile/WFLY-13640_MicroProfile_Reactive_Messaging.adoc
@@ -83,7 +83,7 @@ It must be possible to:
 
 ** MicroProfile Config will be used to configure the connections to the Kafka streams.
 
-*** It should be possible to override (or supplement?) parts of the MicroProfile Config via system properties or environment variables for integration with the Kafka managed service.
+*** Because MicroProfile Config supports systems properties or environmental variable as default ConfigSources with higher ordinals than is given to a microprofile-config.properties packaged in the application, it is possible for a system that controls the server launch to use system properties or environment variables to configure integration with the Kafka managed service.
 
 * Use RSO to create/deal with streams
 * For Reactive Messaging, only constructs which are part of the MicroProfile Reactive Messaging 1.0 specification may be used by the user. See <<spec-api,Spec API>> below.
--- a/microprofile/WFLY-13640_MicroProfile_Reactive_Messaging.adoc
+++ b/microprofile/WFLY-13640_MicroProfile_Reactive_Messaging.adoc
@@ -1,0 +1,183 @@
+= MicroProfile Reactive Messaging with Focus on Apache Kafka
+:author:            Kabir Khan
+:email:             kabir.khan@redhat.com
+:toc:               left
+:icons:             font
+:idprefix:
+:idseparator:       -
+:wfversion:
+
+== Overview
+We want to offer MicroProfile Reactive Messaging 1.0 with Kafka integration.
+
+
+For WildFly {wfversion} the focus will be on a subset of this feature pack, with the focus on Reactive Messaging intergration with Kafka.
+The feature branch is: https://github.com/kabir/wildfly/tree/reactive-port.
+The https://github.com/wildfly-extras/wildfly-mp-reactive-feature-pack[the WildFly Extras MP Reactive Feature Pack] where this work was incubated has a `development` branch where community can evolve to offer experimental connector, Context Propagation etc. on top of what will be offered by this RFE once WildFly {wfversion} is out. The `development` branch essentially contains parts of the original feature pack branch that is not yet considered for WildFly.
+
+For WildFly {wfversion} we will:
+* Include the MicroProfile Reactive Messaging 1.0 specification
+* Limit Reactive Messaging functionality as far as possible to Reactive Messaging 1.0. The 'as far as possible' part will be explained later.
+* Include the SmallRye Reactive Messaging Kafka connector, but only use it via the MicroProfile Reactive Messaging 1.0 API
+* Include the MicroProfile Reactive Streams Operators 1.0.1 (RSO) specification since MicroProfile Reactive Messaging 1.0 depends on that. RSO is also useful for creating and handling streams.
+
+Links to the specifications:
+
+* https://github.com/eclipse/microprofile-reactive-messaging/releases[Reactive Messaging 1.0]
+* https://github.com/eclipse/microprofile-reactive-streams-operators/releases[RSO 1.0.1]
+
+Links to the SmallRye implementations:
+
+* https://github.com/smallrye/smallrye-reactive-messaging[Reactive Messaging]
+* https://github.com/smallrye/smallrye-mutiny/tree/master/reactive-streams-operators[RSO]
+
+Only supporting MicroProfile Reactive Messaging 1.0 means there is *no* support for the https://github.com/smallrye/smallrye-reactive-messaging/blob/master/api/src/main/java/org/eclipse/microprofile/reactive/messaging/Channel.java[@Channel] annotation and the https://github.com/smallrye/smallrye-reactive-messaging/blob/master/api/src/main/java/org/eclipse/microprofile/reactive/messaging/Emitter.java[Emitter] interface. These are essential to be allow user/imperative code push data into, and get data from, the reactive messaging channels. So this initial iteration will not be able to bridge user initiated code (such as a REST API call) and the streams used by Reactive Messaging. The processors and subscribers for Reactive Messaging may however 'call out' and for example save received data in a database.
+
+There are several libraries all implementing the http://www.reactive-streams.org[Reactive Streams] specification. One new implementation is https://github.com/smallrye/smallrye-mutiny[SmallRye Mutiny] which is used under the hood by the SmallRye implementations of RSO and Reactive Messaging. We already use https://github.com/ReactiveX/RxJava/tree/2.x[RxJava 2] as an internal (i.e. it is a private module) dependency for the runtime functionality of some existing WildFly extensions. However, these should not be exposed to users at this time. We should settle on the implementation of RSO, which, although has less functionality than the other two mentioned, is a MicroProfile specification.
+
+In addition for more 'finite' sets of data, Java's CompletionStage is supported.
+
+== Issue Metadata
+
+=== Issue
+
+* https://issues.redhat.com/browse/WFLY-13640
+* https://issues.redhat.com/browse/EAP7-1405
+
+=== Related Issues
+
+* https://issues.redhat.com/browse/EAP7-1518[Quickstart]
+
+=== Dev Contacts
+
+* mailto:{email}[{author}]
+
+=== QE Contacts
+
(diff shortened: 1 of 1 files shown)


microprofile/WFLY-13640_MicroProfile_Reactive_Messaging.adoc
```diff
@@ -0,0 +1,166 @@
...
+
+=== Galleon Layers
+The functionality is split up into layers bringing in modules and subsystems.
+
+The current list of layes can be found https://github.com/wildfly-extras/wildfly-mp-reactive-feature-pack#layers-in-this-feature-pack[here]. Note that the following layers will be removed in the product:
+
+* reactive-messaging-amqp
+* reactive-messaging-mqtt
+
+The 'convenience' layer `reactive-messaging-connectors` (in turn depended on by `microprofile-reactive-all`) will be adjusted not to include these two layers.
```
bstansberry, line 120: I recommend a lot of caution about convenience layers. They are problematic because they are inherently risky for evolution, e.g. the next version adds new mega-heavy connector 'foo' and users are unhappy to suddenly have foo.  Even though you warned them in docs or whatever.

In particular because cautious about having things depend on convenience layers as that hides the evolution risk.

kabir: I will remove these as they are not part of the backport

microprofile/WFLY-13640_MicroProfile_Reactive_Messaging.adoc
```diff
@@ -0,0 +1,203 @@
...
+
+Additionally, a deployment unit processor in the Reactive Messaging subsystem will by default ensure that there are no methods annotated with `@Channel` and `@Emitter` and throw an error if someone tries to use them from their deployment (for example if they compiled against the SmallRye Reactive Messaging API jar insteead of against the MicroProfile). The same applies to
+the usage of annotations from the `io.smallrye.reactive.messaging.annotations` package which are also considered experimental. Note that in WildFly, use of these will be allowed if the `jboss.as.reactive.messaging.experimental` system property is set to
+`true`, as this allows us to test the full functionality of each SmallRye Reactive Messaging upgrade. For product, this allowing
+of experimental features will be totally disabled.
+
+
+
+=== Organisation
+There is an extension/subsystem implementing each of the three specifications mentioned:
```
jstourac, line 144: There are mentioned following specifications here:

* https://github.com/eclipse/microprofile-reactive-messaging/releases - Reactive Messaging 1.0
* https://github.com/eclipse/microprofile-reactive-streams-operators/releases - RSO 1.0.1

what is the third one?

microprofile/WFLY-13640_MicroProfile_Reactive_Messaging.adoc
```diff
@@ -0,0 +1,203 @@
...
+The layers involved are listed below:
+
+==== microprofile-reactive-messaging
+This brings in the MicroProfile Reactive Messaging subsystem, modules and associated functionality.
+
+Layer dependencies:
+
+* cdi
+* microprofile-config
+* microprofile-reactive-streams-operators
```
jstourac, line 167: I suppose that `cdi` dependency is mentioned here on purpose since the layer itself depends on it. Otherwise this is transitively brought in by the `microprofile-reactive-streams-operators` layer.

microprofile/WFLY-13640_MicroProfile_Reactive_Messaging.adoc
```diff
@@ -0,0 +1,203 @@
...
+
+===== microprofile-reactive-messaging-kafka
+This brings in the MicroProfile Reactive Messaging Kafka connector
+which allows reactive messaging to send messsages to and receive messages from Kafka.
```
jstourac, line 171: What is the extra functionality of this layer in comparison with `microprofile-reactive-messaging`? Is it the kafka connector for RM?

jstourac (changes requested):
I've went through this. There seem to be some conflicts to be resolved. Some of the comments are nitpics.

microprofile/WFLY-13640_MicroProfile_Reactive_Messaging.adoc
kabir: That is Context Propagation, which is no longer part of this. It lives in the [feature pack](https://github.com/wildfly-extras/wildfly-mp-reactive-feature-pack) - the `development` branch contains the bits which have not yet 'incubated' to WildFly.

I will change `three` to `two`

microprofile/WFLY-13640_MicroProfile_Reactive_Messaging.adoc
kabir: Removed

microprofile/WFLY-13640_MicroProfile_Reactive_Messaging.adoc
kabir: Yes

microprofile/WFLY-13640_MicroProfile_Reactive_Messaging.adoc
```diff
@@ -0,0 +1,203 @@
...
+
+For WildFly 22 the focus will be on a subset of this feature pack, with the focus on Reactive Messaging intergration with Kafka. (For community the feature pack can evolve to offer more experimental features and Context Propagation for WildFly 22.)
+
+We will:
+
+>>>>>>> b7a909b... Reduce scope for RM 1.0 and Kafka
+* Include the MicroProfile Reactive Messaging 1.0 specification
+* Limit Reactive Messaging functionality as far as possible to Reactive Messaging 1.0. The 'as far as possible' part will be explained later.
+* Include the SmallRye Reactive Messaging Kafka connector, but only use it via the MicroProfile Reactive Messaging 1.0 API
+time for WildFly 21.
```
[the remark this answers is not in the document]
kabir: Cruft - this will be removed once I push :)

microprofile/WFLY-13640_MicroProfile_Reactive_Messaging.adoc
```diff
@@ -0,0 +1,203 @@
+= <INSERT TITLE HERE>
```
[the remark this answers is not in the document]
kabir: Fixed

microprofile/WFLY-13640_MicroProfile_Reactive_Messaging.adoc
```diff
@@ -0,0 +1,183 @@
...
+* [ ] Engineering
+
+* [X] QE
+
+=== Affected Projects or Components
+The changes will be made in WildFly adding a subsystem for each spec.
+
+The functionality of these subsystems will be provided by the SmallRye implementations and their dependencies.
+
+In addition for OpenShift we need to make sure that provides out-of-the box integration with the Kafka managed service (by leveraging integration at the operator level). This  integration is scheduled post-WildFly 23. My current understanding is that we should be able to use environment variables and system properties to indicate to the user application the address and ports of the Kafka managed service.
```
kabir, line 72: @jmesnil Does this look ok as we discussed yesterday?

microprofile/WFLY-13640_MicroProfile_Reactive_Messaging.adoc
```diff
@@ -0,0 +1,183 @@
...
+
+=== Hard Requirements
+It must be possible to:
+
+* Deploy a Reactive Messaging application into WildFly
+* Interact with Kafka streams via the Reactive Messaging 1.0 APIS
+
+** MicroProfile Config will be used to configure the connections to the Kafka streams.
+
+*** It should be possible to override (or supplement?) parts of the MicroProfile Config via system properties or environment variables for integration with the Kafka managed service.
```
kabir, line 86: @jmesnil As discussed yesterday.
I think probably 'supplement' is better since things like the commit strategy etc. should be set via the user application's config. IP or port of Kafka sounds more like external stuff.

If we want to override say the ports, then I believe we will need to do something, as IIRC the deployment microprofile-config.properties normally override the stuff coming from env vars and system properties. In that case we would need to do something to turn that on its head somewhat (possibly something like a ConfigSource looking for system properties and env vars following a certain format, returning them as properties and using a higher ordinal than the deployment microprofile-config.properties one).

microprofile/WFLY-13640_MicroProfile_Reactive_Messaging.adoc
jmesnil: I don't know enough about Kafka on OpenShift to confirm this.

If I read https://strimzi.io/docs/operators/latest/quickstart.html#proc-using-amq-streams-str correctly, EAP will need to know the IP address and port to connect to Kafka with something like: `kafka.bootstrap.servers=localhost:9092`  (from [Quarkus Kafka Guide](https://quarkus.io/guides/kafka))

So when we start the EAP application image, we need to setup an environment variable for it.
The values for the IP address and port can be deduced by looking at the `KafkaTopic` and `Kafka` CRD and check a matching listener.

So if the EAP application is using "my-cluster" topic, we have to look at the `KafkaTopic` with the name "my-topic".
Then in that resource, we look at the `strimzi.io/cluster` label to find the name of the `Kafka` resource that manages this topic.
Finally, we need to find a matching listeners from this `Kafka` resource: 

```
kind: Kafka
metadata:
  name: my-cluster
spec:
  kafka:
    replicas: 1
    listeners:
      - name: plain
        port: 9092
        type: internal
        tls: false
        ...
```

For the IP address, we have to find the service that exposes the matching listeners and determine its IP address.
Once we have that, we have all the info we need to create the `kafka.bootstrap.servers` config property and run EAP.

If we want to hide all this info seeking and service binding from the user, I envision the EAP Operator doing everything for you as long as you tell it that you want to communicate with the `my-topic` topic:

```
spec:
  kafka:
    topic: "my-topic"
    # to be able to determine which listener (and port) to connect to)
    listener: "plain"

microprofile/WFLY-13640_MicroProfile_Reactive_Messaging.adoc
jmesnil: That's not correct: env vars have higher ordinal than microprofile-config.properties (https://github.com/eclipse/microprofile-config/blob/master/spec/src/main/asciidoc/configsources.asciidoc#default-configsources)

So, the deployment can provide "sensible" default value in its microprofile-config.properties  but if we set an env var for this config property, it will take precedence

bstansberry: @kabir @jmesnil So, given the above should the wording of this bullet become:

* Because MicroProfile Config supports systems properties or environmental variable as default ConfigSources with higher ordinals than is given to a microprofile-config.properties packaged in the application, it is possible for system that control the server launch to use system properties or environment variables to configure integration with the Kafka managed service.

kabir: Incorporated Brian's suggestion. @jstourac FYI

jstourac (approved):

PR #329, opened 2020-08-17
[WFLY-13715] Certify Oracle 19c RAC Cluster Support with Oracle UCP

tadamski:


jca/WFLY-13715_Certify_Oracle_19c_RAC_Cluster_Support_with_Oracle_UCP.adoc
```diff
@@ -0,0 +1,51 @@
...
+== Requirements
+
+* customers should be able to take advantage of Oracle UCP features
+* there would be no changes to IronJacamar architecture
+
+== Proposed solution
+* notify OracleConnection about start and end of the request
+* a system property (notify.oracle.connecton) should be set that will switch the notification on
+* if notification is turned on, in org.jboss.jca.core.connectionmanager.pool.mcp.SemaphoreArrayListManagedConnectionPool invoke oracle.jdbc.OracleConnection#beginRequest,oracle.jdbc.OracleConnection#endRequest when connection is started and ended
+
```
maeste, line 48: We should not rely on Oracle specific implementation. We can use the equivalent method of standard java.sqlconnection for Java9+ and require customers to use Java9+
Moreover, considering we are using a standard method, we can even avoid to have system property and just check through reflection the presence of java.sql.Connection#beginRequest and java.sql.Connection#endRequest methods and invoke them IIF they are present (aka we are using JDBC API of Java9+)

tadamski: fixed

PR #330, opened 2020-08-21
[WFLY-13706] New feature pack to make the Keycloak client adapters provisionable using Galleon.

darranl:
As described in the analysis this new feature pack is to be provided from the Keycloak project, however as it's target audience is exclusively users of WildFly this analysis is considering the feature pack from that perspective.

https://issues.redhat.com/browse/WFLY-13706

### DIFF
--- a/elytron/WFLY-13706-keycloak-feature-pack.adoc
+++ b/elytron/WFLY-13706-keycloak-feature-pack.adoc
@@ -159,7 +159,7 @@ adapter subsystem.
                         <name>standalone.xml</name>
                         <model>standalone</model>
                         <layers>
-                            <layer>undertow</layer>
+                            <layer>web-server</layer>
                             <layer>keycloak-client-oidc</layer>
                         </layers>
                     </config>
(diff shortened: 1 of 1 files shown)


elytron/WFLY-13706-keycloak-feature-pack.adoc
```diff
@@ -0,0 +1,316 @@
...
+automatically provisioned?
+
+*TODO?* Should it be possible to omit the layer listing?
+
+*TODO?* Should we eliminate the need to specify the `web-server` layer?  Related to this some further debugging of what
+Keycloak really requires is needed,
+
+=== Provisioned using org.wildfly.plugins:wildfly-jar-maven-plugin
+
+The `wildfly-jar-maven-plugin` only allows a single feature pack location to be specified, for this reason the configuration
```
jfdenise, line 205: @darranl , you can use feature-packs configuration element. You can see an example: https://github.com/jfdenise/wildfly-jar-maven-plugin/blob/advanced-examples/examples/keycloak/pom.xml#L44

darranl: Thanks, will add that as an example as well.

darranl: Going to mark this one as resolved, I have just pushed another change with an inlined example as well so overall we really have four different examples now.

PR #337, opened 2020-09-24
WFLY-13793 Allow for a remote jms queue / topic not to use legacy amq…

soul2zimate:
…1 prefix

EAP7 issue: https://issues.redhat.com/browse/EAP7-1559
WFLY issue: https://issues.redhat.com/browse/WFLY-13793

WFLY PR: https://github.com/wildfly/wildfly/pull/14412

### DIFF
--- a/messaging/WFLY-13793_disable_legacy_amq1_prefix.adoc
+++ b/messaging/WFLY-13793_disable_legacy_amq1_prefix.adoc
@@ -1,0 +1,128 @@
+= [WFLY-13793] Allow for a remote jms queue / topic not to use legacy amq1 prefix
+:author:            Chao Wang
+:email:             chaowan@redhat.com
+:toc:               left
+:icons:             font
+:idprefix:          messaging,jms
+:idseparator:       -
+
+== Overview
+Currently we can disable the use of legacy amq1 prefixes `jms.queue` and `jms.topic` only at the connection factory level but not for queue or topic defined in the model or using JakartaEE annotations.
+The goal of this proposal is to be able to disable the usage of such prefixes at the destination level for external JMS queues and topics.
+
+== Issue Metadata
+
+=== Issue
+
+* https://issues.redhat.com/browse/WFLY-13793[WFLY-13793]
+
+=== Related Issues
+
+* https://issues.redhat.com/browse/EAP7-1559[EAP7-1559]
+* https://issues.redhat.com/browse/EAP7-1244[EAP7-1244]
+
+Previous EAP7-1244 exposed the enable1Prefixes configuration on external pooled connection factories, but it does not allow to disable the use of legacy amq1 prefix for queue or topic defined in the model or using JakartaEE annotations.
+
+=== Dev Contacts
+
+* mailto:chaowan@redhat.com[Chao Wang]
+* mailto:ehugonne@redhat.com[Emmanuel Hugonnet]
+
+=== QE Contacts
+
+tborgato@redhat.com
+
+=== Testing By
+[X] Engineering
+
+[ ] QE
+
+=== Affected Projects or Components
+* WildFly
+
+=== Other Interested Projects
+
+== Requirements
+
+=== Hard Requirements
+
+* Add a `enable-amq1-prefix`  boolean attribute to the https://wildscribe.github.io/WildFly/18.0/subsystem/messaging-activemq/external-jms-queue/[/subsystem=messaging-activemq/external-jms-queue resource] to control the enablement of Artemis 1.x prefix for queue defined on a remote Artemis server. Default is `true` (meaning that name retro-compatibility is enabled). For instance an queue named `testQueue`, by setting this attribute to `false`, it effectively allows you to create an external queue without amq1 prefix `jms.queue`. The created queue will be `testQueue`. Otherwise, it will create an external queue with amq1 prefix `jms.queue`. The created queue will be `jms.queue.testQueue`.
+
+
+* Add a `enable-amq1-prefix`  boolean attribute to the https://wildscribe.github.io/WildFly/18.0/subsystem/messaging-activemq/external-jms-topic/[/subsystem=messaging-activemq/external-jms-topic resource] to control the enablement of Artemis 1.x prefix for topic defined on a remote Artemis server. Default is `true` (meaning that name retro-compatibility is enabled). For instance a topic named `testTopic`, by setting this attribute to `false`, it effectively allows you to create an external topic without amq1 prefix `jms.topic`. The created topic will be `testTopic`. Otherwise, it will create an external topic with amq1 prefix `jms.topic`. The created topic will be `jms.topic.testTopic`.
+
+* For a destination (queue or topic) created by annotation `@JMSDestinationDefinition` on a remote Artemis Server, add a `enable-amq1-prefix` boolean property to control the enablement of Artemis 1.x prefix. If the property is not defined, the default is `true` (meaning that name retro-compatibility is enabled).
+
+[source, java]
+----
+@JMSDestinationDefinition(
+    // explicitly mention a resourceAdapter corresponding to a pooled-connection-factory resource to the remote server
+    resourceAdapter = "remote-artemis",
+    name="java:global/env/myQueue2",
+    interfaceName="javax.jms.Queue",
+    destinationName="myQueue2",
+    properties = {
+        "enable-amq1-prefix=false"
+    }
+)
+----
+
+=== Nice-to-Have Requirements
+
+N/A
+
+=== Non-Requirements
+
+N/A
+
+== Test Plan
+
+* WildFly integration testsuite `ExternalJMSDestinationManagementTestCase`
+
+Prepare Wildfly server messaging-activemq subsystem be able to use internal broker as remote one.
+Create queue `myExternalQueue` and `myExternalTopic` with `enable-amq1-prefix` set to `false`. Both created queues name should not contain amq1 prefix `jms.queue` or `jms.topic`.
+
+The opposite version of the test is: create queue `myExternalLegacyQueue` and `myExternalLegacyTopic` with `enable-amq1-prefix` set to `true`. Both created queue contains amq1 prefix `jms.queue` or `jms.topic`.
+
+
+
+* WildFly integration testsuite `ExternalJMSDestinationDefinitionMessagingDeploymentTestCase`
+
+Prepare Wildfly server messaging-activemq subsystem be able to use internal broker as remote one.
+Create an external JMS queue on a remote broker with the amq1 prefix disabled through JakartaEE annotation `@JMSDestinationDefinition`. i.e.:
+
+[source, java]
+----
+@JMSDestinationDefinitions(
+        value = {
+            @JMSDestinationDefinition(
+                    resourceAdapter = REMOTE_PCF,
+                    name = QUEUE_LOOKUP,
+                    interfaceName = "javax.jms.Queue",
+                    destinationName = QUEUE_NAME,
+                    properties = {"enable-amq1-prefix=false"}
+            ),
+            @JMSDestinationDefinition(
+                    resourceAdapter = REMOTE_PCF,
+                    name = TOPIC_LOOKUP,
+                    interfaceName = "javax.jms.Topic",
+                    destinationName = TOPIC_NAME,
+                    properties = {"enable-amq1-prefix=false"}
+            )
+        }
+)
+----
+
+It's expected to create a queue without amq1 prefix `jms.queue` or `jms.topic`, and be able to produce messages on the queue and receive same messages with MDB listening on that queue name without amq1 prefix `jms.queue` or `jms.topic`.
+
+Otherwise, if `enable-amq1-prefix` is not configured to `false`. It creates queue with amq1 prefix `jms.queue` or `jms.topic`. which is covered in `ExternalJMSDestinationDefinitionLegacyPrefixMessagingDeploymentTestCase`.
+
+== Community Documentation
+
+The feature will be documented in WildFly Admin Guide (in the Messaging Configuration section).
+
+== Release Note Content
+
+WildFly now introduces a new boolean attribute `enable-amq1-prefix` for creating external jms queue / topic without amq1 prefix.
+
+It's allowed to achieve this in the messaging subsystem model or via JakartaEE annotations at runtime.
--- a/messaging/WFLY-13793_disable_legacy_amq1_prefix.adoc
+++ b/messaging/WFLY-13793_disable_legacy_amq1_prefix.adoc
@@ -8,7 +8,26 @@
 
 == Overview
 Currently we can disable the use of legacy amq1 prefixes `jms.queue` and `jms.topic` only at the connection factory level but not for queue or topic defined in the model or using JakartaEE annotations.
-The goal of this proposal is to be able to disable the usage of such prefixes at the destination level for external JMS queues and topics.
+The goal of this proposal is to be able to disable the usage of such prefixes at the destination level for external JMS queues and topics. Thus we would avoid creation of prefixed destinations that will be useless.
+We start by following the procedure described in https://github.com/wildfly/quickstart/tree/main/remote-helloworld-mdb[Remote MDB Quickstart]. Basically what's relevant here is:
+* we start a remote Artemis Broker which is NOT running in compatibility mode (= it doesn't try to add prefixes for all incoming queue/topic names)
+* we create a connection on WildFly to the remote Artemis Broker (called pooled-connection-factory) which overrides the default value of enable-amq1-prefix (which is true) by setting it to enable-amq1-prefix=false: this means also the WildFly Artemis client won't try to add prefixes to all queue/topic names
+* we create an application which defines a queue named HelloWorldMDBQueue which mimics the default behavior by setting *properties = {"enable-amq1-prefix=true"}* (default would be true) on the queue definition:
+
+[source, java]
+----
+@JMSDestinationDefinition(
+    name = "java:/queue/HELLOWORLDMDBQueue",
+    interfaceName = "jakarta.jms.Queue",
+    destinationName = "HelloWorldMDBQueue",
+    properties = {
+        "enable-amq1-prefix=true"
+    }
+)
+----
+
+This results in a destination named 'HelloWorldMDBQueue' containing a single anycast queue named 'HelloWorldMDBQueue' being created on the remote Artemis Broker plus a useless destination named jms.queue.HelloWorldMDBQueue". The goal here is that by setting the properties = {"enable-amq1-prefix=false"} we won't see this useless detination being created.
+
 
 == Issue Metadata
 
@@ -20,6 +39,7 @@ The goal of this proposal is to be able to disable the usage of such prefixes at
 
 * https://issues.redhat.com/browse/EAP7-1559[EAP7-1559]
 * https://issues.redhat.com/browse/EAP7-1244[EAP7-1244]
+* https://issues.redhat.com/browse/EAP7-1792[EAP7-1792]
 
 Previous EAP7-1244 exposed the enable1Prefixes configuration on external pooled connection factories, but it does not allow to disable the use of legacy amq1 prefix for queue or topic defined in the model or using JakartaEE annotations.
 
@@ -46,10 +66,10 @@ tborgato@redhat.com
 
 === Hard Requirements
 
-* Add a `enable-amq1-prefix`  boolean attribute to the https://wildscribe.github.io/WildFly/18.0/subsystem/messaging-activemq/external-jms-queue/[/subsystem=messaging-activemq/external-jms-queue resource] to control the enablement of Artemis 1.x prefix for queue defined on a remote Artemis server. Default is `true` (meaning that name retro-compatibility is enabled). For instance an queue named `testQueue`, by setting this attribute to `false`, it effectively allows you to create an external queue without amq1 prefix `jms.queue`. The created queue will be `testQueue`. Otherwise, it will create an external queue with amq1 prefix `jms.queue`. The created queue will be `jms.queue.testQueue`.
+* Add a `enable-amq1-prefix`  boolean attribute to the https://wildscribe.github.io/WildFly/29.0/subsystem/messaging-activemq/external-jms-queue/[/subsystem=messaging-activemq/external-jms-queue resource] to control the enablement of Artemis 1.x prefix for queue defined on a remote Artemis server. Default is `true` (meaning that name retro-compatibility is enabled). For instance an queue named `testQueue`, by setting this attribute to `false`, it effectively allows you to create an external queue without amq1 prefix `jms.queue`. The created queue will be `testQueue`. Otherwise, it will create an external queue with amq1 prefix `jms.queue`. The created queue will be `jms.queue.testQueue`.
 
 
-* Add a `enable-amq1-prefix`  boolean attribute to the https://wildscribe.github.io/WildFly/18.0/subsystem/messaging-activemq/external-jms-topic/[/subsystem=messaging-activemq/external-jms-topic resource] to control the enablement of Artemis 1.x prefix for topic defined on a remote Artemis server. Default is `true` (meaning that name retro-compatibility is enabled). For instance a topic named `testTopic`, by setting this attribute to `false`, it effectively allows you to create an external topic without amq1 prefix `jms.topic`. The created topic will be `testTopic`. Otherwise, it will create an external topic with amq1 prefix `jms.topic`. The created topic will be `jms.topic.testTopic`.
+* Add a `enable-amq1-prefix`  boolean attribute to the https://wildscribe.github.io/WildFly/29.0/subsystem/messaging-activemq/external-jms-topic/[/subsystem=messaging-activemq/external-jms-topic resource] to control the enablement of Artemis 1.x prefix for topic defined on a remote Artemis server. Default is `true` (meaning that name retro-compatibility is enabled). For instance a topic named `testTopic`, by setting this attribute to `false`, it effectively allows you to create an external topic without amq1 prefix `jms.topic`. The created topic will be `testTopic`. Otherwise, it will create an external topic with amq1 prefix `jms.topic`. The created topic will be `jms.topic.testTopic`.
 
 * For a destination (queue or topic) created by annotation `@JMSDestinationDefinition` on a remote Artemis Server, add a `enable-amq1-prefix` boolean property to control the enablement of Artemis 1.x prefix. If the property is not defined, the default is `true` (meaning that name retro-compatibility is enabled).
 


messaging/WFLY-13793_disable_legacy_amq1_prefix.adoc
```diff
@@ -0,0 +1,91 @@
+= [WFLY-13793] Allow for a remote jms queue / topic not to use legacy amq1 prefix
+:author:            Chao Wang
+:email:             chaowan@redhat.com
+:toc:               left
+:icons:             font
+:idprefix:          messaging,jms
+:idseparator:       -
+
+== Overview
+Currently we can disable the use of legacy amq1 prefixes `jms.queue` and `jms.topic` only at the cnnection factory level but not for queue or topic defined in the model or using JakartaEE annotations.
```
tommaso-borgato, line 10: @soul2zimate This dives directly into the details; I am having difficulties to figure out the whole picture;

Can you add details about the following? 

- which are all the points in the WildFly configurations where it is possible to configure the legacy prefixes? 
- how the setting on the connection factory and on queue and topic work together? is there some kind of overlapping?
- how does a complete configuration using legacy prefixes looks like? (also the Artemis part would be useful)
- how does a complete configuration disabling using legacy prefixes looks like? (also the Artemis part would be useful)

The goal is to have a complete picture of what is the current situation and how it's going to change and, of course, being able to do the setup on my local PC

tommaso-borgato (changes requested):

tommaso-borgato (changes requested):

messaging/WFLY-13793_disable_legacy_amq1_prefix.adoc
```diff
@@ -0,0 +1,91 @@
...
+Previous EAP7-1244 exposed the enable1Prefixes configuration on external pooled connection factories, but it does not allow to disable the use of legacy amq1 prefix for queue or topic defined in the model or using JakartaEE annotations.
+
+=== Dev Contacts
+
+* mailto:chaowan@redhat.com[Chao Wang]
+* mailto:ehugonne@redhat.com[Emmanuel Hugonnet]
+
+=== QE Contacts
+
+TODO
```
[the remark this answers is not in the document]
soul2zimate: QE Contact is added

messaging/WFLY-13793_disable_legacy_amq1_prefix.adoc
```diff
@@ -0,0 +1,91 @@
...
+
+=== Testing By
+// Put an x in the relevant field to indicate if testing will be done by Engineering or QE. Discuss with QE during the Kickoff state to decide this
+[ ] Engineering
```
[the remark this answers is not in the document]
soul2zimate: Added "X"

messaging/WFLY-13793_disable_legacy_amq1_prefix.adoc
```diff
@@ -0,0 +1,91 @@
...
+
+=== Other Interested Projects
+
+== Requirements
+
+* Add a `enable-amq1-prefix`  boolean attribute to the https://wildscribe.github.io/WildFly/18.0/subsystem/messaging-activemq/external-jms-queue/[/subsystem=messaging-activemq/external-jms-queue resource] to control the enablement of Artemis 1.x prefix for queue defined on a remote Artemis server. Default is `true` (meaning that name retro-compatibility is enabled).
+
+* Add a `enable-amq1-prefix`  boolean attribute to the https://wildscribe.github.io/WildFly/18.0/subsystem/messaging-activemq/external-jms-topic/[/subsystem=messaging-activemq/external-jms-topic resource] to control the enablement of Artemis 1.x prefix for topic defined on a remote Artemis server. Default is `true` (meaning that name retro-compatibility is enabled).
+
+* For a destination created by annotation `@JMSDestinationDefinition` on a remote Artemis Server, add a `enable-amq1-prefix` boolean property to control the enablement of Artemis 1.x prefix. If the property is not defined, the default is `true` (meaning that name retro-compatibility is enabled).
```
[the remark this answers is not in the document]
soul2zimate: Yes, reword to "destination (queue or topic)"

soul2zimate:
@tommaso-borgato  I have added some more information about this RFE desired outcome and test expected result. 
@ehsavoie please feel free to add anything else missing here and correct if I am wrong. Thanks,

messaging/WFLY-13793_disable_legacy_amq1_prefix.adoc
```diff
@@ -0,0 +1,126 @@
...
+
+tborgato@redhat.com
+
+=== Testing By
+// Put an x in the relevant field to indicate if testing will be done by Engineering or QE. Discuss with QE during the Kickoff state to decide this
+[X] Engineering
+
+[ ] QE
+
+* TBD
```
marekkopecky, line 41: @soul2zimate  What is this TBD section about?

messaging/WFLY-13793_disable_legacy_amq1_prefix.adoc
```diff
@@ -0,0 +1,126 @@
...
+    destinationName="myQueue2",
+    properties = {
+        "enable-amq1-prefix=false"
+    }
+)
+----
+
+=== Hard Requirements
+
+N/A
```
marekkopecky, line 73: @soul2zimate RFE without any hard requirements doesn't make sense. Please fix it (maybe move text from the beginning of `Requirements` chapter here?)

messaging/WFLY-13793_disable_legacy_amq1_prefix.adoc
tommaso-borgato: @soul2zimate typo "cnnection"

messaging/WFLY-13793_disable_legacy_amq1_prefix.adoc
```diff
@@ -0,0 +1,126 @@
...
+
+=== Non-Requirements
+
+N/A
+
+== Test Plan
+
+* Enginner WildFly integration test `ExternalJMSDestinationManagementTestCase`
```
tommaso-borgato, line 85: @soul2zimate if not already present, please also add the "negative" version of the test: it should test that when `enable-amq1-prefix` is not present (or set to true) the prefix is added

messaging/WFLY-13793_disable_legacy_amq1_prefix.adoc
```diff
@@ -0,0 +1,126 @@
...
+
+Prepare Wildfly server messaging-activemq subsystem be able to use internal broker as remote one.
+Create queue `myExternalQueue` and `myExternalTopic` with `enable-amq1-prefix` set to `false`. Both created queues name should not contain amq1 prefix `jms.queue` or `jms.topic`.
+Create queue `myExternalLegacyQueue` and `myExternalLegacyTopic` with `enable-amq1-prefix` set to `true`. Both created queue contains amq1 prefix `jms.queue` or `jms.topic`.
+
+
+
+* Enginner WildFly integration test `ExternalJMSDestinationDefinitionMessagingDeploymentTestCase`
```
tommaso-borgato, line 93: @soul2zimate if not already present, please also add the "negative" version of the test: it should test that when `enable-amq1-prefix` is not present (or set to true) the prefix is added

messaging/WFLY-13793_disable_legacy_amq1_prefix.adoc
tommaso-borgato: @soul2zimate can you add a quick explain why the attribute at the pooled connection factory level doesn't work? I mean, what is its purpose then and why it doesn't work since everything goes through the pooled connection factory ?

tommaso-borgato (changes requested):

messaging/WFLY-13793_disable_legacy_amq1_prefix.adoc
ehsavoie: this is covered https://github.com/wildfly/wildfly/pull/14412/files#diff-252482f0c3167ecf201ab5fea058fea3442d25f28154a05b8829d8e885616f87R92

messaging/WFLY-13793_disable_legacy_amq1_prefix.adoc
soul2zimate: This is covered. 

`Create queue `myExternalLegacyQueue` and `myExternalLegacyTopic` with `enable-amq1-prefix` set to `true`. Both created queue contains amq1 prefix `jms.queue` or `jms.topic`.`

messaging/WFLY-13793_disable_legacy_amq1_prefix.adoc
soul2zimate: It doesn't work because the managing API of Artemis doesn't take the prefix into account when creating a destination. The pcf prefix only have an affect on the destination which is being listened. And the new external jms queue/topic prefix is configured for the destination is being created.

messaging/WFLY-13793_disable_legacy_amq1_prefix.adoc
soul2zimate: moved it from `Requirements` to `Hard Requirements`

messaging/WFLY-13793_disable_legacy_amq1_prefix.adoc
soul2zimate: it's leftover. removed .

marekkopecky:
`Release Note Content` section is missing here, see  the details in comments in [the template](https://raw.githubusercontent.com/wildfly/wildfly-proposals/main/design-doc-template.adoc)

soul2zimate:
> `Release Note Content` section is missing here, see the details in comments in [the template](https://raw.githubusercontent.com/wildfly/wildfly-proposals/main/design-doc-template.adoc)

I have added the missing `Release Note Content` part at the end.

messaging/WFLY-13793_disable_legacy_amq1_prefix.adoc
```diff
@@ -0,0 +1,128 @@
...
+
+* mailto:chaowan@redhat.com[Chao Wang]
+* mailto:ehugonne@redhat.com[Emmanuel Hugonnet]
+
+=== QE Contacts
+
+tborgato@redhat.com
+
+=== Testing By
+// Put an x in the relevant field to indicate if testing will be done by Engineering or QE. Discuss with QE during the Kickoff state to decide this
```
tommaso-borgato, line 36: @soul2zimate please remove this comment

soul2zimate: comment removed.

messaging/WFLY-13793_disable_legacy_amq1_prefix.adoc
```diff
@@ -0,0 +1,127 @@
...
+
+=== Affected Projects or Components
+* WildFly
+
+=== Other Interested Projects
+
+== Requirements
+
+=== Hard Requirements
+
```
tommaso-borgato, line 48: @soul2zimate IUUC these new attributes have precedence over `messaging-activemq.pooled-connection-factory.remote-artemis:write-attribute(name`: if that's correct can you say it explicitly here?

tommaso-borgato (changes requested):

messaging/WFLY-13793_disable_legacy_amq1_prefix.adoc
tommaso-borgato: @ehsavoie tnks; @soul2zimate can you add this information here? this way I won't ask again on the TP ....

messaging/WFLY-13793_disable_legacy_amq1_prefix.adoc
tommaso-borgato: @soul2zimate can you add this information here? this way I won't ask again on the TP ....

messaging/WFLY-13793_disable_legacy_amq1_prefix.adoc
soul2zimate: I don't see it as a precedence. Before this RFE, the `enable-amq1-prefix` attribute in the `messaging-activemq.pooled-connection-factory` level doesn't control on the remote queue/topic creation. It always creates remote queue/topic with the `jms.queue` / `jms.topic` prefix. I think you have also illustrated that in your [helloworld-mdb example]( https://github.com/tommaso-borgato/helloworld-mdb). The new attribute added in the destination level `messaging-activemq.external-jms-queue` in this RFE give the ability to configure that.

messaging/WFLY-13793_disable_legacy_amq1_prefix.adoc
soul2zimate: I have reworded that queue `myExternalLegacyQueue` and `myExternalLegacyTopic` in the test are made for the opposite test with `enable-amq1-prefix` set to `true`

messaging/WFLY-13793_disable_legacy_amq1_prefix.adoc
soul2zimate: > this is covered https://github.com/wildfly/wildfly/pull/14412/files#diff-252482f0c3167ecf201ab5fea058fea3442d25f28154a05b8829d8e885616f87R92

@ehsavoie The link points to `ExternalJMSDestinationManagementTestCase`, I don't think we have the case when `enable-amq1-prefix` is true in `ExternalJMSDestinationDefinitionMessagingDeploymentTestCase`.

ehsavoie: Maybe you are right, I didn't create a specific deployment for this test case

ehsavoie: I've added ExternalJMSDestinationDefinitionLegacyPrefixMessagingDeploymentTestCase (https://github.com/wildfly/wildfly/pull/14412/files#diff-c1e6f0113d3eae573c1878dc135c7f63954abe9f6d8e6482c8a631fe22bb469f) to cover this

soul2zimate: Thank you, I have added this new `ExternalJMSDestinationDefinitionLegacyPrefixMessagingDeploymentTestCase` in the `Test Plan` section.

messaging/WFLY-13793_disable_legacy_amq1_prefix.adoc
tommaso-borgato: @soul2zimate yes, but there is also the **send/receive phase**: what if, for example, `messaging-activemq.pooled-connection-factory.enable-amq1-prefix=true`  and  `external-jms-queue name.enable-amq1-prefix=false` (or vice-versa)? which of the two setting has precedence?

tommaso-borgato (changes requested):

messaging/WFLY-13793_disable_legacy_amq1_prefix.adoc
```diff
@@ -0,0 +1,91 @@
...
+=== Affected Projects or Components
+* WildFly
+
+=== Other Interested Projects
+
+== Requirements
+
+* Add a `enable-amq1-prefix`  boolean attribute to the https://wildscribe.github.io/WildFly/18.0/subsystem/messaging-activemq/external-jms-queue/[/subsystem=messaging-activemq/external-jms-queue resource] to control the enablement of Artemis 1.x prefix for queue defined on a remote Artemis server. Default is `true` (meaning that name retro-compatibility is enabled).
+
```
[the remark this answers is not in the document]
tommaso-borgato: @soul2zimate tnks for adding info; please resolve discussions when you are done

messaging/WFLY-13793_disable_legacy_amq1_prefix.adoc
```diff
@@ -0,0 +1,91 @@
...
+* Add a `enable-amq1-prefix`  boolean attribute to the https://wildscribe.github.io/WildFly/18.0/subsystem/messaging-activemq/external-jms-topic/[/subsystem=messaging-activemq/external-jms-topic resource] to control the enablement of Artemis 1.x prefix for topic defined on a remote Artemis server. Default is `true` (meaning that name retro-compatibility is enabled).
+
```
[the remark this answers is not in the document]
tommaso-borgato: @soul2zimate tnks for adding info; in general please resolve discussions when you are done

messaging/WFLY-13793_disable_legacy_amq1_prefix.adoc
```diff
@@ -0,0 +1,128 @@
...
+
+* Add a `enable-amq1-prefix`  boolean attribute to the https://wildscribe.github.io/WildFly/18.0/subsystem/messaging-activemq/external-jms-topic/[/subsystem=messaging-activemq/external-jms-topic resource] to control the enablement of Artemis 1.x prefix for topic defined on a remote Artemis server. Default is `true` (meaning that name retro-compatibility is enabled). For instance a topic named `testTopic`, by setting this attribute to `false`, it effectively allows you to create an external topic without amq1 prefix `jms.topic`. The created topic will be `testTopic`. Otherwise, it will create an external topic with amq1 prefix `jms.topic`. The created topic will be `jms.topic.testTopic`.
+
+* For a destination (queue or topic) created by annotation `@JMSDestinationDefinition` on a remote Artemis Server, add a `enable-amq1-prefix` boolean property to control the enablement of Artemis 1.x prefix. If the property is not defined, the default is `true` (meaning that name retro-compatibility is enabled).
```
tommaso-borgato, line 54: @soul2zimate same question as here https://github.com/wildfly/wildfly-proposals/pull/337/files#r708021748: what happens when the two setting don't match?

tommaso-borgato (changes requested):

messaging/WFLY-13793_disable_legacy_amq1_prefix.adoc
```diff
@@ -0,0 +1,128 @@
...
+== Test Plan
+
+* WildFly integration testsuite `ExternalJMSDestinationManagementTestCase`
+
+Prepare Wildfly server messaging-activemq subsystem be able to use internal broker as remote one.
+Create queue `myExternalQueue` and `myExternalTopic` with `enable-amq1-prefix` set to `false`. Both created queues name should not contain amq1 prefix `jms.queue` or `jms.topic`.
+
+The opposite version of the test is: create queue `myExternalLegacyQueue` and `myExternalLegacyTopic` with `enable-amq1-prefix` set to `true`. Both created queue contains amq1 prefix `jms.queue` or `jms.topic`.
+
+
```
tommaso-borgato, line 87: @soul2zimate we also need a test to check what happens when both `messaging-activemq.pooled-connection-factory.enable-amq1-prefix` and `messaging-activemq.external-jms-queue.enable-amq1-prefix` are set and they have different vaues

tommaso-borgato (changes requested):

messaging/WFLY-13793_disable_legacy_amq1_prefix.adoc
```diff
@@ -0,0 +1,91 @@
...
+=== Non-Requirements
+
+N/A
+
+== Test Plan
+First connect WildFly to an remote Artemis server.
+
+Create an external JMS queue on a remote broker with the amq1 prefixe disabled from /subsystem=messaging-activemq/external-jms-queue resource, and connect to it to produce messages on the queue and receive same messages with a MDB listening that queue name without amq1 prefix.
+
```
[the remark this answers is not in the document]
tommaso-borgato: @soul2zimate tnks for the added info

messaging/WFLY-13793_disable_legacy_amq1_prefix.adoc
```diff
@@ -0,0 +1,128 @@
...
+                    name = TOPIC_LOOKUP,
+                    interfaceName = "javax.jms.Topic",
+                    destinationName = TOPIC_NAME,
+                    properties = {"enable-amq1-prefix=false"}
+            )
+        }
+)
+----
+
+It's expected to create a queue without amq1 prefix `jms.queue` or `jms.topic`, and be able to produce messages on the queue and receive same messages with MDB listening on that queue name without amq1 prefix `jms.queue` or `jms.topic`.
```
tommaso-borgato, line 116: @soul2zimate we also need a test to check what happens when both `messaging-activemq.pooled-connection-factory.enable-amq1-prefix` and `@JMSDestinationDefinition.enable-amq1-prefix` are set and they have different values

tommaso-borgato (changes requested):

messaging/WFLY-13793_disable_legacy_amq1_prefix.adoc
```diff
@@ -0,0 +1,91 @@
...
+
+N/A
+
+== Test Plan
+First connect WildFly to an remote Artemis server.
+
+Create an external JMS queue on a remote broker with the amq1 prefixe disabled from /subsystem=messaging-activemq/external-jms-queue resource, and connect to it to produce messages on the queue and receive same messages with a MDB listening that queue name without amq1 prefix.
+
+Create an external JMS queue on a remote broker with the amq1 prefixe disabled from annotation `@JMSDestinationDefinition`, and connect to it to produce messages on the queue and receive same messages with a MDB listening that queue name without amq1 prefix.
+
```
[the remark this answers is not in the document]
tommaso-borgato: @soul2zimate tnks for the added info

messaging/WFLY-13793_disable_legacy_amq1_prefix.adoc
```diff
@@ -0,0 +1,128 @@
...
+Otherwise, if `enable-amq1-prefix` is not configured to `false`. It creates queue with amq1 prefix `jms.queue` or `jms.topic`. which is covered in `ExternalJMSDestinationDefinitionLegacyPrefixMessagingDeploymentTestCase`.
+
+== Community Documentation
+
+The feature will be documented in WildFly Admin Guide (in the Messaging Configuration section).
+
+== Release Note Content
+
+WildFly now introduces a new boolean attribute `enable-amq1-prefix` for creating external jms queue / topic without amq1 prefix.
```
tommaso-borgato, line 126: @soul2zimate we should also specify what happens at send/receive time

tommaso-borgato (changes requested):

messaging/WFLY-13793_disable_legacy_amq1_prefix.adoc
ehsavoie: We will honor both: we will create a destination address with the name provided and pass it to the RA. The RA will treat the address with its own configuration (so by default it would append the prefixes). Then of course the broker itself will do some more changes according to its configuration.

ehsavoie: I think that if the 2 settings don't match we might want to warn the user that his configuration looks inconsistent.

ehsavoie: @jbertram can you confirm this ?

tommaso-borgato: IUUC, in the case that I've depicted, we are creating a queue without prefix e.g. 'myQueue' (honoring just `messaging-activemq.external-jms-queue.enable-amq1-prefix=false`) .... when later sending a message, then both `messaging-activemq.external-jms-queue.enable-amq1-prefix=false` and `messaging-activemq.pooled-connection-factory.enable-amq1-prefix=true` are involved: the first doesn't add anything while the second will add `jms.queue` thus producing a message for queueu `jms.queue.myQueue`

is that correct?

tommaso-borgato (changes requested):

messaging/WFLY-13793_disable_legacy_amq1_prefix.adoc
tommaso-borgato: @soul2zimate I guess we can close this thread and go on above

messaging/WFLY-13793_disable_legacy_amq1_prefix.adoc
tommaso-borgato: @soul2zimate this test would limit its scope to destination creation,why don't we add also sending some messages to the destination? this would make clear what is the value that we need on `messaging-activemq.pooled-connection-factory.enable-amq1-prefix` to complete the process successfully: I mean creating, sending ad receiving a message to the same queue making sure e.g. we don't create queue `myQueue` and the try to send a message to queue `jms.queue.MyQueue`

tommaso-borgato (changes requested):

messaging/WFLY-13793_disable_legacy_amq1_prefix.adoc
tommaso-borgato: @soul2zimate same as above: after creating the destination we should also check we can send/receive from that very same destination e.g. we create `MyQueue` and then we send/receive from `MyQueue` (not `jms.queue.MyQueue`)

tommaso-borgato (changes requested):

messaging/WFLY-13793_disable_legacy_amq1_prefix.adoc
soul2zimate: In @ehsavoie 's tests, [ExternalJMSDestinationManagementTestCase](https://github.com/wildfly/wildfly/pull/14412/files#diff-252482f0c3167ecf201ab5fea058fea3442d25f28154a05b8829d8e885616f87R60) tests management operations and check the created destinations.

And the " sending/receiving some messages to the destination" is covered in [ExternalJMSDestinationDefinitionMessagingDeploymentTestCase](https://github.com/wildfly/wildfly/pull/14412/files#diff-c0f01ceb72a3c4f68c2adcff2e7d1f8db8d39f174238f06f93f9031db72bed0bR75) via a proper deployment with MDB and Servlet in order to create, send ad receive messages.

messaging/WFLY-13793_disable_legacy_amq1_prefix.adoc
ehsavoie: If Artemis is configured with auto-create-queue to true, the queue is created, else the message is not delivered (but i think you would get an error as the queue doesn't exist). ActiveMQNonExistentQueueException[errorType=QUEUE_DOES_NOT_EXIST message=AMQ229017: Queue jms.queue.myQueue does not exist]

messaging/WFLY-13793_disable_legacy_amq1_prefix.adoc
```diff
@@ -0,0 +1,128 @@
...
+:author:            Chao Wang
+:email:             chaowan@redhat.com
+:toc:               left
+:icons:             font
+:idprefix:          messaging,jms
+:idseparator:       -
+
+== Overview
+Currently we can disable the use of legacy amq1 prefixes `jms.queue` and `jms.topic` only at the connection factory level but not for queue or topic defined in the model or using JakartaEE annotations.
+The goal of this proposal is to be able to disable the usage of such prefixes at the destination level for external JMS queues and topics.
```
tommaso-borgato, line 11: @soul2zimate I realized that also the final behavior is not clear to me and that we could probably add some info here; 

Let me use an example: let's say we have a queue defined in the model like :

```
<external-jms-queue name=myExternalQueue" entries="java:jboss/exported/jms/queue/myExternalQueue" enable-amq1-prefix="true"/>
```

and that the PCF is configured with the opposite setting:

```
/subsystem=messaging-activemq/pooled-connection-factory=remote-artemis:write-attribute(name="enable-amq1-prefix", value="false")
```

we said that PFC has no effect when creating the queue hence I guess the queue gets created with name `jms.queue.myExternalQueue`;

what happens when sending a message using the PFC? will it try to remove the `jms.queue.` prefix?

is the following configuration equivalent? is it even correct?: 

```
<external-jms-queue name=jms.quque.myExternalQueue" entries="java:jboss/exported/jms/queue/myExternalQueue" enable-amq1-prefix="true"/>
```

tommaso-borgato (changes requested):

messaging/WFLY-13793_disable_legacy_amq1_prefix.adoc
ehsavoie: The PCF doesn't remove prefixes, it just don't append them

soul2zimate: > is the following configuration equivalent? is it even correct?:
>
><external-jms-queue name=jms.quque.myExternalQueue" entries="java:jboss/exported/jms/queue/myExternalQueue" enable-amq1-prefix="true"/>

The remote queue will be consisted of the prefix `jms.queue` + queue name `jms.quque.myExternalQueue`, so   it will create a remote queue with name `jms.queue.jms.queue.myExternalQueue`. But I don't think this is best practice to do so (having two `jms.queue` in the front)

tommaso-borgato: I was just asking because it wasn't clear to me how this `enable-amq1-prefix` works: IIUC when set to true it always adds the prefix even if it's already present, when set to false it doesn't remove the prefix even if already preset, correct?

ehsavoie: @tommaso-borgato you are talking about different things. In WildFly we are only manipulating the address used to be passed to Artemis RA or API: so it out of the scope of this RFE how this RA works. The 'only' place where we really do something significant is at the queue creation command which is but a subset of the current RFE. This feature is mostly about beeing able to call and use a queue called **myQueue** on a remote Artemis without having to use an external jndi context.

tommaso-borgato: @ehsavoie I try and summarize the current status: feature has "accidentally" been merged without the AD being approved nor the pre-check being granted....
Now we can either: revert that change or try and complete the works here; IIRC, we stopped because it was unclear how the whole thing worked and we could not picture it out within the time we had available back then;

IIRC, we were talking about the different `points` where the queue/topic name can be manipulated (= the legacy amq1 prefix can be added/removed):
- inside the deployed application JAR/WAR/EAR: e.g. Jakarta EE annotation (can we also use deployment descriptor? other?)
- inside the model when defining the item: e.g. external-jms-queue
- inside the model when defining the how to communicate with the internal/external broker: e.g. pooled-connection-factory

is that all? could you start by refining and/or completing this list?

I would then move on with the remaining questions:
- what's the priority for the `points` mentioned above?
- where can the legacy amq1 prefix be added/removed? can it just be added?
- what difference does it make if the broker is external/internal?
- what difference does it make if the queue/topic already exists or not?
- what happens if the queue/topic doesn't exist? is it auto-created? is the prefix added?

tommaso-borgato (changes requested):

ehsavoie:
@tommaso-borgato 
this concerns only 'external" queues and topics as the internal EAP broker enforces the use of the legacy prefixes for locally defined destinations. 
We need to enable this at the model and at the deployment level.

You have different ways to access an external broker that's why we need the attribute at different level. It all depends on how you are obtaining the destination:
 - locally defined and injected: here you need it at the communication level and at the destination level
 - remote JNDI call on the broker: here you need it at the communication level
 - usage of Artemis API and destination address: here you need it at the communication level

The legacy prefix is here to ensure retro-compatibility with the destination addresses defined in HornetQ and Artemis 1.x. See [Fully Qualified Queue Names](https://activemq.apache.org/components/artemis/documentation/latest/address-model.html) to understand the impact of those prefixes.
Setting the attribute at the communication level isn't sufficient as WildFly is enforcing the use of the legacy prefixes on the destination address.
We can configure the broker to 'ignore' the prefixes automatically but this won't be applied at the creation of the destinations by WildFly. So while we would be using the JMS queue **myqueue** we would still create a *jms.queue.myqueue**.
A queue can be autocreated by the broker depending on the broker configuration, by WildFly can do it also if the broker won't allow it. We need to ensure that the address that WildFly is using matches that of the broker. The setting of the prefix is not only done by Wildfly but also by the Artemis client and resource adapter.

messaging/WFLY-13793_disable_legacy_amq1_prefix.adoc
tommaso-borgato: @ehsavoie just copying an pasting your answer here to make the thread consistent:

this concerns only 'external" queues and topics as the internal EAP broker enforces the use of the legacy prefixes for locally defined destinations.
We need to enable this at the model and at the deployment level.

You have different ways to access an external broker that's why we need the attribute at different level. It all depends on how you are obtaining the destination:

    locally defined and injected: here you need it at the communication level and at the destination level
    remote JNDI call on the broker: here you need it at the communication level
    usage of Artemis API and destination address: here you need it at the communication level

The legacy prefix is here to ensure retro-compatibility with the destination addresses defined in HornetQ and Artemis 1.x. See [Fully Qualified Queue Names](https://activemq.apache.org/components/artemis/documentation/latest/address-model.html) to understand the impact of those prefixes.
Setting the attribute at the communication level isn't sufficient as WildFly is enforcing the use of the legacy prefixes on the destination address.
We can configure the broker to 'ignore' the prefixes automatically but this won't be applied at the creation of the destinations by WildFly. So while we would be using the JMS queue myqueue we would still create a jms.queue.myqueue*.
A queue can be autocreated by the broker depending on the broker configuration, by WildFly can do it also if the broker won't allow it. We need to ensure that the address that WildFly is using matches that of the broker. The setting of the prefix is not only done by Wildfly but also by the Artemis client and resource adapter.

tommaso-borgato (changes requested):

tommaso-borgato:
@ehsavoie _the internal EAP broker enforces the use of the legacy prefixes for locally defined destinations._ why is it configured this way? IIUC, always adding the prefix means that both `jms.queue.SomeQueue` and `SomeQueue` in EAP would be translated to `jms.queue.SomeQueue` when sent to Artemis; since Artemis is forces using compatibility mode, then it would always strip the prefix; result is that `jms.queue.SomeQueue` and `SomeQueue` in EAP would always map to address `SomeQueue` in Artemis;

Is that correct?

ehsavoie:
@tommaso-borgato Artemis compatibility mode doesn't apply to Artemis management messages: when WildFly is creating the destination it forces the prefix in the destination address. This is what this is all about: changing the create message WildFly sends to have the prefix or not.
All those changes are targeting destination used for a remote broker. 
Also you are wrong here: embedded Artemis 2.x in WildFly forces legacy mode: the prefixes are always present. There is no broker.xml in embedded mode.
Documentation doesn't state that because it has always been the case in WildFly or EAP: <jms-queue name="foo"> will create a queue at address jms.queue.foo. 
Also things a re bit more complex than that as we also need to be able to have address matching on the client side and not only on the server side. We need a way to configure both the client and the remote broker. Until now this was coarse grained. The goal is to be able to define that at the destination level.

tommaso-borgato:
@ehsavoie I try and summarize here the outcome from our chat and experiments;

We start by following the procedure described in https://github.com/wildfly/quickstart/tree/main/remote-helloworld-mdb; 
Basically what's relevant here is:
- we start a remote Artemis Broker which is NOT running in compatibility mode (= it doesn't try to add prefixes for all incoming queue/topic names)
- we create a connection on WildFly to the remote Artemis Broker (called `pooled-connection-factory`) which overrides the default value of `enable-amq1-prefix` (which is true) by setting it to `enable-amq1-prefix=false`: this means also the WildFly Artemis client won't try to add prefixes to all queue/topic names
- we create an application which defines a queue named `HelloWorldMDBQueue` which overrides the default behavior by setting `properties = {"enable-amq1-prefix=false"}` (default would be true) on the queue definition:
   ```java
@JMSDestinationDefinition(
            name = "java:/queue/HELLOWORLDMDBQueue",
            interfaceName = "jakarta.jms.Queue",
            destinationName = "HelloWorldMDBQueue",
            properties = {"enable-amq1-prefix=false"} // if you remove this, a useless jms.queue.HelloWorldMDBQueue is created
        )
   ```

This results in a destination named `HelloWorldMDBQueue` containing a single anycast queue named `HelloWorldMDBQueue` being created on the remote Artemis Broker;

Everything is consistent: our java annotation defines a destination named `HelloWorldMDBQueue` and the remote Artemis Broker, contains a destination with the same name, no prefix added;

This is what happens after the feature provided by this RFE has been used: it is setting `properties = {"enable-amq1-prefix=false"}`;

Now, remove `properties = {"enable-amq1-prefix=false"}` from `@JMSDestinationDefinition` and redeploy the application: you'll notice that:
- everything continues to work: messages are routed to Artemis destination `HelloWorldMDBQueue`  
- **a useless destination named `jms.queue.HelloWorldMDBQueue` is created on the remote Artemis Broker

Summary: with the feature provided by this RFE we can get rid of that useless destination

@ehsavoie why don't we add the former example to the AD? it would help understanding what it's all about

tommaso-borgato:
@ehsavoie I also tried using queues defined in model:

I you define an external queue in WildFly model without defining `enable-amq1-prefix`:
```
/subsystem=messaging-activemq/external-jms-queue=myExternalQueue:add(entries=[java:jboss/exported/jms/queue/myExternalQueue])
```

When you later use that queue in your Java code:
```
@Resource(lookup = "java:jboss/exported/jms/queue/myExternalQueue")
    private transient Queue queue5;
```

Here is what happens:
- a useless `myExternalQueue` destination is created on the remote Artemis broker
- a destination named `jms.queue.myExternalQueue` is created on the remote Artemis broker: messages are correctly routed to this queue

Thanks to the feature provided by this RFE, we can get rid of the useless `myExternalQueue` destination  by setting `enable-amq1-prefix` either to true or false:
```
/subsystem=messaging-activemq/external-jms-queue=myExternalQueue:add(entries=[java:jboss/exported/jms/queue/myExternalQueue], enable-amq1-prefix=true)
```
in this case, just a destination named `jms.queue.myExternalQueue` is created on the remote Artemis broker
or
```
/subsystem=messaging-activemq/external-jms-queue=myExternalQueue:add(entries=[java:jboss/exported/jms/queue/myExternalQueue], enable-amq1-prefix=false)
```
in this case, just a destination named `myExternalQueue` is created on the remote Artemis broker;

@ehsavoie do you think we can add also this scenario to the AD? this way we cover both annotation and model defined destinations;

tommaso-borgato:
@ehsavoie yet another thing for some future:

## Proposal for a future RFE

Given:

    Artemis Remote Brokers don't run in compatibility mode (compatibility mode off = don't add legacy prefixes to names - generally enabled by adding "anycastPrefix=jms.queue.;multicastPrefix=jms.topic." to any acceptor url) by default both on bare metal and on OpenShift
    Artemis Broker Embedded in WildFly, by default runs in compatibility mode (compatibility mode on = add legacy prefixes to names)

We could make the behavior of the Artemis Broker Embedded in WildFly consistent with bare metal and OpenShift: the Artemis Broker Embedded in WildFly would not run in compatibility mode any more;

This might be also the occasion to adjust the behavior of the other related components accordingly in order to make the user experience more simple and consistent across different Artemis options (embedded, remote on bare metal, remote on OpenShift):

    @jakarta.jms.JMSDestinationDefinition annotation
    external-jms-queue model object
    pooled-connection-factory model object

messaging/WFLY-13793_disable_legacy_amq1_prefix.adoc
```diff
@@ -0,0 +1,145 @@
...
+The goal of this proposal is to be able to disable the usage of such prefixes at the destination level for external JMS queues and topics. Thus we would avoid creation of prefixed destinations that will be useless.
+We start by following the procedure described in https://github.com/wildfly/quickstart/tree/main/remote-helloworld-mdb[Remote MDB Quickstart]. Basically what's relevant here is:
+* we start a remote Artemis Broker which is NOT running in compatibility mode (= it doesn't try to add prefixes for all incoming queue/topic names)
+* we create a connection on WildFly to the remote Artemis Broker (called pooled-connection-factory) which overrides the default value of enable-amq1-prefix (which is true) by setting it to enable-amq1-prefix=false: this means also the WildFly Artemis client won't try to add prefixes to all queue/topic names
+* we create an application which defines a queue named HelloWorldMDBQueue which overrides the default behavior by setting *properties = {"enable-amq1-prefix=true"}* (default would be true) on the queue definition:
```
tommaso-borgato, line 15: "which overrides the default behavior" --> "which mimics the default behavior"

we actually override the default later by setting enable-amq1-prefix=false

messaging/WFLY-13793_disable_legacy_amq1_prefix.adoc
```diff
@@ -0,0 +1,145 @@
...
+[source, java]
+----
+@JMSDestinationDefinition(
+    name = "java:/queue/HELLOWORLDMDBQueue",
+    interfaceName = "jakarta.jms.Queue",
+    destinationName = "HelloWorldMDBQueue"
+)
+----
+
+This results in a destination named 'HelloWorldMDBQueue' containing a single anycast queue named 'HelloWorldMDBQueue' being created on the remote Artemis Broker plus a useless destination named jms.queue.HelloWorldMDBQueue". The goal here is that by setting the *properties = {"enable-amq1-prefix=true"}* we won't see this useless detination being created.
```
tommaso-borgato, line 26: I think here we should say "The goal here is that by setting the *properties = {"enable-amq1-prefix=false"}* we won't see this useless detination being created."

tommaso-borgato (changes requested):

tommaso-borgato:
@soul2zimate after you give a look at my latest 2 comments I think we can finally approve this AD

messaging/WFLY-13793_disable_legacy_amq1_prefix.adoc
ehsavoie: :+1:

messaging/WFLY-13793_disable_legacy_amq1_prefix.adoc
ehsavoie: :+1:

soul2zimate:
updated as per the last QE comment. Thanks you.

messaging/WFLY-13793_disable_legacy_amq1_prefix.adoc
```diff
@@ -0,0 +1,145 @@
...
+* we create a connection on WildFly to the remote Artemis Broker (called pooled-connection-factory) which overrides the default value of enable-amq1-prefix (which is true) by setting it to enable-amq1-prefix=false: this means also the WildFly Artemis client won't try to add prefixes to all queue/topic names
+* we create an application which defines a queue named HelloWorldMDBQueue which mimics the default behavior by setting *properties = {"enable-amq1-prefix=true"}* (default would be true) on the queue definition:
+
+[source, java]
+----
+@JMSDestinationDefinition(
+    name = "java:/queue/HELLOWORLDMDBQueue",
+    interfaceName = "jakarta.jms.Queue",
+    destinationName = "HelloWorldMDBQueue"
+)
```
tommaso-borgato, line 23: @soul2zimate , @ehsavoie  since we explicitly mentioned it, shouldn't we add `properties = {"enable-amq1-prefix=true"}` here even if it is the default?

soul2zimate: Yes, it's fine to explicitly add it. 
@ehsavoie May I get your blessing on this as well ?

ehsavoie: you have :anatomical_heart:

soul2zimate: Updated, Once this is approved. I will squash commits.

tommaso-borgato (approved):

PR #341, opened 2020-11-02
[WFCORE-5178] Native support for OIDC

fjuma:
This proposal is an initial draft. More updates will be added based on discussions or any issues found during the implementation.

https://issues.redhat.com/browse/WFCORE-5178

### DIFF
--- a/elytron/WFCORE-5178_native-support-for-oidc.adoc
+++ b/elytron/WFCORE-5178_native-support-for-oidc.adoc
@@ -1,0 +1,178 @@
+= WFCORE-5178 Native support for OpenID Connect
+:author:            Farah Juma
+:email:             fjuma@redhat.com
+:toc:               left
+:icons:             font
+:idprefix:
+:idseparator:       -
+
+== Overview
+
+The ability to secure deployments using https://openid.net/connect/[OpenID Connect] (OIDC)
+is currently provided by installing a Keycloak
+https://www.keycloak.org/docs/latest/securing_apps/#_jboss_adapter[client adapter].
+When Elytron was still in development for WildFly 11, the specifications in use by Keycloak
+around OIDC were still very much in a state of ongoing development. Thus, at the time, it made
+sense for Keycloak to handle this integration. Since the OIDC specifications are now stable,
+it makes sense for individual projects to handle their own OIDC integration. This RFE is to
+add native support for OIDC to Elytron. This means that the Keycloak client adapter will
+no longer be needed to secure applications deployed to WildFly using OIDC. This will also
+greatly simplify other cases where OIDC is used (e.g., bootable JAR).
+
+== Issue Metadata
+
+=== Issue
+
+* https://issues.redhat.com/browse/WFCORE-5178[WFCORE-5178]
+
+=== Related Issues
+
+* https://issues.redhat.com/browse/ELY-2034[ELY-2034]
+* https://issues.redhat.com/browse/WFLY-14017[WFLY-14017]
+* https://issues.redhat.com/browse/EAP7-1540[EAP7-1540]
+
+=== Dev Contacts
+
+* mailto:{email}[{author}]
+
+=== QE Contacts
+
+* TBD
+
+=== Testing By
+// Put an x in the relevant field to indicate if testing will be done by Engineering or QE.
+// Discuss with QE during the Kickoff state to decide this
+
+[ ] Engineering
+
+[ ] QE
+
+* TBD
+
+=== Affected Projects or Components
+
+* WildFly Elytron
+* WildFly Core
+* WildFly
+
+=== Other Interested Projects
+
+== Requirements
+
+=== Hard Requirements
+
+==== HTTP mechanism implementation
+
+A new `HttpServerAuthenticationMechanism` implementation, similar to `org.keycloak.adapters.elytron.KeycloakHttpServerAuthenticationMechanism`,
+will be added to WildFly Elytron to provide support for OIDC, as specified in the https://openid.net/specs/openid-connect-core-1_0.html[OpenID Connect Core 1.0 spec].
+
+Like the existing Keycloak client adapter, this new `HttpServerAuthenticationMechanism` implementation will only
+support the https://openid.net/specs/openid-connect-core-1_0.html#CodeFlowAuth[Authorization Code Flow] and
+Authorization Code Flow with https://tools.ietf.org/html/rfc7636[Proof Key for Code Exchange] (PKCE).
+
+==== Token storage
+
+The new HTTP mechanism implementation will be able to make use of a session token store and a cookie token store
+to store account info (i.e., access tokens, ID tokens, and refresh tokens).
+
+The session token store will be used by default but it should be possible to specify that the cookie token store
+should be used instead.
+
+The new HTTP mechanism implementation will refresh the access token when it is expired. Since the Keycloak client adapter
+provides an option to refresh the token on every request, we should provide this option as well but make
+sure to document that the use of this option is discouraged since it can have a significant impact on performance.
+For more details about what will be documented, see the description of the `always-refresh-token` option
+https://www.keycloak.org/docs/latest/securing_apps/#_refresh_token_each_req[here].
+
+==== New elytron-oidc-client subsystem
+
+A new subsystem, `elytron-oidc-client` will be added to WildFly. Its schema will be similar to the existing
+https://github.com/keycloak/keycloak/blob/master/adapters/oidc/wildfly/wildfly-subsystem/src/main/resources/schema/wildfly-keycloak_1_1.xsd[keycloak-adapter-subsystem].
+The new subsystem will install a `DeploymentUnitProcessor` which will detect if OIDC is required for a deployment.
+
+Like with the existing Keycloak client adapter, the OIDC configuration will be specified via the deployment
+or via subsystem configuration. For the deployment case, the OIDC configuration will be specified in a JSON
+configuration file in the WEB-INF directory along with the `auth-method`, `OIDC`, specified in the `web.xml`.
+For the subsystem case, the `auth-method` won't need to be specified in the `web.xml` file, it will be set
+automatically.
+
+Once a deployment has been processed, if its `auth-method` is `OIDC`, the OIDC integration will be activated.
+Otherwise, no activation will occur and the deployment will continue to use the existing security capabilities
+provided by WildFly.
+
+If the OIDC integration is activated for a deployment, then appropriate module dependencies will be added to
+the deployment (e.g., a dependency on the new Elytron OIDC module). This will ensure the activation of the
+OIDC HTTP mechanism implementation for the deployment. The deployment will make use of a "Virtual Security Domain",
+similar to what was implemented for https://github.com/wildfly/wildfly-proposals/blob/master/microprofile/WFLY-12609_MicroProfile_JWT_1_1.adoc#hard-requirements[Microprofile JWT].
+This means that it will no longer be necessary to pre-configure SecurityDomains in the Elytron subsystem,
+as was required with the Keycloak client adapter.
+
+It should be possible to make use of different OpenID providers with WildFly. While working on the
+implementation for this RFE, we should make sure that we aren't including Keycloak-specific configuration, if
+possible.
+
+==== New galleon layer for OIDC support
+
+A new Galleon layer, `oidc`, should be added to allow a server to be provisioned with OIDC support.
+
+=== Nice-to-Have Requirements
+
+* https://openid.net/specs/openid-connect-rpinitiated-1_0.html[RP-Initiated Logout] is an optional
+OIDC specification that defines how a relying party (RP) can request that an OpenID provider logs
+out the end-user. The Keycloak client adapter provides support for this so this is something we
+should provide as well but we could possibly look at that as a follow-up to this RFE if necessary.
+
+* The existing Keycloak adapter can also be used to secure the admin console. This requires some
+additional work so it might be done as a follow-up task.
+
+* *Migration*: For the most part, the same subsystem attributes and JSON configuration options that could
+be used with the `keycloak-adapter-subsystem` and the `keycloak.json` file will still work with the
+`elytron-oidc-subsystem` and the `oidc.json` file. The main difference is the actual subsystem name and
+the name of the JSON configuration file. That means that the migration tool could potentially detect an
+existing `keycloak` subsystem configuration and then change its name to the new subsystem
+name instead (and update any attributes/elements if needed). For the deployment side, the migration tool
+could detect deployments with `auth-method` set to KEYCLOAK and with a keycloak.json configuration file and
+update the `auth-method` to OIDC and rename the configuration file to oidc.json.
+
+=== Non-Requirements
+
+This integration will not be supporting the use of pre-configured SecurityDomains. The purpose of OIDC
+is to allow a client to verify the identity of an end-user based on authentication that is performed
+by an authorization server. Information about the end-user can be obtained from ID tokens. Scope
+information (i.e., role information) can be obtained from access tokens. Thus, we do not need
+pre-configured security resources in the Elytron subsystem.
+
+Support for the following optional OIDC specifications are currently non-requirements. These
+are also not provided by the Keycloak Java client adapter:
+
+* https://openid.net/specs/openid-connect-discovery-1_0.html[Discovery]
+* https://openid.net/specs/openid-connect-registration-1_0.html[Dynamic Registration]
+* https://openid.net/specs/oauth-v2-multiple-response-types-1_0.html[OAuth 2.0 Multiple Response Types]
+* https://openid.net/specs/oauth-v2-form-post-response-mode-1_0.html[OAuth 2.0 Form Post Response Mode]
+* https://openid.net/specs/openid-connect-session-1_0.html[Session Management]
+* https://openid.net/specs/openid-connect-frontchannel-1_0.html[Front-Channel Logout]
+* https://openid.net/specs/openid-connect-backchannel-1_0.html[Back-Channel Logout]
+* https://openid.net/specs/openid-connect-federation-1_0.html[OpenID Connect Federation]
+
+== Test Plan
+
+This will require some discussion with QE to determine the best approach for testing. We will
+need to decide if it is acceptable to add dependencies on Keycloak (or another OIDC provider)
+to our testsuites or if we need to mock an OIDC provider.
+
+== Community Documentation
+
+A new section will be added to the Elytron documentation that describes:
+
+ * How to activate the OIDC integration for a deployment
+ * The new `elytron-oidc-client` subsystem and how to configure it
+ * The OIDC configuration properties that can be specified via a JSON file or via subsystem configuration
+ * How to use the new Galleon `oidc-client` layer
+
+It would be good to have a quickstart for this. This can be handled under its own RFE.
+
+== Release Note Content
+
+WildFly now provides the ability to secure deployments using OpenID Connect, without needing
+to make use of the Keycloak client adapter. It is now possible to make use of other OpenID
+Connect providers in addition to Keycloak.


elytron/WFCORE-5178_native-support-for-oidc.adoc
```diff
@@ -0,0 +1,164 @@
...
+==== New elytron-oidc subsystem
+
+A new subsystem, `elytron-oidc` will be added to WildFly. Its schema will be similar to the existing
+https://github.com/keycloak/keycloak/blob/master/adapters/oidc/wildfly/wildfly-subsystem/src/main/resources/schema/wildfly-keycloak_1_1.xsd[keycloak-adapter-subsystem].
+The new subsystem will install a `DeploymentUnitProcessor` which will detect if OIDC is required for a deployment.
+
+Like with the existing Keycloak client adapter, the OIDC configuration will be specified via the deployment
+or via subsystem configuration. For the deployment case, the OIDC configuration will be specified in a JSON
+configuration file in the WEB-INF directory along with the `auth-method`, `OIDC`, specified in the `web.xml`.
+For the subsystem case, the `auth-method` won't need to be specified in the `web.xml` file, it will be set
```
OndrejKotek, line 94: Will the subsystem overrides any non-`OIDC` value in case it's specified in the `auth-method`?

fjuma: The subsystem will override a non-OIDC value if the deployment has been configured as a `secure-deployment` in the subsystem configuration (like the Keycloak subsystem did).

elytron/WFCORE-5178_native-support-for-oidc.adoc
```diff
@@ -0,0 +1,164 @@
...
+
+The new HTTP mechanism implementation will be able to make use of a session token store and a cookie token store
+to store account info (i.e., access tokens, ID tokens, and refresh tokens).
+
+The session token store will be used by default but it should be possible to specify that the cookie token store
+should be used instead.
+
+The new HTTP mechanism implementation will refresh the access token when it is expired. Since the Keycloak client adapter
+provides an option to refresh the token on every request, we should provide this option as well but make
+sure to document that the use of this option is discouraged since it can have a significant impact on performance.
```
kstekovi, line 83: Can you please more describe performance impact? It will have impact on EAP itself or on authorization server?

elytron/WFCORE-5178_native-support-for-oidc.adoc
```diff
@@ -0,0 +1,164 @@
...
+
+==== New elytron-oidc subsystem
+
+A new subsystem, `elytron-oidc` will be added to WildFly. Its schema will be similar to the existing
+https://github.com/keycloak/keycloak/blob/master/adapters/oidc/wildfly/wildfly-subsystem/src/main/resources/schema/wildfly-keycloak_1_1.xsd[keycloak-adapter-subsystem].
+The new subsystem will install a `DeploymentUnitProcessor` which will detect if OIDC is required for a deployment.
```
kstekovi, line 89: It will be part of Elytron which is in EAP it doesn't require installation like Keycloak? Just needs to be configured properly this  `elytron-oidc` subsystem. In configuration of EAP or in deployment itself.

elytron/WFCORE-5178_native-support-for-oidc.adoc
```diff
@@ -0,0 +1,164 @@
...
+to store account info (i.e., access tokens, ID tokens, and refresh tokens).
+
+The session token store will be used by default but it should be possible to specify that the cookie token store
+should be used instead.
+
+The new HTTP mechanism implementation will refresh the access token when it is expired. Since the Keycloak client adapter
+provides an option to refresh the token on every request, we should provide this option as well but make
+sure to document that the use of this option is discouraged since it can have a significant impact on performance.
+
+==== New elytron-oidc subsystem
```
kstekovi, line 85: It will be public module in Elytron? will be depend on some new private modules?

elytron/WFCORE-5178_native-support-for-oidc.adoc
```diff
@@ -0,0 +1,164 @@
...
+
+It should be possible to make use of different OpenID providers with WildFly. While working on the
+implementation for this RFE, we should make sure that we aren't including Keycloak-specific configuration, if
+possible.
+
+==== New galleon layer for OIDC support
+
+A new Galleon layer, `oidc`, should be added to allow a server to be provisioned with OIDC support.
+
+=== Nice-to-Have Requirements
```
kstekovi, line 116: The authentication and authorization is only for deployment or can be used also for server management?

elytron/WFCORE-5178_native-support-for-oidc.adoc
```diff
@@ -0,0 +1,164 @@
...
+
+==== New elytron-oidc subsystem
+
+A new subsystem, `elytron-oidc` will be added to WildFly. Its schema will be similar to the existing
+https://github.com/keycloak/keycloak/blob/master/adapters/oidc/wildfly/wildfly-subsystem/src/main/resources/schema/wildfly-keycloak_1_1.xsd[keycloak-adapter-subsystem].
+The new subsystem will install a `DeploymentUnitProcessor` which will detect if OIDC is required for a deployment.
+
+Like with the existing Keycloak client adapter, the OIDC configuration will be specified via the deployment
+or via subsystem configuration. For the deployment case, the OIDC configuration will be specified in a JSON
+configuration file in the WEB-INF directory along with the `auth-method`, `OIDC`, specified in the `web.xml`.
```
kstekovi, line 93: Is possible use server variables in configuration of OIDC?

elytron/WFCORE-5178_native-support-for-oidc.adoc
```diff
@@ -0,0 +1,164 @@
...
+A new Galleon layer, `oidc`, should be added to allow a server to be provisioned with OIDC support.
+
+=== Nice-to-Have Requirements
+
+https://openid.net/specs/openid-connect-rpinitiated-1_0.html[RP-Initiated Logout] is an optional
+OIDC specification that defines how a relying party (RP) can request that an OpenID provider logs
+out the end-user. The Keycloak client adapter provides support for this so this is something we
+should provide as well but we could possibly look at that as a follow-up to this RFE if necessary.
+
+=== Non-Requirements
```
OndrejKotek, line 123: Do we want to deal with migration paths from the Keycloak client adapter to the new subsystem?

elytron/WFCORE-5178_native-support-for-oidc.adoc
```diff
@@ -0,0 +1,164 @@
...
+The session token store will be used by default but it should be possible to specify that the cookie token store
+should be used instead.
+
+The new HTTP mechanism implementation will refresh the access token when it is expired. Since the Keycloak client adapter
+provides an option to refresh the token on every request, we should provide this option as well but make
+sure to document that the use of this option is discouraged since it can have a significant impact on performance.
+
+==== New elytron-oidc subsystem
+
+A new subsystem, `elytron-oidc` will be added to WildFly. Its schema will be similar to the existing
```
OndrejKotek, line 87: What about a bit more descriptive name for the subsystem like `elytron-oidc-client`? (The same for the Galleon layer, e.g. `oidc-client`.)

elytron/WFCORE-5178_native-support-for-oidc.adoc
fjuma: I'll update this to add a link to the [Keycloak](https://www.keycloak.org/docs/latest/securing_apps/#_refresh_token_each_req) documentation where the `always-refresh-token` option is described.

elytron/WFCORE-5178_native-support-for-oidc.adoc
fjuma: That sounds better, will update the analysis document to reflect the new names.

elytron/WFCORE-5178_native-support-for-oidc.adoc
fjuma: Installation won't be needed since the new `elytron-oidc-client` subsystem will be part of WildFly/EAP. Configuration can either be added directly in the new subsystem or it can also be included as part of the deployment.

elytron/WFCORE-5178_native-support-for-oidc.adoc
fjuma: Yes, it will be possible to configure this via the subsystem directly.

elytron/WFCORE-5178_native-support-for-oidc.adoc
fjuma: This is for securing applications that are deployed to the server. Securing the admin console will require some additional work so that might be done as part of a follow-up task.

elytron/WFCORE-5178_native-support-for-oidc.adoc
fjuma: The same subsystem attributes and JSON configuration options that could be used with the `keycloak-adapter-subsystem` and the `keycloak.json` file will still work with the `elytron-oidc-client` subsystem and the `oidc.json` file. The only difference is the actual subsystem name and the name of the JSON configuration file and auth-method. 

That means that the migration tool could potentially detect an existing `keycloak` subsystem configuration and then change its name to the new subsystem name instead. I can add this as a nice to have.

For the deployment side, do you think it would make sense for the new subsystem to support both the old JSON configuration file name and `auth-method` (i.e., keycloak.json and `<auth-method>KEYCLOAK</auth-method>` in addition to the new JSON configuration file name and `auth-method` (i.e., `oidc.json` and `<auth-method>OIDC</auth-method>`). That way, a user wouldn't need to update their deployment configuration to switch to the new approach?

OndrejKotek: I don't know how much do we want to push this feature.

In case it's just an alternative to the Keycloak client adapter and the adapter support will be kept, the migration step in the tool should be rather an optional step (to be explicitly enabled). As the migration tool enables to migrate also non-standard modules/layers, it should be also possible do disable the migration of any Keycloak client adapter related modules.

In case we want to push this feature, or we know that the WF Keycloak client adapter will be deprecated soon (e.g. because of this feature), we should provide the migration support and enable it by default.

I agree that it would be great to support Keycloak-secured deployments without a change. However supporting Keycloak configuration and `auth-method` in deployments by the new subsystem seems a bit dirty and could bring some complexity to handle cases when having both the subsystems in use (`keycloak-adapter-subsystem` and `elytron-oidc-client`). Maybe it could be possible and sufficient if the migration tool changed the the `auth-method` and the JSON configuration file name during a deployment migration.

fjuma: Thanks, @OndrejKotek.

The idea is that we won't need the Keycloak client adapter anymore, only this feature will be used.

Having the migration tool update the `auth-method` and JSON configuration file name during a deployment migration is a good idea.

I'll update the document to reflect this.

OndrejKotek: @fjuma Will we allow the two subsystems to work in parallel?

@emmartins Is it possible to adjust the migration tool to update `auth-method` in `web.xml` and rename a file during a deployment migration? I suppose we haven't performed a similar step/operations so far, but I also don't see an obstacle to have such thing.

fjuma: > @fjuma Will we allow the two subsystems to work in parallel?

Nothing would prevent them from working in parallel. But the idea will be to switch fully to the new subsystem so the Keycloak adapter won't need to be produced anymore at some point.

elytron/WFCORE-5178_native-support-for-oidc.adoc
kstekovi: will be this subsystem configured in domain mode by default? like in standalone modes?

fjuma: The subsystem will only be configured in standalone mode by default (similar to the Keycloak adapter subsystem).

elytron/WFCORE-5178_native-support-for-oidc.adoc
```diff
@@ -0,0 +1,178 @@
...
+For more details about what will be documented, see the description of the `always-refresh-token` option
+https://www.keycloak.org/docs/latest/securing_apps/#_refresh_token_each_req[here].
+
+==== New elytron-oidc-client subsystem
+
+A new subsystem, `elytron-oidc-client` will be added to WildFly. Its schema will be similar to the existing
+https://github.com/keycloak/keycloak/blob/master/adapters/oidc/wildfly/wildfly-subsystem/src/main/resources/schema/wildfly-keycloak_1_1.xsd[keycloak-adapter-subsystem].
+The new subsystem will install a `DeploymentUnitProcessor` which will detect if OIDC is required for a deployment.
+
+Like with the existing Keycloak client adapter, the OIDC configuration will be specified via the deployment
```
kstekovi, line 93: What have higher priority in configuration? JSON configuration in deployment should override XML subsystem configuration in EAP?

fjuma: Like with the Keycloak client adapter, if the deployment has been configured as a `secure-deployment` in the subsystem, then that will take precedence. If the deployment has not been configured as `secure-deployment` in the subsystem, then the deployment's configuration file will be used instead.

elytron/WFCORE-5178_native-support-for-oidc.adoc
```diff
@@ -0,0 +1,178 @@
...
+
+If the OIDC integration is activated for a deployment, then appropriate module dependencies will be added to
+the deployment (e.g., a dependency on the new Elytron OIDC module). This will ensure the activation of the
+OIDC HTTP mechanism implementation for the deployment. The deployment will make use of a "Virtual Security Domain",
+similar to what was implemented for https://github.com/wildfly/wildfly-proposals/blob/master/microprofile/WFLY-12609_MicroProfile_JWT_1_1.adoc#hard-requirements[Microprofile JWT].
+This means that it will no longer be necessary to pre-configure SecurityDomains in the Elytron subsystem,
+as was required with the Keycloak client adapter.
+
+It should be possible to make use of different OpenID providers with WildFly. While working on the
+implementation for this RFE, we should make sure that we aren't including Keycloak-specific configuration, if
```
OndrejKotek, line 111: Are there any Keycloak-specific configuration options we want to include? If it's the case, why do want those?

elytron/WFCORE-5178_native-support-for-oidc.adoc
```diff
@@ -0,0 +1,178 @@
...
+should provide as well but we could possibly look at that as a follow-up to this RFE if necessary.
+
+* The existing Keycloak adapter can also be used to secure the admin console. This requires some
+additional work so it might be done as a follow-up task.
+
+* *Migration*: For the most part, the same subsystem attributes and JSON configuration options that could
+be used with the `keycloak-adapter-subsystem` and the `keycloak.json` file will still work with the
+`elytron-oidc-subsystem` and the `oidc.json` file. The main difference is the actual subsystem name and
+the name of the JSON configuration file. That means that the migration tool could potentially detect an
+existing `keycloak` subsystem configuration and then change its name to the new subsystem
```
OndrejKotek, line 132: What about attributes like `client-key-password`, `client-keystore`, `client-keystore-password`, `realm-public-key`, `truststore`, `truststore-password`? We should transform those to leverage (Elytron) capabilities, like client `ssl-context` and/or `credential-store`.

elytron/WFCORE-5178_native-support-for-oidc.adoc
```diff
@@ -0,0 +1,178 @@
...
+== Community Documentation
+
+A new section will be added to the Elytron documentation that describes:
+
+ * How to activate the OIDC integration for a deployment
+ * The new `elytron-oidc-client` subsystem and how to configure it
+ * The OIDC configuration properties that can be specified via a JSON file or via subsystem configuration
+ * How to use the new Galleon `oidc-client` layer
+
+It would be good to have a quickstart for this. This can be handled under its own RFE.
```
OndrejKotek, line 172: @fjuma Can you create and add the new qs RFE please?

elytron/WFCORE-5178_native-support-for-oidc.adoc
fjuma: There are some options (like "realm" and "auth-server-base-url" and some clustering-related options like "register-node-at-startup") that only make sense with the Keycloak provider. I've kept these so that it's still possible to configure things with Keycloak easily.

The comment in the document here about avoiding Keycloak-specific things was mostly just to ensure that other OpenID providers could be used in addition to Keycloak.

elytron/WFCORE-5178_native-support-for-oidc.adoc
OndrejKotek: In the last build, I see the subsystem is not configured in `standalone-microprofile*.xml`. Is it intentional?

Will be the subsystem available in OpenShift configuration by default, or will there be a property to enable it?

elytron/WFCORE-5178_native-support-for-oidc.adoc
OndrejKotek: To avoid confusion, what about adding a prefix for such options ("keycloak-"), or move those to a separate vendor-specific configuration (even a generic map?).

When I work with Keycloak client, it's hard for me to understand which options are relevant to me. There are a lot of them, badly documented in some cases.

elytron/WFCORE-5178_native-support-for-oidc.adoc
fjuma: That's not intentional :) I've updated my branch so you should see the subsystem in the `standalone-microprofile*.xml` files now.

Since the subsystem is in the default configs, it should be available in the OpenShift config by default too.

elytron/WFCORE-5178_native-support-for-oidc.adoc
fjuma: Right now, it's possible to specify these attributes in either the oidc.json file or in the subsystem configuration.

The advantage of the first approach is that all of the configuration is specified in the deployment itself without requiring any server configuration.

I could look into the possibility of updating the `elytron-oidc-client` subsystem configuration to allow references to an Elytron `ssl-context` / `credential-store` to be specified. For the deployment case, perhaps the SSL config could be specified via a wildfly-config.xml file in the deployment.

I'll need to look into this further. This would likely need to be a follow-up task though since the original plan was to make use of the same options the Keycloak adapter provides.

elytron/WFCORE-5178_native-support-for-oidc.adoc
fjuma: An advantage of using the same option names as the existing Keycloak adapter is that the Keycloak web console can be used to generate the JSON content or the subsystem config for the user so they can easily update their deployment or subsystem config. If we were to add a prefix to these option names or move them into a separate map, then it wouldn't be possible to use the Keycloak web console to generate the config for user (unless that is updated as well).

Instead of adding a prefix to these option names or moving them into a separate map, what if I update the description of these options to emphasize that these are Keycloak-specific options?

elytron/WFCORE-5178_native-support-for-oidc.adoc
fjuma: Sure, will just discuss this one on our team call first tomorrow and then will create it.

elytron/WFCORE-5178_native-support-for-oidc.adoc
OndrejKotek: In case we want to keep such functionality/compatibility, documentation and good descriptions in management model / XSD are the only way. Improvement of descriptions would help a lot for sure.

elytron/WFCORE-5178_native-support-for-oidc.adoc
OndrejKotek: Maybe the questions was whether it will be possible to use some kind of expressions even in `oidc.json`.

fjuma: Ah, that makes sense. Yes, it should be possible to use expressions in `oidc.json`.

elytron/WFCORE-5178_native-support-for-oidc.adoc
OndrejKotek: > The idea is that we won't need the Keycloak client adapter anymore, only this feature will be used.

Are Keycloak developers aware of this idea, so that they could drop the support on their side? A KEYCLOAK Jira issue?

fjuma: They are aware that we are working on this. We'll need to first deprecate the existing adapter and then remove later on.

emmartins: @OndrejKotek @fjuma The migration tool does not touches deployments, it just provides the feature to copy deployments to the new server. Any app/deployment migration issue should be handled by Migration Toolkit for Applications (MTA) and our Migration Guide.

emmartins: Wrt the subsystem (name) migration, agree that if supported by the tool such migration task it should be off by default.

elytron/WFCORE-5178_native-support-for-oidc.adoc
```diff
@@ -0,0 +1,178 @@
...
+This integration will not be supporting the use of pre-configured SecurityDomains. The purpose of OIDC
+is to allow a client to verify the identity of an end-user based on authentication that is performed
+by an authorization server. Information about the end-user can be obtained from ID tokens. Scope
+information (i.e., role information) can be obtained from access tokens. Thus, we do not need
+pre-configured security resources in the Elytron subsystem.
+
+Support for the following optional OIDC specifications are currently non-requirements. These
+are also not provided by the Keycloak Java client adapter:
+
+* https://openid.net/specs/openid-connect-discovery-1_0.html[Discovery]
```
OndrejKotek, line 148: Is this (partially?) provided and leveraged when using `provider-url`?

elytron/WFCORE-5178_native-support-for-oidc.adoc
```diff
@@ -0,0 +1,178 @@
...
+implementation for this RFE, we should make sure that we aren't including Keycloak-specific configuration, if
+possible.
+
+==== New galleon layer for OIDC support
+
+A new Galleon layer, `oidc`, should be added to allow a server to be provisioned with OIDC support.
+
+=== Nice-to-Have Requirements
+
+* https://openid.net/specs/openid-connect-rpinitiated-1_0.html[RP-Initiated Logout] is an optional
```
OndrejKotek, line 120: Is this included? If not, do we want a follow-up RFE?

elytron/WFCORE-5178_native-support-for-oidc.adoc
```diff
@@ -0,0 +1,178 @@
...
+OIDC specification that defines how a relying party (RP) can request that an OpenID provider logs
+out the end-user. The Keycloak client adapter provides support for this so this is something we
+should provide as well but we could possibly look at that as a follow-up to this RFE if necessary.
+
+* The existing Keycloak adapter can also be used to secure the admin console. This requires some
```
OndrejKotek, line 125: Is this included? If not, do we want a follow-up RFE?

elytron/WFCORE-5178_native-support-for-oidc.adoc
fjuma: The discovery specification is actually something that an OpenID provider implements. I'll remove it from the list here since our implementation is concerned with the client side (i.e., relying party config). 

In general though, the OpenID Connect spec mentions that metadata about an OpenID provider is usually obtained via discovery. The `provider-url` is a URL for an OpenID provider that can be used to discover information about it via the `provider-url/.well-known/openid-configuration` url.

elytron/WFCORE-5178_native-support-for-oidc.adoc
fjuma: This will be a follow-up RFE. I'm going to create an RFE this week to track the follow-up items (i.e., RP-initiated logout, the ability to secure the admin console, etc.).

elytron/WFCORE-5178_native-support-for-oidc.adoc
fjuma: Yes, this will be done as a follow-up task, as mentioned in the above comment.

elytron/WFCORE-5178_native-support-for-oidc.adoc
fjuma: Created https://issues.redhat.com/browse/EAP7-1796 / https://issues.redhat.com/browse/WFLY-15260

darranl:
I just added the hold label as this one needs a backport, I think better to keep open just in case backporting requires some additions.

bstansberry:
I'm going to merge this as I link to these in the WildFly release announcements.

PR #304, opened 2020-03-16
[WFCORE-4135] Support for read only server configuration directory

yersan:
A proposal to use read only server configuration dir

Jira issue: https://issues.redhat.com/browse/WFCORE-4135

PR #342, opened 2020-11-05
[WFCORE-4135] Support for read only server configuration directory

yersan:
This supersedes https://github.com/wildfly/wildfly-proposals/pull/304
A proposal to use read only server configuration dir

Jira issue: https://issues.redhat.com/browse/WFCORE-4135

Hello @OndrejKotek , as discussed I dropped from the analysis the notes about the required changes for cloud images. If you are fine with the current content and has no further questions, would you mind approving it? thanks

### DIFF
--- a/ee/WFCORE-4135_Support_for_ read_only_server_config_dir.adoc
+++ b/ee/WFCORE-4135_Support_for_ read_only_server_config_dir.adoc
@@ -1,0 +1,93 @@
+= Support for read only server configuration directory
+:author:            Yeray Borges
+:email:             yborgess@redhat.com
+:toc:               left
+:icons:             font
+:idprefix:
+:idseparator:       -
+
+== Overview
+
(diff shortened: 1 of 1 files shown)


OndrejKotek:
@yersan Can you also add `Community Documentation` and `Release Note Content` sections please?

This seems to deserve a brief documentation. If it's the case than also WildFly project is affected.

yersan:
Thanks for the review @OndrejKotek , I added the mentioned sections. I don't believe we will need a follow up Jira on WildFly to add any documentation since this RFE is more as an enhancement on the uses of `--read-only-server-config` command line parameter. It is assumed if you have a read-only configuration directory, you would need to add `--read-only-server-config` to your server launch because you have a read-only configuration file as well.

OndrejKotek:
@yersan Thank you for the update, I have no other suggestions. Approved by QE.

PR #343, opened 2020-11-10
[CLOUD-3843] Support bare metal Galleon layers in S2i workflow

yersan:
Jira issue: https://issues.redhat.com/browse/CLOUD-3843

### DIFF
--- a/openshift/CLOUD-3843_Support_bare_metal_Galleon_layers_in_S2i.adoc
+++ b/openshift/CLOUD-3843_Support_bare_metal_Galleon_layers_in_S2i.adoc
@@ -1,0 +1,85 @@
+= CLOUD-3843: Support bare metal Galleon layers in S2i workflow.
+:author:            Yeray Borges
+:email:             yborgess@redhat.com
+:toc:               left
+:icons:             font
+:idprefix:
+:idseparator:       -
+
+== Overview
+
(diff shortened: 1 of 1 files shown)


openshift/CLOUD-3843_Support_bare_metal_Galleon_layers_in_S2i.adoc
```diff
@@ -0,0 +1,85 @@
...
+
+== Test Plan
+
+* Behave tests added to repository https://github.com/jboss-container-images/jboss-eap-7-openshift-image. For each Galleon layer we want to convert into a decorator, a behave test will verify that the decorator layer can be combined with a base layer and the server starts without errors.
+* OpenShift integration testing.
+
+== Community Documentation
+
+* There is no need to document the Galleon layers since they are already documented at [WildFly Galleon layers] https://docs.wildfly.org/21/Galleon_Guide.html#wildfly_galleon_layers.
+* This RFE will ensure that the https://github.com/wildfly/wildfly-s2i/[Wildfly S2I] community documentation contains an updated reference to the available Galleon layers on its README.md file.
```
bstansberry, line 81: I like the way that README.md already is. It references the list of layers in the main WildFly docs, and then notes some others that are not part of main WildFly.  If we knew that some particular layers were a problem in s2i then we could note that in the README but I otherwise wouldn't want to start maintaining a separate list of layers.

PR #344, opened 2020-11-11
[WFLY-14062] CVE-2020-25633 fix

ronsigal:


jamezp:
I'm not really sure we need a proposal for this. It looks like maybe a change in behavior, but for bug purposes.

ronsigal:
Hey @jamezp,

I wasn't sure so I went ahead and did it. I can confirm that it's a behavior change to fix a security problem. But if I can avoid the feature process, so much the better. 

I sent @bstansberry email last week, but he was on PTO, so it fell through the cracks. I'm going to tag @asoldano here also, since I asked him to create an EAP7 Feature Request JIRA for me.

Thanks, @jamezp.

bstansberry:
@ronsigal This is not a feature so this isn't needed.

ronsigal:
Thanks, @bstansberry. I really wasn't sure, so I wanted to play it safe. Note that we've added a RESTEasy configuration parameter to restore the original behavior, and, presumably, will need to update the jaxrs subsystem accordingly.

ronsigal:
Hey @bstansberry, I've just read your recent email. So, given the new RESTEasy configuration parameter, my understanding is that we should go back to considering this change a feature, right?

ronsigal:
The EAP7 issue has been changed to https://issues.redhat.com/browse/JBEAP-20563.

1. Can I close this analysis document?
2. More generally, is there documentation on the process for JBEAP issues?

PR #345, opened 2020-11-12
[WFCORE-5170] Certificate Revocation Lists Proposal

SoniaZaldana:
https://issues.redhat.com/browse/WFCORE-5170

### DIFF
--- a/elytron/WFCORE-5170-multiple-certificate-revocation-lists.adoc
+++ b/elytron/WFCORE-5170-multiple-certificate-revocation-lists.adoc
@@ -1,0 +1,187 @@
+= [WFCORE-5170] Certificate Revocation Lists
+:author:            Sonia Zaldana Calles
+:email:             szaldana@redhat.com
+:toc:               left
+:icons:             font
+:idprefix:
+:idseparator:       -
+
+== Overview
+
+Certificate Revocation Lists (CRLs) contain a list of certificates that have been revoked
+by the issuing Certificate Authority before their expected expiration date, and therefore should no longer
+be trusted. Elytron currently supports configuring one Certificate Revocation List.
+
+However, if several Certificate Authorities are used, there is no way to configure several
+CRL files. This task is to enable support for more than one Certificate Revocation List in Elytron.
+
+== Issue Metadata
+
+=== Issue
+
+* https://issues.redhat.com/browse/WFCORE-5170[WFCORE-5170]
+
+=== Related Issues
+
+* https://issues.redhat.com/browse/WFLY-14007[WFLY-14007]
+* https://issues.redhat.com/browse/ELY-2033[ELY-2033]
+* https://issues.jboss.org/browse/EAP7-1548[EAP7-1548]
+
+
+=== Dev Contacts
+
+* mailto:{email}[{author}]
+
+=== QE Contacts
+
+N/A
+
+=== Testing By
+// Put an x in the relevant field to indicate if testing will be done by Engineering or QE.
+// Discuss with QE during the Kickoff state to decide this
+* [ X ] Engineering
+
+* [ ] QE
+
+=== Affected Projects or Components
+
+* WildFly, WildFly-Core and Elytron
+
+=== Other Interested Projects
+
+N/A
+
+== Requirements
+
+=== Hard Requirements
+
+* Add a new attribute ``certificate-revocation-lists`` in the ``trust-manager`` resource in the Elytron subsystem:
+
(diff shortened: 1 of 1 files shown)


elytron/WFCORE-5170-multiple-certificate-revocation-lists.adoc
```diff
@@ -0,0 +1,123 @@
...
+
+N/A
+
+== Requirements
+
+=== Hard Requirements
+
+* Add a new attribute ``certificate-revocation-lists`` in the ``trust-manager`` resource in
+the Elytron subsystem. It should accept a list of certificate revocation list objects
+as follows:
```
fjuma, line 62: Would be good to mention the optional attributes for the certificate revocation list objects (`path` and `relative-to`).

elytron/WFCORE-5170-multiple-certificate-revocation-lists.adoc
```diff
@@ -0,0 +1,123 @@
...
+
```
fjuma, line 63: Would be good to mention that as before, it should be possible to specify a certificate revocation list object without specifying the `path` or `relative-to` attribute (for the case where CRLs are obtained from distribution points referenced in certificates).

fjuma (changes requested):

elytron/WFCORE-5170-multiple-certificate-revocation-lists.adoc
SoniaZaldana: Added

elytron/WFCORE-5170-multiple-certificate-revocation-lists.adoc
SoniaZaldana: I have now added the case where CRLs are obtained from the distribution points.

elytron/WFCORE-5170-multiple-certificate-revocation-lists.adoc
```diff
@@ -0,0 +1,127 @@
...
+** Alternatively, the ``trust-manager`` must be able to configure CRLs using certificate revocation
+list objects (which override the CRLs referenced in the distribution points). These objects have the
+attributes ``path`` and ``relative-to`` which specify the path to the CRL and the
+base path to the CRL respectively. These objects can be configured follows:
+
+    /subsystem=elytron/trust-manager=myTrustManager:write-attribute(name=certificate-revocation-lists, value=[{path=/path/to/CRL_FILE.crl.pem, relative-to=/SOME/DIRECTORY},
+    {path=/path/to/CRL_FILE2.crl.pem, relative-to=/SOME/DIRECTORY}])
+
+* The attribute ``certificate-revocation-list`` in the ``trust-manager`` resource
```
mchoma, line 75: I wonder can by `certificate-revocation-list` transformed into `certificate-revocation-lists` automatically by transformers? Is it good idea? (This is just question. I don't know what is best practice in subsystem development)

SoniaZaldana: Hi Martin, subsystem transformers usually transform new attributes to older ones for backwards compatibility and not the other way around. In previous similar cases, the approach taken was to simply deprecate the existing attribute. Please let me know if you have any other concerns.

darranl: I think we will likely be better keeping the two attributes and making them mutually exclusive - the problem with the attributes that can take multiple values is the syntax in the CLI is more complex so I think retaining both makes sense.

SoniaZaldana: Sounds good. I'll update the proposal.

fjuma: +1, making them mutually exclusive makes sense.

fjuma (approved):

elytron/WFCORE-5170-multiple-certificate-revocation-lists.adoc
```diff
@@ -0,0 +1,189 @@
...
+* mailto:{email}[{author}]
+
+=== QE Contacts
+
+TBD
+
+=== Testing By
+// Put an x in the relevant field to indicate if testing will be done by Engineering or QE.
+// Discuss with QE during the Kickoff state to decide this
+* [ ] Engineering
```
fjuma, line 42: @SoniaZaldana Since this was tested by Engineering, this section can now be updated. Thanks.

PR #346, opened 2020-11-16
[WFLY-14074] Proposal for Normalization of principal propagation/injection across elytron and legacy

Skyllarr:
https://issues.redhat.com/browse/WFLY-14074

### DIFF
--- a/elytron/WFLY-14074-normalization-of-principal-propagation.adoc
+++ b/elytron/WFLY-14074-normalization-of-principal-propagation.adoc
@@ -1,0 +1,111 @@
+= WFLY-14074 Normalize principal propagation/injection across elytron and legacy
+:author:            Diana Krepinska Vilkolakova
+:email:             dvilkola@redhat.com
+:toc:               left
+:icons:             font
+:idprefix:
+:idseparator:       -
+
+== Overview
+
+This issue seeks to provide configuration option for switching between functionality introduced with WFLY-12537 and functionality that was present before.
+
+=== Background
+
+`Secured EJB` refers to EJB that has any security metadata configured at the EJB level. `Unsecured EJB` is EJB without such security metadata.
+
+Incoming run-as identity refers to the security identity that was active for the context before the run-as* annotation was applied. For further information please see https://docs.wildfly.org/19/Developer_Guide.html#run-as[RunAsPrincipal annotation documentation].
+
+If Elytron was used before the issue WFLY-12537 was resolved, then the security identity of an unsecured EJB was always anonymous if there was no incoming run-as identity.
+
+This behaviour was inconsistent with legacy behaviour. The key difference is that Elytron did not differentiate between remote and local EJB. The identity is anonymous in legacy only if mentioned unsecured EJB was called from remote context. For local EJB the identity is the current security identity obtained from security domain.
+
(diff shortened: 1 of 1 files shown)


elytron/WFLY-14074-normalization-of-principal-propagation.adoc
```diff
@@ -0,0 +1,107 @@
...
+
+N/A
+
+=== Non-Requirements
+
+N/A
+
+== Implementation Plan
+
+The configuration option for the EJB deployments could be added to `application-security-domain` component in the `ejb3` subsystem. The parameter could be called `legacy-compliance` or more specifically `legacy-compliance-principal-injection`.
```
fjuma, line 85: Since a user would likely want to use the same behaviour (i.e., either the pre-WFLY-12537 behaviour or the post-WFLY-12537 behaviour) for all cases, instead of needing to configure this for each `application-security-domain` resource, it might be better to either introduce a new element in the EJB subsystem for this or perhaps introduce a system property to configure this.

chengfang: There is a container element `application-security-domains`. Maybe we can have an attribute under it to apply to all application-security-domain resources.

Skyllarr: @chengfang I will try to go that route and add it as an attribute to `application-security-domains` element. Thank you! And what do you think should be the default value for it?

chengfang: @Skyllarr I prefer the 2nd option (Have previous (legacy) behaviour by default and provide a config option that allows people to switch to the new behaviour), and clearly document it in both community doc and product doc.

Skyllarr: @chengfang  Thank you, yes it does make sense to use previous legacy configuration as default. I will specify this in the proposal. 

I had a conversation with @darranl about a plan to place this new configuration option `legacy-compliance-principal-propagation` as an attribute to `application-security-domains` container element. He suggested to rather place it in the child `application-security-domain` element instead, as this is kind of policy that could be applied to individual deployment. But mostly that the `application-security-domain` resources that are present in the `jb3` and `undertow` subsystem will probably be removed in the future. 

Given there are currently no configuration options in the parent `application-security-domains` resource, it might be better to not add any there before the removal. The child `application-security-domain` resource already has other configurations like `enable-jacc` and `security-domain`. So if we add `legacy-compliance-principal-propagation` option there as well, it could be handled together with those during the removal.

chengfang:
@Skyllarr LGTM.

elytron/WFLY-14074-normalization-of-principal-propagation.adoc
OndrejKotek: @Skyllarr Can you update the implementation plan with the conclusions, and the release note content then please?

Skyllarr: @OndrejKotek Updated. Thanks

OndrejKotek:
@Skyllarr Thank you for the update. The analysis is clear and straightforward, I have no questions or suggestions. Approved by QE.

PR #349, opened 2020-11-18
[WFLY-14087]: Document how to configure a datasource to take advantag…

ehsavoie:
…e of new Artemis pool support.

Jira: https://issues.redhat.com/browse/WFLY-14087

### DIFF
--- a/messaging/WFLY-14087_documentation_connection_pool_configuration.adoc
+++ b/messaging/WFLY-14087_documentation_connection_pool_configuration.adoc
@@ -1,0 +1,71 @@
+= Document how to configure a datasource to take advantage of new Artemis pool support
+:author:            Emmanuel Hugonnet
+:email:             ehugonne@redhat.com
+:toc:               left
+:icons:             font
+:idprefix:          jms,messaging
+:idseparator:       -
+
+== Overview
+
+Since Apache Artemis 2.16 the JDBC Jjournal has been updated to properly use a connection poool. This proposal is to take that change into account and look into the consequences form WildFly integration point of view.
+
+== Issue Metadata
+
+=== Issue
+
+* https://issues.jboss.org/browse/WFLY-14087[WFLY-14087]
+
+=== Related Issues
+
+* https://issues.jboss.org/browse/WFLY-13226[WFLY-13226]
+* https://issues.apache.org/jira/browse/ARTEMIS-2823[ARTEMIS-2823]
+
+=== Dev Contacts
+
+* mailto:{email}[{author}]
+
+=== QE Contacts
+
+* mailto:mkopecky@redhat.com[Marek Kopecky]
+
+=== Testing By
+// Put an x in the relevant field to indicate if testing will be done by Engineering or QE. 
+// Discuss with QE during the Kickoff state to decide this
+* [ ] Engineering
+
+* [X] QE
+
+=== Affected Projects or Components
+
+* WildFly
+* Apache Artemis
+
+=== Other Interested Projects
+
+== Requirements
+
+With Apache Artemis 2.16, proper support for a JDBC pool was introduced. While apache Artemis uses https://commons.apache.org/proper/commons-dbcp/configuration.html[Apache Commons DBCP], it allows a better support for other pools too.
+This change requires to provide proper configuration of the Ironjacamar Datasource for good perfomance. Thus we need to document this in our community doc.
+Take into account https://github.com/apache/activemq-artemis/pull/3204#issuecomment-696183450[this comment on the PR].
+
+=== Hard Requirements
+
+WildFly was already providing a connection pool to Artemis but now we need to ensure that it is still working properly with our test suite, eventually reconfiguring the pool to match the tests behaviour.
+We need to check that this new behaviour from Artemis doesn't introduce regression.
+We also need to provide an up to date documentation to help our users with the configuration of this pool.
+
+=== Nice-to-Have Requirements
+
+N/A
+
+=== Non-Requirements
+
+N/A
+
+== Test Plan
+
+Reuse the existing test suite
+
+== Community Documentation
+The feature will be documented in WildFly Admin Guide (in the Messaging Configuration section).


messaging/WFLY-14087_documentation_connection_pool_configuration.adoc
```diff
@@ -0,0 +1,62 @@
...
+
+* WildFly
+* Apache Artemis
+
+=== Other Interested Projects
+
+== Requirements
+
+With Apache Artemis 2.16, proper support for a JDBC pool was introduced. While apache Artemis uses https://commons.apache.org/proper/commons-dbcp/configuration.html[Apache Commons DBCP], it allows a better support for other pools too.
+This change requires to provide proper configuration of the Ironjacamar Datasource for good perfomance. Thus we need to document this in our community doc.
```
marekkopecky, line 45: > This change requires to provide proper configuration of the Ironjacamar Datasource for good perfomance.

Can you please elaborate the details about this configuration?

messaging/WFLY-14087_documentation_connection_pool_configuration.adoc
```diff
@@ -0,0 +1,62 @@
...
+
+=== Hard Requirements
+
+N/A
```
marekkopecky, line 49: Can you please elaborate new changes that this RFE adds? Maybe new management operations? New management model resources?

messaging/WFLY-14087_documentation_connection_pool_configuration.adoc
```diff
@@ -0,0 +1,62 @@
...
+
+* https://issues.jboss.org/browse/WFLY-13226[WFLY-13226]
+* https://issues.apache.org/jira/browse/ARTEMIS-2823[ARTEMIS-2823]
+
+=== Dev Contacts
+
+* mailto:{email}[{author}]
+
+=== QE Contacts
+
```
marekkopecky, line 27: It's me: mkopecky@redhat.com

messaging/WFLY-14087_documentation_connection_pool_configuration.adoc
```diff
@@ -0,0 +1,62 @@
...
+=== Nice-to-Have Requirements
+
+N/A
+
+=== Non-Requirements
+
+N/A
+
+== Test Plan
+
```
marekkopecky, line 60: What is necessary to test from the dev's point of view? Full test plan will be prepared separately by QE, but this section in the analysis should contains key points that was find out during analysis preparation.

messaging/WFLY-14087_documentation_connection_pool_configuration.adoc
marekkopecky: This is still not fixed

messaging/WFLY-14087_documentation_connection_pool_configuration.adoc
marekkopecky: Hard requirements section is still empty

messaging/WFLY-14087_documentation_connection_pool_configuration.adoc
```diff
@@ -0,0 +1,65 @@
+= Document how to configure a datasource to take advantage of new Artemis pool support
+:author:            Emmanuel Hugonnet
+:email:             ehugonne@redhat.com
+:toc:               left
+:icons:             font
+:idprefix:          jms,messaging
+:idseparator:       -
+
+== Overview
+
```
marekkopecky, line 10: Overview is missing

messaging/WFLY-14087_documentation_connection_pool_configuration.adoc
marekkopecky: TP section in analysis is still empty

messaging/WFLY-14087_documentation_connection_pool_configuration.adoc
ehsavoie: No because it depends on the usage. Since the pool is going to be used for paging, binding, large messages and with the HA (aka lock lease, node manager state, etc.)

marekkopecky: This usage needs to be elaborated in this document.

Originally, Artemis took one DB connection from the EAP pool of connections (from datasource subsystem) during the startup. So there is extra-high probability, that there is at least one connection available.
EAP 7.3 is using artemis 2.10, but it was upgraded to Artemis 2.16. And Artemis 2.16 is using the connection pool differently. Artemis doesn't take one connection during the startup, but take a connection once it is needed. But it seems to be possible now that the pool don't need to contains any available connection if throughput is high. If the operation needing a connection now hangs, then to me this seems like a regression. Anyway, this case needs to be clearly described in the analysis.

There is some concern on artemis side ( https://github.com/apache/activemq-artemis/pull/3204#issuecomment-696183450 ) with the suggestion that the pool should be configured properly. This analysis needs to define, what "configured properly" means (set max-pool-size high enough?). I don't think adding critical analyzer for this helps (although I'm not so familiar with it yet) because the default value in EAP seems to be LOG which means that a log message will be printed saying that broker is not responsive.

marekkopecky: This usage needs to be elaborated in this document.

Originally, Artemis took one DB connection from the EAP pool of connections (from datasource subsystem) during the startup. So there is extra-high probability, that there is at least one connection available.
EAP 7.3 is using artemis 2.10, but it was upgraded to Artemis 2.16. And Artemis 2.16 is using the connection pool differently. Artemis doesn't take one connection during the startup, but take a connection once it is needed. But it seems to be possible now that the pool don't need to contains any available connection if throughput is high. If the operation needing a connection now hangs, then to me this seems like a regression. Anyway, this case needs to be clearly described in the analysis.

There is some concern on artemis side ( https://github.com/apache/activemq-artemis/pull/3204#issuecomment-696183450 ) with the suggestion that the pool should be configured properly. This analysis needs to define, what "configured properly" means (set max-pool-size high enough?). I don't think adding critical analyzer for this helps (although I'm not so familiar with it yet) because the default value in EAP seems to be LOG which means that a log message will be printed saying that broker is not responsive.

ehsavoie: Well, if no connection is available from the pool I would think that this means that before you would be hanging also ince the connection would already be in used.
The critical analyzer isn't a solution, in the end it wasn't added. Also please note that you can control the critical anlyzer in EAP via https://github.com/wildfly/wildfly/pull/13745

marekkopecky (approved):
I'm approving this PR

PR #357, opened 2021-01-05
[WFLY-14261] MP Health 3.0 proposal

xstefank:
https://issues.redhat.com/browse/WFLY-14261

### DIFF
--- a/microprofile/WFLY-14261_upgrade-microprofile-health-3.0.adoc
+++ b/microprofile/WFLY-14261_upgrade-microprofile-health-3.0.adoc
@@ -1,0 +1,79 @@
+= Upgrade MicroProfile Health to 3.0
+:author:            Martin Stefanko
+:email:             mstefank@redhat.com
+:toc:               left
+:icons:             font
+:keywords:          microprofile,health,observability
+:idprefix:
+:idseparator:       -
+
+== Overview
+
+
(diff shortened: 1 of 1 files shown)


microprofile/WFLY-14261_upgrade-microprofile-health-3.0.adoc
```diff
@@ -0,0 +1,79 @@
...
+[ ] QE
+
+== Requirements
+
+* Upgrade MicroProfile Health to 3.0
+* Upgrade SmallRye Health to 3.0.0
+* Pass all MicroProfile Health tests in WildFly integration test suite
+* Pass the MicroProfile Health 3.0 TCK without failures
+* Adjust the tests to work as needed
+* Adjust QuickStarts to address backward compatibility issues (tracked in https://issues.redhat.com/browse/WFLY-14398[WFLY-14398])
```
mchoma, line 47: If quickstart change is part of this RFE, we would need change of quickstart test as well?

xstefank: only change required in QS was this - https://github.com/wildfly/quickstart/pull/470 so use standalone-microprofile.xml for arquillian tests. Do you think that precheck or QS testsuite is necessary in this regard?

PR #364, opened 2021-01-20
Add a section to the analysis document template to help clarify what …

bstansberry:
…types of WildFly installations are relevant for the feature.

The template by default indicates that all are relevant.


bstansberry:
@kabir fyi

PR #369, opened 2021-01-29
[WFCORE-5261] Add a common script configuration to be shared across t…

jamezp:
…he various scripts.

https://issues.redhat.com/browse/WFCORE-5261
https://issues.redhat.com/browse/WFLY-14399

### DIFF
--- a/scripts/WFCORE-5261-common-conf.adoc
+++ b/scripts/WFCORE-5261-common-conf.adoc
@@ -1,0 +1,104 @@
+= Add Common Script Configuration File
+:author:            James R. Perkins
+:email:             jperkins@redhat.com
+:toc:               left
+:icons:             font
+:idprefix:
+:idseparator:       -
+
+== Overview
+
+There is a desire to have a common script configuration file, `common.conf`. This will be called before the standalone
+or domain script configuration file. The other scripts will also invoke the common script configuration file. It allows
+things like common environment variables to be set by users regardless of the script being used.
+
+== Issue Metadata
+
+=== Issue
+
+* https://issues.redhat.com/browse/WFCORE-5261[WFCORE-5261]
+* https://issues.redhat.com/browse/WFLY-14399[WFLY-14399]
+
+=== Related Issues
+
+* https://issues.redhat.com/browse/EAP7-1251[EAP7-1251]
+
+=== Dev Contacts
+
+* mailto:{email}[{author}]
+
+=== QE Contacts
+
+* mailto:spriadka@redhat.com[Simon Priadka]
+
+=== Testing By
+
+* [x] Engineering
+
+* [ ] QE
+
+=== Affected Projects or Components
+
+* WildFly Core
+* WildFly
+* Any project or component which invokes script files
+
+=== Other Interested Projects
+
+=== Relevant Installation Types
+* [x] Traditional standalone server (unzipped or provisioned by Galleon)
+
+* [x] Managed domain
+
+* [x] OpenShift s2i
+
+* [ ] Bootable jar
+
+== Requirements
+
+=== Hard Requirements
+
+* The common script should be invoked before the standard script configuration files.
+* This will need to work on all supported platforms.
+** `common.conf`
+** `common.conf.bat`
+** `common.conf.ps1`
+* The environment variable `COMMON_CONF` can be used to override the path to the file.
+** By default, `$JBOSS_HOME/bin/common.conf` will be used.
+* All scripts in the `$JBOSS_HOME/bin` directory will invoke this script configuration file.
+** Note this includes scripts in WildFly Core and WildFly.
+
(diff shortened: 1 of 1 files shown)


scripts/WFCORE-5261-common-conf.adoc
```diff
@@ -0,0 +1,102 @@
...
+** By default, `$JBOSS_HOME/bin/common.conf` will be used.
+* All scripts in the `$JBOSS_HOME/bin` directory will invoke this script configuration file.
+
+=== Nice-to-Have Requirements
+
+=== Non-Requirements
+
+* The `common.conf`, `common.conf.bat` and `common.conf.ps1` files do not need to be present in default distribution.
+  Documentation will be added to indicate these files can be created. However, we should not add these files by default
+  as we shouldn't add more clutter to the standard bin directory.
```
spriadka, line 75: Will runtime scripts take into account this custom conf as well?

scripts/WFCORE-5261-common-conf.adoc
```diff
@@ -0,0 +1,102 @@
...
+:toc:               left
+:icons:             font
+:idprefix:
+:idseparator:       -
+
+== Overview
+
+There is a desire to have a common script configuration file, `common.conf`. This will be called before the standalone
+or domain script configuration file. The other scripts will also invoke the common script configuration file. It allows
+things like common environment variables to be set by users regardless of the script being used.
```
spriadka, line 13: Can you please specify the scope of things that will be taken into account of this custom configuration file? Will it be only environment variables?

scripts/WFCORE-5261-common-conf.adoc
jamezp: If I'm understanding you correctly yes. All scripts will invoke the `common.conf`, `common.conf.bat` or `common.conf.ps1` if they exist. By all scripts I mean the ones in the `$JBOSS_HOME/bin` directory. Looking at it we may need to do some work in WildFly as well as there are some scripts there that do not exist in WildFly Core.

scripts/WFCORE-5261-common-conf.adoc
jamezp: User could really put anything in there that can be executed. From the description of the RFE it seems to mostly discuss environment variables. However, we can't really validate or control what is placed in there.

scripts/WFCORE-5261-common-conf.adoc
```diff
@@ -0,0 +1,104 @@
...
+* The environment variable `COMMON_CONF` can be used to override the path to the file.
+** By default, `$JBOSS_HOME/bin/common.conf` will be used.
+* All scripts in the `$JBOSS_HOME/bin` directory will invoke this script configuration file.
+** Note this includes scripts in WildFly Core and WildFly.
+
+=== Nice-to-Have Requirements
+
+=== Non-Requirements
+
+* The `common.conf`, `common.conf.bat` and `common.conf.ps1` files do not need to be present in default distribution.
```
OndrejKotek, line 75: In case COMMON_CONF is defined but the referenced file does not exist or is not accessible, will there be a warning informing a user about such suspicious situation?

jamezp: I did not implement it that way, but that is a good idea.

scripts/WFCORE-5261-common-conf.adoc
```diff
@@ -0,0 +1,104 @@
...
+* Any project or component which invokes script files
+
+=== Other Interested Projects
+
+=== Relevant Installation Types
+* [x] Traditional standalone server (unzipped or provisioned by Galleon)
+
+* [x] Managed domain
+
+* [x] OpenShift s2i
```
mchoma, line 53: Do we really want to expose custom scripts into runtime? So far we have only custom s2i scripts in build time (that is safe). But introducing custom scripts into runtime open whole new set of problems. As immutability of image. Consistency of EAP pods in cluster. When will be change applied? ...

Also in Openshift we have native way for setting environment variables - ConfigMaps. I think we should recommend this and not come with own proprietary solution.

I read the linked customer report and it was not specifically targeted for OpenShift environment.

Can we avoid advertising this feature in OpenShift for customers?

@jmesnil @luck3y WDYT?

jamezp: We essentially already do this with the `standalone.conf`. The only difference between that and the `common.conf` is where in the `standalone.sh` it runs and the fact that it will run on all scripts. I only marked it here because I thought the end user would be able to add it. However I don't think it's something we should provide in the image so maybe I'm wrong.

scripts/WFCORE-5261-common-conf.adoc
snelluli1: @jamezp: Do we have this implemented now? If yes, what error message does the system provide?

PR #373, opened 2021-02-10
[WFCORE-4360] Support expression resolution backed by a CredentialStore.

darranl:
This pull request supersedes https://github.com/wildfly/wildfly-proposals/pull/213

This analysis discusses the resources and utilities that will be provided to support the encryption of expressions in the management model.

https://issues.jboss.org/browse/EAP7-677
https://issues.jboss.org/browse/WFCORE-4360

At this point in time unless there are problems in the late stages if implementation it is not expected that the details will diverge too fat from this analysis.

### DIFF
--- a/elytron/WFCORE-4360-CredentialStore_Expression_Resolution.adoc
+++ b/elytron/WFCORE-4360-CredentialStore_Expression_Resolution.adoc
@@ -193,7 +193,7 @@ We will not support the decryption of expressions other than the support within 
 
 Base64 representations will be used for both the export and import of the generated keys and for the expression value to be passed to the expression resolver - generally these details should be opaque to the end user however we may include the description within the documentation allowing for others to make use of the format.  We should consider that in the future alternative representations may be required either to support alternative key types or to support alternative representations so the representations will also be versioned.
 
-The general structure of the first representation will be (illestrated as an array): -
+The general structure of the first representation will be (illustrated as an array): -
 
     { 'E', 'L', 'Y', (VERSION), (TYPE), VALUE... }
 
--- a/elytron/WFCORE-4360-CredentialStore_Expression_Resolution.adoc
+++ b/elytron/WFCORE-4360-CredentialStore_Expression_Resolution.adoc
@@ -156,13 +156,18 @@ The `--import-secret-key` action will take the alias to store the key as an argu
 
  * `--key` - The key to import encoded using Base64.
 
-If the `--key` option is not specified the user will be prompted to enter the key interactively, this will avoid the key being seen in the current running processes and avoid the key being cached in the users command line history.
+If the `--key` option is not specified the user will be prompted to enter the key interactively, this will avoid the key being seen in the current running
+processes and avoid the key being cached in the users command line history.
 
-NOTE: For `generate-secret-key` and `import-secret-key` if an entry already exists for that alias / algorithm combination the behaviour will be the same as the `add` action in relation to replacing the entry or reporting an error.  (To be checked)
+NOTE: For `generate-secret-key` and `import-secret-key` if an entry already exists for that alias / algorithm combination the entry in the credential store
+will be replaced.
 
-For this specific enhancement where an algorithm is required for the `SecretKey` we will use the hard coded value of `AES`, if later enhancement add support for further key types then it may be necessary for an optional `algorithm` argument / parameter to be added to each of the above commands / operations.  Should an `algorithm` argument be added at a later point it would be optional and default to `AES` for backwards compatibility.
+For this specific enhancement where an algorithm is required for the `SecretKey` we will use the hard coded value of `AES`, if later enhancement add support
+for further key types then it may be necessary for an optional `algorithm` argument / parameter to be added to each of the above commands / operations.  Should
+an `algorithm` argument be added at a later point it would be optional and default to `AES` for backwards compatibility.
 
-It is not feasible for all actions to have a short form so these new actions will have a long form only, within the tool we should try and restrict the use of the short form for parameters only.
+It is not feasible for all actions to have a short form so these new actions will have a long form only, within the tool we should try and restrict the use of
+the short form for parameters only.
 
 ==== Key manipulation - Management Operations
 
--- a/elytron/WFCORE-4360-CredentialStore_Expression_Resolution.adoc
+++ b/elytron/WFCORE-4360-CredentialStore_Expression_Resolution.adoc
@@ -109,6 +109,55 @@ As with the command line utility it is expected each of these will need to take 
 
 NOTE: The credential store also offers a programatic API, should the tooling and operations be insufficient for an end user manual population of the credential store may remain an option.
 
+==== Command Line Encryption
+
+Whilst the Elytron tool does contain a `credential-store` command the encryption of values is really a special operation so a new command `encryption` will be added to the tool.
+
+When using the new `encryption` command it will be necessary to reference an existing `alias` within an existing `credential-store` to obtain an instance of the `SecretKey` that will be used for encryption of a plain text String, this command will make use of the same attributes as used by the `credential-store` command to identify the store and alias to use to load the key.  This is important so that using different commands in the same command line utility is consistent.
+
+The encryption of String instances will make use of `javax.crypto.Cipher` APIs, although the only `SecretKey` type supported will be `AES` various transformations will be supported, a transformation can be specified in the format: -
+
+    Algorithm Name/Algorithm Mode/Padding
+
+The support algorithm names will be: -
+
+ * `AES`
+ * `AES_128`
+ * `AES_192`
+ * `AES_256`
+
+The final three names restrict the algorithm to the specified key size.
+
+The supported modes will be: -
+
+ * `CBC`
+ * `CTR`
+ * `GCM`
+ 
+The following padding will be supported: -
+
+ * `NoPadding`
+ * `PKCS5Padding`
+
+Different permutations of name, mode, and padding require different `ParameterSpec` instances to configure the `Cipher`, by defining the supported set in advance selection of the correct `ParameterSpec` can be more deterministic.
+
+If no cipher is specified the tool will default to `????`.
+
+The new `encryption` command will support two actions: -
+
+ * `--encrypt` - To encrypt a value and return the result Base64 encoded.
+ * `--create-expression` - To encrypt a value and return it as a complete expression.
+ 
+In both cases the value to be encrypted can be passed in with the action or if not specified the user will be prompted to enter the value.  In the case of `--create-expression` the user can also optionally specify the `resolver` that will be used to create the expression String.
+
+The following arguments can also be specified to both actions: -
+
+ * `--iv-length` - The size in bytes of the initialisation vector, valid range <12,16> with a default of 12.
+ * `--authentication-tag-length` - The length of the authentication tag in bits, only applicable for the `GCM` mode.  Default 128.
+
+The tool will not add any support for the decryption of expressions or their values.
+
+
 ==== Expression Resolver
 
 Presently WildFly Core supports an expression resolver that can delegate to a Vault configuration and if that is not available fall back to use either system properties or environment variables.  This will be updated to make use of the `CapabilityRegistry` and attempt to lookup a capability using a pre-defined constant (To be defined) which exposes an expression resolver runtime API.  The expression resolver looked up using a capability will be used first to attempt to resolve any expression, if this is successful the resolved value will be used - if not we will fallback to the existing behaviour of using the vault and if that fails using a system property or environment variable.
@@ -138,22 +187,7 @@ The return value of the operation will be the complete expression ready to be us
 
 We will not support the decryption of expressions other than the support within the management model to resolve existing expressions.
 
-==== Command Line Encryption
 
-Whilst the Elytron tool does contain a `credential-store` command the encryption of values is really a special operation so a new command `encryption` will be added to the tool.
-
-This command will duplicate the parameters used by the `credential-store` command to configure the location and initialisation of the underlying credential store, additionally it will take an argument to specify the alias from the credential store to load the `SecretKey` to use for encryption.
-
-Additional arguments to the tool will be used to specify the algorithm to use for encryption as well as any additional parameters requires such as a salt or iteration count.
-
-The new `encryption` command will support two actions: -
-
- * `--encrypt` - To encrypt a value and return the result Base64 encoded.
- * `--create-expression` - To encrypt a value and return it as a complete expression.
- 
-In both cases the value to be encrypted can be passed in with the action or if not specified the user will be prompted to enter the value.  In the case of `--create-expression` the user can also optionally specify the `resolver` that will be used to create the expression String. 
-
-The tool will not add any support for the decryption of expressions or their values.
 
 ==== Base64 Representations
 
@@ -173,7 +207,7 @@ As version `1` of this implementation only supports `AES` `SecretKey` instances 
 
 An encrypted expression will be encoded with a type of 2: -
 
-    { 'E', 'L', 'Y', (VERSION), 2, DATA... }
+    { 'E', 'L', 'Y', (VERSION), 2, IV Length, Initialisation Vector, Cipher Text }
 
 At the moment only a single version of the representation will be supported, in the future the commands and operations that generated the encoded values may take a parameter to specify which version to generate however that is not required within the first implementation.  Additionally later versions of the tooling and operations should always attempt to use the oldest version possible to encode a value allowing newer tools to encode values for use by older versions.
 
--- a/elytron/WFCORE-4360-CredentialStore_Expression_Resolution.adoc
+++ b/elytron/WFCORE-4360-CredentialStore_Expression_Resolution.adoc
@@ -87,6 +87,20 @@ This enhancement is only in relation to expression resolution within the applica
 
 Expression resolution will only be supported against attributes that already support expression resolution, this enhancement will not perform a review of which attributes support expression resolution and will not be changing any attributes to support expression resolution.
 
+== Implementation Plan
+
+A large portion of this enhancement is going to be in relation to the addition of the expression resolver within the management model and demonstrating the viability of the approach for this reason the implementation will be divided into two stages.
+
+=== Hard Requirements
+
+A first pass of the implementation will focus exclusively on the defined "Hard Requirements" this will result in an end to end delivery of the enhancement.
+
+=== Nice-to-Have Requirements
+
+The nice to have requirements are effectively additional enhancements that can be added to this work such as alternative key types etc..
+
+After the hard requirements have been implemented we will make a decision as to which of the nice to have requirements we will also include in this work, a part of this decision process will take into account any upcoming feature freeze deadlines.  This enhancement can be considered complete once the hard requirements have been implemented, if any of the nice to have requirements have not been included in the enhancement we will consider raising future feature requests to further enhance this in a later stage.
+
 == Test Plan
 
 Any utilities added to the Wildfly Elytron project will also be accompanied with their own unit tests.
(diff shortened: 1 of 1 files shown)


elytron/WFCORE-4360-CredentialStore_Expression_Resolution.adoc
```diff
@@ -0,0 +1,419 @@
...
+
+The overall intent of this enhancement is for the Elytron subsystem to provide a configurable expression resolver resource to make use of an `AES`
+`SecretKey` loaded from a `CredentialStore` to decrypt expressions in the management model which have been previously encrypted using the same key.  To
+meet this requirement some additional functionality will also be required to hande the management of the keys.
+
+It is a requirement within this enhancement that multiple secret keys can be used for the decryption of expressions, these keys could be loaded from the same
+credential store or from separate credential stores - this relates to the "chicken and egg" problem which will be described later.
+
+The initial integration of this enhancement will not support specifying custom `java.security.Provider` instances as the key type and transformation will be
+using pre-defined defaults - however the expression resolver will be initialised after the Elytron subsystem has performed initialisation and registration of
```
OndrejKotek, line 85: What about attributes of the Elytron subsystem and provider-loader that allow expressions? Do I understand it well that those will be processed just with the general resolver but the new credential store related resolver will not be applied to those?

darranl: On this point there is going to be a limit to how far we can go.

Within the implementation the act of resolving an expression could trigger some initialisation - i.e. we need a credential store and we need the providers registered.  This initialisation could trigger the need for expression resolution.

We must be able to support the scenario where the clear-text password of one credential store is encrypted using the secret key of another credential store so if there is any recursion triggered for expression resolution that must be supported.

If however the recursion leads to an unresolvable cycle that will have to fail and lead to an error.  A cycle should be easy to detect i.e. if on first use of "resolverOne" the initialisation triggers something else requiring "resolverOne" we know we have a cycle - on a ThreadLocal we will likely maintain a Set of requested resolvers which would enable us to detect a longer chain forming a cycle.

If recursion is triggered alternative expression resolution using system properties should remain possible.

elytron/WFCORE-4360-CredentialStore_Expression_Resolution.adoc
```diff
@@ -0,0 +1,420 @@
...
+
+ * `generate-secret-key`
+ * `export-secret-key`
+ * `import-secret-key`
+ 
+Within the management model operations are self describing so making use of dedicated operations makes it easier for tooling to present a meaningful UI to end
+users automatically.
+
+The `credential-store` resource already contains a `remove-alias` operation however this assumes the type of the credential is `PasswordCredential` a new
+attribute will be added to this operation `entry-type` which will default to `PasswordCredential` but will also accept `SecretKeyCredential` as a value
```
OndrejKotek, line 170: Does this mean that we will allow to store two entries of different types under one alias?

darranl: Yes, the credential store APIs and implementation have always supported multiple credential types under the same alias - it is just that up until recently we only supported PasswordCredential.  

Within this RFE anything that needs a SecretKey will be assuming the type is SecretKeyCredential.

The support for storing keys used for SSH will also have been using this feature to store different credential types under the same alias.

elytron/WFCORE-4360-CredentialStore_Expression_Resolution.adoc
```diff
@@ -0,0 +1,420 @@
...
+
+ * `alias` - The alias to use to store the imported key.
+ * `key` - The encoded representation of the secret key.
+
+=== Expression Resolver Resource
+
+Presently WildFly Core supports an expression resolver that can delegate to a Vault configuration and if that is not available fall back to use either system
+properties or environment variables.  This will be updated to make use of the `CapabilityRegistry` and attempt to lookup a capability using a predefined
+constant `org.wildfly.controller.expression-resolver` which exposes an expression resolver runtime API.  The expression resolver looked up using a capability
+will be used after first attempting to resolve the expression as a Vault expression to resolve any expression.
```
OndrejKotek, line 204: Do I understand it correctly that the new ordering will be: Vault, capabilities, system properties, environment variables?

darranl: I would need to double check the exact order but 100% this new capaility based resolution is also immediately after Vault.  

Also this should not impact any existing recursion that is (or is not) possible with vault.

elytron/WFCORE-4360-CredentialStore_Expression_Resolution.adoc
```diff
@@ -0,0 +1,420 @@
...
+
+The `--export-secret-key` action will return a single Base64 value which is the key in it's encoded form.
+
+The `--import-secret-key` action will take the alias to store the key as an argument, it will also take the following additional option: -
+
+ * `--key` - The key to import encoded using Base64.
+
+If the `--key` option is not specified the user will be prompted to enter the key interactively, this will avoid the key being seen in the current running processes and avoid the key being cached in the users command line history.
+
+NOTE: For `generate-secret-key` and `import-secret-key` if an entry already exists for that alias / algorithm combination the behaviour will be the same as the `add` action in relation to replacing the entry or reporting an error.  (To be checked)
```
OndrejKotek, line 148: I have tried the `add` action and it replaces an existing entry. On the other hand the corresponding management operation reports an error. Will we keep this discrepancy also for the new actions and operations?

elytron/WFCORE-4360-CredentialStore_Expression_Resolution.adoc
```diff
@@ -0,0 +1,420 @@
...
+By default the output of the `encrypt` action will just output the token representing the encrypted value, if the parameter `--summary` is also supplied to
+the command the output will also illustrate how this could be represented as an expression.  However the command line tool will not be aware of the management
+model configuration so any expression representation will be for guidance only.
+
+==== Management Operation
+
+The `expression=encryption` resource will also contain an operation `create-expression` the purpose of this operation being to take a clear text value and
+using one of the defined `resolvers` output an expression that can be used elsewhere in the management model.  The operation will just require two parameters:
+
+ * `resolver` - The name of the defined resolver to use to encrypt the data, if the `expression=resolver` resource has the `default-resolver` attribute
```
OndrejKotek, line 262: expression=resolver -> expression=encryption ?

elytron/WFCORE-4360-CredentialStore_Expression_Resolution.adoc
```diff
@@ -0,0 +1,420 @@
...
+model configuration so any expression representation will be for guidance only.
+
+==== Management Operation
+
+The `expression=encryption` resource will also contain an operation `create-expression` the purpose of this operation being to take a clear text value and
+using one of the defined `resolvers` output an expression that can be used elsewhere in the management model.  The operation will just require two parameters:
+
+ * `resolver` - The name of the defined resolver to use to encrypt the data, if the `expression=resolver` resource has the `default-resolver` attribute 
+ defined this parameter can be omitted.
+ * `data` - The clear text value to be encrypted.
```
OndrejKotek, line 264: What about `clear-text` instead of `data`? That would correspond with the command line tool option.

elytron/WFCORE-4360-CredentialStore_Expression_Resolution.adoc
darranl: I think keep it.  

IMO the management operation sounds safer as it prevents an accidental overwrite but I am not sure if we want to change the tool in that way at this point - if we did we could probably have an ELY issue opened straight away as it in relation to existing behaviour.

elytron/WFCORE-4360-CredentialStore_Expression_Resolution.adoc
darranl: +1 I will change that,

elytron/WFCORE-4360-CredentialStore_Expression_Resolution.adoc
darranl: I will quickly clean these up now, I had a couple of iterations on naming this resource.

elytron/WFCORE-4360-CredentialStore_Expression_Resolution.adoc
```diff
@@ -0,0 +1,420 @@
...
+
+The return value of the operation will be the complete expression ready to be used elsewhere in the mode.  As this resource is aware of it's complete
+configuration the result will be usable as returned, caution may however be required if an expression is being prepared on one process or profile to be used
+on another in case there are differences in the resolver configuration.
+
+We will not support the decryption of expressions other than the support within the management model to resolve existing expressions.
```
OndrejKotek, line 270: What about `:resolve-expression(expression=${ENC:Resolver:ENCRYPTED_DATA})`, will it work?

darranl: Actually yes, I believe that call should work.

elytron/WFCORE-4360-CredentialStore_Expression_Resolution.adoc
```diff
@@ -0,0 +1,420 @@
...
+Any utilities added to the Wildfly Elytron project will also be accompanied with their own unit tests.
+
+The major testing of this enhancement will happen within the WildFly Core project allowing us to test in the same location it is implemented.
+
+== Community Documentation
+
+The community documentation will be enhanced to include details of the new feature.
+
+The documentation must sufficiently describe the encryption process to a level which would allow users and third parties to create their own tooling to generate the expressions.  They may wish to accomplish this using Java however they may also choose to do so in alternative languages provided the required cipher algorithms are available.
+
```
OndrejKotek, line 420: [Release Note Content](https://github.com/wildfly/wildfly-proposals/blob/master/design-doc-template.adoc#release-note-content) is missing.

elytron/WFCORE-4360-CredentialStore_Expression_Resolution.adoc
```diff
@@ -0,0 +1,420 @@
...
+Within WildFly Core the general `ExpressionResolver` support will be updated to detect the availability of an expression resolver capability exposing
+a runtime API for expression resolution.
+
+Also within WildFly Core the Elytron subsystem will be updated with new resources as required, resources will also contain operations both for key management
+and operations for the encryption of clear text strings for use as expressions in the management model.
+
+Additionally all documentation lives within the WildFly project so updates will be applied there accordingly.
+
+=== Other Interested Projects
+
```
OndrejKotek, line 72: [Relevant Installation Types](https://github.com/wildfly/wildfly-proposals/blob/master/design-doc-template.adoc#relevant-installation-types) is missing.

darranl: I will add this but there will be no exclusions, this will be a core feature applicable in all installation types.

elytron/WFCORE-4360-CredentialStore_Expression_Resolution.adoc
darranl: I will add this one but this is just for Brian for community content, this is not part of the RFE process.

elytron/WFCORE-4360-CredentialStore_Expression_Resolution.adoc
```diff
@@ -0,0 +1,420 @@
...
+==== Command Line Encryption
+
+It was initially considered if a new top level command should be added to the WildFly Elytron Tool to support the encryption of clear text values, however as
+this enhancement is only making use a single transformation we do not have a need for many additional configuration parameters beyond those required to
+initialise and access the underlying credential store.
+
+The `credential-store` command will be updated with an additional action `encrypt`, this in turn will make use of two parameters:
+
+ * `--alias` - The alias of the `SecretKey` stored within the credential store.
```
OndrejKotek, line 247: According to the implementation, an alias is taken as the argument of the action `encrypt`. The implementation is consistent.

elytron/WFCORE-4360-CredentialStore_Expression_Resolution.adoc
```diff
@@ -0,0 +1,420 @@
...
+The following operations will be added to the `credential-store` resource:
+
+ * `generate-secret-key`
+ * `export-secret-key`
+ * `import-secret-key`
+ 
+Within the management model operations are self describing so making use of dedicated operations makes it easier for tooling to present a meaningful UI to end
+users automatically.
+
+The `credential-store` resource already contains a `remove-alias` operation however this assumes the type of the credential is `PasswordCredential` a new
```
OndrejKotek, line 169: There is also the `read-aliases` operation. The `entry-type` parameter should be available also in this case, or the output should be structured to cover all the entry types.

elytron/WFCORE-4360-CredentialStore_Expression_Resolution.adoc
darranl: Added.

elytron/WFCORE-4360-CredentialStore_Expression_Resolution.adoc
darranl: Done

elytron/WFCORE-4360-CredentialStore_Expression_Resolution.adoc
darranl: Done

elytron/WFCORE-4360-CredentialStore_Expression_Resolution.adoc
darranl: For now the credential store already supports two different entry types and this missing information is already an issue in the present implementation, this RFE is adding support for a third entry type.

I agree we should improve our read-aliases support but I think it should be a part of a small follow up RFE as it ends up with it's own set of compatibility issues.

Firstly the credential store APIs do not expose this information so changes to the SPI could impact existing implementations.

I would suggest the read-aliases operation should have a new parameter "show-type" to switch on showing type information, there should also be a second paramter "types" so the caller can specify the types to return in the output.

I don't mind getting this raised into a new RFE unless you really want to push that this should be solved for WildFly 23.

elytron/WFCORE-4360-CredentialStore_Expression_Resolution.adoc
```diff
@@ -0,0 +1,420 @@
...
+attribute will be added to this operation `entry-type` which will default to `PasswordCredential` but will also accept `SecretKeyCredential` as a value
+allowing credentials of type `SecretKeyCredential` to be removed from the credential store.  The `secret-key-credential-store` resource will also have a
+`remove-alias` operation added but as this credential store only supports a single credential type the type parameter will be omitted.
```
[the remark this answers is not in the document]
darranl: I have added this, it should also have the reload operation although like the existing resource I need to implement that/

darranl:
@OndrejKotek I have had another pass through, other than the "read-aliases" operation which you may want to discuss further I think I have covered all the points you raised.

elytron/WFCORE-4360-CredentialStore_Expression_Resolution.adoc
OndrejKotek: Thank you, this can be resolved.

elytron/WFCORE-4360-CredentialStore_Expression_Resolution.adoc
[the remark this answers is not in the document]
OndrejKotek: `credential-store-resource` -> `credential-store` resource ?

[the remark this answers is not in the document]
darranl: Done

elytron/WFCORE-4360-CredentialStore_Expression_Resolution.adoc
OndrejKotek: Thank you, this can be resolved.

elytron/WFCORE-4360-CredentialStore_Expression_Resolution.adoc
OndrejKotek: Thanks a lot for the explanation. Maybe it's worth mentioning the solution here. In case you consider this just an implementation detail, this conversation can be resolved.

elytron/WFCORE-4360-CredentialStore_Expression_Resolution.adoc
OndrejKotek: Agree.

The note can be updated and this conversation can be resolved.

elytron/WFCORE-4360-CredentialStore_Expression_Resolution.adoc
darranl: I think this specific detail is an implementation detail, but this discussion has reminded me that we should list the relative-to paths that we can support.  As expression resolution is early we can not use user defined paths but we should support the standard set e.g. jboss.server.config.dir so I should list those.

elytron/WFCORE-4360-CredentialStore_Expression_Resolution.adoc
darranl: This one is done.

elytron/WFCORE-4360-CredentialStore_Expression_Resolution.adoc
OndrejKotek: I have no strong arguments against moving this to a follow-up. Can create a new RFE for this please?

darranl: I will create that now then we can close this discussion.

elytron/WFCORE-4360-CredentialStore_Expression_Resolution.adoc
OndrejKotek: Thank you for the info, this can be resolved.

elytron/WFCORE-4360-CredentialStore_Expression_Resolution.adoc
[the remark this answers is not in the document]
OndrejKotek: Thank you for the update, this can be resolved.

elytron/WFCORE-4360-CredentialStore_Expression_Resolution.adoc
darranl: Here is a new RFE to handle this https://issues.redhat.com/browse/EAP7-1643

elytron/WFCORE-4360-CredentialStore_Expression_Resolution.adoc
OndrejKotek: This is covered by https://issues.redhat.com/browse/WFWIP-373 now, the conversation can be resolved.

elytron/WFCORE-4360-CredentialStore_Expression_Resolution.adoc
```diff
@@ -0,0 +1,459 @@
...
+`javax.crypto.SecretKeyFactory` as the keys can be generated using random key material directly using the `javax.crypto.spec.SecretKeySpec` class we should
+still use the `SecretKeyFactory` in case certified providers are added for key generation.  The generated `SecretKey` will be stored in a `CredentialStore`.
+
+==== Credential Store
+
+==== PropertiesCredentialStore / secret-key-credential-store
+
+A new credential store type will be implemented, this credential store will be dedicated to the storage of `SecretKeyCredential` instances which will be stored
+in a properties file.  Unlike other credential store types this implementation will not be adding any additional protection to the keys it contains, this
+credential store is primarily intended to solve the problem of how we provide the first secret key to the application server process without relying on a hard
```
OndrejKotek, line 111: I have just realized the an external source can be used for the master password, which is not documented upstream, see https://issues.redhat.com/browse/WFLY-14589

For the documentation (analysis) purposes, what is the preferred way how to secure the master password? For example, I am not sure whether `secret-key-credential-store` could be used for a FIPS compliant configuration.

darranl: Yes using CMD or EXT to obtain a password to unlock a credential store does remain an option so if a user wants they don't have to use the secret-key-credential-store - overall they have a few options now to decide the most appropriate approach for their initial secrets.

PR #379, opened 2021-03-04
[CLOUD-3928], Evolve support for custom Galleon feature-packs usage during s2i

jfdenise:


jmesnil (approved):

openshift/CLOUD_3928_Custom_galleon_layers.adoc
```diff
@@ -0,0 +1,113 @@
...
+
+* Define new env variable: `GALLEON_DIR=<path relative to app root dir that contains Galleon content>`. Default being `galleon`. If the referenced directory doesn't
+exist, the build phase aborts.
+
+ * A settings.xml file can be located in the `${GALLEON_DIR: galleon}` dir and used during provisioning. Any local repository configured in such settings are ignored. 
+If no settings.xml is found, a default settings.xml is used.
+
+* Define a new env variable: `GALLEON_CUSTOM_FEATURE_PACKS_MAVEN_REPO=<absolute path to a maven repo dir containing custom galleon feature-packs>`
+By default we are looking into the `<src root dir>/${GALLEON_DIR:galleon}/repository`. Any path can be set for `GALLEON_CUSTOM_FEATURE_PACKS_MAVEN_REPO`, this allows to mount a volume containing the maven repo.
+During s2i build, the content of the maven repository is copied to the local maven repository present in the image used to provision a server.
```
jbliznak, line 80: Does that mean `copied over`, ie. GALLEON_CUSTOM_FEATURE_PACKS_MAVEN_REPO can be used to override galleon repo available in builder image?

openshift/CLOUD_3928_Custom_galleon_layers.adoc
```diff
@@ -0,0 +1,113 @@
...
+If the referenced directory doesn't exist, the build phase aborts.
```
jbliznak, line 81: How does this work together with this requirement?
> By default we are looking into the `<src root dir>/${GALLEON_DIR:galleon}/repository`.

openshift/CLOUD_3928_Custom_galleon_layers.adoc
```diff
@@ -0,0 +1,113 @@
...
+=== Hard Requirements
+
+* Define new env variable: `GALLEON_DIR=<path relative to app root dir that contains Galleon content>`. Default being `galleon`. If the referenced directory doesn't
+exist, the build phase aborts.
+
+ * A settings.xml file can be located in the `${GALLEON_DIR: galleon}` dir and used during provisioning. Any local repository configured in such settings are ignored. 
+If no settings.xml is found, a default settings.xml is used.
+
+* Define a new env variable: `GALLEON_CUSTOM_FEATURE_PACKS_MAVEN_REPO=<absolute path to a maven repo dir containing custom galleon feature-packs>`
+By default we are looking into the `<src root dir>/${GALLEON_DIR:galleon}/repository`. Any path can be set for `GALLEON_CUSTOM_FEATURE_PACKS_MAVEN_REPO`, this allows to mount a volume containing the maven repo.
```
jbliznak, line 79: Could the default `<src root dir>/${GALLEON_DIR:galleon}/repository` or `GALLEON_CUSTOM_FEATURE_PACKS_MAVEN_REPO` be used together with provisioning.xml?

openshift/CLOUD_3928_Custom_galleon_layers.adoc
```diff
@@ -0,0 +1,113 @@
...
+
+=== Hard Requirements
+
+* Define new env variable: `GALLEON_DIR=<path relative to app root dir that contains Galleon content>`. Default being `galleon`. If the referenced directory doesn't
+exist, the build phase aborts.
+
+ * A settings.xml file can be located in the `${GALLEON_DIR: galleon}` dir and used during provisioning. Any local repository configured in such settings are ignored. 
+If no settings.xml is found, a default settings.xml is used.
+
+* Define a new env variable: `GALLEON_CUSTOM_FEATURE_PACKS_MAVEN_REPO=<absolute path to a maven repo dir containing custom galleon feature-packs>`
```
jbliznak, line 78: Does it have to be absolute? If I want to bundle the repository within application code, do I need to find out where and how the application get cloned to determine the absolute path? 

I am thinking about allow absolute path or the one relative to the `<src root dir>`, WDYT?

openshift/CLOUD_3928_Custom_galleon_layers.adoc
jfdenise: No, this repo is only for custom feature-packs, the local repo can't be set.

openshift/CLOUD_3928_Custom_galleon_layers.adoc
jfdenise: If GALLEON_CUSTOM_FEATURE_PACKS_MAVEN_REPO is set. We don't look at `<src root dir>/${GALLEON_DIR:galleon}/repository` in this case.

openshift/CLOUD_3928_Custom_galleon_layers.adoc
jfdenise: That is one or the other. If GALLEON_CUSTOM_FEATURE_PACKS_MAVEN_REPO is set, we use it, otherwise we look in the default path.

openshift/CLOUD_3928_Custom_galleon_layers.adoc
jfdenise: If you want it to be relative, then you can use GALLEON_DIR=foo/bar . Then we will expect to find a `repository` directory inside <app src>/foo/bar

openshift/CLOUD_3928_Custom_galleon_layers.adoc
jbliznak: so in maven terminology this repo will serve as another remote repo for the build? eg. something like:
```xml
<profile>
      <id>galleon-custom-fp-repo-profile</id>
      <repositories>
        <repository>
          <id>galleon-custom-fp-repo</id>
          <url>file://${env.GALLEON_CUSTOM_FEATURE_PACKS_MAVEN_REPO}</url>
          <releases>
            <enabled>true</enabled>
          </releases>
          <snapshots>
            <enabled>true</enabled>
          </snapshots>
        </repository>
      </repositories>
</profile>
```

jfdenise: We keep the builder image local repo, we simply copy into it the custom repository content.

openshift/CLOUD_3928_Custom_galleon_layers.adoc
jbliznak: that's a possibility too, ok :)

openshift/CLOUD_3928_Custom_galleon_layers.adoc
jbliznak: That's what I asked originally, so I suppose that the answer there should have been yes.
In other words, is the content of `GALLEON_CUSTOM_FEATURE_PACKS_MAVEN_REPO` somehow restricted or it can contain any maven artifacts? Let's suppose I would like to use it to provide all artifacts needed to build an app (not just FPs) to perform offline build - is that supported scenario?

jfdenise: It is unsupported scenario, we will only document that this is the place where to put Galleon related artifacts, not application artifacts.

jbliznak: Thanks.

Btw, what about shorter name of the variable? eg. just `GALLEON_MAVEN_REPO` or `GALLEON_CUSTOM_MAVEN_REPO`?

jfdenise: I agree that the name is long but it also makes it clear that the repo is supposed to contain only custom galleon feature-packs.

jbliznak: ok, that was last question, thanks for your patience.

jbliznak (approved):

openshift/CLOUD_3928_Custom_galleon_layers.adoc
```diff
@@ -0,0 +1,113 @@
...
+By default we are looking into the `<src root dir>/${GALLEON_DIR:galleon}/repository`. Any path can be set for `GALLEON_CUSTOM_FEATURE_PACKS_MAVEN_REPO`, this allows to mount a volume containing the maven repo.
+During s2i build, the content of the maven repository is copied to the local maven repository present in the image used to provision a server.
+If the referenced directory doesn't exist, the build phase aborts.
+
+* Define a new env variable: `GALLEON_PROVISION_FEATURE_PACKS=<comma separated list of custom galleon feature-packs GAV>`. The order is important, the feature-packs
+are added to the provisioning context in the order of the list. If the one of the GAV is missing G, A or V, the build phase aborts.
+This env variable can only be used in conjunction with the existing `GALLEON_PROVISION_LAYERS` that can now contain both WildFly s2i layers and custom layers.
```
jbliznak, line 85: Current state is
>  GALLEON_PROVISION_LAYERS env var has precedence over provisioning.xml

So does this line imply also the following?
> GALLEON_PROVISION_FEATURE_PACKS having precedence over provisioning.xml

jfdenise: The GALLEON_PROVISION_FEATURE_PACKS only makes sense if GALLEON_PROVISION_LAYERS is set. So we can say yes, but not alone.

PR #381, opened 2021-03-15
WIP

ehsavoie:


core/yaml.adoc
```diff
@@ -0,0 +1,125 @@
...
+
+=== Testing By
+// Put an x in the relevant field to indicate if testing will be done by Engineering or QE. 
+// Discuss with QE during the Kickoff state to decide this
+* [ ] Engineering
+
+* [ ] QE
+
+=== Affected Projects or Components
+
```
jmesnil, line 41: which projects are affected? I assume at least `wildfly-core` and `wildlfy-jar-maven-plugin`?

core/yaml.adoc
```diff
@@ -0,0 +1,125 @@
...
+=== Other Interested Projects
+
+=== Relevant Installation Types
+// Remove the x next to the relevant field if the feature in question is not relevant
+// to that kind of WildFly installation
+* [x] Traditional standalone server (unzipped or provisioned by Galleon)
+
+* [x] Managed domain
```
jmesnil, line 49: are you sure?

core/yaml.adoc
```diff
@@ -0,0 +1,125 @@
...
+
+* [x] OpenShift s2i
```
jmesnil, line 51: are you sure?

core/yaml.adoc
```diff
@@ -0,0 +1,125 @@
...
+* [ ] QE
+
+=== Affected Projects or Components
+
+=== Other Interested Projects
+
+=== Relevant Installation Types
+// Remove the x next to the relevant field if the feature in question is not relevant
+// to that kind of WildFly installation
+* [x] Traditional standalone server (unzipped or provisioned by Galleon)
```
jmesnil, line 47: are you sure?

core/yaml.adoc
```diff
@@ -0,0 +1,125 @@
...
+* [x] Managed domain
+
+* [x] OpenShift s2i
+
+* [x] Bootable jar
+
+== Requirements
+
+WildFly will be able to take a list of YAML files from the command line to redefine its exisiting configuration.
+The operations will be applied in the order the files are defined and after the initial operations defined by the XML configuration.
```
jmesnil, line 58: > The operations
I assume that these "operations" are WildFly management operations created from the YAML files input?

core/yaml.adoc
```diff
@@ -0,0 +1,125 @@
...
+The YAML configuration should be *idempotent* so that we can start and restart with the same command line without failure nor resulting with a different final configuration.
+The YAML structure should follow the management model structure. We may use YAML local tags to support operations syntax.
```
jmesnil, line 60: I'm not sure to understand what that means?

core/yaml.adoc
```diff
@@ -0,0 +1,125 @@
...
+For each entry, it will check if the resource is already created by the XML, otherwise it will add a new operation to the list of boot operations to add the required resource.
+If the resource already exist then it will add a new operation to the list of boot operations to update the attributes.
+
+=== Hard Requirements
+
+* Support for multiple files.
+External files will take precedence over the XML configuration ones. Then order will follow the command line.
+
+* If the resource is already defined then all we have to du is updating its attrbiutes using the `write-attribute` operation.
+* If the YAML part define an empty resource and the resource already exists: log a WARNING, otherwise try to add it.
```
jmesnil, line 71: why do you log a WARN and do not fail?

core/yaml.adoc
```diff
@@ -0,0 +1,125 @@
...
+* If the YAML is incorrect: like defining an non-existing attribute or child resource we should fail.
+
+=== Nice-to-Have Requirements
+
+* Undefine attribute: using a YAML tag !undefine. For example:
```
jmesnil, line 76: off-topic but I'm looking at Helm Charts and their convention is to use `nil` to remove something (this could be applied both to resources and attributes here)

core/yaml.adoc
```diff
@@ -0,0 +1,125 @@
...
+
+* [x] OpenShift s2i
+
+* [x] Bootable jar
+
+== Requirements
+
+WildFly will be able to take a list of YAML files from the command line to redefine its exisiting configuration.
+The operations will be applied in the order the files are defined and after the initial operations defined by the XML configuration.
+The YAML configuration should be *idempotent* so that we can start and restart with the same command line without failure nor resulting with a different final configuration.
```
jmesnil, line 59: I'm not sure to understand what that means. If my YAML is adding a resource to a list, is it correct?

Does the persisted management model includes the resources from the YAML? If that's not the case, every time we start the server, it will start from a clean state and we can add the resource to the list?

core/yaml.adoc
```diff
@@ -0,0 +1,125 @@
...
+wildfly-configuration:
+    subsystem:
+        microprofile-jwt-smallrye: !remove
+----
+
+* Allow support for adding element to list and supporting index: this might impact idempotence.
+
+=== Non-Requirements
+
+The goal of the YAML files is to be able to customize an existing configuration. It is not here to replace the existing configuration support with XML.
```
jmesnil, line 99: do we want to constraint the management subtrees that can be impacted by this feature:

* [ ] core-service
* [ ] deployment
* [ ] deployment-overlay
* [ ] extension
* [ ] interface
* [ ] path
* [ ] socket-binding-group
* [ ] subsystem
* [ ] system-property

core/yaml.adoc
```diff
@@ -0,0 +1,125 @@
...
+As such we won't add support to add extensions for the server as this might require modules which can be missing.
+As this configuration extension is for a server only this won't be supported in domain mode.
```
jmesnil, line 101: for a *standalone* server

jmesnil (changes requested):

core/yaml.adoc
```diff
@@ -0,0 +1,125 @@
...
+The aim of this proposal is to be able to inject external configuration in a YAML format to WildFly in order to customize an existing configuration.
+It is out of the scope of this proposal to provide support to define a complete server configuration in YAML: as such you won't be able to add a new extension using a YAML file.
+We also want this to be *idempotent*: that means that starting and restarting the server with the same command line should produce the same result and shouldn't fail.
+This feature is expected to be used as base for Bootable JAR implementation, it is not expected to be called directly by user (TODO get confirmation on this from Jeff and Brian and on how to 'hide' the feature).
+
+== Issue Metadata
+
+=== Issue
+
+* https://issues.redhat.com/browse/WFCORE[WFCORE-XXXX]
```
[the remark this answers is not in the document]
ehsavoie: none currently

core/yaml.adoc
ehsavoie: Only wildfly-core, bootable jar only if it wants to take advantage of that

core/yaml.adoc
ehsavoie: Because it doesn't do anything, failing seems a bit 'harsh' and would add the resource if it was missing from the XML.

core/yaml.adoc
ehsavoie: But null is a supported value in some attributes.

core/yaml.adoc
ehsavoie: No I was thinking of having several YAML files as arguments.
The persisted management model can be the one updated after applying the YAML, resetting an attribute value to the same value would don nothing, since the resources would already be created the add wouldn't occur, same for the  delete.

core/yaml.adoc
ehsavoie: Good point

core/yaml.adoc
ehsavoie: +1

core/yaml.adoc
ehsavoie: I updated the doc on this

core/yaml.adoc
ehsavoie: tags are specific yaml instructions  to (re)define a type. I don't want to introduce a schema an a namespace for them.

core/yaml.adoc
ehsavoie: Yes

ehsavoie:
Superseded by ttps://github.com/wildfly/wildfly-proposals/pull/384

PR #384, opened 2021-03-26
[WFCORE-5343]: Proposal for YAML support

ehsavoie:
Supersedes https://github.com/wildfly/wildfly-proposals/pull/381

### DIFF
--- a/core/WFCORE-5343_Support_YAML_files_for_configuration_customization.adoc
+++ b/core/WFCORE-5343_Support_YAML_files_for_configuration_customization.adoc
@@ -1,0 +1,256 @@
+= Support for YAML files to customize WildFly configuration
+:author:            Emmanuel Hugonnet
+:email:             ehugonne@redhat.com
+:toc:               left
+:icons:             font
+:idprefix:
+:idseparator:       -
+
+== Overview
+
+The aim of this proposal is to be able to inject external configuration in a YAML format to WildFly in order to customize an existing configuration.
+It is out of the scope of this proposal to provide support to define a complete server configuration in YAML: as such you won't be able to add a new extension using a YAML file.
+We also want this to be *idempotent*: that means that starting and restarting the server with the same command line should produce the same result and shouldn't fail.
+This feature is expected to be used by users of full WildFly installation.
+Also given that the YAML files are used to cusomize an XML configuration, they are expected to work on the current server implementation.
+This feature might interfere with the support for CLI script on boot as those configuration changes will be applied before the CLI boot scripts so the server might not be fully configured at that time (like for example in WildFly s2i images).
+
+== Issue Metadata
+
+=== Issue
+
+* https://issues.redhat.com/browse/WFCORE-5343[WFCORE-5343]
+* https://issues.redhat.com/browse/WFCORE-6503[WFCORE-6503]
+
+=== Related Issues
+
+* https://issues.redhat.com/browse/EAP7-1677[EAP7-1677]
+* https://issues.redhat.com/browse/EAP7-1877[EAP7-1877]
+* https://issues.redhat.com/browse/EAP7-1922[EAP7-1922]
+* https://issues.redhat.com/browse/WFLY-13978[WFLY-13978]
+* https://issues.redhat.com/browse/WFCORE-5324[WFCORE-5324]
+
+=== Dev Contacts
+
+* mailto:{email}[{author}]
+* mailto:jdenise@redhat.com[Jean-François Denise]
+
+=== QE Contacts
+
+* mailto: mnovak@redhat.com[Miroslav Novak]
+
+=== Testing By
+// Put an x in the relevant field to indicate if testing will be done by Engineering or QE. 
+// Discuss with QE during the Kickoff state to decide this
+* [ ] Engineering
+
+* [X] QE
+
+=== Affected Projects or Components
+
+`wildfly-core` and `wildlfy-jar-maven-plugin`.
+`wildfly-s2i` images may have issues since the server configuration is going to be modified after the yaml file have been applied.
+
+=== Other Interested Projects
+
+=== Relevant Installation Types
+// Remove the x next to the relevant field if the feature in question is not relevant
+// to that kind of WildFly installation
+* [x] Traditional standalone server (unzipped or provisioned by Galleon)
+
+* [] Managed domain
+
+* [] OpenShift s2i
+
+* [] Bootable jar
+
+== Requirements
+
+WildFly will be able to take a list of YAML files from the command line to redefine its exisiting configuration.
+
+The operations will be applied in the order the files are defined and after the initial operations defined by the XML configuration.
+The YAML configuration should be *idempotent* so that we can start and restart with the same command line without failure nor resulting with a different final configuration.
+The YAML structure should follow the management model structure, the YAML tree should match the management model resource tree.
+
+We may use YAML local tags to support operations syntax.
+So at boot, WildFly will load and parse its XML configuration, then add all the required extensions. It will then parse the YAML files, looking for a `wildfly-configuration` element as root for the configuration updates.
+
+For each entry, it will check if the resource is already created by the XML, otherwise it will add a new operation to the list of boot operations to add the required resource.
+If the resource already exists then it will add a new operation to the list of boot operations to update the attributes.
+
+If we support more operations (like delete) then we will update the current list of operation to reflect that. This means that if an attribute is updated in a YAML file, then the resource is deleted in a second one then the first operation won't be executed.
+
+This feature differs from what was done in https://issues.redhat.com/browse/WFCORE-5324[WFCORE-5324] because while we are customizing the configuration, the result is not persisted.
+Please note that the scripts would be applied after the YAML files so those changes might be done on a server that is not in its final configuration state (like for example in WildFly s2i images). In order to address that, if we detect that we are running in admin-mode with cli scripts to be applied we won't apply the YAML changes until the reload.
+The changes are applied on boot and not after an initial boot, and thus don't require a *reload*, thus we should have better performance.
+Also we don't provide the logic semantic that you have with the Jboss CLI scripts but on the other hand since we don't persist the resulting configuration you can reuse the same command and you don't have to check for pre-existing resources.
+
+=== Hard Requirements
+
+* Support for multiple files.
+External files will take precedence over the XML configuration ones. Then order will follow the command line: 
+- If the resource doesn't already exist: add the resource with all the specified attributes of the YAML section.
+- If the resource is already defined then all we have to do is update its attributes using the `write-attribute` operation.
+- If the YAML part defines an empty resource and the resource already exists: log a WARNING, otherwise try to add it.
+- If the YAML is incorrect: like defining an non-existing attribute or child resource we should fail.
+
+* Undefine attribute: using a YAML tag !undefine. For example:
+----
+wildfly-configuration:
+    subsystem:
+        logging:
+          console-handler:
+            CONSOLE: 
+              named-formatter: COLOR_PATTERN
+              level: !undefine
+----
+Using a null value is not 'doable' as we have attribute where the 'null' value is significant. This is the proper way to 'remove' an attribute value.
+
+* Allow setting null if a value is empty and the attribute is null-significant.
+* Allow support to remove resources using a YAML tag !remove. For example:
+----
+wildfly-configuration:
+    subsystem:
+        microprofile-jwt-smallrye: !remove
+----
+The remove operation would remove all the resource children related operations. Thus removing a subsystem will ensure that any operation related to a resource of said subsystem won't be applied at all during the boot, even those defined beforehand. Of course if the resource is defined afterwards it will get added again.
+Note that the `!remove` operation is only for resources and not for attributes. If you want to 'remove' an attribute the proper way is to `!undefine` it.
+
+
+* Allow support for adding elements to list and supporting indexes.
+A YAML sequence of the elements to be added is expected.
+----
+  permission-set:
+    default-permissions: 
+      permissions: !list-add 
+        - class-name: org.wildfly.transaction.client.RemoteTransactionPermission
+          module: org.wildfly.transaction.client
+          target-name: "*"
+          index: 0
+----
+* Paths can be absolute, relative to the current execution directory or relative to the standalone configuration directory.
+* Support *unmanaged* deployments as those don't have an impact on anything except the configuration. 
+* Managed deployments are not supported and any description of such a deployment must fail.
+* Allow to define a size limit for YAML files using `-Dorg.wildfly.configuration.extension.yaml.codepoint.limit=`.
+
+=== Non-Requirements
+
+The goal of the YAML files is to be able to customize an existing configuration. It is not here to replace the existing configuration support with XML. As such we won't support part of the management model.
+
+Only those elements would be *supported*:
+
+- core-service
+- interface
+- socket-binding-group
+- subsystem
+- system-property
+- deployment: to add *unmanaged* deployments to the server.
+
+That means that at least those entries would be *ignored*:
+
+ - extension: to add extension to the server as this might require modules which can be missing.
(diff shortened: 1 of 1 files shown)


core/WFCORE-5343_Support_YAML_files_for_configuration_customization.adoc
```diff
@@ -0,0 +1,144 @@
...
+So at boot, WildFly will load and parse its XML configuration, then add all the required extensions. It will then parse the YAML files, looking for a `wildfly-configuration` element as root fo the configuration updates.
+For each entry, it will check if the resource is already created by the XML, otherwise it will add a new operation to the list of boot operations to add the required resource.
+If the resource already exist then it will add a new operation to the list of boot operations to update the attributes.
+
+=== Hard Requirements
+
+* Support for multiple files.
+External files will take precedence over the XML configuration ones. Then order will follow the command line.
+
+* If the resource is already defined then all we have to du is updating its attrbiutes using the `write-attribute` operation.
```
jmesnil, line 75: typo s/du/do
s/attrbiutes/attributes

core/WFCORE-5343_Support_YAML_files_for_configuration_customization.adoc
```diff
@@ -0,0 +1,144 @@
...
+* If the YAML part define an empty resource and the resource already exists: log a WARNING, otherwise try to add it.
```
jmesnil, line 76: I suppose that the `:add` operation will include all the attributes specified in the YAML section, right?

core/WFCORE-5343_Support_YAML_files_for_configuration_customization.adoc
```diff
@@ -0,0 +1,144 @@
...
+----
+wildfly-configuration:
+    subsystem:
+        logging:
+          console-handler:
+            CONSOLE: 
+              named-formatter: COLOR_PATTERN
+              level: !undefine
+----
+Using a null value is not 'doable' as we have attrbiute where the 'null' value is significant.
```
jmesnil, line 91: s/attrbiute/attribute

core/WFCORE-5343_Support_YAML_files_for_configuration_customization.adoc
```diff
@@ -0,0 +1,144 @@
...
+
+* Undefine attribute: using a YAML tag !undefine. For example:
+----
+wildfly-configuration:
+    subsystem:
+        logging:
+          console-handler:
+            CONSOLE: 
+              named-formatter: COLOR_PATTERN
+              level: !undefine
```
jmesnil, line 89: aiui a YAML tag is used to identified a type. Is that syntax valid YAML?

core/WFCORE-5343_Support_YAML_files_for_configuration_customization.adoc
```diff
@@ -0,0 +1,144 @@
...
+
+* Allow setting null if value is empty and attribute is null-significant.
+* Allow support to remove resource using a YAML tag !remove. For example:
+----
+wildfly-configuration:
+    subsystem:
+        microprofile-jwt-smallrye: !remove
+----
+
+* Allow support for adding element to list and supporting index: this might impact idempotence.
```
jmesnil, line 101: how so? If we don't persist the standalone XML resulting from the YAML configuration, we would always add the element from the same list at boot time.

core/WFCORE-5343_Support_YAML_files_for_configuration_customization.adoc
```diff
@@ -0,0 +1,144 @@
...
+
+=== Non-Requirements
+
+The goal of the YAML files is to be able to customize an existing configuration. It is not here to replace the existing configuration support with XML.
+As such we won't support part of the management model. Those elements would be filtered out:
```
jmesnil, line 106: We should rather use an allow list than a disallow list to filter elements.

* `subsystem`
* `system-properties`
* `socket-binding-group`
* `interface` 
* something else?

core/WFCORE-5343_Support_YAML_files_for_configuration_customization.adoc
```diff
@@ -0,0 +1,144 @@
...
+ - deployment-overlay: to add deployment-overlays to the server as this require more that just some configuration.
+ - path: since those should already have been defined when the YAML files are parsed.
+ 
+As this configuration extension is for a *standalone server* only this *won't be supported in domain mode*.
+
+Because we need to support expressions, it migt be complex to get a JSON schema describing the YAML files for IDE completion. Also such a schema would be really big and would be really difficult to maintain or verify.
+
+== Implementation Plan
+
+If we are processing YAML files then the server is in *read-only* mode: that makes idempotence easy to acheive. We could persist the resulting configuration after boot so that we wouldn't be in read-only mode but then what is the purpose of the YAML files after the 1st execution ?
```
jmesnil, line 119: s/acheive/achieve

core/WFCORE-5343_Support_YAML_files_for_configuration_customization.adoc
```diff
@@ -0,0 +1,144 @@
...
+What we are going to do is use the boot operations to define what to do with what is described in the YAML files.
+In the YAML tree we can check if the node is matching a resource or an attribute.
+If it is a resource then we check if there are boot operations for this address, if such is the case then we have to process the attributes, otherwise that means that we need to create the resource.
+To create a resource we will look into the YAML subtree for the attributes that are defined, and add an `add` operation for it.
+If the resource is created then we look for it attributes that are writeable in the YAML subtree and add a `write-attribute` operation for each of them. That means that we need to have every element in a list since we are not using `list-add` in this case.
+Then we process the sub-resources if any exist.
+We should be able to get the effective XML file to validate the resut of our YAML configuraiton.
```
jmesnil, line 127: s/configuraiton/configuration

jmesnil, line 127: when are you planning to validate the XML? against which schema?

core/WFCORE-5343_Support_YAML_files_for_configuration_customization.adoc
ehsavoie: Well it depends on persisting the resulting configuration or not. It wasn't clear in the beginning, this is no longer the case.

core/WFCORE-5343_Support_YAML_files_for_configuration_customization.adoc
ehsavoie: No, I meant to use the :read-config-as-xml operation to get the effective XML.

core/WFCORE-5343_Support_YAML_files_for_configuration_customization.adoc
```diff
@@ -0,0 +1,164 @@
...
+:author:            Emmanuel Hugonnet
+:email:             ehugonne@redhat.com
+:toc:               left
+:icons:             font
+:idprefix:
+:idseparator:       -
+
+== Overview
+
+The aim of this proposal is to be able to inject external configuration in a YAML format to WildFly in order to customize an existing configuration.
```
mchoma, line 11: We see overlap between this feature and https://issues.redhat.com/browse/WFCORE-5324.

Can you elaborate in proposal what does it add compared to https://issues.redhat.com/browse/WFCORE-5324. Is it just different format? I assume underneath both cli and yaml ways end in DMR so behaviour, starting times and limitations would be same.

Can you elaborate on that as well?

ehsavoie: It is different in that the operations from the YAML files are executed on boot in addition to the one from the XML file. So this means that we should have better performance, we don't require a reload of the server and we don't change the resulting XML configuration since we are in "read-only" mode on that part.

ehsavoie: Also this isn't a script: no if/else etc. 
Also you can reuse the same command line in a loop without having to have the logic in your script to manage existing resources.
Last but not least, this is for all versions of WildFly: bootable and bare metal.

mchoma: Can I achieve `better performance` and `no reload requirement` and `idempotency` with CLI as well? Or is it consequence of limited scope of yaml files? Still I assume `CLI way` must be superset of `yaml way`, because in the end yaml must use same capabilities as CLI. Am I missing something?

So I have cli way which is applicable everywhere; bootable jar, bare metal, openshift, domain. And I can express anything in CLI. 

I will have yaml file with much more limited usage (just bootable jar, bare metal, just subset of CLI capabilities). 

Why should I use yaml file? When I will be writing yaml files I will need to try cli command before transform it into yaml snippet anyway. Isn't it better to directly compose cli script?

I agree yaml file is easy to read. But something similar can be achieved with formatting of CLI commands. ( I know it is not yaml :) )
```
/authentication=classic/login-module=Remoting:add( \
    code=Remoting, \
    flag=optional, \
    module-options=[ \
        password-stacking=useFirstPass \
    ] \
)
```

ehsavoie: It would work on OpenShift ;-) Just domain is not supported.
You won't test using CLI script, in the same way you don't test using CLI to write or configure a feature in Galleon.
CLI script requires a reload: so boot-time is longer. Like I wrote the cost is almost just the cost of parsing the YAML.
In your case you are missing the test to check if the resource already exists. Also you would have to create all the tree, while here we would create them for you.

jfdenise: @mchoma, yaml syntax, in the context of bootable JAR, is to help transitioning from Thorntail to WildFly Bootable JAR. The object model it exposed can be seen as simpler than CLI syntax. CLI implies a start in admin and a reload. We apply CLI commands (that are more than DMR) on top of a running server. Yaml will be more efficient. Having both is powerful. The order of application is: 1) yaml files 2) CLI script. With CLI you can call some management operations not exposed by YAML. I suspect that yaml will be much more used than CLI with bootable JAR. 
I don't know of its usage outside of bootable JAR. If yaml is usable outside of Bootable JAR, we could have an issue with Openshift s2i images. The YAML being applied first, the server would be not yet fully configured by the CLI script generated by our launch scripts (that currently relies on the same mechanism as the one used in WFCORE-5324).

rpelisse:
>I agree yaml file is easy to read. But something similar can be achieved with formatting of CLI commands. ( I know it is not yaml :) )

That's a big issue for integration into automation tool (such as Ansible). The CLI is not idempotent and thus we have customers writing pages and pages of CLI script while they could express their configuration using this features.

>I don't know of its usage outside of bootable JAR.

I'm part of an initiative to better integrate middleware product into Ansible (one use case being helping customers to replace their automation based on JON by Ansible) and this feature would help lot. Currently, we use another tool, called JCliff, to help Ansible managed Wildfly or EAP. It's quite cumbersome and far less efficient than this feature.

mchoma:
> The CLI is not idempotent 
Can be with `if` conditions. But is more talkative for sure.

> this feature would help a lot.
Will you expose in ansible yaml as is? Wouldn't there be any "facade" in ansible playbooks which can hide CLI talkativness? Isn't it problem for you yaml syntax is subset of CLI? Wouldn't you end in situation you will have to use CLI underneath anyway in some cases? My point is don't you need yaml have same power as CLI when used in Ansible?

rpelisse:
@mchoma I built a prototype based on @ehsavoie work and I have yet to find an use case where I need to use the CLI directly (apart from app deployment). I simply deploy the YAML configuration as a template, so I don't really need a facade. I actually find the direct mapping between CLI syntax and the YAML very comfortable. I also think people using Ansible and used to its YAML will be happy to the template has a close syntax.

> But is more talkative for sure.

I think you have largely underestimated the issue here. If I take my prototype, here is what I configured : 
- adding a new JDBC driver (postgresql)
- deploying a new datasource
- configuring mod_cluster
- adding a JMS queue
- set some system properties

Pretty basic, right? With this feature, all of this is configured in one place within Ansible : the template file producing the YAML configuration. The config is loaded and executed when the server boots in a matter of milliseconds.

If the configuration evolves after the installation was performed, Ansible will detect the changes in the resulting file produced by the template, so Ansible will trigger a server restart to apply the change.

Now, let's do the same with CLI. First, I need to wrap CLI into a somewhat reusable Ansible task (akin to [that](https://github.com/ansible-middleware/ansible_collections_jcliff/blob/master/roles/wildfly_common_criteria/tasks/jboss_cli.yml)), then I need to call this generic task for each item... This requires several exchange between Ansible and the server and takes way more time than what this feature offer. And let's not even discussed what happens if one query fails and dealing with the configuration that requires server reload...

OK, let's say you have done of all of that and that is somehow working for the installation part. Now someone comes and change the configuration. Now, you need to make CLI script able to detect that there is a differences between the requested configuration and the state of the server....

So it's not more talkative. It's pages of custom scripts versus one configuration file.

ehsavoie:
Also I would say that the YAML is less verbose than a cli script as you only describe some configuration instead of describing the operations to be applied to create the final configuration. It is a very different perspective.

core/WFCORE-5343_Support_YAML_files_for_configuration_customization.adoc
```diff
@@ -0,0 +1,174 @@
...
+What we are going to do is use the boot operations to define what to do with what is described in the YAML files.
+In the YAML tree we can check if the node is matching a resource or an attribute.
+If it is a resource then we check if there are boot operations for this address, if such is the case then we have to process the attributes, otherwise that means that we need to create the resource.
+To create a resource we will look into the YAML subtree for the attributes that are defined, and add an `add` operation for it.
+If the resource is created then we look for it attributes that are writeable in the YAML subtree and add a `write-attribute` operation for each of them. That means that we need to have every element in a list since we are not using `list-add` in this case.
+Then we process the sub-resources if any exist.
+We should be able to get the effective XML file (using the `read-config-as-xml` operation) to validate the resut of our YAML configuration.
```
xstefank, line 156:
```suggestion
We should be able to get the effective XML file (using the `read-config-as-xml` operation) to validate the result of our YAML configuration.
```

xstefank (approved):
LGTM

zhantemirov:
The AD is complete enough for the Jira's current state (Planning).
Maybe, it would be useful to have having one little thing (backward compatibility) described - Darran has touched this topic in [his commentary](https://issues.redhat.com/browse/EAP7-1877?focusedId=22901504&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-22901504).

@ehsavoie WDYT?

Otherwise LGTM.

ehsavoie:
@zhantemirov updated as not supported

mnovak1 (commented):
LGTM, just typo/formatting problems.

mnovak1 (approved):

core/WFCORE-5343_Support_YAML_files_for_configuration_customization.adoc
```diff
@@ -0,0 +1,255 @@
...
+ * YamlExtensionTestCase#testSimpleYamlWithReload checks that the YAML changes are not persisted.
+ * YamlExtensionTestCase#testSimpleYamlWithCliBootOps: checks that YAML and cli scripts are working properly together.
+ * YamlExtensionTestCase#testSimpleYaml#testDeploymentYaml: checks that adding unmanaged deployments via YAML works.
+ * YamlExtensionTestCasetestServerStartFailedForManagedDeployment: checks that managed deployments can't be added via YAML.
+ * YamlExtensionTestCase#testAddingExtensionPathDeploymentOverlayYamlLogsWarnings: checks that path, extension and deployment-overlay can't be added via YAML.
+
+
+== Community Documentation
+
+https://www.wildfly.org/news/2022/04/26/YAML-configuration-extension/
```
yersan, line 242: @ehsavoie I would say this should be https://github.com/wildfly/wildfly/pull/17970. having the article linked is good as well

yersan (approved):

yersan:
Thanks to everyone involved in this

PR #386, opened 2021-04-19
[WFLY-14347] Allow to configure module for custom validation classes

tadamski:
https://issues.redhat.com/browse/WFLY-14347
https://issues.redhat.com/browse/EAP7-1189

### DIFF
--- a/jca/WFLY-14347_Allow_to_configure_module_for_custom_validation_classes.adoc
+++ b/jca/WFLY-14347_Allow_to_configure_module_for_custom_validation_classes.adoc
@@ -1,0 +1,72 @@
+= WFLY-14347 Allow to configure module for custom validation classes
+:author:            Tomasz Adamski
+:email:             tadamski@redhat.com
+:toc:               left
+:icons:             font
+:keywords:          connector, validation, custom modules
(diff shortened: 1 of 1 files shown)


darranl:
Hi @tadamski just taking a look at your analysis some of the sections from the template seem to be missing https://github.com/wildfly/wildfly-proposals/blob/main/design-doc-template.adoc

I wanted to ask what you thought about the testing requirements for this, do you think test cases can be developed in the WildFly testsuite for this to enable us to mark as tested by engineering or do you think this may need more involved collaboration with QE?

msve:
Hey I checked the analysis and it looks good from QE perspective. Since the additional capability is relatively minor (allowing something already existing to be loaded from modules), I don't think additional QE tests are required, the tests included in the main PR for the RFE will do (though it would be nice to have it documented a bit, and it will still required a test plan).

tadamski:
@darranl @msve I have updated the proposal to cover the tests - I have provided an integration test as a part of PR (https://github.com/wildfly/wildfly/pull/14598). I also added the note about community docs which I believe are not needed.

bstansberry (approved):

PR #387, opened 2021-04-21
Add a section to the analysis template to emsure security considerations are identified early.

darranl:


jamezp (approved):
This looks good and makes sense to me.

bstansberry (approved):

rhusar:
@darranl Is this intended to be more of a reminder to self to consider the security implications of a feature or is this content meant to be consumed by someone? Regardless, this is a good idea.

darranl:
@rhusar Overall I think it will be a combination.

Firstly as you say it acts as a reminder to double check security implementations as a new feature is developed, I am going to add some additional guidance in this repo for the kinds of things to think about but not intended to be a verbose check list as each feature will be different.

Regarding consumption of the information anyone else involved in the feature team should also consider this information, during testing we may want to verify any assumptions and during documentation this my identify areas to call out specifically related to security.  Finally we have had some features in the past where the security team come along at the very end and ask "have you considered X or Y?" which can end up delaying a feature being merged so we also want to try and get involved with any security review / questions early to get that kind of thing out of the way.

PR #390, opened 2021-05-04
WFCORE-4868 Introduce aliases for standard configuration files

michpetrov:
Issue: [EAP7-1700](https://issues.redhat.com/browse/EAP7-1700) / [WFCORE-4868](https://issues.redhat.com/browse/WFCORE-4868)

### DIFF
--- a/controller/WFCORE-4868_configuration-files-aliases.adoc
+++ b/controller/WFCORE-4868_configuration-files-aliases.adoc
@@ -1,0 +1,95 @@
+---
+categories:
+ - server
+---
+= [Community] Aliases for standard configuration files
+:author:            Michal Petrov
+:email:             mpetrov@redhat.com
+:toc:               left
+:icons:             font
+:idprefix:
+:idseparator:       -
+
+== Overview
+
+Introduce aliases for standard configuration files
+
+This will allow users to reference the configuration by the suffix or a shorter name instead of typing out the filename.
+
+.Full list of aliases:
+|===
+|  standalone.xml variant | Alias
+
+| *-full.xml              | full
+
+| *-ha.xml                | ha
+
+| *-full-ha.xml           | fha, full-ha
+
+| *-load-balancer.xml     | lb, load-balancer
+
+| *-microprofile.xml      | mp, microprofile
+
+| *-microprofile-ha.xml   | mpha, microprofile-ha
+|===
+
+== Issue Metadata
+
+=== Issue
+
+* https://issues.redhat.com/browse/WFCORE-4868
+
+=== Stability Level
+* [ ] Experimental
+
+* [ ] Preview
+
+* [x] Community
+
+* [ ] default
+
+=== Dev Contacts
+
+* mailto:{email}[{author}]
+* mailto:yborgess@redhat.com[Yeray Borges]
+* mailto:ehugonne@redhat.com[Emmanuel Hugonnet]
+
+=== QE Contacts
+
+=== Testing By
+* [x] Engineering
+
+* [ ] QE
+
+=== Affected Projects or Components
+WildFly
+
+=== Other Interested Projects
+
+=== Relevant Installation Types
+* [x] Traditional standalone server (unzipped or provisioned by Galleon)
+
+* [ ] Managed domain
+
+* [ ] OpenShift s2i
+
+* [ ] Bootable jar
+
+== Requirements
+
+=== Hard Requirements
+
+Provide a set of aliases for the files distributed with WildFly that can be used with the commandline argument `--server-config|-c` instead of of a full filename.
+
+=== Nice-to-Have Requirements
+
+=== Non-Requirements
+
+== Test Plan
+Unit test will be added to server.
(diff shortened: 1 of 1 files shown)


controller/WFCORE-4868_configuration-files-aliases.adoc
```diff
@@ -0,0 +1,98 @@
+---
+categories:
+ - controller
+---
+= Aliases for standard configuration files
```
yersan, line 5: @michpetrov We need to add the  [Community] word here in the feature title, that will help when this proposal is published on wildfly.org

controller/WFCORE-4868_configuration-files-aliases.adoc
```diff
@@ -0,0 +1,98 @@
...
+
+| *-microprofile-ha.xml  | mpha
+|===
+
+== Issue Metadata
+
+=== Issue
+
+* https://issues.redhat.com/browse/WFCORE-4868
+* https://issues.redhat.com/browse/EAP7-1700
```
yersan, line 39: The EAP7 reference should be removed from this doc, since this is community RFE

yersan:
> It is not clear to me where this aliases list live.
So far it is only available for the configuration files we provide, so it is hardcoded in the code, maybe they could be extracted in an automatic and predictable way, but that is not available for the first cut yet.

> i would love to make it editable so I could add my own aliases :) Also does it only concerns sandlone configuration files ? 
That could be an evolution of this one. Right now is only prepared to supply aliases of the configuration files we provide by default. Those are the standalone variants.

controller/WFCORE-4868_configuration-files-aliases.adoc
```diff
@@ -0,0 +1,100 @@
...
+| *-full-ha.xml          | fha
+
+| *-load-balancer.xml    | lb
+
+| *-microprofile.xml     | mp
+
+| *-microprofile-ha.xml  | mpha
+|===
+
+Special values for `-c` remain in effect, e.g. `-c boot` references the standalone.boot.xml file (if present) and it is not possible to reference a custom file named `standalone-boot.xml` this way.
```
yersan, line 35: I am more inclined to keep the `boot` (and `initial`, v* and so on) untouched as they are. They should not be treated as possible aliases. If I use `-c boot`, this `boot` should not be treated as an alias, it should have the same meaning as before.

controller/WFCORE-4868_configuration-files-aliases.adoc
```diff
@@ -0,0 +1,100 @@
...
+* [ ] Preview
+
+* [x] Community
+
+* [ ] default
+
+=== Dev Contacts
+
+* mailto:{email}[{author}]
+
```
yersan, line 55: @michpetrov This section describes the feature team. 

Please add you as a developer, me as SME, and Emmanuel Hugonnet as an outside perspective person.

controller/WFCORE-4868_configuration-files-aliases.adoc
michpetrov: That's why I'm specifying that `boot` _cannot_ be used as an alias for a file named "standalone-boot.xml"

yersan: Ahh, got it @michpetrov , I wrongly read this statement, thanks!

yersan:
> Updated.
> 
> Just to clarify and perhaps I need to make the distinction clearer. I want this to work for the file suffixes, and a set of strings that are meant for the files we distribute.
> 
> I.e. "standalone-load-balancer.xml" can be referenced by "lb" (which will be hardcoded, and we could make this extensible in the future) but it could also be referenced by "load-balancer" (which is not hardcoded).

@michpetrov You are proposing to also enable other aliases besides the ones that can be applied to the files we distribute. So if the user has `standalone-custom.xml` they can reference it using just `-c custom`. That's more versatile, but I don't see it too useful at all, instead of creating `standalone-custom.xml` you could create `custom.xml`, even more, if you plan to restrict all to [`standalone` as the base name](https://github.com/wildfly/wildfly-core/pull/5903#discussion_r1529138040). Creating the shorter file seems simpler and avoids any weird behavior when you have `standalone-custom.xml` and `custom.xml` both on your configuration file. In such a case, if you use `-c custom` it is a bit confusing the expected logic about which file will be finally picked up. I would rather go with the simplest option which would be to translate with aliases the files we distribute only. In the future, if the need arises, we can think of how to allow the users to create custom aliases. I have no strong opinion on that, though, but I am more inclined to treat only the files we distribute.

If we go as it is not, could you also clarify this case in the proposal and define what would be the final logic applied?

darranl:
FYI if we are targeting community the "Nice-to-have" requirements should be converted to non-requirements or moved to a future-work section

yersan (approved):

ehsavoie (approved):

bstansberry:
Thanks @michpetrov, @yersan and @ehsavoie!

PR #391, opened 2021-05-04
Proposal for WFCORE-5184, check server and deployment state at boot time

jfdenise:


bootable-jar/WFCORE-5184_Check_Boot.adoc
```diff
@@ -0,0 +1,100 @@
...
+* This new feature doesn't introduce a timeout to bound the server start time (that can be very long for large and complex application).
+* In case the server never reaches these states, the Bootable JAR doesn't take any decision and waits until the server is stopped (by any mean).
+* When the server is running the following checks are operated:
+** If boot errors are detected, the process abort
+** If a deployment exists and has a FAILED status, the process is aborted.
+* If no invalid conditions are met, the bootable JAR advertises that it is up and running.
+* If an invalid condition is met, the bootable JAR advertises the problem and exits (error code == 1).
+* It could happen that the server is aborting (Fatal server error), in this case, no action is taken by the Bootable JAR. 
+The server exits automatically (as done today).
+* Once the server has booted successfully, no more check is operated.
```
jamezp, line 78: I'm not sure I follow this. Do you mean it's extracted as normal?

Somewhat along these lines if the boot fails should the extracted server be deleted/cleaned up?

bootable-jar/WFCORE-5184_Check_Boot.adoc
```diff
@@ -0,0 +1,100 @@
...
+* Evolve wildfly-jar-maven-plugin tests with some new tests to cover cases where the JAR should be aborted by this new support.
+* Evolve QE testsuite with new tests.
+
+== Community Documentation
+
+The WildFly doc will cover the new argument. The argument is also auto-documented in the bootable JAR help content.
+
+== Release Note Content
+
+Not candidate for release notes.
```
jamezp, line 100: Given a change in the default behavior I wonder if we should just make a minor note here.

jfdenise: Good point, I added a note.

bootable-jar/WFCORE-5184_Check_Boot.adoc
jfdenise: I added a note to specify that the nominal cleanup is applied when the boot is aborted (actually System.exit(1) is called by the bootable JAR runtime so shutdown hook is called). 
The WIP btranch is there BTW: https://github.com/wildfly/wildfly-core/compare/master...jfdenise:WFCORE-5184

jfdenise: I also added a note that says that the deployment status is not monitored once the server has booted. This is what I meant

bootable-jar/WFCORE-5184_Check_Boot.adoc
```diff
@@ -0,0 +1,100 @@
...
+* [x] Bootable jar
+
+== Requirements
+
+The JAR is auto-checking the server and deployment states at boot time in order to abort the process in case of errors.
+
+=== Hard Requirements
+
+* The boot checks are enabled by default. This is a change of behavior that needs to be documented.
+* Introduce `--check-boot=[true|false]` in order to disable the checks.
```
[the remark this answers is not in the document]
jfdenise: You mean a -Dorg.wildfly.bootable.check.boot=true|false? I am wandering if tha tis really required if we have a --check-boot argument?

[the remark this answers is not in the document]
jamezp: That could work. What I more meant is something that could be stored in the config so something like:
```
/system-property=org.wildfly.bootable.check.boot:add(value=false)
```

I'd have to look at the WIP, which I will :), to see if this would even work. Not critical really just an idea.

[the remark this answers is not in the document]
jfdenise: Ah, I got it. I think that it would work, we could get this property prior to do the check.

bstansberry:
I haven't looked at this in detail, but I'll just say that this should be a general behavior of the server not specific to bootable jar.  I'm pretty sure I filed a WFCORE for that a year or two ago.

Perhaps bootable jar would have some subtly different behavior or the initial implementation in bootable jar will differ, but the overall long term design here should be general purpose.

I don't think the long (or short) term solution should involve the complex checks described here. The server boot should simply fail on any boot operation step failure, regardless of OperationContext.Stage in which that step executes. Currently it only fails for Stage.MODEL or capability resolution failures before the start of Stage.RUNTIME.

Note that admin-only needs different semantics, which may differ for bootable jar, although the KISS principle might suggest they should not differ. One use case for admin-only is to correct your server configuration. That's not really applicable to bootable jar but to treat bootable differently there'd have to be a significant use case where the general behavior is wrong.

jfdenise:
@bstansberry , thank-you. I linked 2 issues I found that seem to reflect your comment. Having that in the server would have a major impact on this proposal.

bstansberry:
I'll say a bit more about the long vs short term as I understand there is a desire for a solution for bootable jar in the short term while a more general fix may take more time.

First I think the user API (i.e. command line params) for this should be consistent with what we would do for https://issues.redhat.com/browse/WFCORE-436 and https://issues.redhat.com/browse/WFCORE-2219. That doesn't mean it has to account for everything that those would add, but whatever those would add should reuse what is done for this, and expand upon it if necessary. Most likely that means WFCORE-436 and this use the same param, and then WFCORE-2219 adds something.

In terms of the implementation of the check, a short term check would need to produce results consistent with what would happen in the long term, so the long term solution can be brought in later as an implementation detail.

I think an issue with the above would be --admin-only mode, which is more forgiving of errors, but my understanding of the proposal is this feature would not be valid with --admin-only. If disallowing that combination can be enforced, that can be ok, as that is a constraint that can be relaxed in the future without breaking compatibility.

jfdenise:
@bstansberry , I am wandering if we will manage to offer today a solution that would work on top of the future support without any changes in behavior. I am wandering if we should simply delay this feature until WildFly offers a solution for it. Bootable JAR will then benefit from it.

PR #397, opened 2021-06-04
[WFLY-14854] Add Support for OpenTelemetry tracing to WildFly

jasondlee:
https://issues.redhat.com/browse/WFLY-14854?

### DIFF
--- a/observability/WFLY-14854_add_opentelemetry_support.adoc
+++ b/observability/WFLY-14854_add_opentelemetry_support.adoc
@@ -1,0 +1,76 @@
+ = Add OpenTelemetry Tracing Support to WildFly
+:author:            Jason Lee
+:email:             jasondlee@redhat.com
+:toc:               left
+:icons:             font
+:idprefix:
+:idseparator:       -
+
+== Overview
+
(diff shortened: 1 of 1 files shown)


jasondlee:
@jmesnil, @bstansberry, @xstefank, @jamezp : I've started the analysis doc for adding Otel support to WildFly. It is a WIP at the moment, but feedback would still be appreciated. Please note and comment on the directory in which this doc is located. Brian and I weren't sure where, exactly, this should live, so we opted to put it here and get this group's feedback.

observability/WFLY-14854_add_opentelemetry_support.adoc
```diff
@@ -0,0 +1,83 @@
...
+== Requirements
+
+=== Hard Requirements
+
+* The server/domain administrator should be able to configure OpenTelemetry export options
+* Jakarta REST endpoints should be automatically traced
+* Adherence to the Observability Standard for Runtimes
+* Context propagation should work for both incoming and outgoing request (e.g., a `ClientRequestFilter`)
+* Use of the OpenTracing shim to allow existing OpenTracing apps to continue to function without modification
+* Removal of OpenTracing support?
```
xstefank, line 55: For at least the time we would like to claim MP compatibility we shouldn't remove the OpenTracing support. Note that this is not the current plan for at least after MP 5 which will switch to jakarta.* namespace. I like the idea with the shim that would have OTel underneath.

xstefank (approved):
I think the observability is a correct directory.

jasondlee:
This PR was targeted for the wrong repo, so I'm going to close it and create a new one with the correct target. All comments and have been addressed/incorporated.

PR #398, opened 2021-06-07
[WFLY-14869] galleon feature pack to support MicroProfile LRA

ochaloup:
https://issues.redhat.com/browse/WFLY-14869

### DIFF
--- a/transactions/WFLY-14869-microprofile-lra.adoc
+++ b/transactions/WFLY-14869-microprofile-lra.adoc
@@ -1,0 +1,129 @@
+= Support for MicroProfile LRA
+:author:            Ondrej Chaloupka
+:email:             ochaloup@redhat.com
+:toc:               left
+:icons:             font
+:idprefix:
+:idseparator:       -
+:keywords:          transaction, saga, lra, microprofile
+
+== Overview
+
+The goal of this RFE is providing the Galleon feature pack for WildFly
+which integrates the https://github.com/jbosstm/narayana/tree/master/rts/lra[Narayana implementation]
+of the https://github.com/eclipse/microprofile-lra[MicroProfile LRA specification].
+
+The MicroProfile LRA is a transaction processing pattern well suited for microservice development
+based on the saga pattern. The motivation of such processing is summarized in the https://github.com/eclipse/microprofile-lra/blob/1.0/README.adoc#motivation[MicroProfile Spec readme].
+
+
+== Issue Metadata
+
+=== Issue
+
+* https://issues.redhat.com/browse/WFLY-14869
+
+=== Related Issues
+
+* https://issues.redhat.com/browse/EAP7-1496
+
+=== Dev Contacts
+
+* mailto:{mkunwar@redhat.com}[Mayank Kunwar]
+* mailto:{ochaloup@redhat.com}[{author}]
+
+=== QE Contacts
+
+* mailto:{szhantem@redhat.com}[Sultan Zhantemirov]
+
+=== Testing By
+
+* [ ] Engineering
+* [X] QE
+
+
+=== Affected Projects or Components
+
+* Transactions
+* WildFly
+
+=== Other Interested Projects
+
+* MicroProfile
+* Galleon
+* WildFly Operator
+
+=== Relevant Installation Types
+
+* [x] Traditional standalone server (unzipped or provisioned by Galleon)
+* [ ] Managed domain
+* [ ] OpenShift s2i
+* [x] Bootable jar
+
+== Requirements
+
+=== Hard Requirements
+
+The feature pack will integrate the LRA specification as the Galleon feature pack.
+The Galleon layer will be available for provisioning.
+The provisioned feature pack will provide all functionality defined in LRA specification for users in WildFly.
+
+The feature pack will be split to two Galleon layers (i.e., two separate subsystems).
+
+The first layer provides a capability of the LRA participant - bundling the Narayana client library capable
+to parse the annotations from LRA specification and work with JAX-RS and non-JAX-RS bindings.
+
+The second layer provides the capability of the LRA coordinator - a server which provides
+a HTTP endpoints where application may register as participants and then the coordinator
+calls them back (via HTTP calls) when LRA finishes.
+
+The feature pack will be placed amongst other MicroProfile WildFly feature packs
+under https://github.com/wildfly/wildfly/tree/master/microprofile[the WildFly GitHub repository].
+
+Each MicroProfile spec is optional to allow for exclusion during provisioning.
+
+The LRA feature packs will be dependent on the `transactions` subsystem. The LRA uses the same object store types
+as the JTA/JTS transaction uses, it requires the recovery manager thread to be running and requires transaction
+specifics like unique node-identifier per running application server (WildFly).
+
+We need to provide a way how to list unfinished (as well failed) LRA records in the Narayana object store.
+As said before the LRA coordinator is based on the Narayana core (`transactions` subsystem within WildFly)
+the LRA records and JTA records shares the same storage and recovery processing.
+The LRA records will be listed under `/subsystem=transactions/log-store=log-store/lras=...`
+and the operations (e.g. a need for probing) will be consistent to what was develope for JTA transaction types
+under `/subsystem=transactions/log-store=log-store/transactions=...`.
+
+=== Nice-to-Have Requirements
+
+Create a quickstart on LRA on how to use LRA with WildFly for https://github.com/wildfly/quickstart
+
+=== Non-Requirements
+
+NONE
+
+== Test Plan
+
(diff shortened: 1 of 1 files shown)


ochaloup:
@mmusgrov @xstefank @jmfinelli I started the design doc for creating of the galleon feature pack for the MicroProfile LRA. Please share your opinions (or provide a PR on top of this :-) about the topic.

transactions/WFLY-14869-microprofile-lra.adoc
```diff
@@ -0,0 +1,105 @@
...
+
+* MicroProfile
+* Galleon
+
+=== Relevant Installation Types
+
+* [x] Traditional standalone server (unzipped or provisioned by Galleon)
+* [ ] Managed domain
+* [ ] OpenShift s2i
+* [ ] Bootable jar
```
xstefank, line 63: we should definitely target also at least bootable jar

transactions/WFLY-14869-microprofile-lra.adoc
```diff
@@ -0,0 +1,105 @@
...
+
+NONE
+
+== Test Plan
+
+1. Testing LRA functionality within WildFly
++
+The LRA functionality tests are part of the Narayana LRA codebase. The WildFly will be running with LRA TCK module (https://github.com/jbosstm/narayana/tree/5.11.2.Final/rts/lra/test/tck)
+and with the Narayana integration test module (https://github.com/jbosstm/narayana/tree/5.11.2.Final/rts/lra/test/basic).
+2. Testing the GAlleon layer provisioning
```
xstefank, line 93:
```suggestion
2. Testing the Galleon layer provisioning
```

xstefank (approved):

transactions/WFLY-14869-microprofile-lra.adoc
```diff
@@ -0,0 +1,106 @@
...
+:toc:               left
+:icons:             font
+:idprefix:
+:idseparator:       -
+:keywords:          transaction, saga, lra, microprofile
+
+== Overview
+
+The goal of this RFE is providing the Galleon feature pack for WildFly
+which integrates the https://github.com/jbosstm/narayana/tree/5.11.2.Final/rts/lra[Narayana implementation]
```
mmusgrov, line 13: Can change to subsequent versions if we need to update the narayana code?

transactions/WFLY-14869-microprofile-lra.adoc
mmusgrov: And will it be easy to target the other installation types once we get some feedback on its use in traditional standalone server?

mmusgrov (approved):
Looks great.

transactions/WFLY-14869-microprofile-lra.adoc
```diff
@@ -0,0 +1,106 @@
...
+
+=== Non-Requirements
+
+NONE
+
+== Test Plan
+
+1. Testing LRA functionality within WildFly
++
+The LRA functionality tests are part of the Narayana LRA codebase. The WildFly will be running with LRA TCK module (https://github.com/jbosstm/narayana/tree/5.11.2.Final/rts/lra/test/tck)
```
jmfinelli, line 92: [Some work](https://issues.redhat.com/browse/JBTM-3427) towards this has been already carried out but it is not completed yet

jmfinelli (approved):
Looks great to me as well :-)
Thanks @ochaloup

transactions/WFLY-14869-microprofile-lra.adoc
xstefank: I believe the feature pack should be reusable in the bootable jar. The bootable jar will be mostly preferred packaging for microservices-based EAP installations with XP. I don't think it would be wise to delay this work.

transactions/WFLY-14869-microprofile-lra.adoc
ochaloup: thanks

transactions/WFLY-14869-microprofile-lra.adoc
ochaloup: I believe the point about bootable jar is valid and it would be good to add it to the scope.

transactions/WFLY-14869-microprofile-lra.adoc
ochaloup: My point was to show where the LRA implementation can be found. I wanted to have a static link that will be working for anybody coming to the document in the next years despite of the changes in the Narayana repository.
The link can be changed to anything where the LRA code will be placed.

xstefank (approved):

transactions/WFLY-14869-microprofile-lra.adoc
mmusgrov: okay, except that [linking to master](https://github.com/jbosstm/narayana/tree/master/rts/lra) would server the same purpose.

ochaloup:
With the general agreement I think I can move it to standard PR. I hope to work on some PoC next week and then comment more.
As well I need to find somebody from QE to look at this but they are overhelmed with work for the EAP 7.4.0 release and will be available only after it's done.

transactions/WFLY-14869-microprofile-lra.adoc
ochaloup: I see, I have a different point of thinking but as it's better to use the `master` for you I'm changing it.

transactions/WFLY-14869-microprofile-lra.adoc
ochaloup: +1

bstansberry:
@ochaloup This is meant to be shipped in WF itself, yes? If so it goes in the 'wildfly' feature pack. A separate feature pack is used for things that have a distinct software lifecycle.

mmusgrov:
@bstansberry We want LRA is be available in standard WilfFly so we will want an entry for it under the microprofile directory.
We also want it to be a standard galleon layer so that users can include LRA when they provision their servers.
Is that what you mean by "wildfly feature pack"?

ochaloup:
I updated the text with information that this is a 'wildfly' feature pack which will go to WildFly GitHub repository as a other microprofile specifications - https://github.com/wildfly/wildfly/tree/master/microprofile. This is what @bstansberry asked about.

The second update is that we specify to split the functionality to two separate Galleon layers - one with functionality of the LRA client the other with functionality of the LRA coordinator.

xstefank:
@ochaloup shouldn't we be consistent with Quarkus and just do the participant? Just thinking aloud.

ochaloup:
@xstefank I see I was not thinking this way so far. In fact until now I thought that the Quarkus extension will be coupling the coordinator and the client. I do understand that the approach changed.

From your perspective, do you think we should not permit to run the coordinator and the client coupled in one JVM and having only one option for deployment which will be client in WildFly and the coordinator somewhere in a separate service which will be delivered as probably "a Quarkus based jar"/"docker image on top of jar"?

xstefank:
Personally, I don't have stong preference. I would like the decoupled approach when the coordinator is a standalone microservice and thus standalone problem but I can understand that if the drive comes from distributed JTA like transactions then you want to have coordinator in the service itself. Still the LRA participants and LRA coordinator are a decoupled concepts in my mind so I would suggest having a requirement to run coordinator externally like Jaeger, Prometheus, or Kafka.

mmusgrov:
We don't need to be consistent with quarkus and we definitely do need a subsystem that includes the coordinator.

transactions/WFLY-14869-microprofile-lra.adoc
```diff
@@ -0,0 +1,118 @@
...
+=== Hard Requirements
+
+The feature pack will integrate the LRA specification as the Galleon feature pack.
+The Galleon layer will be available for provisioning.
+The provisioned feature pack will provide all functionality defined in LRA specification for users in WildFly.
+
+The feature pack will be split to two Galleon layers (i.e., two separate subsystems).
+
+The first layer provides a capability of the LRA client - an application library that provides
+CDI annotations to manage the scope of the LRA and programmatic API to manage the LRAs.
```
mmusgrov, line 77: I don't think we want to support a programmatic API just yet, the MP LRA spec only provides annotation support, the programmatic API won't be available until spec version 1.x. Also the client API is for our internal use only and we must not encourage users to use it.

mmusgrov (changes requested):

transactions/WFLY-14869-microprofile-lra.adoc
ochaloup: Absolutely correct. I removed that statement from the proposal.

ochaloup:
I added the QE contact and added a paragraph that we need a way how to list unfinished LRA records.

xstefank:
I've taken over this RFE from Ondra. I will create a new analysis document.

PR #399, opened 2021-06-07
[WFLY-12610] adding recovery process operation and attributes to transaction subsystem

ochaloup:
https://issues.redhat.com/browse/WFLY-12610

### DIFF
--- a/transactions/WFLY-12610-transaction-recovery-process-operation.adoc
+++ b/transactions/WFLY-12610-transaction-recovery-process-operation.adoc
@@ -1,0 +1,137 @@
+= Adding management operations to process transaction recovery
+:author:            Ondrej Chaloupka
+:email:             ochaloup@redhat.com
+:toc:               left
+:icons:             font
+:idprefix:
+:idseparator:       -
+:keywords:          transactions,recovery
+
+== Overview
+
+The transaction subsystem obtains an operation to execute transaction recovery manually via JBoss CLI.
+On top of that the transaction subsystem obtains configuration settings which configures the transaction recovery processing
+to speed up resolving the prepared unfinished transactions from the Narayana object store.
+
+== Issue Metadata
+
+=== Issue
+
+* https://issues.redhat.com/browse/WFLY-12610[WFLY-12610]
+
+=== Related Issues
+
+* https://issues.redhat.com/browse/JBEAP-20714[JBEAP-20714]
+
+=== Dev Contacts
+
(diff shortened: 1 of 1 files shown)


transactions/WFLY-12610-transaction-recovery-process-operation.adoc
```diff
@@ -0,0 +1,137 @@
...
+|operation
+|Execute the transaction recovery processing. Runtime operation.
+
+|orphan-safety-interval
+|attribute
+|Defines a time after which the record at remote resource without counterpart in the object store is considered as orphan and rolled-back, in milliseconds. Runtime, no reload needed.
+
+|transactions-enabled
+|attribute
+|When true a new transaction can be started, when false starting a new transaction is forbidden. Runtime, no reload needed.
```
mmusgrov, line 91: I would be wary of adding a `transactions-enabled` operation since that could be abused or inadvertently configured. If you embed or move the functionality into another operation that is sensitive to the starting of new transactions then that would be safer. Is there another operation where that could be done?

Otherwise the feature looks useful.

mmusgrov (commented):
I had just one concern, otherwise the proposal looks good.

transactions/WFLY-12610-transaction-recovery-process-operation.adoc
ochaloup: Thanks for the feedback. I see, having the special attribute to disable starting transaction could not be appropriate in all contexts.
I would propose to have a chance to disable starting new transactions when the server goes suspended - i.e. adding a new attribute to `suspend` operation, like `:suspend(transactions-enabled=false)`

ochaloup: @mmusgrov I discussed the possibilities of the `:suspend` operation with @kabir and I found that the `:suspend` operation is a generic WF core operation that is not permitted to be used directly from WildFly subsystems.
I would use a similar way of having a special attribute which activates the action only on suspend.
It would mean to rename the attribute to `transactions-enabled-on-suspend` (would you have a better name?) with value `true/false`. The functionality will be to disable transactions only when `:suspend` is called and enabled when container is resumed.
What do you think about such approach?

mmusgrov: I think the approach is good (and the attribute name describes its intent).

mmusgrov (approved):
Thanks

PR #400, opened 2021-06-15
[WFCORE-5464] Upgrade DMR to be able to resolve expressions from env …

kabir:
…vars

https://issues.redhat.com/browse/WFCORE-5464

### DIFF
--- a/management/WFCORE-5464_Check_Env_Vars_On_Expression_Resolution.adoc
+++ b/management/WFCORE-5464_Check_Env_Vars_On_Expression_Resolution.adoc
@@ -1,0 +1,104 @@
+= Check Environment Variables When Resolving Expressions
+:author:            Kabir Khan
+:email:             kkhan@redhat.com
+:toc:               left
+:icons:             font
+:idprefix:
+:idseparator:       -
+
+== Overview
+As the cloud use-cases rely more on container environment variables vars than system properties, we should allow for transparent resolution of properties from env vars as well as the current system property mechanism.
+
+For example, taking following xml
+
+[source, xml]
+----
+<element some-attribute="${test.my-property:default}"/>
+----
+
+At present you need to start the server with `-Dtest.my-property=hello` (or via a management model entry for the system property) to replace the value at runtime. This RFE proposes that if the env var `TEST_MY_PROPERTY=hola` is set when starting the server, we may also use the value of the environment variable. To keep this similar to how MP Config does the same thing, the system property will take precedence over the environment variable if both are defined.
+
+
+== Issue Metadata
+
+=== Issue
+
+* https://issues.redhat.com/browse/WFCORE-5464[WFCORE-5464 - Add environment variables as a source for model expression resolution]
+
+=== Related Issues
+
+* https://issues.redhat.com/browse/DMR-45[DMR-45 - Resolve env variables without requiring a env. prefix]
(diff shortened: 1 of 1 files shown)


management/WFCORE-5464_Check_Env_Vars_On_Expression_Resolution.adoc
```diff
@@ -0,0 +1,101 @@
...
+
+== Requirements
+
+=== Hard Requirements
+
+We will also check environment variables when resolving expressions, inspired by how MicroProfile Config (https://github.com/eclipse/microprofile-config/blob/2.0/spec/src/main/asciidoc/configsources.asciidoc[specification]) does it. However, we will do something a bit lighter weight.
+
+This means that if the expression can be resolved by both a system property and an environment variable, the system property takes precedence, as per the `Default ConfigSources` section of the MicroProfile Config specification. The other ConfigSources mentioned are not relevant for this use case.
+
+When converting between the expression name and the environment variable we will only use option 3 of the `Environment Variables Mapping Rules` from the MicroProfile Config specification. The aim is to avoid having to do too much string manipulation and too many lookups when resolving expressions, and to my knowledge environment variables available on OpenShift (which is the main target for this RFE) follow the `UPPERCASE_WITH_UNDERSCORES` pattern for environment variable names.
```
jmesnil, line 71: this naming convention is usually called SCREAMING_SNAKE_CASE 😆

jmesnil (approved):

vsalbaba:
I think there is some danger in reading an existing environment variable by mistake after upgrading.

If I have an environment variable `COMMON_VAR_NAME=foo` already in use
and I have `${common-var-name:bar}` in wildfly configuration,
and I expect the value to resolve to default,
when I upgrade to new wildfly version
then the value of `${common-var-name:bar}` will be sudenly `foo`

I think we should at least add warning about possible conflicts into the update notes or docs

kabir:
@vsalbaba Good catch, I have updated the documentation PR https://github.com/wildfly/wildfly/pull/14376, and added a "Other Documentation Concerns" to this analysis document

vsalbaba:
> I think there is some danger in reading an existing environment variable by mistake after upgrading.

I've had one more issue raised about this, it would be helpful to log which attributes were overridden from an environment variable.

This would be imho a nice solution to "how to debug that my attributes suddenly changed after wildfly update" problem.

kabir:
Regarding the log question, we've discussed a bit internally and feel that is a quite a corner case scenario. Turning on debug logging isn't really much easier than inspecting the environment, and that the documentation should suffice.

Also, the resolution happens in jboss-dmr which is a zero dependency library (its [pom](https://github.com/jbossas/jboss-dmr/blob/master/pom.xml)), and I doubt @dmlloyd would accept adding the jboss logging dependency to it.

dmlloyd:
I think it would probably be OK to add logging to this project via `jboss-logging`.  That's my usual one exception to having no dependencies.  Anyway it is likely that any run time environment which contains DMR would also contain `jboss-logging`.

kabir:
@jmesnil @dmlloyd @vsalbaba I've added the logging to https://github.com/wildfly/wildfly-proposals/pull/400 and updated https://github.com/wildfly/wildfly-core/pull/4633 to account for it

PR #403, opened 2021-06-25
Proposal for WFLY-14936, New WildFly S2I images

jfdenise:
Proposal for new WildFly s2i and runtime images.

### DIFF
--- a/openshift/WFLY-14936_New_S2I_Image.adoc
+++ b/openshift/WFLY-14936_New_S2I_Image.adoc
@@ -1,0 +1,234 @@
+= New WildFly S2I image
+:author:           Jean-Francois Denise
+:email:             jdenise@redhat.com
+:toc:               left
+:icons:             font
+:idprefix:
+:idseparator:       -
+
+== Overview
+
+Today, we are relying on the WildFly s2i builder image to build a WildFly application image. 
+There is a strong coupling inside this image that implies releasing a new image for each new WildFly release. 
+This approach is not flexible. We should be able to build an application image for any WildFly release 
+(even for a locally built SNAPSHOT) without having to wait for the WildFly s2i builder image to be deployed.
+
+In addition, the WildFly server located inside the builder image, enforces a set of configuration API (a set of environment variables) 
+and bash launch scripts. At startup time, a complex CLI script generation that impacts the server startup time is mandatorily executed. 
+Having this level of configurability should be a user choice and not be enforced by the WildFly s2i image.
+
+To solve these pain points, we are removing the server and its configuration aspects from the image. In order for the launch scripts and server configuration adjustments 
+to be offered, we are introducing a new cloud Galleon feature-pack to be provisioned in combination with WildFly Galleon feature-pack.
+
+The WildFly s2i image becomes a generic image allowing to properly install and execute any provisioned WildFly server.
+ As a developer one should be able in max 2 steps to deploy to the cloud an app. Same for WildFly developers that should 
+deploy a locally built WildFly SNAPSHOT in one step (as done with WildFly Bootable JAR today). 
+
+==== Responsabilities of the WildFly s2i builder image
+
+* Provide JDK and maven tooling
+* Provides automatic and configurable JVM configuration at server startup (same as the openjdk builder image)
+* Understand the execution constraints for the server and adapt the server launch processing (e.g.: start the server in a way that would allow a server killed by POD to terminate properly).
+* Provides nice defaults for the minimal set of server configuration items (e.g.: jboss.node.name derived from hostname, public interface address, management address)
+* Install the server in the image.
+* Support s2i build from source and binary.
+* Handle incremental build.
+* Built application Image must be runnable but also usable inside a docker chained build to construct 
+   a final application image (that doesn't contain any build tooling). 
+
+==== Responsabilities of the WildFly runtime image
+
+* Provides Java JDK. We are including the JDK in order for the image to contain JDK debug tooling.
+* Provides automatic and configurable JVM configuration at server startup (same as the openjdk builder image)
+* Understand the execution constraints for the server and adapt the server launch processing (e.g.: start the server in a way that would allow a server killed by POD to terminate properly).
+* Provides nice defaults for the minimal set of server configuration items (e.g.: jboss.node.name derived from hostname, public interface address, management address)
+
+==== A Maven plugin to provision and configure WildFly
+
+The steps that have been removed from the WildFly s2i build processing are now aggregated inside a 
+single point of configuration inside the WildFly Maven plugin that is evolved with https://issues.redhat.com/browse/WFMP-138[WFMP-138]. 
+The plugin handles Galleon provisioning, user CLI scripts execution, ability to enrich server content and application(s) deployment. 
+
+This plugin used in conjunction with this new WildFly s2i image is all what is needed to offer a flexible experience. 
+We gain the flexibility to provision/configure a WildFly instance to answer specific requirements instead of 
+relying on a fixed set of features packaged inside a WildFly s2i builder image.
+
+N.B.: The image is not strongly coupled with this Maven plugin. For source build, any plugin that outputs a server installation can be used during S2I source build.
+For binary build, any server installation can be the source of binary build.
+
+==== Legacy s2i workflow
+
+The image still supports s2i build that don't require the usage of the WildFly Maven plugin. This support is deprecated and could be removed in a 
+future version of the images. 
+
+Some `GALLEON_XXX` env variables have to be set to enable the legacy s2i workflow. 
+When the legacy workflow is enabled, the server is first provisioned then the application deployment built during s2i is copied to the provisioned server deployments directory.
+
+===== Galleon provisioning based on env variables
+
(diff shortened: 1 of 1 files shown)


jfdenise:
@jmesnil , S2I V2 proposal

openshift/WFLY-14936_New_S2I_Image.adoc
```diff
@@ -0,0 +1,204 @@
...
+is also done in the master branch.
+Once the deprecated images are no more released, a legacy branch will be created, the v2 branch will be merged into master. At this point, no more "legacy" content will be released.
+
+=== Hard Requirements
+
+==== WildFly s2i builder image
+
+===== Building
+
+* The builder image is built from ubi8-minimal image.
```
jmesnil, line 125: as the image is runnable, would it make sense to base it instead on `openjdk11` + maven tooling?
This would be quite symmetric with the runtime image:
* WildFly S2I Builder Image -> based on openjdk
* WildfFly S2I Runtime image -> based on openjdk-jre

openshift/WFLY-14936_New_S2I_Image.adoc
```diff
@@ -0,0 +1,204 @@
...
+* Image execution can be configured with Java related environment variables. https://github.com/jboss-openshift/cct_module/tree/master/jboss/container/java/jvm/api[JVM API]
+* The image entry-point offers the following features:
+** Discover the WildFly launcher script to be called: ```standalone.sh, standalone-cloud.sh or openshift-launch.sh```. 
+N.B.: In case of openshift-launch.sh (legacy entry point), no configuration is applied. Configuration of the server is delegated 
+to the ```openshift-launch.sh``` script that handles it all.
+** Computes the JVM options to be conveyed to the server
+** Handles POD termination with CLI shutdown. Can be disabled with ```CLI_GRACEFUL_SHUTDOWN=true``` env variable.
+** Set default value for management (0.0.0.0) and public (value of ```hostname -i``` command) interfaces. 
+    The default values can be specified with ```SERVER_PUBLIC_BIND_ADDRESS``` and  ```SERVER_MANAGEMENT_BIND_ADDRESS```.
+** Enables the server statistics. Can be overridden with ```SERVER_ENABLE_STATISTICS=false```.
```
jmesnil, line 148: once https://issues.redhat.com/browse/WFCORE-5464 is merged, we will be able to do that "natively" in WildFly using the `WILDFLY_STATISTICS_ENABLED` env var instead

jmesnil (commented):
that looks great!

openshift/WFLY-14936_New_S2I_Image.adoc
```diff
@@ -0,0 +1,204 @@
...
+* The runtime image is built from openjdk JRE 11 image.
+* Has the same versioning that the S2I Builder image.
+* It is deployed on quay.io: ```quay.io/wildfly/wildfly-runtime-jre11```
+* It contains the user jboss
+* Image produced from this runtime image is runnable.
+* It contains the same entry-point present in the builder image.
+* An example of Docker file to install a server inside the runtime image:
+
+```
+FROM wildfly/wildfly-runtime-ubi8:dev
```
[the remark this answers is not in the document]
jfdenise: Thanks, fixed.

openshift/WFLY-14936_New_S2I_Image.adoc
```diff
@@ -0,0 +1,204 @@
...
+COPY --chown=jboss:root vanilla-wildfly $JBOSS_HOME
```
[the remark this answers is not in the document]
jfdenise: I added a chained build example. The servetr in the builder image is /opt/wildfly, tmp/target has been deleted during s2i build.

openshift/WFLY-14936_New_S2I_Image.adoc
```diff
@@ -0,0 +1,204 @@
...
+
+To solve these pain points, we are removing the server and its configuration aspects from the image. 
+
+The WildFly s2i image becomes a generic image allowing to properly install and execute any provisioned WildFly server.
+ As a developer one should be able in max 2 steps to deploy to the cloud an app. Same for WildFly developers that should 
+deploy a locally built WildFly SNAPSHOT in one step (as done with WildFly Bootable JAR today). 
+
+==== Responsabilities of the WildFly s2i builder image
+
+* Provide JDK and maven tooling
```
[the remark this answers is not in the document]
jfdenise: I think that it would be hard to achieve. We are using cct modules and s2i modules for the s2i framework. We could investigate it but my guess is that it would not come in this RFE.

openshift/WFLY-14936_New_S2I_Image.adoc
```diff
@@ -0,0 +1,204 @@
...
+* Provide automatic and configurable JVM configuration at server startup (same as the openjdk builder image)
```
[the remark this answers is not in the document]
jfdenise: Correct, added a runtime responsibilities chapter.

openshift/WFLY-14936_New_S2I_Image.adoc
```diff
@@ -0,0 +1,204 @@
...
+* Understand the execution constraints for the server and adapt the server launch processing (e.g.: start the server in a way that would allow a server killed by POD to terminate properly).
```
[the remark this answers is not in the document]
jfdenise: Correct, added a runtime responsibilities chapter.

openshift/WFLY-14936_New_S2I_Image.adoc
```diff
@@ -0,0 +1,204 @@
...
+* Provide nice defaults for the minimal set of server configuration items (e.g.: jboss.node.name derived from hostname, public interface address, management address)
```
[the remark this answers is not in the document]
jfdenise: Correct, added a runtime responsibilities chapter.

openshift/WFLY-14936_New_S2I_Image.adoc
jfdenise: Openjdk builder image has its own implementation of S2I that can't be overridden in a child image.

openshift/WFLY-14936_New_S2I_Image.adoc
jfdenise: That is a good point. Do we want the statistics enabled by default when running in the image? So have -Dwildfly.statistics-enabled=[true by default but overridden by SERVER_ENABLE_STATISTICS] passed in all cases?
If that is not the case, we can remove this "default" and let the user uses the SERVER_ARGS env variable to provide it explicitly (explicitly SERVER_ARGS=-Dwildfly.statistics-enabled=true) or use WFCORE-5464.
I had the feeling that we wanted it by default in cloud context.

openshift/WFLY-14936_New_S2I_Image.adoc
```diff
@@ -0,0 +1,228 @@
...
+
+==== Responsabilities of the WildFly s2i builder image
+
+* Provide JDK and maven tooling
+* Provides automatic and configurable JVM configuration at server startup (same as the openjdk builder image)
+* Understand the execution constraints for the server and adapt the server launch processing (e.g.: start the server in a way that would allow a server killed by POD to terminate properly).
+* Provides nice defaults for the minimal set of server configuration items (e.g.: jboss.node.name derived from hostname, public interface address, management address)
+* Install the server in the image.
+* Support s2i build from source and binary.
+* Support custom S2i build operated from JKube (see: https://issues.redhat.com/browse/WFLY-14936[WFLY-14938]).
```
bstansberry, line 34: This link is wrong; doesn't point to WFLY-14938

openshift/WFLY-14936_New_S2I_Image.adoc
```diff
@@ -0,0 +1,228 @@
...
+```
+
+Chained build from builder image
+```
+FROM quay.io/wildfly/wildfly-runtime-jre11:latest
+COPY --from=builder --chown=jboss:root /opt/wildfly $JBOSS_HOME
+RUN chmod -R ug+rwX $JBOSS_HOME
+```
+
+==== WildFly Cloud legacy Galleon feature-pack
```
mchoma, line 190: @jfdenise Imagine I have existing s2i applications relying on environment properties. How exactly should look like migration to this new way? Is there somewhere any example I can have a look? With my experience from bootable jar approach I assume I will have to adjust pom.xml of application to build server using maven plugin. In that pom.xml I will have feature pack for wildfly server version which I want to use and also `wildfly-cloud-legacy-galleon-pack` as I want to be backward compatible and use environment properties as I was used to.

If my assumptions are right, are those app changes applicable by PM as there is big focus on easy migration between versions. Ideally without app changes. Or is it something which can be covered by migration tool?

jfdenise: @mchoma , yes, what you describe is what a user would need to do. You can see an example: https://github.com/wildfly/wildfly-s2i/blob/v2/test/test-app-clustering/pom.xml
Then behave tests that uses env variables: https://github.com/wildfly/wildfly-s2i/blob/v2/wildfly-builder-image/tests/features/jgroups.feature#L4
This does require a change in application pom.xml.

PR #404, opened 2021-06-25
[WFCORE-5489] As a developer, I want to override management attribute…

jmesnil:
… values using environment variables

JIRA: https://issues.redhat.com/browse/WFCORE-5489

Signed-off-by: Jeff Mesnil <jmesnil@redhat.com>

### DIFF
--- a/management/WFCORE-5489_override_attribute_value_from_env_var.adoc
+++ b/management/WFCORE-5489_override_attribute_value_from_env_var.adoc
@@ -1,0 +1,161 @@
+= WFCORE-5489 - As a developer, I want to override management attribute values using environment variables
+:author:            Jeff Mesnil
+:email:             jmesnil@redhat.com
+:toc:               left
+:icons:             font
+:idprefix:
+:idseparator:       -
+
+== Overview
+
+WildFly on the cloud relies on container environment variables to customize its configuration to the container platform.
+
+It is difficult to change the value of a management attribute from the standalone configuration if the attribute does not provide a default value with an expression that can be resolved at runtime.
+
+For example, the user wants to update the `proxy-address-forwarding` attribute on the `/subsystem=undertow/server=default-server/http-listener=default` resource.
+
+It is not possible to achieve this with our existing configuration as the value is not specified in standalone.xml and defaults to `false`.
+If we want to allow our users to change this attribute, we must explicitely set its value with an expression that defaults to its default value:
+
+[source]
+----
+/subsystem=undertow/server=default-server/http-listener=default:write-attribute(name=proxy-address-forwarding, value=${undertow.default.http-listener.proxy-address-forwarding:false)
+----
+
+[NOTE]
+====
+This is an actual use case as we are changing the value of this attribute when a Bootable jar is targeting a cloud deployment in https://github.com/wildfly-extras/wildfly-jar-maven-plugin/blob/master/plugin/src/main/resources/org/wildfly/plugins/bootablejar/maven/cloud/openshift-undertow-script.cli 
+====
+
+With that change, the user can now use the expression resolution to change the value of any attributes.
+
+However this implies for *any* attribute that can be modified by the user at deployment time to explicity set its value to a resolvable expression with a default.
+This is not achievable while keeping our configuration readable and maintainable.
+
+Instead this RFE proposes to introduce a generic mechanism to let user override any __simple__ management attribute using environment variables. Complex attributes (of type `OBJECT`, `LIST`, `PROPERTY`) are not impacted by this change.
+
+For the same example above, the user could set an env var when starting WildFly to change the value of this `proxy-address-forwarding` attribute with the env var `SUBSYSTEM_UNDERTOW_SERVER_DEFAULT_SERVER_HTTP_LISTENER_DEFAULT__PROXY_ADDRESS_FORWARDING`
+
+[source]
+----
+$ export SUBSYSTEM_UNDERTOW_SERVER_DEFAULT_SERVER_HTTP_LISTENER_DEFAULT__PROXY_ADDRESS_FORWARDING=true
+$ ./bin/standalone.sh
+----
+
+If he runs WildFly on the Cloud, he needs to set this *env var* on its pod container without any modificaiton to the application image.
+
+This RFE will simplify the integration of services with WildFly with information that is only known during deployment (such as URLs, credentials, etc.)
+
+== Resource and attribute mapping
+
+The mapping is straightforward. The name of the env var is based on the address of the resource and the name of the attribute:
+
+1. take the address of the resource
+** `/subsystem=undertow/server=default-server/http-listener=default`
+2. remove the leading slash (`/`)
+** `subsystem=undertow/server=default-server/http-listener=default`
+3. append *two underscores* (`__`) and the name of the attribute:
+** `subsystem=undertow/server=default-server/http-listener=default__proxy-address-forwarding`
+4. Replace all non-alphanumeric characters with an underscore (`_`) and put it in upper case
+** `SUBSYSTEM_UNDERTOW_SERVER_DEFAULT_SERVER_HTTP_LISTENER_DEFAULT__PROXY_ADDRESS_FORWARDING`
+
+
+== Issue Metadata
+
+=== Issue
+
+* https://issues.redhat.com/browse/WFCORE-5489[As a developer, I want to override management attribute values using environment variables]
+
+=== Related Issues
+
+* https://issues.redhat.com/browse/WFCORE-5464[WFCORE-5464 - Add environment variables as a source for model expression resolution]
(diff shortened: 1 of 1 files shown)


jmesnil:
@kabir fiy, as it relates to https://issues.redhat.com/browse/WFCORE-5464 which also improves the configuration flexibility of WildFly

@jfdenise fiy, as this could help simplify some of our cloud deployment shenanigans (eg. the `<cloud>` CLI scripts as mentioned as an use case)

management/WFCORE-5489_override_attribute_value_from_env_var.adoc
```diff
@@ -0,0 +1,138 @@
...
+====
+This is an actual use case as we are changing the value of this attribute when a Bootable jar is targeting a cloud deployment in https://github.com/wildfly-extras/wildfly-jar-maven-plugin/blob/master/plugin/src/main/resources/org/wildfly/plugins/bootablejar/maven/cloud/openshift-undertow-script.cli 
+====
+
+With that change, the user can now use the expression resolution to change the attribute value.
+
+However this implies for *any* attribute that can be modified by the user at deployment time to explicity set its value to a resolvable expression with a default.
+This is not achievable while keeping our configuration readable and maintainable.
+
+Instead this RFE proposes to introduce a generic mechanism to let user override *any* management attribute using environment variables.
```
yersan, line 35: Would these overridden values coming from the environment values be persisted on the server configuration files?

I'm not sure how this is going to be implemented, but keep in mind that the configuration scripts use bash to read existing configurations from the server configuration files. I would say any overrides will be done always after configuring the server via shell scripts because if they are done before without being persisted in the server configuration files, the script execution will never take them into account. Maybe this situation deserves a hard requirement describing it.

jmesnil: The idea behind this feature is to affect the in-memory management model without affecting its storage.

The main use case is for cloud deployment where the configuration is read once when the pod is started.
However, I'll clarify the requirements when we use it on bare metal.

jmesnil: @yersan I've clarified the requirement. Could you please review it?

yersan: Thanks @jmesnil , yes, I think the hard requirement about the cloud deployment makes it more clear.

The launch scripts will read the initial server configuration without taking into account values overridden by these env variables. The environment variables will override the configuration generated by the launch scripts.

management/WFCORE-5489_override_attribute_value_from_env_var.adoc
```diff
@@ -0,0 +1,152 @@
...
+
+== Test Plan
+
+Tests will be added to the WildFly Core testsuite to ensure that overriding an attribute value with an environment variable works as expected.
+
+Tests will also verify the mapping between the (resource, address) tuple and the name of the environment variable to look up.
+
+== Community Documentation
+
+The WildFly Admin and Developer Guides will be enhanced to showcase the new mechanism.
```
bstansberry, line 127: Where?

bstansberry (changes requested):

management/WFCORE-5489_override_attribute_value_from_env_var.adoc
bstansberry: Ok, https://github.com/wildfly/wildfly/pull/14418 tells me where.

management/WFCORE-5489_override_attribute_value_from_env_var.adoc
```diff
@@ -0,0 +1,152 @@
...
+
+Everytime an attribute value is validated and set in the management model, we look if there is a corresponding environment value that overrides it *before* doing any expression resolution or correction.
+
+This change will be done in `org.jboss.as.controller.AttributeDefinition#validateAndSet`. Any attribute definition that does not use this method to validate and set their values will not be taken into account.
+
+This feature is activated by a presence of an environment variable: `WILDFLY_OVERRIDING_ENV_VARS`.
+
+As the target of this feature is application images running on the cloud, we will make it the default behaviour for the S2I images by setting this `WILDFLY_OVERRIDING_ENV_VARS` in the builder and runtime images.
+
+For cloud deployment with Bootable Jar, the users will have to set this env var in addtion to any env var that will overrides attribute values.
```
bstansberry, line 110: I think noting some items related to persistence would be good:

1) If an attribute value is determined from an environment variable, the next time the configuration is persisted, that value from the environment variable will be persisted.
2) Until something triggers persistence of the configuration file (which may never happen), the configuration file will not reflect the current running configuration.

These aren't really 'requirements', they are more a description of the intended behavior, such that doing something else is *not* a requirement.  I'll add an explicit related non-requirement below.

management/WFCORE-5489_override_attribute_value_from_env_var.adoc
```diff
@@ -0,0 +1,152 @@
...
+
+=== Nice-to-Have Requirements
+
+=== Non-Requirements
+
+* This feature is not activated by default in other installation types than OpenShift S2I
+* It is not possible to override management attributes value using Java System Properties.
```
bstansberry, line 117: * It is not a requirement that an attribute value being determined from an environment variable triggers persistence of the configuration in order to make its state match the current running configuration.
* Following from this it is not a requirement that files in the configuration history dir reflect environment variable values unless those history files record configuration state that was persisted for some other reason. For example, a standalone.xml.initial and standalone.xml.boot would not record information that wasn't in the config file that was read on the initial or most recent boot.

bstansberry (changes requested):
I've made some comments on this but they don't imply any changes to the feature implementation; they're more about spelling out what will happen.

management/WFCORE-5489_override_attribute_value_from_env_var.adoc
jmesnil: added your comment verbatim :)

management/WFCORE-5489_override_attribute_value_from_env_var.adoc
jmesnil: I've added some items about persistence

management/WFCORE-5489_override_attribute_value_from_env_var.adoc
```diff
@@ -0,0 +1,144 @@
...
+== Resource and attribute mapping
+
+The mapping is straightforward. The name of the env var is based on thethe address of the resource and the name of the attribute:
+
+1. take the address of the resource
+** `/subsystem=undertow/server=default-server/http-listener=default`
+2. remove the leading slash (`/`)
+** `subsystem=undertow/server=default-server/http-listener=default`
+3. append *two underscores* (`__`) and the name of the attribute:
+** `subsystem=undertow/server=default-server/http-listener=default__proxy-address-forwarding`
```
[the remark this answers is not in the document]
jmesnil: +1, I've explicitly mentioned this works for simple attributes only (and updated the wilfdly-core accordingly to verify the type of the attribute before overriding the value: https://github.com/wildfly/wildfly-core/blob/7c284dbf42d7931362699a1349e441bff20a92ca/controller/src/main/java/org/jboss/as/controller/AttributeDefinition.java#L515

management/WFCORE-5489_override_attribute_value_from_env_var.adoc
```diff
@@ -0,0 +1,144 @@
...
+$ ./bin/standalone.sh
+----
+
+If he runs WildFly on the Cloud, he needs to set this *env var* on its pod container without any modificaiton to the application image.
+
+This RFE will simplify the integration of services with WildFly with information that is only known during deployment (such as URLs, credentials, etc.)
+
+== Resource and attribute mapping
+
+The mapping is straightforward. The name of the env var is based on thethe address of the resource and the name of the attribute:
```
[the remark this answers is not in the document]
jmesnil: fixed, thanks

management/WFCORE-5489_override_attribute_value_from_env_var.adoc
```diff
@@ -0,0 +1,144 @@
...
+
+== Requirements
+
+=== Hard Requirements
+
+Everytime an attribute value is validated and set in the management model, we look if there is a corresponding environment value that overrides it *before* doing any expression resolution or correction.
+
+This change will be done in `org.jboss.as.controller.AttributeDefinition#validateAndSet`. Any attribute definition that does not use this method to validate and set their values will not be taken into account.
+
+The fact of populating the management model with the overridden attribute value implies that, going forward, the persisted value will be the overriden value.
```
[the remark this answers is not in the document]
jmesnil: I've added a item about it

jmesnil:
@bstansberry I've updated the analysis, doc and code based on your feedback.
The main change is that this feature is not explicitly working only for simple attributes. Complex attributes are not taken into account in the code. (the community doc and analysis PR have been updated according to this change in the implementation).

bstansberry (approved):

PR #405, opened 2021-06-27
[WFCORE-5466] Analysis for the removal of legacy vault integration from WildFly Core and WildFly.

darranl:
https://issues.redhat.com/browse/WFCORE-5466

### DIFF
--- a/security/WFCORE-5466_Remove_Legacy_Vault_Integration.adoc
+++ b/security/WFCORE-5466_Remove_Legacy_Vault_Integration.adoc
@@ -1,0 +1,312 @@
+= WFCORE-5466 Remove Legacy Vault Integration
+:author:            Darran Lofthouse
+:email:             darran.lofthouse@redhat.com
+:toc:               left
+:icons:             font
+:idprefix:
+:idseparator:       -
+
+== Overview
+
+At the start for JBoss AS7 the secure handling of credentials in the management model was handled
+by the PicketBox vault allowing expressions in the management model to be resolved against the 
+vault.
+
+During WildFly 11 with the addition of support for WildFly Elytron an alternative solution was
+added providing a credential store which can be used to resolve credentials.  This was later
+enhanced to also support the use of encrypted expressions in the management model in general.
+
+This task is focused on the removal of the PicketBox based vault integration.
+
+== Issue Metadata
+
+=== Issue
+
+* https://issues.redhat.com/browse/WFCORE-5466[WFCORE-5466 - Remove Vault support entirely from WildFly Core]
+
+=== Related Issues
+
+* https://issues.redhat.com/browse/WFCORE-5485[WFCORE-5485 - Bump the CLI configuration schema to version 4]
+* https://issues.redhat.com/browse/WFLY-14912[WFLY-14912 - Remove Vault support entirely from WildFly]
+* https://issues.redhat.com/browse/EAP7-1745[EAP7-1745 - Remove Vault support]
+* https://issues.redhat.com/browse/WFLY-14943[WFLY-14943 - Once all legacy security removal is in perform a full refactor]
+
+=== Dev Contacts
+
+* mailto:{email}[{author}]
+
+=== QE Contacts
+
+TBD
+
+=== Testing By
+// Put an x in the relevant field to indicate if testing will be done by Engineering or QE. 
+// Discuss with QE during the Kickoff state to decide this
+* [x] Engineering
+
+* [x] QE
+
+As a removal RFE this does not require new feature test development, however both teams will need
+to review existing testsuites for tests dependent on the Vault integration.  Identified tests will
+either need deleting or converting to a credential store equivalent.
+
+=== Affected Projects or Components
+
+* WildFly Core
+* WildFly
+
+=== Other Interested Projects
+
+* [x] External Testsuites
+
+* [ ] Quickstarts
+
+=== Relevant Installation Types
+
+* [x] Traditional standalone server (unzipped or provisioned by Galleon)
+
+* [x] Managed domain
+
+* [x] OpenShift s2i
+
+* [x] Bootable jar
+
+== Requirements
+
+As a removal task, this section is describing the steps being undertaken rather than a set of
+"requirements".
+
+=== Hard Requirements
+
+The command line vault utility scripts `vault.bat`, `vault.ps1`, and `vault.sh` are all removed
+from the server.
+
+The next version of the management model is updated to remove all vault resources.  Any use of
+vault resources in the latest version will result in a usual unexpected element error.
+
+Where configurations are read using older model versions if they contain a vault definition this
+will be logged as a warning and the vault configuration dropped at the next write.  This is to
+allow older configurations to still be handled in admin mode.
+
+Vault expressions are not prevented from occurring within the management model for two reasons:
+
+* The WildFly Elytron expression encryption can use a similar format.
+* Domains will need to handle configurations containing vault expressions for older host controllers.
+
+The vault elements in the WildFly CLI config are removed from the next version of the schema.  The
+presence of the element in ALL versions of configuration will result in a parsing error being
+reported.  The parsing is more abrubt in the CLI as this is only used for configuring SSL so will
+need immediate attention.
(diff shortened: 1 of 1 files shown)


security/WFCORE-5466_Remove_Legacy_Vault_Integration.adoc
```diff
@@ -0,0 +1,312 @@
...
+The vault elements in the WildFly CLI config are removed from the next version of the schema.  The
+presence of the element in ALL versions of configuration will result in a parsing error being
+reported.  The parsing is more abrubt in the CLI as this is only used for configuring SSL so will
+need immediate attention.
+
+All vault integration code is deleted from WildFly Core reducing the dependence on PicketBox.
+
+All vault test utilities are deleted and removed from the testsuite.
+
+* TODO - Describe impact on Galleon layers.
```
bstansberry, line 105: I believe the relationship to Galleon (in general, not just layers) is:

1) There are ee-feature-pack and servlet-feature-pack 'vault' layers (latter no longer relevant to the main WF dist). But these provide no configuration; they only exist to allow users to ensure that the picketbox package and its deps are provisioned in contexts where they otherwise would not be. So these could be removed altogether (the right thing to do I think, but would break configs that for some reason directly specified these layers), or just turned essentially into no-ops.

2) There's a vaulttools package used to ensure vault related scripts are provisioned. This can be dropped.

security/WFCORE-5466_Remove_Legacy_Vault_Integration.adoc
```diff
@@ -0,0 +1,311 @@
...
+This RFE is not quite sufficient to remove dependencies on PicketBox from WildFly Core.
+
+The legacy security subsystem also supports it's own vault definition, this is outside the
+scope of the task and will be handled with the removal of the subsystem.
+
+The core management vault resources are not transformed so configuration does not need to be
+retained for managing older host controllers.
+
+The following items identified during the removal of Vault integration will be handled
+independently in their own follow up tasks:
```
OndrejKotek, line 123: WFLY-14919 and WFCORE-5490 look like RFEs. Do we want to treat those as RFEs or will those be part of another RFE? Do we need/want those in Alpha?

darranl: I think both of those should be a follow up for later, at this point the priority is getting the old code out and then we can follow up.  We should probably clone both to JBEAPs for now with a target release of 8.0.0.GA so we have them in a single location and don't forget.

security/WFCORE-5466_Remove_Legacy_Vault_Integration.adoc
```diff
@@ -0,0 +1,311 @@
...
+
+This RFE is not quite sufficient to remove dependencies on PicketBox from WildFly Core.
+
+The legacy security subsystem also supports it's own vault definition, this is outside the
+scope of the task and will be handled with the removal of the subsystem.
+
+The core management vault resources are not transformed so configuration does not need to be
+retained for managing older host controllers.
+
+The following items identified during the removal of Vault integration will be handled
```
OndrejKotek, line 122: > Encrypted strings stored in the password vault can be used in an application’s source code.

Do we have/want this functionality with credential stores? Or, will it be part of WFLY-14919?

darranl: Where is that one coming from?  The credential store APIs are already supported so applications can initialise a credential store themselves if they want to access credentials. 

 Additionally they can use a wildfly-config.xml which uses a credential store for it's passwords.

security/WFCORE-5466_Remove_Legacy_Vault_Integration.adoc
```diff
@@ -0,0 +1,311 @@
...
+either need deleting or converting to a credential store equivalent.
+
+=== Affected Projects or Components
+
+* WildFly Core
+* WildFly
+
+=== Other Interested Projects
+
+* [x] External Testsuites
```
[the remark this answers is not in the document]
darranl: I am not aware of any, notification was sent out to the wildfly-dev list in June that this removal is happening.

security/WFCORE-5466_Remove_Legacy_Vault_Integration.adoc
OndrejKotek: https://access.redhat.com/documentation/en-us/red_hat_jboss_enterprise_application_platform/7.4/html/how_to_configure_server_security/securely_storing_credentials#use_an_encrypted_sensitive_string_in_an_application

darranl: I believe that would be a part of WFLY-14919 - i.e. as I understand it it is the DUPs in the EE subsystem that add vault handling into the deployment process should be for deployment descriptors and annotations i.e. they operate on the resulting meta data.

I have updated the Jira to make sure it includes annotations.

security/WFCORE-5466_Remove_Legacy_Vault_Integration.adoc
```diff
@@ -0,0 +1,311 @@
...
+
+=== Hard Requirements
+
+The command line vault utility scripts `vault.bat`, `vault.ps1`, and `vault.sh` are all removed
+from the server.
+
+The next version of the management model is updated to remove all vault resources.  Any use of
+vault resources in the latest version will result in a usual unexpected element error.
+
+Where configurations are read using older model versions if they contain a vault definition this
```
OndrejKotek, line 87: Can you plan and add test cases for these please?

security/WFCORE-5466_Remove_Legacy_Vault_Integration.adoc
```diff
@@ -0,0 +1,311 @@
...
+will be logged as a warning and the vault configuration dropped at the next write.  This is to
+allow older configurations to still be handled in admin mode.  This same approach will be used for
+appclient parsing.
+
+Vault expressions are not prevented from occurring within the management model for two reasons:
```
OndrejKotek, line 92: Do we plan to provide an automated way to migrate vault expression occurrences into credential store based values and encrypted expressions?

security/WFCORE-5466_Remove_Legacy_Vault_Integration.adoc
```diff
@@ -0,0 +1,311 @@
...
+|WildFly
+|org.jboss.as.test.integration.messaging.jms.definitions.JMSResourceDefinitionsTestCase
+|Reduced
+
+|WildFly
+|org.jboss.as.test.integration.messaging.jms.context.VaultedInjectedJMSContextTestCase
+|Reduced
+
+|WildFly
+|org.jboss.as.test.integration.security.vault.ExternalPasswordCommandsTestCase
```
OndrejKotek, line 243: I cannot find an alternative test case for credential stores. We should add a followup task for this.

OndrejKotek: The test case covers also usage of Java class as an external source of master password. There seems to be no alternative for credential store - do we want to provide such feature?

https://access.redhat.com/documentation/en-us/red_hat_jboss_enterprise_application_platform/7.4/html-single/how_to_configure_server_security/index#obtain_keystore_password_from_external_source `{CLASS[@jboss_module_spec]}classname[:ctorargs]`

security/WFCORE-5466_Remove_Legacy_Vault_Integration.adoc
```diff
@@ -0,0 +1,311 @@
...
+|org.jboss.as.server.deployment.DeploymentAddHandlerTestCase
+|Tweaked
+
+|WildFly Core
+|org.jboss.as.server.test.InterfaceManagementUnitTestCase
+|Tweaked
+
+|WildFly Core
+|org.jboss.as.server.parsing.SystemPropertiesParsingTest
+|Reduced
```
bstansberry, line 172: @darranl I think this should either be Removed or Tweaked (to use credential store exp). This test is for https://issues.redhat.com/browse/WFCORE-4736 and removing the expressions makes it pointless.

PR #408, opened 2021-07-07
[WFLY-14947] Create analysis document for Micrometer Support (Observability metrics policy)

jasondlee:
https://issues.redhat.com/browse/WFLY-14947

### DIFF
--- a/observability/WFLY-14947-Implement_observability_policy_metrics.adoc
+++ b/observability/WFLY-14947-Implement_observability_policy_metrics.adoc
@@ -1,0 +1,71 @@
+ = Implement Observability Policy for Metrics
+:author:            Jason Lee
+:email:             jasondlee@redhat.com
+:toc:               left
+:icons:             font
+:idprefix:
+:idseparator:       -
+
+== Overview
+
+While WildFly/EAP already supports basic metrics in the base WildFly/EAP system, as well as MicroProfile Metrics in the WildFly MicroProfile configuration and EAP XP, there is an effort in Red Hat Middleware Runtimes to implement a consistent observability policy across all Runtime offerings. In the Java space, the library of choice, and the emerging de facto standard for Java developers is https://micrometer.io[Micrometer].  This RFE will add support for Micrometer to WildFly/EAP and adhere to the policy decided upon by the various stakeholders.
+
+== Issue Metadata
+
+=== Issue
+
+* https://issues.redhat.com/browse/WFLY-14947
+
+=== Related Issues
+
+* https://issues.redhat.com/browse/EAP7-1686
+
+=== Dev Contacts
+
+* mailto:{email}[{author}]
+
+=== QE Contacts
+
+=== Testing By
+* [X] Engineering
+* [X] QE
+
+=== Affected Projects or Components
+
+=== Other Interested Projects
+
+=== Relevant Installation Types
+* [x] Traditional standalone server (unzipped or provisioned by Galleon)
+* [x] Managed domain
+* [x] OpenShift s2i
+* [x] Bootable jar
+
+== Requirements
+
+=== Hard Requirements
+
+* Feature parity with existing "base metrics"
+* Application registry available during injection
+* Prometheus endpoint that provides application and jvm statistics as allowed by RBAC (if enabled)
+* Deprecate/remove/replace existing base metrics?
+
+=== Nice-to-Have Requirements
+
+* MicroProfile Metrics-friendly API wrapper (similar to what Quarkus offers)
+
+=== Non-Requirements
+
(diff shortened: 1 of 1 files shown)


observability/WFLY-14947-Implement_observability_policy_metrics.adoc
```diff
@@ -0,0 +1,71 @@
...
+* [X] Engineering
+* [X] QE
+
+=== Affected Projects or Components
+
+=== Other Interested Projects
+
+=== Relevant Installation Types
+* [x] Traditional standalone server (unzipped or provisioned by Galleon)
+* [x] Managed domain
```
jmesnil, line 39: Please check but I don't think we support metrics on domain mode.

jmesnil (commented):
It's not clear to me what the scope of that RFE is.

Is it adding application metrics with MicroMeter and exposes them to the existing `metriccs` HTTP endpoint?
Or are you planning to upgrade the existing `metrics` subsystem and use MicroMeter underneath for all metrics (including JVM and WildFly management metrics)?

jasondlee:
That's a discussion I need to have with you and Brian, probably. My assumption so far is that this would replace the current "base metrics", in which case this document should make that clear. Now that Brian is back, I'll try to grab some time to get his take.

observability/WFLY-14947-Implement_observability_policy_metrics.adoc
jasondlee: > Please check but I don't think we support metrics on domain mode.

OK. That was probably a copy and paste from a template or prior document. Will verify if it's needed, but I think you're right.

bstansberry: If wanted that should be a separate RFE. It brings with a big set of different considerations that would need to involve a lot of different people.

observability/WFLY-14947-Implement_observability_policy_metrics.adoc
```diff
@@ -0,0 +1,71 @@
...
+* [x] OpenShift s2i
+* [x] Bootable jar
+
+== Requirements
+
+=== Hard Requirements
+
+* Feature parity with existing "base metrics"
```
bstansberry, line 47: Let's come up with some terms for different types of metrics and record them here.

In MP Metrics we have 3 types

1) base
2) vendor
3) application

"Base" there not referring to the non-MP subsystem, but to a (IIRC) spec defined set that all impls would expose. JVM stuff I believe. Is there some equivalent concept in Micrometer-land? If not overall, is there some common set that projects we cooperate closely with like Infinispan, Keycloak would use if they did Micrometer?

Regardless of the answer to that some concept to distinguish JVM metrics from others (at least for discussion) seems good.

"Vendor" then is things not under the control of the application developer. In our case there are two conceptual subcategories

a) Metrics exposed via the management API.
b) Metrics that components we integrate might want to expose via Micrometer that are not mediated by the management layer.

I think analysis terminology for all of these is needed, and we need requirements spelled out for how we will handle all of them.

observability/WFLY-14947-Implement_observability_policy_metrics.adoc
```diff
@@ -0,0 +1,71 @@
...
+* Application registry available during injection
+* Prometheus endpoint that provides application and jvm statistics as allowed by RBAC (if enabled)
+* Deprecate/remove/replace existing base metrics?
+
```
bstansberry, line 51: What requirements are there related to output format, particularly metric naming. AIUI what we do now for vendor metrics is not what is wanted, so we need requirements re what is wanted.

observability/WFLY-14947-Implement_observability_policy_metrics.adoc
jasondlee: Agreed. The term "base" is very confusing in a WildFly context, and I've to replace it. I don't have a suggestion off hand (other than maybe "system"). I'm not aware of anything that, say, Keycloak or Infinispan are using, but conversation with them (and others in Runtimes) makes sense,  I think, as we'd like our efforts to be rowing in the same direction.

observability/WFLY-14947-Implement_observability_policy_metrics.adoc
jasondlee: The observability policy simply states "Every runtime should expose a Prometheus endpoint", which would imply that we adhere to this: https://prometheus.io/docs/practices/naming/ As I understand things, Micrometer's Prometheus exporter covers some of that, though there are parts (e.g., "must have a single unit (i.e. do not mix seconds with milliseconds, or seconds with bytes)") that may be harder to enforce?

PR #410, opened 2021-07-09
Proposal for WFCORE-5500, Evolve CLI control flow

jfdenise:
Set of enhancements to remove limitations in CLI scripts to cope with complex and dynamic server management scenarios.

### DIFF
--- a/cli/WFCORE-5500_Evolved_Control_Flow.adoc
+++ b/cli/WFCORE-5500_Evolved_Control_Flow.adoc
@@ -1,0 +1,284 @@
+= CLI control flow improvements 
+:author:           Jean-Francois Denise
+:email:             jdenise@redhat.com
+:toc:               left
+:icons:             font
+:idprefix:
+:idseparator:       -
+
+== Overview
+
+Today the CLI suffers from limitations when using workflow commands (if/else,for and try/catch). These limitations don't allow to 
+write CLI scripts to deal with complex server configurations. With the introduction of Galleon, the content of server configuration is much more dynamic 
+and writing CLI scripts to adapt to the server configuration is required.
+
+For example, it is a common practice to check for the existence of multiple subsystems, to iterate array of resources and 
+take decisions according to the configuration (or env variables) state. Current CLI support handles such cases in a very limited way.
+
+We are here not defining a new programming language (that is really out of scope of the CLI command line utility) but we 
+are evolving existing commands to solve the major blocking points that forbid its usage in complex scenario.
+
+== Issue Metadata
+
+=== Issue
+
+* https://issues.redhat.com/browse/WFCORE-5500[WFCORE-5500]
+
+=== Related Issues
+
+* Completion issue with control flow commands: https://issues.redhat.com/browse/WFCORE-3489[WFCORE-3489]
+
+* Variables handling: https://issues.redhat.com/browse/WFCORE-5388[WFCORE-5388]
+
+=== Dev Contacts
+
+* mailto:{email}[{author}]
+
+=== QE Contacts
+
+* TBD
+
+=== Testing By
+// Put an x in the relevant field to indicate if testing will be done by Engineering or QE. 
+// Discuss with QE during the Kickoff state to decide this
+* [ ] Engineering
+
+* [ ] QE
+
+=== Affected Projects or Components
+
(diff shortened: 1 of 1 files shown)


jfdenise:
@yersan @kabir , FYI.

cli/WFCORE-5500_Evolved_Control_Flow.adoc
```diff
@@ -0,0 +1,284 @@
...
+          run-batch
+        done
+     done
+  end-if
+end-if
+```
+
+=== ```if``` command new operators for Array.
+
+* Add the ability to test the size of an array by using the operator ```#=```
```
jmesnil, line 96: out of curiosity is the operand `#=` based on an existing syntax?

cli/WFCORE-5500_Evolved_Control_Flow.adoc
```diff
@@ -0,0 +1,284 @@
...
+* Add the ability to test if an array contains a String element by using the operator ```*=```
```
jmesnil, line 97: how much are we limited syntactically? Could we name that operator `contains`?

```
if (result contains "ManagementDomain") of /subsystem=elytron:read-child-names(child-type=security-domain)
   /subsystem=elytron/security-domain=ManagementDomain:remove()
 end-if 
```

cli/WFCORE-5500_Evolved_Control_Flow.adoc
```diff
@@ -0,0 +1,284 @@
...
+```
+
+=== ```if``` command ModelNode path to take into account arrays.
+
+* ModelNode path takes into account array '['<index>']'.
+
+For example:
+
+```
+if result.realms[0].realm==ApplicationRealm of /subsystem=elytron/security-domain=ApplicationDomain:read-resource
```
jmesnil, line 114: should we always test is the size of the array is ok ?

cli/WFCORE-5500_Evolved_Control_Flow.adoc
```diff
@@ -0,0 +1,284 @@
...
+* A single variable can be used.
+* A path inside a variable can be used (eg: ```$foo.bar[0].attribute```).
+* The keyword ```this``` in a condition references the current variable.
+
+For example:
+
+```
+# An array of security domain names
+for domain in /subsystem=elytron:read-children-names(child-type=security-domain)
+  if this=="ApplicationDomain" of $domain
```
jmesnil, line 132: in that simple case, could it be `if $domain == "ApplicationDomain"` ?

jmesnil (commented):
Another enhancements I would like to see in the CLI is the ability to ignore a failed operation by appending `?` to the name of the operation.

This would greatly simply all the testing we do to make sure that resources exists before removing them.

So, instead of using the pattern

```
if (outcome != success) of /extension=foo:read-resource
  /extension= foo:remove()
end-if
```

We would have the following:

```
/extension=foo:remove?()
```

cli/WFCORE-5500_Evolved_Control_Flow.adoc
kabir: I see the idea, but also struggle a bit with it :) 
I think what I struggle a bit with is that we don't really know the order the results will be returned in. But that could just be the example which confuses me

jfdenise:
@jmesnil , the idea of the '?' to ignore failure is a great idea! I updated the proposal with this new feature, that, as you said, greatly simplify writing CLI commands.

cli/WFCORE-5500_Evolved_Control_Flow.adoc
jfdenise: It is inspired by bash that uses # as prefix to array to retrieve the size.

cli/WFCORE-5500_Evolved_Control_Flow.adoc
jfdenise: That is a limitation, we need special character for the operator.

cli/WFCORE-5500_Evolved_Control_Flow.adoc
jfdenise: @jmesnil , the condition would be evaluated to false if the array index is incorrect. Does that answer your point?

PR #422, opened 2021-09-01
Proposal for WFLY-14934, WildFly server Maven plugin

jfdenise:


maven/WFLY-14934_Server_Plugin.adoc
```diff
@@ -0,0 +1,155 @@
...
+=== Relevant Installation Types
+// Remove the x next to the relevant field if the feature in question is not relevant
+// to that kind of WildFly installation
+* [ ] Traditional standalone server (unzipped or provisioned by Galleon)
+
+* [ ] Managed domain
+
+* [x] OpenShift s2i
+
+* [x] Maven-controlled standalone server
```
jmesnil, line 62: Let's emphasise that this proposal *adds* a new type of installation

maven/WFLY-14934_Server_Plugin.adoc
```diff
@@ -0,0 +1,155 @@
...
+* extra-server-content-dirs: A list of directories to copy content to the provisioned server
+* feature-pack-location: The WildFly Galleon feature-pack location
+* feature-packs: A list of feature-pack configurations to install
+* hollow-server: Create a server that doesn’t contain application
+* layers: A list of Galleon layers to provision.
+* log-time: Whether to log provisioning time at the end
+* off-line: Whether to use offline mode when the plugin resolves an artifact
+* output-dir-name: By default the generated server is 'server'
+* plugin-options: Arbitrary Galleon options used when provisioning the server
+* project-build-dir: Project build dir.
```
jmesnil, line 102: what's the use of that configuration?

maven/WFLY-14934_Server_Plugin.adoc
```diff
@@ -0,0 +1,155 @@
...
+* provisioning-file: The path to the provisioning.xml file to use
```
jmesnil, line 103: is there a default path?

maven/WFLY-14934_Server_Plugin.adoc
```diff
@@ -0,0 +1,155 @@
...
+* record-state: Whether to record provisioned state in .galleon directory.
+* skip: To skip execution
+
+==== Plugin artifacts
+
+===== WildFly Server Maven Plugin
+
+The maven plugin GAV is ```org.wildfly.plugins:wildfly-server-maven-plugin:1.0.0.Final```.
+The source repository is expected to be ```http://github.com/wildfly-extras/wildfly-server-plugin```
```
jmesnil, line 112: should it be `http://github.com/wildfly-extras/wildfly-server-maven-plugin` to be consistent?

maven/WFLY-14934_Server_Plugin.adoc
```diff
@@ -0,0 +1,155 @@
...
+
+===== WildFly Bootable JAR and WildFly Server plugins shared artifacts
+
+In order to avoid duplicating some logic common with WildFly Bootable JAR Maven plugin, the ```org.wildfly.plugins:wildfly-jar-maven-plugin-common:1.0.0.Final``` and ```org.wildfly.plugins:wildfly-jar-maven-tests-common:1.0.0.Final```
+Maven artifacts are defined. 
+Both WildFly Bootable JAR and WildFly Server Maven plugins depend on these artifacts.
+The source repository is expected to be ```http://github.com/wildfly-extras/wildfly-jar-maven-plugin-common```
```
jmesnil, line 119: if we create a new common libray for `wildfly-server-maven-plugin` and `wildfly-jar-maven-plugin`, let's name it `wildfly-common-maven-plugin` instead of `wildfly-jar-maven-plugin-common`

Have we also considered having all these Maven modules inside a single GitHub repository:

```
+ http://github.com/wildfly-extras/wildfly-maven-plugins
   + common -> `org.wildfly.plugins:wildfly-comon-maven-plugin`
   + server -> `org.wildfly.plugins:wildfly-server-maven-plugin`
   + jar -> `org.wildfly.plugins:wildfly-jar-maven-plugin`
```

maven/WFLY-14934_Server_Plugin.adoc
```diff
@@ -0,0 +1,155 @@
...
+* Add new QE tests.
+
+== Community Documentation
+
+* A Maven plugin https://github.com/jfdenise/wildfly-server-maven-plugin/wiki[community documentation] will be defined.
+* Community examples will be defined.
+
+== Release Note Content
+
+Yes.
```
jmesnil, line 155: What would be the content of the release notes for this feature?

maven/WFLY-14934_Server_Plugin.adoc
jfdenise: @jmesnil you are right, un-needed configuration point.

maven/WFLY-14934_Server_Plugin.adoc
jfdenise: Yes, the wiki covers the detail: https://github.com/jfdenise/wildfly-server-maven-plugin/wiki#provisioningFile

maven/WFLY-14934_Server_Plugin.adoc
jfdenise: Typo, that is http://github.com/wildfly-extras/wildfly-server-maven-plugin

maven/WFLY-14934_Server_Plugin.adoc
jfdenise: In order to un-couple the life cycles of the plugins, I was thinking to separate the repositories and have each repo to have its own versioning and life cycle. Fixing a bug in bootable JAR shouldn't imply a bump in server plugin.
The bootable JAR is released for each new WildFly Major (mainly for examples but we have a bootable JAR runtime in WF that evolved sometime in sync with the plugin).
For the WildFly server plugin, I hope that we can uncouple the plugin life cycle from WF life cycle. We shouldn't release a new one for new WF Major. This plugin is very generic.

jamezp:
What is the difference between this and a combo of the `org.jboss.galleon:galleon-maven-plugin` and `org.wildfly.plugins:wildfly-maven-plugin`? It sounds like you could do the same thing with those two.

jfdenise:
@jamezp , yes, we could compose plugins to achieve a similar output. An integrated plugin would offer a simpler setup. Perhaps the main difference is the automatic deployment of the application?

jamezp:
@jfdenise It could be. I'm just wondering if we should either update the plugins or if we really want 4 separate plugins for everything.

maven/WFLY-14934_Server_Plugin.adoc
jmesnil: ok

jfdenise:
@jamezp , yes, integrating part of this support inside the wildfly-maven-plugin would reduce the number of artifacts. I have worked on an example that achieve the same results with 3 plugins: galleon, resources and wildfly-maven-plugin.
With the wildfly-server-plugin: https://github.com/jfdenise/wildfly-server-maven-plugin/blob/compose-plugins-example/examples/jaxrs/pom.xml
With the sequence of plugins: https://github.com/jfdenise/wildfly-server-maven-plugin/blob/compose-plugins-example/examples/jaxrs-legacy/pom.xml
The current state is that we can achieve the same result. The pom.xml configuration is more complex (perhaps more error prone), we would possibly need an "offline deploy" capability (to avoid calling deploy CLI command), and the ability to resolve or not expressions when calling CLI.
I think that the galleon maven plugin direct usage exposes more Galleon concepts that are more hidden with the "integrated plugin".

Perhaps evolving the existing wildfly-maven-plugin  instead of defining a new plugin, as you suggest, is the way to go. I realize that we would have an overlap between the wildfly-server-maven-plugin and wildfly-plugin (that is well known and largely used) that could be confusing.

jamezp (approved):
This looks good to me.

jmesnil (approved):

maven/WFLY-14934_Server_Plugin.adoc
```diff
@@ -0,0 +1,186 @@
...
+
+* cli-sessions: A list of CLI execution session. Each session can contain: a list of CLI scripts, a list of commands, a list of properties files, 
+a list of java options, the ability to resolve or not expressions locally (false by default) and system properties for the CLI forked process. 
+NB: CLI commands are automatically executed in the context of an embedded server.
+* extra-server-content-dirs: A list of directories to copy content to the provisioned server inside a forked process.
+* filename: The name of the artifact to deploy inside the project build directory. By default a lookup is done to retrieve the primary artifact to deploy.
+* name: The name of the deployment. By default the file name is used.
+* runtime-name: The runtime name of the deployment. By default the file name is used. NB: In order to deploy 
+the deployment inside the root context, ```ROOT.war``` runtime name is to be used.
+* server-config: The server config to use when executing CLI scripts and deploying artifact. Is ```standalone.xml``` by default if no ```server-groups``` have
```
jmesnil, line 142: I'm not sure to understand what it would mean to run `mvn package` with a `server-groups`. 

The first use case for this goal is in the cloud context where WildFly always runs in standalone mode. I wonder if focusing on that use case might not be better (and add support for domain mode when we widen the scope of this plugin)

jmesnil (commented):
TL;DR I'm still approving this proposal because that the right thing to do but I don't consider the list of options definitive. They will be better addressed in the code PR and then we will update this dev analysis.

I generated the doc for the wildfly-maven-plugin with @jfdenise PoC and I find the plugin configuration too confusing with its flat hierarchy.

As an example, we have 3 different ways to run JBoss CLI commands

* `<commands>` are run on a running WildFlyServer with the `execute-commands` goal
* `<scripts>` is a file containing CLI commands that are also run with the `execute-commands` goal
* `<cli-sessions>` is a file containing CLI commands that are run during the `package` goal when a server is provisioned.

I think we should either rename some of the new parameters to make them more explicit (eg. `<packaging-scripts>` instead `<cli-sessions>` or introduce some hierarchy in the configuration:

```
<configuration>
  <scripts>...</scripts> -> existing scripts executed during execute-commands
  <packaging>
    <scripts>...</scripts> -> scripts executed when a server is packaged
  </packaging>
</configuration>
```

In a similar note, I'm not sure we should stick with the `featurePacks` API from Galleon.
Do we really need to have users configure feature packs with the full set of options (coming from the `galleon-maven-plugin`? 
If there are more specific/complex use of Galleon when the server is provisioned, shouldn't the user switch to define a proper `provisioning.xml`?

maven/WFLY-14934_Server_Plugin.adoc
jmesnil: This is an existing parameter that would be impacted by the `package` goal. It's not clear to me how it interact with the `server-configuration/model & name`

maven/WFLY-14934_Server_Plugin.adoc
```diff
@@ -0,0 +1,186 @@
...
+a list of java options, the ability to resolve or not expressions locally (false by default) and system properties for the CLI forked process. 
+NB: CLI commands are automatically executed in the context of an embedded server.
+* extra-server-content-dirs: A list of directories to copy content to the provisioned server inside a forked process.
+* filename: The name of the artifact to deploy inside the project build directory. By default a lookup is done to retrieve the primary artifact to deploy.
+* name: The name of the deployment. By default the file name is used.
+* runtime-name: The runtime name of the deployment. By default the file name is used. NB: In order to deploy 
+the deployment inside the root context, ```ROOT.war``` runtime name is to be used.
+* server-config: The server config to use when executing CLI scripts and deploying artifact. Is ```standalone.xml``` by default if no ```server-groups``` have 
+been provided.  Is ```domain.xml``` by default if ```server-groups``` have been provided.
+* server-groups: Domain mode only, the list of server-group to use when deploying deployment.
```
jmesnil, line 144: This is an existing parameter that would be impacted by the `package` goal. It's not clear to me how it interact with the `server-configuration/model & name`

maven/WFLY-14934_Server_Plugin.adoc
```diff
@@ -0,0 +1,186 @@
...
+* stdout: Indicates how ```stdout``` and ```stderr``` should be handled for the forked CLI process.
```
jmesnil, line 145: `<stdout>` is not descriptive enough. Aiui, this controls Gallen output. Maybe `<galleon-output>` instead?

maven/WFLY-14934_Server_Plugin.adoc
```diff
@@ -0,0 +1,186 @@
...
+==== Plugin requirements
+
+* The Maven plugin produces inside the project target dir a directory (named ```server``` by default) that contains the server and any deployment.
+
+==== WildFly Maven Plugin new goals
+
+* ```wildfly:package```: Galleon provisioning of a server, CLI script execution, copy content to the server, deploy application (if present). 
+* ```wildfly:provision```: Galleon provisioning of a server.
+
+In addition to these 2 new goals, we are replacing the existing automatic download and unzip of wildfly-dist by a Galleon provisioning of
```
jmesnil, line 109: This change should be backwards compatible. In the absence of any provisioning/packaging specific parameters, a WildFly Full server must be provisioned.

If any provisioning/packaging parameters is specified, the equivalent of the `provision & package` goals must be executed before the existing goals are invoked.

maven/WFLY-14934_Server_Plugin.adoc
```diff
@@ -0,0 +1,186 @@
...
+WildFly. This change impacts ```run```, ```start``` and ```execute-commands``` goals).
+ 
+==== Galleon related Plugin configuration items
+
+These configuration items are common to the ```provision``` and ```package``` goals. They are used to configure 
+Galleon to provision a server.
+
+* feature-packs: A list of feature-pack to install.
+* log-time: Whether to log provisioning time at the end.
+* off-line: Whether to use offline mode when the plugin resolves an artifact.
```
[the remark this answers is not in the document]
jamezp: Minor nit, but if we change it IMO it should be `offline` with or without the suffix to be more consistent with `offline` option for the `execute-commands` goal.

maven/WFLY-14934_Server_Plugin.adoc
jamezp: My assumption here is this is what the `execute-commands` uses and `stdout` was used for consistency.

maven/WFLY-14934_Server_Plugin.adoc
```diff
@@ -0,0 +1,186 @@
...
+In addition to these 2 new goals, we are replacing the existing automatic download and unzip of wildfly-dist by a Galleon provisioning of 
+WildFly. This change impacts ```run```, ```start``` and ```execute-commands``` goals).
+ 
+==== Galleon related Plugin configuration items
+
+These configuration items are common to the ```provision``` and ```package``` goals. They are used to configure 
+Galleon to provision a server.
+
+* feature-packs: A list of feature-pack to install.
+* log-time: Whether to log provisioning time at the end.
```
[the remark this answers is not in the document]
jfdenise: +1

maven/WFLY-14934_Server_Plugin.adoc
```diff
@@ -0,0 +1,186 @@
...
+* off-line: Whether to use offline mode when the plugin resolves an artifact.
+* plugin-options: Arbitrary Galleon options used when provisioning the server.
```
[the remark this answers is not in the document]
jfdenise: @jmesnil +1 with galleon-options

maven/WFLY-14934_Server_Plugin.adoc
```diff
@@ -0,0 +1,186 @@
...
+* provision-directory-name: By default the server is generated in ```target/server``` directory. 
+The plugin offers a configuration item to provide the name of a directory relative to the project ```target``` directory in which the server will be installed.
+* provisioning-file: The path to the provisioning.xml file to use, by default ```<project base dir>/galleon/provisioning``` is used. If some feature-packs have been set,
```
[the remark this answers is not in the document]
jfdenise: Thanks will fix.

maven/WFLY-14934_Server_Plugin.adoc
```diff
@@ -0,0 +1,186 @@
...
+* record-state: Whether to record provisioned state in .galleon directory. When false, the file ```.wildfly-maven-plugin-provisioning.xml``` is generated in the server home directory.
+It contains Galleon provisioning informations used to provision this server.
+* server-configuration: A configuration contains the included and excluded Galleon layers to build the configuration.
+A configuration is identified by a ```model``` (default to ```standalone```) and a ```name``` (default to ```standalone.xml```).
+ 
+==== ```package``` goal specific configuration items
+
+These configuration items names, when applicable, are re-using the names that are in use in existing WildFly Maven plugin goals.
+
+* cli-sessions: A list of CLI execution session. Each session can contain: a list of CLI scripts, a list of commands, a list of properties files,
```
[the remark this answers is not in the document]
jfdenise: +1.

maven/WFLY-14934_Server_Plugin.adoc
jfdenise: Yes, re-using the same option for CLI output.

jmesnil: ok, never mind :)

maven/WFLY-14934_Server_Plugin.adoc
jfdenise: @jmesnil I am wandering if we should rename server-configuration (the galleon configuration that contains layers) to provisioning-config

maven/WFLY-14934_Server_Plugin.adoc
honza-kasik: If I understood the debate correctly: If we are going to update `run`, `start` and `execute-commands`, they must work even when user won't make any provisioning configuration. Is that correct? Is updating these commands necessary?

maven/WFLY-14934_Server_Plugin.adoc
```diff
@@ -0,0 +1,186 @@
...
+ 
+==== ```package``` goal specific configuration items
+
+These configuration items names, when applicable, are re-using the names that are in use in existing WildFly Maven plugin goals.
+
+* cli-sessions: A list of CLI execution session. Each session can contain: a list of CLI scripts, a list of commands, a list of properties files, 
+a list of java options, the ability to resolve or not expressions locally (false by default) and system properties for the CLI forked process. 
+NB: CLI commands are automatically executed in the context of an embedded server.
+* extra-server-content-dirs: A list of directories to copy content to the provisioned server inside a forked process.
+* filename: The name of the artifact to deploy inside the project build directory. By default a lookup is done to retrieve the primary artifact to deploy.
```
honza-kasik, line 138: Is this intended also for the out-of-project artifacts?

maven/WFLY-14934_Server_Plugin.adoc
```diff
@@ -0,0 +1,186 @@
...
+* name: The name of the deployment. By default the file name is used.
+* runtime-name: The runtime name of the deployment. By default the file name is used. NB: In order to deploy 
+the deployment inside the root context, ```ROOT.war``` runtime name is to be used.
+* server-config: The server config to use when executing CLI scripts and deploying artifact. Is ```standalone.xml``` by default if no ```server-groups``` have 
+been provided.  Is ```domain.xml``` by default if ```server-groups``` have been provided.
```
honza-kasik, line 143: With builder image in mind (per Context chapter), why is the managed domain mode considered for executing CLI scripts?

maven/WFLY-14934_Server_Plugin.adoc
jfdenise: @jmesnil , server-configuration/model and name is removed, I replaced by layers config element. We are keeping server-config with a simple default value "standalone.xml". domai nsupport being removed.

maven/WFLY-14934_Server_Plugin.adoc
jfdenise: @jmesnil , server-groups and server-configuration removed.

maven/WFLY-14934_Server_Plugin.adoc
jfdenise: @honza-kasik , I added a chapter for run and start command. We want to keep the original behavior for these goals (install a server if none is found), execute-commands will not provision a server. https://issues.redhat.com/browse/WFMP-130 tracks the changes on the existing goals.

maven/WFLY-14934_Server_Plugin.adoc
```diff
@@ -0,0 +1,186 @@
...
+A configuration is identified by a ```model``` (default to ```standalone```) and a ```name``` (default to ```standalone.xml```).
+ 
+==== ```package``` goal specific configuration items
+
+These configuration items names, when applicable, are re-using the names that are in use in existing WildFly Maven plugin goals.
+
+* cli-sessions: A list of CLI execution session. Each session can contain: a list of CLI scripts, a list of commands, a list of properties files, 
+a list of java options, the ability to resolve or not expressions locally (false by default) and system properties for the CLI forked process. 
+NB: CLI commands are automatically executed in the context of an embedded server.
+* extra-server-content-dirs: A list of directories to copy content to the provisioned server inside a forked process.
```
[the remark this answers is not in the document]
jfdenise: I added some details to answer your question.

maven/WFLY-14934_Server_Plugin.adoc
jfdenise: @honza-kasik , yes good point, the spec was too wide, domain is no more in the scope for packaging. I removed configuration item and mentioned that in "Non Requirements".

maven/WFLY-14934_Server_Plugin.adoc
jfdenise: I made it clearer that the look-up is only done inside the project build dir.

jfdenise:
@jamezp @jmesnil @honza-kasik I made changes to reflect the latest discussions and feedbacks. The impact is quite major (renaming of configuration items, simplification of galleon configuration, removal of domain support in the package goal, default behavior for the ```provision``` goal.

maven/WFLY-14934_Server_Plugin.adoc
```diff
@@ -0,0 +1,208 @@
...
+* excluded-layers: A set of layers to exclude when building-up a custom configuration.
+
+==== ```provision``` goal specific behavior
+
+Calling ```provision``` goal without any Galleon configuration items will provision the same default server 
+that we are provisioning when calling ```run``` and ```start``` goals.
+
+==== ```package``` goal specific configuration items and behavior
+
+A Galleon configuration is expected, if no galleon configuration is found, the goal execution fails.
```
honza-kasik, line 139: Just to be sure: This is about "Galleon related Plugin configuration item" and not only about providing value to `provisioning-file` option.

maven/WFLY-14934_Server_Plugin.adoc
```diff
@@ -0,0 +1,208 @@
...
+be located inside the ```<provisioned server>/standalone/configuration/foo.properties```.
+* filename: The name of the artifact to deploy. A look-up is done inside the project build directory. 
+By default a lookup is done to retrieve the primary artifact to deploy. If no file is found, no deployment is operated.
+* name: The name of the deployment. By default the file name is used.
+* runtime-name: The runtime name of the deployment. By default the file name is used. NB: In order to deploy 
+the deployment inside the root context, ```ROOT.war``` runtime name is to be used.
+* server-config: The server config to use when executing CLI scripts and deploying artifact. Is ```standalone.xml``` by default.
+* stdout: Indicates how ```stdout``` and ```stderr``` should be handled for the forked CLI process. 
+By default the output is redirected to the current process. ```none``` to disable output, ```System.out``` or ```System.err``` to redirect to the current process. 
+Any other value is assumed to be the path to a file and the out/err will be written there.
```
honza-kasik, line 158: There will be a single file at the end with output of all CLI executions from list provided in `packaging-scripts`?

honza-kasik (commented):
Thank you, I just have two last questions hopefully...

maven/WFLY-14934_Server_Plugin.adoc
jfdenise: @honza-kasik good point, I made it clear.

maven/WFLY-14934_Server_Plugin.adoc
jfdenise: @honza-kasik , very good point, I made it clear that all output is appended to the same file.

maven/WFLY-14934_Server_Plugin.adoc
```diff
@@ -0,0 +1,208 @@
...
+
+=== Hard Requirements
+
+==== Plugin requirements
+
+* The Maven plugin produces inside the project target dir a directory (named ```server``` by default) that contains the server and any deployment.
+
+==== WildFly Maven Plugin new goals
+
+* ```wildfly:package```: Galleon provisioning of a server, CLI script execution, copy content to the server, deploy application (if present).
```
jmesnil, line 108: I'm not sure if we want to put that in this proposal but the key aspect for our new S2I images is that we will rely on this `package` goal only.
It makes sense to introduce a `provision` goal to split the provisioning from the packaging of WildFly but that's not something that will be called by S2I.

maven/WFLY-14934_Server_Plugin.adoc
```diff
@@ -0,0 +1,186 @@
...
+
+* feature-packs: A list of feature-pack to install.
+* log-time: Whether to log provisioning time at the end.
+* off-line: Whether to use offline mode when the plugin resolves an artifact.
+* plugin-options: Arbitrary Galleon options used when provisioning the server.
+* provision-directory-name: By default the server is generated in ```target/server``` directory. 
+The plugin offers a configuration item to provide the name of a directory relative to the project ```target``` directory in which the server will be installed.
+* provisioning-file: The path to the provisioning.xml file to use, by default ```<project base dir>/galleon/provisioning``` is used. If some feature-packs have been set, 
+the provisioning file is being ignored.
+* record-state: Whether to record provisioned state in .galleon directory. When false, the file ```.wildfly-maven-plugin-provisioning.xml``` is generated in the server home directory.
```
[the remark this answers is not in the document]
jmesnil: @jfdenise are you ok to update this name too?

maven/WFLY-14934_Server_Plugin.adoc
```diff
@@ -0,0 +1,208 @@
...
+will provision the same default server that we are provisioning when calling ```run``` and ```start``` goals.
+
+==== ```package``` goal specific configuration items and behavior
+
+A Galleon configuration is expected, if no galleon configuration is found, the goal execution fails.
+
+The configuration items names, when applicable, are re-using the names that are in use in existing WildFly Maven plugin goals.
+
+* packaging-scripts: A list of execution of CLI scripts and commands. Each execution can contain: a list of CLI scripts, a list of commands, a list of properties files, 
+a list of java options, the ability to resolve or not expressions locally (false by default) and system properties for the CLI forked process.
```
jmesnil, line 144: `the ability to resolve or not expressions locally (false by default)` Afaict this one would only be effective for a list of CLI scripts. When I played with your POC any expression in a list of commands directly inside the `pom.xml` are resolved *by Maven* before they are passed to the CLI forked process.

I've not been able to prevent expression resolution by Maven...
Could you confirm please?

maven/WFLY-14934_Server_Plugin.adoc
```diff
@@ -0,0 +1,208 @@
...
+==== Galleon related Plugin configuration items
+
+These configuration items are common to the ```provision``` and ```package``` goals. They are used to configure 
+Galleon to provision a server.
+
+* feature-packs: A list of feature-pack to install.
+* log-provisioning-time: Whether to log provisioning time at the end.
+* offline-provisioning: Whether to use offline mode when the plugin resolves an artifact.
+* galleon-options: Arbitrary Galleon options used when provisioning the server.
+* provisioning-dir: By default the server is generated in ```target/server``` directory.
```
jmesnil, line 123: Neat-picking but:

* provisioning-dir: the name of the directory relative to the Maven project build directory when the server is provisioned. It defaults to `server` and will result in the server being provisioned in (`${project.build.directory}/server` which is resolved to `target/server` by default)

maven/WFLY-14934_Server_Plugin.adoc
```diff
@@ -0,0 +1,208 @@
...
+* stdout: Indicates how ```stdout``` and ```stderr``` should be handled for the forked CLI process. 
+By default the output is redirected to the current process. ```none``` to disable output, ```System.out``` or ```System.err``` to redirect to the current process. 
+Any other value is assumed to be the path to a file and the out/err will be written there. All CLI executions output is appended to the file.
+
+==== ```run``` and ```start``` goals impact
+
+* If jboss-home has not been set and no server if found in ```<buildDir>/<provisioning-dir>``` a server is provisioned.
+* The groupId, artifactId, classifier, type configuration items are removed.
+* The ```version``` configuration item is kept allowing to identify a WildFly version to install.
+* The ```provisioning-dir``` allows to specify where to (find/provision) the server, defaults to ```<buildDir>/server```
```
jmesnil, line 165: >, defaults to ```<buildDir>/server```

can be removed as the prop is already specified above

maven/WFLY-14934_Server_Plugin.adoc
```diff
@@ -0,0 +1,208 @@
...
+the deployment inside the root context, ```ROOT.war``` runtime name is to be used.
+* server-config: The server config to use when executing CLI scripts and deploying artifact. Is ```standalone.xml``` by default.
+* stdout: Indicates how ```stdout``` and ```stderr``` should be handled for the forked CLI process. 
+By default the output is redirected to the current process. ```none``` to disable output, ```System.out``` or ```System.err``` to redirect to the current process. 
+Any other value is assumed to be the path to a file and the out/err will be written there. All CLI executions output is appended to the file.
+
+==== ```run``` and ```start``` goals impact
+
+* If jboss-home has not been set and no server if found in ```<buildDir>/<provisioning-dir>``` a server is provisioned.
+* The groupId, artifactId, classifier, type configuration items are removed.
```
jmesnil, line 163: this would need to be added to the plugin release notes

maven/WFLY-14934_Server_Plugin.adoc
```diff
@@ -0,0 +1,208 @@
...
+* The ```version``` configuration item is kept allowing to identify a WildFly version to install.
+* The ```provisioning-dir``` allows to specify where to (find/provision) the server, defaults to ```<buildDir>/server```
+
+==== ```execute-commands``` goal impact
+
+* The ability to provision a server if no server installation found is removed.
```
jmesnil, line 169: this would need to be added to the plugin release notes

maven/WFLY-14934_Server_Plugin.adoc
```diff
@@ -0,0 +1,208 @@
...
+
+==== WildFly Maven plugin versioning
+
+A new major version would contain these new features.
+
+==== Requirements for cloud integration
+
+* The plugin can be used in a "cloud build" context. In particular, the plugin must be usable with:
```
jmesnil, line 177: well, my comment about the `package` goal should be better put into that section (and maybe move that `Requirements for cloud integration` section with the main requirement section)

maven/WFLY-14934_Server_Plugin.adoc
jfdenise: Thanks, done.

maven/WFLY-14934_Server_Plugin.adoc
jfdenise: Thanks, done.

maven/WFLY-14934_Server_Plugin.adoc
jfdenise: Thanks, done.

maven/WFLY-14934_Server_Plugin.adoc
jfdenise: Thanks, done.

maven/WFLY-14934_Server_Plugin.adoc
jfdenise: Thanks, done.

maven/WFLY-14934_Server_Plugin.adoc
jfdenise: Confirmed, added a note for CLI scripts.

maven/WFLY-14934_Server_Plugin.adoc
jfdenise: Ok, done.

maven/WFLY-14934_Server_Plugin.adoc
```diff
@@ -0,0 +1,213 @@
...
+
+* The ability to provision a server if no server installation is found is removed.
+
+==== WildFly Maven plugin versioning
+
+A new major version would contain these new features.
+
+==== Impact on quickstarts 
+
+* A set of quickstarts (List TBD) will be refactored to use this plugin in both bare-metal and cloud.
```
honza-kasik, line 188: Can this TBD be resolved?

maven/WFLY-14934_Server_Plugin.adoc
```diff
@@ -0,0 +1,213 @@
...
+
+==== WildFly Maven Plugin new goals
+
+* ```wildfly:package```: Galleon provisioning of a server, CLI script execution, copy content to the server, deploy application (if present). 
+* ```wildfly:provision```: Galleon provisioning of a server.
+
+In addition to these 2 new goals, we are replacing the existing automatic download and unzip of wildfly-dist by a Galleon provisioning of 
+WildFly. This change impacts ```run``` and ```start```goals).
+ 
+==== Galleon related Plugin configuration items
```
honza-kasik, line 125: Why are the configuration options not same as in the bootable JAR plugin? Example:

| Bootable JAR | Provisioning              |
|--------------|---------------------------|
| log-time     | log-provisioning-time     |
| offline      | offline-provisioning      |
| record-state | record-provisioning-state |
| cli-sessions | packaging-scripts         |

maven/WFLY-14934_Server_Plugin.adoc
```diff
@@ -0,0 +1,213 @@
...
+// Discuss with QE during the Kickoff state to decide this
+* [ ] Engineering
+
+* [X] QE
+
+=== Affected Projects or Components
+
+* https://github.com/wildfly/wildfly-maven-plugin/[WildFly Maven plugin]
+
+* https://github.com/eclipse/jkube[JKube]
```
honza-kasik, line 47: How will be the JKube project affected?

maven/WFLY-14934_Server_Plugin.adoc
```diff
@@ -0,0 +1,213 @@
...
+
+* NONE
+
+=== Non-Requirements
+
+* No support for domain mode when packaging an application.
+
+== Test Plan
+
+* Add new Maven plugin tests in plugin source repository.
```
honza-kasik, line 200: What kind of tests will be added there? What are the risks?

maven/WFLY-14934_Server_Plugin.adoc
```diff
@@ -0,0 +1,213 @@
...
+* Add new QE tests.
```
honza-kasik, line 201: Since this is upstream analysis, are upstream QE tests expected?

maven/WFLY-14934_Server_Plugin.adoc
```diff
@@ -0,0 +1,213 @@
...
+The existing https://github.com/wildfly/wildfly-maven-plugin/[WildFly Maven plugin] has been designed to interact (mainly) with a running server instance. 
+We are here proposing to evolve the capabilities of this plugin for one to be able to build a fully configured server containing application deployment.
+ 
+This proposal adds a new kind of installation that is operated and controlled by the pom.xml file.  The pom.xml file contains all the information needed
+to control the server and application content and versioning.
+
+=== Requirements for cloud integration
+
+* The new ``package`` goal that this proposal adds to the Maven plugin is the goal to use during s2i build. 
+Other goals (e.g.: ``provisioning``, ``execute-commands``, ...) are not expected to be used.
```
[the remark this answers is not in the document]
jmesnil: `execute-commands` is not added, it already exist.
`provision` is added as it is a prerequisite for the `package` goal and it splits the provisioning of the server from its packaging (local modification to the installation)

maven/WFLY-14934_Server_Plugin.adoc
jmesnil: Not in the context of RFE.

maven/WFLY-14934_Server_Plugin.adoc
jmesnil: please read https://github.com/wildfly/wildfly-proposals/pull/422#discussion_r713113618 and subsequents comments

maven/WFLY-14934_Server_Plugin.adoc
jmesnil: @jfdenise wdyt? The test plan for this Maven plug-in would necessitate only test in this core repo.

Integration tests would have to be added later in the wildfly-s2i projects later

maven/WFLY-14934_Server_Plugin.adoc
honza-kasik: @jmesnil So now it is "descriptive name" vs "cross-plugin compatible name"?

maven/WFLY-14934_Server_Plugin.adoc
honza-kasik: @jmesnil Thank you. What I meant is we shouldn't approve analysis including TODOs/TBDs. If quickstarts are not subject of this proposal, it should be IMHO included.

maven/WFLY-14934_Server_Plugin.adoc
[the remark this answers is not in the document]
honza-kasik: Ah, thanks! Though what's the point of exposing the `provision` goal?

maven/WFLY-14934_Server_Plugin.adoc
jmesnil: Since when "cross-plugin compatible name" is a thing? It is more important that the parameters added to this maven-wildlfy-plugin are consistent with existing parameters of the same plugin.

maven/WFLY-14934_Server_Plugin.adoc
jmesnil: Tests that assert behaviours of the `package` and `provision` goal. What do you mean by risks?

maven/WFLY-14934_Server_Plugin.adoc
jmesnil: @jfdenise actually, as the first release focuses on cloud deployment, we should not mention that it will also be use for bare-metal.
Can you also please remove the `(List TBD)` as the list of refactored quickstarts will not be defined by this RFE. We will define that lists when we have S2I images that leverages this plugin.

maven/WFLY-14934_Server_Plugin.adoc
jfdenise: The would have to add some support to jkube (as we did for bootable JAR) in order for jkube to support the plugin usage and new s2i V2 image.

maven/WFLY-14934_Server_Plugin.adoc
jfdenise: Good point, I updated the proposal.

maven/WFLY-14934_Server_Plugin.adoc
jfdenise: Ok, yes. Fixed.

maven/WFLY-14934_Server_Plugin.adoc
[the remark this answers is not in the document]
jfdenise: The provision goal is a zero config goal (with a default Galleon config that provisions a full WildFly). That is the way to build a naked server without deployment.

maven/WFLY-14934_Server_Plugin.adoc
honza-kasik: Thank you @jmesnil ! "Consistency" was the word I should've used. The reason why I asked this is that users who already are using bootable JAR plugin might stumble upon this plugin too. They might also use them in the same `pom.xml`. I can imagine it would be confusing for some of them to set two "similar" plugins differently...

maven/WFLY-14934_Server_Plugin.adoc
honza-kasik: E.g.: Are there any known (to development) scenarios which are covered by plugin tests and needs to be expanded in QE tests?

maven/WFLY-14934_Server_Plugin.adoc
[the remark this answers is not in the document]
honza-kasik: Thank you @jfdenise ! I just don't understand why we bother with this goal when analysis mentions:

> ...are not expected to be used.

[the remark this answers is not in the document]
jfdenise: This is the chapter "Requirements for cloud integration". In a cloud context they are not expected to be used.

maven/WFLY-14934_Server_Plugin.adoc
honza-kasik: WDYT about introducing the "new" names in bootable JAR plugin in future @jmesnil @jfdenise?

jmesnil: There is no requirement to "backport" the name of the bootable jar plugin

maven/WFLY-14934_Server_Plugin.adoc
jfdenise: I added a description of the 2 new tests we are adding

maven/WFLY-14934_Server_Plugin.adoc
```diff
@@ -0,0 +1,213 @@
...
+
+* The ability to provision a server if no server installation is found is removed.
+
+==== WildFly Maven plugin versioning
+
+A new major version would contain these new features.
+
+==== Impact on quickstarts 
+
+* A set of quickstarts will be refactored to use this plugin on the cloud.
```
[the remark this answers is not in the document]
jfdenise: @honza-kasik we don't a list of quickstarts, I am wandering if we should remove this mention and cover impact on quickstarts in another future RFE?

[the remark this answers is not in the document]
honza-kasik: @jfdenise +1 That sounds to me like the optimal solution! Mentioning that the edit of quickstarts is subject of other proposal would be great. The best would be having the proposal already in JIRA.

maven/WFLY-14934_Server_Plugin.adoc
honza-kasik: @jfdenise Good point, thanks. I created a new WFWIP tracker https://issues.redhat.com/browse/WFWIP-407

maven/WFLY-14934_Server_Plugin.adoc
```diff
@@ -0,0 +1,215 @@
...
+
+* https://github.com/wildfly/wildfly-maven-plugin/[WildFly Maven plugin]
+
+* https://github.com/eclipse/jkube[JKube]
+
+* https://github.com/wildfly/wildfly-s2i[WildFly s2i]
+
+* https://github.com/wildfly/quickstart[WildFly quickstarts]
+
+* ODO dev files.
```
honza-kasik, line 53: How they will be affected? Will be there additional RFE for this?

maven/WFLY-14934_Server_Plugin.adoc
honza-kasik: This means this proposal also adds support of jkube? Is this support different in product? If it is, can this be mentioned in product analysis please?

maven/WFLY-14934_Server_Plugin.adoc
```diff
@@ -0,0 +1,215 @@
...
+
+* [X] QE
+
+=== Affected Projects or Components
+
+* https://github.com/wildfly/wildfly-maven-plugin/[WildFly Maven plugin]
+
+* https://github.com/eclipse/jkube[JKube]
+
+* https://github.com/wildfly/wildfly-s2i[WildFly s2i]
```
honza-kasik, line 49: Isn't s2i affected by following proposal and not this one?

maven/WFLY-14934_Server_Plugin.adoc
```diff
@@ -0,0 +1,215 @@
...
+
+* https://github.com/wildfly/quickstart[WildFly quickstarts]
```
honza-kasik, line 51: Quickstarts are now subject of following proposal per 

>  A set of quickstarts should be refactored to use this plugin on the cloud. This should be covered by another proposal.

maven/WFLY-14934_Server_Plugin.adoc
```diff
@@ -0,0 +1,215 @@
...
+Calling ```provision``` goal without any Galleon configuration items (no feature-pack provided nor provisioning.xml file) 
+will provision the same default server that we are provisioning when calling ```run``` and ```start``` goals.
+
+==== ```package``` goal specific configuration items and behavior
+
+A Galleon configuration is expected, if no galleon configuration is found, the goal execution fails.
+
+The configuration items names, when applicable, are re-using the names that are in use in existing WildFly Maven plugin goals.
+
+* packaging-scripts: A list of execution of CLI scripts and commands. Each execution can contain: a list of CLI scripts, a list of commands, a list of properties files,
```
honza-kasik, line 154: What is the difference between `a list of CLI scripts` and `a list of commands`? In what combinations these can be used?

maven/WFLY-14934_Server_Plugin.adoc
jfdenise: I suspect yes, in other RFE.

maven/WFLY-14934_Server_Plugin.adoc
jfdenise: User as freedom to inline commands in the pom and/or reference CLI scripts. They can choose their own granularity for scripts vs commands.

maven/WFLY-14934_Server_Plugin.adoc
jmesnil: Changes to quickstarts will not be covered by this proposal. They will be *eventually* affected when we work on new S2I images that uses this plug-in.

maven/WFLY-14934_Server_Plugin.adoc
honza-kasik, line 53: @jfdenise This should be IMHO removed since it should be part of other proposal. Opening new thread to make this visible.

maven/WFLY-14934_Server_Plugin.adoc
```diff
@@ -0,0 +1,215 @@
...
+to control the server and application content and versioning.
+
+=== Requirements for cloud integration
+
+* The new ``package`` goal that this proposal adds to the Maven plugin is the goal to use during s2i build. 
+Other goals (e.g.: ``provisioning``, ``execute-commands``, ...) are not expected to be used.
+* The plugin can be used in a "cloud build" context. In particular, the plugin must be usable with:
+** WildFly s2i V2 builder image for s2i in binary and source (see: https://issues.redhat.com/browse/WFLY-14936[WFLY-14934])
+** JKube s2i and docker binary builds (see https://issues.redhat.com/browse/WFLY-14938[WFLY-14938])
+** Other technologies such as Buildpacks.
```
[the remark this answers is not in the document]
honza-kasik: So can this be reworded please?

maven/WFLY-14934_Server_Plugin.adoc
jmesnil: that's fine. This is not an exhaustive list of components that will be modified by this proposal. It lists components that will be affected by this feature.

jfdenise:
@honza-kasik , all your comments should be integrated now. Thanks.

maven/WFLY-14934_Server_Plugin.adoc
```diff
@@ -0,0 +1,217 @@
...
+* The new ``package`` goal that this proposal adds to the Maven plugin is the goal to use during s2i build. 
+Other goals (e.g.: ``provisioning``, ``execute-commands``, ...) are not expected to be used.
+
+* The plugin can be used in a "cloud build" context. In particular, the plugin must be usable with:
+** WildFly s2i V2 builder image for s2i in binary and source (see: https://issues.redhat.com/browse/WFLY-14936[WFLY-14934])
+** JKube s2i and docker binary builds (see https://issues.redhat.com/browse/WFLY-14938[WFLY-14938])
+* The generated server can be installed and run successfully inside a WildFly runtime image v2 defined 
+   in https://issues.redhat.com/browse/WFLY-14936[WFLY-14936].
+
+The integration of this plugin for cloud integration will affect other projects:
```
honza-kasik, line 104: I suggest to add an explicit note that changes in mentioned projects are not a part of this RFE and that they can be improved by leveraging this RFE to make things finally clear. Preferably in/around "The integration of this plugin for cloud integration will affect other projects:" including " Eventually, other technologie such as BuildPacks will be able to leverate this plugin to build cloud images for WildFly.".

maven/WFLY-14934_Server_Plugin.adoc
```diff
@@ -0,0 +1,217 @@
...
+
+== Issue Metadata
+
+=== Issue
+
+* https://issues.redhat.com/browse/WFLY-14934[WFLY-14934]
+
+=== Related Issues
+
+* Deprecation of some existing wildfly-maven-plugin features: https://issues.redhat.com/browse/WFMP-130[WFMP-130]
```
jmesnil, line 22: @jfdenise could you please add https://issues.redhat.com/browse/WFMP-138 to the list of related issues?

jfdenise: Thanks, added.

PR #434, opened 2021-10-06
Proposal for WFLY-15421, OIDC support for WildFly s2i

jfdenise:


jfdenise:
@fjuma @darranl , a proposal for oidc subsystem usage in a cloud context.

openshift/WFLY-15421_OIDC_for_WildFly_s2i_v2.adoc
```diff
@@ -0,0 +1,297 @@
...
+=== Relevant Installation Types
+// Remove the x next to the relevant field if the feature in question is not relevant
+// to that kind of WildFly installation
+* [ ] Traditional standalone server (unzipped or provisioned by Galleon)
+
+* [ ] Managed domain
+
+* [x] OpenShift s2i
+
+* [ ] Bootable jar
```
jmesnil, line 57: Is this also impacted?
Could a Bootable Jar image uses the `elytron-oidc-client` layer?

openshift/WFLY-15421_OIDC_for_WildFly_s2i_v2.adoc
```diff
@@ -0,0 +1,297 @@
...
+clients into the authorization server.
+
+The existing `SSO_*` env variables used for OIDC are usable for Keycloak OIDC integration. 
+It helps transition smoothly to use the native OIDC support. 
+This proposal defines a mapping between the `SSO_*` and `OIDC_*` env variables.
+
+The env variable `OIDC_PROVIDER_NAME=<name>` MUST be set when using `OIDC_*` env variables. When using `SSO_*` env variables 
+the provider name `keycloak` is internally used.
+
+A mix of `SSO_*` and `OIDC_*` env variables, although nor recommended, would work,
```
jmesnil, line 113: s/nor/not

openshift/WFLY-15421_OIDC_for_WildFly_s2i_v2.adoc
```diff
@@ -0,0 +1,297 @@
...
+
+In order for war deployments to be discovered they must have the `OIDC` auth-method in web.xml. 
+Deployments that contain a `WEB-INF/oidc.json` descriptor file are ignored.
+
+NB: A single provider can be configured for a running server. The discovered deployments are all registered inside the same OIDC provider.
+
+We are offering a new set of env variables `OIDC_*` to be used to configure the subsystem, enable the discovery of deployments and registration of 
+clients into the authorization server.
+
+The existing `SSO_*` env variables used for OIDC are usable for Keycloak OIDC integration.
```
jmesnil, line 106: We should also deprecate all SSO_* env variables, wdyt?

openshift/WFLY-15421_OIDC_for_WildFly_s2i_v2.adoc
```diff
@@ -0,0 +1,297 @@
...
+
+This registration is enabled only for the `keycloak` provider. Other providers would need the client to be already 
+registered inside the authorization server.
+
+===== Routes discovery
+
+* Route discovery (if no OIDC_HOSTNAME_HTTP(S) is set) is bound to openshift. 
+* This feature can’t be used on other cloud system.
+* The route discovery  implies that complex JSON structure be parsed.  `jq` (small runtime to parse json returned by openshift service)  is expected to be 
+installed in the WildFly s2i v2 image.
```
jmesnil, line 265: In which image? the builder or the runtime image?

openshift/WFLY-15421_OIDC_for_WildFly_s2i_v2.adoc
```diff
@@ -0,0 +1,297 @@
...
+* Add new Behave tests 
+* Add new Openshift QE tests.
+
+== Community Documentation
+
+* WildFly Openshift community doc is required.
+
+== Release Note Content
+
+Yes. Could be integrated in the release notes  that cover WildFly s2i v2.
```
jmesnil, line 297: This section should actually contain the content of the release note that we can copy/paste :)

openshift/WFLY-15421_OIDC_for_WildFly_s2i_v2.adoc
jfdenise: This proposal is not bound to bootable JAR, it is wildfly-s2i specific. The introduction of the galleon layer is in https://github.com/wildfly/wildfly-proposals/pull/341

openshift/WFLY-15421_OIDC_for_WildFly_s2i_v2.adoc
jfdenise: Thanks, fixed.

openshift/WFLY-15421_OIDC_for_WildFly_s2i_v2.adoc
jfdenise: Correct, fixed.

openshift/WFLY-15421_OIDC_for_WildFly_s2i_v2.adoc
jfdenise: Both, fixed.

openshift/WFLY-15421_OIDC_for_WildFly_s2i_v2.adoc
```diff
@@ -0,0 +1,298 @@
...
+new env variables will only be added in the `OIDC_*` namespace. We have a first example `OIDC_PROVIDER_SSL_REQUIRED` 
+that would be new without a mapping in the `SSO_*` namespace.
+
+Example of configuration using new OIDC env variables for Keycloak:
+
+```
+OIDC_PROVIDER_NAME=keycloak
+OIDC_PROVIDER_URL=http://keycloak/auth/realms/Wildfly
+OIDC_USER_NAME=foo
+OIDC_USER_PASSWORD=foo
```
fjuma, line 124: Are `OIDC_USER_NAME` and `OIDC_USER_PASSWORD` the client id and client secret?

`OIDC_CLIENT_ID` and `OIDC_CLIENT_SECRET` might be better variable names.

fjuma: Actually, I also see `OIDC_SECURE_DEPLOYMENT_SECRET` below.

How are the `OIDC_USER_NAME`, `OIDC_USER_PASSWORD`, and `OIDC_SECURE_DEPLOYMENT_SECRET` variables used.

jfdenise: The NAME and PASSWORD are credetials used to retrieve the token from the keycloak server (token required to then create a new remote client). These pieces of information are not in the subsystem configuration.
The SECRET is the actual secret set in the secure deployment configuration. This piece of information is also sent to the server when creating/registering the client in the remote keycloak server.

fjuma (approved):
Thanks, @jfdenise! This looks good to me.

openshift/WFLY-15421_OIDC_for_WildFly_s2i_v2.adoc
```diff
@@ -0,0 +1,306 @@
...
+:email:             jdenise@redhat.com
+:toc:               left
+:icons:             font
+:idprefix:
+:idseparator:       -
+
+== Overview
+
+For WildFly s2i V2 we are removing the keycloack OIDC and SAML client adapters present in the s2i image.
+The only supported integration with keycloak would be based on the `elytron-oidc-client` subsystem and
```
OndrejKotek, line 12: This a bit imprecise as the SAML part is out of scope of this RFE and is handled within the scope of WFLY-16306.

openshift/WFLY-15421_OIDC_for_WildFly_s2i_v2.adoc
```diff
@@ -0,0 +1,306 @@
...
+|N
+|False by default
+
+|OIDC_PROVIDER_SSL_REQUIRED
+|NONE, new env variable (see CLOUD-3529 )
+|Is https enabled in the realm. By default external (private and local address no https).
+|N
+|external
+
+|OIDC_PROVIDER_TRUSTSTORE
```
OndrejKotek, line 205: Is it possible to make it somehow easier for the users, for example having just a variable referencing an OpenShift secret with the truststore (or even just trusted certificate(s))?

openshift/WFLY-15421_OIDC_for_WildFly_s2i_v2.adoc
```diff
@@ -0,0 +1,306 @@
...
+
+We are offering a new set of env variables `OIDC_*` to be used to configure the subsystem, enable the discovery of deployments and registration of 
+clients into the authorization server.
+
+The existing `SSO_*` env variables used for OIDC are usable for Keycloak OIDC integration. 
+It helps transition smoothly to use the native OIDC support. 
+This proposal defines a mapping between the `SSO_*` and `OIDC_*` env variables.
+The `SSO_*` env variables are deprecated.
+
+The env variable `OIDC_PROVIDER_NAME=<name>` MUST be set when using `OIDC_*` env variables. When using `SSO_*` env variables
```
kstekovi, line 111: Is this really needed? it is jut for identification which set of env variables to use. cannot be checked which variable are filled and use them? when I will use the RH-SSO then i should use also `keycloak` as name of provider? I think just SSO_PROVIDER should be enough and don't introduced OIDC variables.  When is filled SSO_PROVIDER then SSO_REAL and SSO_URL is not mandatory to use.

openshift/WFLY-15421_OIDC_for_WildFly_s2i_v2.adoc
```diff
@@ -0,0 +1,306 @@
...
+```
+OIDC_PROVIDER_NAME=other
+OIDC_PROVIDER_URL=http://foo/specific/url/for/provider
+OIDC_HOSTNAME_HTTP=my-app-myproject.192.168.42.233.nip.io
+```
+
+===== Env variables mapping table 
+
+|===
+|Env var |SSO|Details|Required|Defaults
```
kstekovi, line 149: I think new set of variable with prefix OIDC just make configuration more complex. Is not enough add variable for OIDC_PROVIDER_SSL_REQUIRED with prefix SSO instead OIDC?

openshift/WFLY-15421_OIDC_for_WildFly_s2i_v2.adoc
jfdenise: I think that it is a good idea. It would avoid to mount the secret trustore in the pod. We could see it as a new RFE with the introduction of OIDC_PROVIDER_TRUSTORE_SECRET and OIDC_PROVIDER_CERTIFICATE_SECRET.

What I understand is that we would use the secret name, query the Openshift API to retrieve the Truststore secret from the current namespace then setup a local truststore file locally. I think that it would be doable. The question I am wandering are the authorizations required for a user to be able to retrieve a secret. 
I did a quick attempt using openshift sandbox and I was not able to get the secrets from the pod but this would require more investigations.

openshift/WFLY-15421_OIDC_for_WildFly_s2i_v2.adoc
jfdenise: The provider name is required to create and configure the elytron-oidc-client provider.
In the future we can imagine other kind of providers for which we would have to adjust some logic in the launch scripts to  implement the interaction with the provider (eg: add new client). The name would be used as a provider identifier.

openshift/WFLY-15421_OIDC_for_WildFly_s2i_v2.adoc
jfdenise: The SSO_XXX are kept for backward compatibility with Keycloak/SSO adapters. The OIDC_XXX are introduced as new env variable that are in use with the elytron-oidc-client. elytron-oidc-client is an open adapter that allows to connect to other oidc provider than RH SSO/Keycloak.

openshift/WFLY-15421_OIDC_for_WildFly_s2i_v2.adoc
jfdenise: Thank-you, I removed the SAML mention.

openshift/WFLY-15421_OIDC_for_WildFly_s2i_v2.adoc
OndrejKotek: I have not enough knowledge in this area. I just remember how painful was to configure certificates for EAP+RH-SSO on OpenShift for me a year or two ago. If such feature is worth the effort I vote for the new RFE.

I have noticed there is JGROUPS_ENCRYPT_SECRET environment variable. Maybe we can get partially inspired from the underlying solution there. See https://access.redhat.com/documentation/en-us/red_hat_jboss_enterprise_application_platform/7.4/html-single/getting_started_with_jboss_eap_for_openshift_online/index#configuring_sym_encrypt

jfdenise: JGROUPS_ENCRYPT_SECRET is not used as you think it is. It is the env variable to provide the secret in the template: https://github.com/jboss-container-images/jboss-eap-openshift-templates/blob/eap74-dev/templates/eap74-https-s2i.json#L720
A user still needs to provide the name and directory: JGROUPS_ENCRYPT_KEYSTORE and JGROUPS_ENCRYPT_KEYSTORE_DIR. So actually when using helm charts, these env variables are no more in the context. We should fix that in some way.

openshift/WFLY-15421_OIDC_for_WildFly_s2i_v2.adoc
kstekovi: The provider name must be some exact name to configure elytron-oidc-client? for example `rh-sso` or `keycloak`. Will be supported also some non RH identity providers? Or the provider name cannot be null or empty to activate elytron-oidc-client configuration?

openshift/WFLY-15421_OIDC_for_WildFly_s2i_v2.adoc
kstekovi: I wanted to point out the SSS_XXX and OIDC_XXX variables are almost same. So the elytron-oidc-client could be configured by SSO_XXX varibles. then the OIDC_PROVIDER_NAME should have some specific value to use Keycloak adapter and different one for elytron-oidc-client. For example as it is used in web.xml. auth-method must be OIDC to use elytron-oidc-client and KEYCLOAK for Keycloak adapter.

openshift/WFLY-15421_OIDC_for_WildFly_s2i_v2.adoc
jfdenise: @kstekovi , in the future, we can imagine to introduce new providers yes. But for now we are limiting this feature to keycloak/sso

kstekovi: OpenID is a standard so it shouldn't matter which provider is used. cannot be used some general value like oidc?

`keycloak` is for SSO_XXX and keycloak adapter and `sso` is for OIDC_XXX and elytron-oidc-client?

jfdenise: The env variable OIDC_XXX cover the standard, so yes these env variables will be used whatver the provider. Some logic (to interact with the management interface of the provider to create client) could be specific.
SSO_XXX are there to help migration from legacy Keycloak/SSO oidc adapter to elytron-oidc-client. OIDC_XXX env are used to configure the elytron-oidc-client subsystem whatever the oidc provider.

openshift/WFLY-15421_OIDC_for_WildFly_s2i_v2.adoc
kstekovi: resolved in https://github.com/wildfly/wildfly-proposals/pull/434#discussion_r970853665

openshift/WFLY-15421_OIDC_for_WildFly_s2i_v2.adoc
kstekovi: Thank you. It make sense have both configuration divided into two sets as you write into AD.

openshift/WFLY-15421_OIDC_for_WildFly_s2i_v2.adoc
```diff
@@ -0,0 +1,306 @@
...
+* If the war file is named ``ROOT.war``, and no ``<module-name>`` has been set, the client-id is ``root``.
+* Otherwise the name of the deployment file (without .war extension) is used as client-id value.
+* In addition, in case the env variable ``APPLICATION_NAME`` is set, the value of the client-id is prefixed by ``$APPLICATION_NAME-``
+
+===== Dynamic registration of clients
+
+This registration is enabled only for the `keycloak` provider. Other providers would need the client to be already 
+registered inside the authorization server.
+
+===== Routes discovery
```
kstekovi, line 269: There are required some extra permissions to load the hosts from OpenShift. Could be mentioned here?

jfdenise: @kstekovi , thank-you I added the role and oc commands required to allow for route discovery.

kstekovi: Thank you. just small notes to yours commands. 
The name of role can be whatever. \<your-name-of-role\> could be better. 
The name of the user isn't `view` \<your-user-name\> could be better.

openshift/WFLY-15421_OIDC_for_WildFly_s2i_v2.adoc
```diff
@@ -0,0 +1,313 @@
...
+===== Client ID computation
+
+The client-id attribute is computed during deployment discovery:
+
+* If the web.xml file contains a ``<module-name>``, the client-id value is the value of this module-name.
+* If the war file is named ``ROOT.war``, and no ``<module-name>`` has been set, the client-id is ``root``.
+* Otherwise the name of the deployment file (without .war extension) is used as client-id value.
+* In addition, in case the env variable ``APPLICATION_NAME`` is set, the value of the client-id is prefixed by ``$APPLICATION_NAME-``
+
+===== Dynamic registration of clients
```
kstekovi, line 264: The user specified in `OIDC_USER_NAME` must have specific permission assigned to be able create the client in RH-SSO. Could you also mention this permission? Client Role `create-client` from group `realm-management`.

PR #435, opened 2021-10-08
[WFLY-15452] - modify ajp-listener to allow specifying pattern for aj…

baranowb:
…p request attributes
Issue: https://issues.redhat.com/browse/WFLY-15452

### DIFF
--- a/undertow/WFLY-15452_ajp-listener_allowed_attr_pattern.adoc
+++ b/undertow/WFLY-15452_ajp-listener_allowed_attr_pattern.adoc
@@ -1,0 +1,130 @@
+= [Preview]modify ajp-listener to allow specifying pattern for ajp request attributes
+:author:            Bartosz Baranowski
+:email:             bbaranow@redhat.com
+:toc:               left
+:icons:             font
+:idprefix:
+:idseparator:       -
+
+== Overview
+
+Since UNDERTOW-1667 one can set additional AJP request attribute parsing permission via env variable. However there is no way to set it in WFLY config/model. This RFE's goal is to make it possible.
+
+== Issue Metadata
+
+=== Issue
+
+* https://issues.redhat.com/browse/WFLY-15452[WFLY-15452]
+
+=== Related Issues
+
+* https://issues.redhat.com/browse/UNDERTOW-1667[UNDERTOW-1667]
+* https://issues.redhat.com/browse/UNDERTOW-1977[UNDERTOW-1977]
+* https://issues.redhat.com/browse/WFLY-15453[WFLY-15453]
+
+=== Stability Level
+// Choose the planned stability level for the proposed functionality
+* [ ] Experimental
+
+* [X] Preview
+
+* [ ] Community
+
+* [ ] default
+
+=== Dev Contacts
+
+* mailto:{email}[{author}]
+
+=== QE Contacts
+
+* mailto:msvehla@redhat.com[Martin Svehla]
+
+=== Testing By
+// Put an x in the relevant field to indicate if testing will be done by Engineering or QE. 
+// Discuss with QE during the Kickoff state to decide this
+* [ ] Engineering
+
+* [X] QE
+
+=== Affected Projects or Components
+
+* undertow
+
+=== Other Interested Projects
+
+=== Relevant Installation Types
+// Remove the x next to the relevant field if the feature in question is not relevant
+// to that kind of WildFly installation
+* [x] Traditional standalone server (unzipped or provisioned by Galleon)
+
+* [x] Managed domain
+
+* [x] OpenShift s2i
+
+* [x] Bootable jar
+
+== Requirements
+
+=== Hard Requirements
+
+* Being able to configure pattern via model/xml.
+[literal]
+<subsystem xmlns="urn:jboss:domain:undertow:14.0" default-server="some-server" default-servlet-container="myContainer" default-virtual-host="default-virtual-host" instance-id="some-id" statistics-enabled="true">
+   ...
+   <server default-host="other-host" name="some-server" servlet-container="myContainer">
+      ...
+      <ajp-listener ...  allowed_request_attr_pattern="(?:apple|banana)"  .../>
+      ...
+   </server>
+   ...
+</subsystem>
+
+Parameters will be present in undertow server element(for standalone: /subsystem=undertow/server=default-server/ajp-listener=myListener):
+* allowed_request_attr_pattern
+** Default: null
+** Type: String(regex - java.util.regex.Pattern)
+
+=== Nice-to-Have Requirements
+
+=== Non-Requirements
+
+== Backwards Compatibility
+
+Possibly. Subsystem transformers should be able to handle it.
+
+=== Default Configuration
+
+No change.
+
+=== Importing Existing Configuration
+
+No steps should suffice, as it would mean defaulting to 'null', which is default value in undertow source.
+
+=== Deployments
+
+Not affected.
+
+=== Interoperability
+
+Not affected.
+
+== Implementation Plan
+
+Done.
+
+== Security Considerations
+
+Possibly. UNDERTOW-1667 is a CVE, so this RFE should be documented well, in order to warn users of potential exposure.
+
+== Test Plan
+
+Unit tests should cover new functionality(there is already test case covering AjpListener).
+
+== Community Documentation
+
+Task for WFLY documentation already exist - WFLY-15453. HOwever, this is model change and there is model reference doc generated, so its unclear which approach is better?
+
+== Release Note Content
+
+Allow configuration of AJP request attribute pattern with model entry, rather than only via system property.


undertow/WFLY-15452_ajp-listener_allowed_attr_pattern.adoc
```diff
@@ -0,0 +1,121 @@
+= <INSERT TITLE HERE>
```
fl4via, line 1: It appears you have the title at the wrong line :-)

undertow/WFLY-15452_ajp-listener_allowed_attr_pattern.adoc
```diff
@@ -0,0 +1,121 @@
...
+:author:            Modify ajp-listener to allow specifying AJP_ALLOWED_REQUEST_ATTRIBUTES_PATTERN
+:email:             bbaranow@redhat.com
+:toc:               left
+:icons:             font
+:idprefix:
+:idseparator:       -
+
+== Overview
+
+Since UNDERTOW-1667 one can set additional AJP request attribute parsing permission via env variable. However there is no way to set it in WFLY config nor directly in UndertowOptions. Later is being addressed in UNDERTOW-1978, to which this RFE is a follow up.
```
fl4via, line 11: We need to update this, since the UndertowOptions is also already available without any PRs for UNDERTOW-1978

undertow/WFLY-15452_ajp-listener_allowed_attr_pattern.adoc
```diff
@@ -0,0 +1,121 @@
...
+=== Other Interested Projects
+
+=== Relevant Installation Types
+// Remove the x next to the relevant field if the feature in question is not relevant
+// to that kind of WildFly installation
+* [x] Traditional standalone server (unzipped or provisioned by Galleon)
+
+* [x] Managed domain
+
+* [~x] OpenShift s2i
```
fl4via, line 54: Is this a maybe?

undertow/WFLY-15452_ajp-listener_allowed_attr_pattern.adoc
```diff
@@ -0,0 +1,121 @@
...
+
+* [ ] Bootable jar
+
+== Requirements
+
+=== Hard Requirements
+
+* Being able to configure pattern via XML configuration file.
```
fl4via, line 62: via model, it could be directly in the XML file or it could be via CLI

undertow/WFLY-15452_ajp-listener_allowed_attr_pattern.adoc
```diff
@@ -0,0 +1,121 @@
...
+
+=== Nice-to-Have Requirements
+
+=== Non-Requirements
+
+== Backwards Compatibility
+
+Possibly. Subsystem transformers should be able to handle it.
+
+=== Default Configuration
```
fl4via, line 72: I think default configuration refers to what will be the default configuration of the server, can someone vouch if I'm right?
So, by default, are we having a allowed_request_attr_pattern in the server? Nope, because we don't have an ajp-listener. Still, the example of the xml could be valid in Hard Requirements, and very helpful for the test plan

undertow/WFLY-15452_ajp-listener_allowed_attr_pattern.adoc
```diff
@@ -0,0 +1,121 @@
...
+
+Not affected.
+
+=== Interoperability
+
+Not affected.
+
+== Implementation Plan
+
+This RFE depends on linked undertow RFE, which will provide UndertowOptions and handling.
```
fl4via, line 98: This needs to be updated, as the RFE is already in Undertow

undertow/WFLY-15452_ajp-listener_allowed_attr_pattern.adoc
```diff
@@ -0,0 +1,121 @@
...
+
+== Security Considerations
+
+Possibly. UNDERTOW-1667 is a CVE, so this RFE should be documented well, in order to warn users of potential exposure.
```
fl4via, line 102: Right, I think it is a good idea to see what was made of CVE original documentation and associate this with that. If we don't have any current documentation for the CVE we should, since it involves configuration of the server, the user must be aware of that.

fl4via (changes requested):

undertow/WFLY-15452_ajp-listener_allowed_attr_pattern.adoc
baranowb: yeah.

undertow/WFLY-15452_ajp-listener_allowed_attr_pattern.adoc
baranowb: yep

undertow/WFLY-15452_ajp-listener_allowed_attr_pattern.adoc
bstansberry: @fl4via

I prefer "standard configuration" when describing the standalone/host/domain.xml files we include in our standard distribution zip.

"Default" is ambiguous because many attributes have default values but those are only relevant is a resource with is configured. Plus, our standard configurations may have a value for an attribute that is different from its default value. So I prefer to only use 'default' to refer to default attribute behavior.

undertow/WFLY-15452_ajp-listener_allowed_attr_pattern.adoc
```diff
@@ -0,0 +1,132 @@
+= [Preview]modify ajp-listener to allow specifying pattern for ajp request attributes
+:author:            Modify ajp-listener to allow specifying AJP_ALLOWED_REQUEST_ATTRIBUTES_PATTERN
```
PrarthonaPaul, line 2: This should be your name :)

undertow/WFLY-15452_ajp-listener_allowed_attr_pattern.adoc
```diff
@@ -0,0 +1,132 @@
...
+
+* [x] OpenShift s2i
+
+* [x] Bootable jar
+
+== Requirements
+
+=== Hard Requirements
+
+* Being able to configure pattern via model/xml.
```
PrarthonaPaul, line 71: It would be good to add the details of the new ajp-listener attribute. What type is it? What are the default values? What do each of those values mean in terms of functionality? Does this attribute need another attribute to be configured?

undertow/WFLY-15452_ajp-listener_allowed_attr_pattern.adoc
PrarthonaPaul: Perhaps the default value (of the new attribute) could be undefined as in when defining the attribute definition, you don't use the `SimpleAttributeDefinitionBuilder.setDefaultValue()` function. And the meaning of that is that there is no `allowed_request_attr_pattern`, even is `ajp-listener` is defined. Or if there is a default value recommended by the specification (or a value that is commonly used by other such solutions), then we can use that as the default value.

And as for the standard configuration, if the `ajp-listener` is not defined by default, the value for `allowed_request_attr_pattern` would also be undefined regardless of what we used for `SimpleAttributeDefinitionBuilder.setDefaultValue()` 

WDYT? @fl4via @baranowb @bstansberry

undertow/WFLY-15452_ajp-listener_allowed_attr_pattern.adoc
```diff
@@ -0,0 +1,132 @@
...
+
+Possibly. Subsystem transformers should be able to handle it.
+
+=== Default Configuration
+
+No change.
+
+=== Importing Existing Configuration
+
+No steps should suffice, as it would mean defaulting to 'null', which is default value in undertow source.
```
PrarthonaPaul, line 97: Minor comment, it would be good to be more specific about what "it" refers to.

undertow/WFLY-15452_ajp-listener_allowed_attr_pattern.adoc
```diff
@@ -0,0 +1,132 @@
...
+
+Done.
+
+== Security Considerations
+
+Possibly. UNDERTOW-1667 is a CVE, so this RFE should be documented well, in order to warn users of potential exposure.
+
+== Test Plan
+
+Unit tests should cover new functionality(there is already test case covering AjpListener).
```
PrarthonaPaul, line 117: Same comment as https://github.com/wildfly/wildfly-proposals/pull/444 regarding tests.

undertow/WFLY-15452_ajp-listener_allowed_attr_pattern.adoc
```diff
@@ -0,0 +1,132 @@
...
+
+== Community Documentation
+
+Task for WFLY documentation already exist - WFLY-15453. HOwever, this is model change and there is model reference doc generated, so its unclear which approach is better?
```
PrarthonaPaul, line 121: Perhaps we can add details abotu the new attribute here: https://docs.wildfly.org/32/Admin_Guide.html#ajp-listener?

undertow/WFLY-15452_ajp-listener_allowed_attr_pattern.adoc
```diff
@@ -0,0 +1,132 @@
...
+
+== Release Note Content
+////
```
PrarthonaPaul, line 124: Please add a short synopsis here that would be added to the release note to inform the reader about the feature.

PrarthonaPaul (changes requested):
Hi @baranowb
The proposal looks good! I have added some minor comments, please feel free to let me know if you have any questions.

undertow/WFLY-15452_ajp-listener_allowed_attr_pattern.adoc
baranowb: "importing existing configuration". This just means that default value will do nothing.

undertow/WFLY-15452_ajp-listener_allowed_attr_pattern.adoc
baranowb: Thats the goal behind WFLY-15453

PrarthonaPaul: Perfect! Thanks

PrarthonaPaul:
Looks good to me! 
Thanks @baranowb

undertow/WFLY-15452_ajp-listener_allowed_attr_pattern.adoc
bstansberry: @PrarthonaPaul looking at https://github.com/wildfly/wildfly/pull/14789/files I see the attribute has no default value, so it looks like what you describe is what was implemented.

AIUI the effect of that is if the ajp-listener is configured with this feature in place the listener behavior is no different from before; i.e. undertow will reject unknown request attributes. So this feature has no 'Default configuration' compatibility concerns.

undertow/WFLY-15452_ajp-listener_allowed_attr_pattern.adoc
bstansberry: @baranowb WFLY-15453 was resolved long ago, so I'm unclear on what the plan is.

undertow/WFLY-15452_ajp-listener_allowed_attr_pattern.adoc
PrarthonaPaul: sounds good! Thanks @bstansberry

fl4via: +1 @bstansberry @PrarthonaPaul

fl4via (approved):

PrarthonaPaul (approved):

bstansberry:
Thanks @baranowb @fl4via and @PrarthonaPaul

undertow/WFLY-15452_ajp-listener_allowed_attr_pattern.adoc
```diff
@@ -0,0 +1,130 @@
...
+
+* https://issues.redhat.com/browse/UNDERTOW-1667[UNDERTOW-1667]
+* https://issues.redhat.com/browse/UNDERTOW-1977[UNDERTOW-1977]
+* https://issues.redhat.com/browse/WFLY-15453[WFLY-15453]
+
+=== Stability Level
+// Choose the planned stability level for the proposed functionality
+* [ ] Experimental
+
+* [X] Preview
```
istraka, line 29: Is this correct? IMO we are aiming for `dafault`  with EAP7-1836.

undertow/WFLY-15452_ajp-listener_allowed_attr_pattern.adoc
```diff
@@ -0,0 +1,130 @@
...
+** Default: null
+** Type: String(regex - java.util.regex.Pattern)
+
+=== Nice-to-Have Requirements
+
+=== Non-Requirements
+
+== Backwards Compatibility
+
+Possibly. Subsystem transformers should be able to handle it.
```
emmartins, line 94: @baranowb @bstansberry may we (Migration feature team) assume that if there is any need for migration this will be handled by the parsers/transformers. "Possibly" is not really reassuring that this feature team is fully handling migration on its own.

emmartins: @baranowb @bstansberry ^^

undertow/WFLY-15452_ajp-listener_allowed_attr_pattern.adoc
```diff
@@ -0,0 +1,130 @@
...
+      <ajp-listener ...  allowed_request_attr_pattern="(?:apple|banana)"  .../>
+      ...
+   </server>
+   ...
+</subsystem>
+
+Parameters will be present in undertow server element(for standalone: /subsystem=undertow/server=default-server/ajp-listener=myListener):
+* allowed_request_attr_pattern
+** Default: null
+** Type: String(regex - java.util.regex.Pattern)
```
istraka, line 86: What should be an outcome when custom attribute doesn't match the pattern? Should it be ignored, but still processed (200) or should the listener return 403?

PR #440, opened 2021-10-20
WFLY-15525 Limit number of records loaded by the JDBC job repository

TomasHofman:
https://issues.redhat.com/browse/WFLY-15525
https://issues.redhat.com/browse/EAP7-1821

### DIFF
--- a/batch/WFLY-15525_Limit_number_of_records_loaded_by_jdbc_store.adoc
+++ b/batch/WFLY-15525_Limit_number_of_records_loaded_by_jdbc_store.adoc
@@ -97,7 +97,7 @@ JBeret and Wildfly code bases need to be modified.
 Currently, the JobRepository interface has method `List<Long> getJobExecutionsByJob(String jobName)`, which returns a list
 of all job execution IDs for given jobName present in the repository.
 
-The proposal is to add a new method `List<Long> getJobExecutionsByJob(String jobName, Long maxRecords)`, that would return
+The proposal is to add a new method `List<Long> getJobExecutionsByJob(String jobName, Integer maxRecords)`, that would return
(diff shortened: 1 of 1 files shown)


batch/WFLY-15525_Limit_number_of_records_loaded_by_jdbc_store.adoc
```diff
@@ -0,0 +1,145 @@
...
+
+JBeret and Wildfly code bases need to be modified.
+
+=== JBeret Core
+
+Currently, the JobRepository interface has method `List<Long> getJobExecutionsByJob(String jobName)`, which returns a list
+of all job execution IDs for given jobName present in the repository.
+
+The proposal is to add a new method `List<Long> getJobExecutionsByJob(String jobName, Long maxRecords)`, that would return
+a list with maximum size of `maxRecords`.
```
chengfang, line 101: Did you mean `int maxRecords`?

What's the range of this attribute? min = 1, max = some reasonably big int?

TomasHofman: I actually wanted `Integer maxRecords`, where null value would mean no limit. It would make implementation little bit simpler, as the original method could just delegate to `getJobExecutionsByJob(jobName, null)`.

I think that min = 1 is OK, but not sure if we should set max. If the user wants to be able to list executions in Web Console, he should stick to couple of hundreds max. If the user wants to be able to load executions via CLI (and e.g. process with `grep`), much higher limits are OK. So I don't know where to draw the line, and for that reason I would not define max.

msve:
The analysis looks good from QE perspective, no comments here

chengfang (approved):

PR #442, opened 2021-10-29
[WFLY-15286] Migrate RESTEasy from 4.x to 5.x

jamezp:
https://issues.redhat.com/browse/WFLY-15286


jamezp:
This is no longer being done. While the same things apply here to the RESTEasy 6.1 upgrade, that is part of the Jakarta EE 10 compliance so I'm closing this.

PR #443, opened 2021-11-01
WFLY-7628 Distributed EJB timers.

pferraro:
https://issues.redhat.com/browse/WFLY-7628

### DIFF
--- a/clustering/ejb/WFLY-7628_Distributed_EJB_Timers.adoc
+++ b/clustering/ejb/WFLY-7628_Distributed_EJB_Timers.adoc
@@ -9,8 +9,10 @@
 == Overview
 
 Currently, WildFly supports 2 mechanisms for persistent EJB timers: a file-based implementation and a JDBC-based implementation.
-Of these, only the JDBC-based implementation is suitable for use in a cluster.   However, this implementation does not scale well to large to medium sized clusters and relies on a single point of failure.
-WildFly needs an persistent timer service implementation suitable for use in a cluster is both highly-available and reasonably scalable.
+Of these, only the JDBC-based implementation is suitable for use in a cluster.
+However, this implementation does not scale well in large to medium sized clusters and relies on a single point of failure due to the blocking nature of the mechanism use to determine which cluster member should schedule a given timer expiration.
+WildFly needs an persistent timer service implementation suitable for use in a cluster that is both highly-available and reasonably scalable.
+This feature will add a new timer service implementation that uses the consistent hash of an Infinispan cache to determine which timers a given cluster member should manage.
 
 == Issue Metadata
 
@@ -20,7 +22,7 @@ WildFly needs an persistent timer service implementation suitable for use in a c
 
 === Related Issues
 
-* https://issues.redhat.com/browse/WFLY-7628[WFLY-7628]
+* https://issues.redhat.com/browse/WFLY-14953[WFLY-14953]
 
 === Dev Contacts
 
@@ -29,21 +31,22 @@ WildFly needs an persistent timer service implementation suitable for use in a c
 === QE Contacts
 
 * mailto:msvehla@redhat.com[Martin Svehla]
+* mailto:fburzigo@redhat.com[Fabio Burzigotti]
 
 === Testing By
-// Put an x in the relevant field to indicate if testing will be done by Engineering or QE. 
-// Discuss with QE during the Kickoff state to decide this
+
 * [ ] Engineering
 
 * [x] QE
 
 === Affected Projects or Components
 
+Clustering, EJB
+
 === Other Interested Projects
 
 === Relevant Installation Types
-// Remove the x next to the relevant field if the feature in question is not relevant
-// to that kind of WildFly installation
+
 * [x] Traditional standalone server (unzipped or provisioned by Galleon)
 
 * [x] Managed domain
@@ -56,87 +59,108 @@ WildFly needs an persistent timer service implementation suitable for use in a c
 
 === Hard Requirements
 
-* Introduce a proper SPI for a TimerService within org.jboss.as.ejb3.timerservice.spi, to facilitate multiple implementations.
-** Adapt the existing TimerResource runtime attributes/operations to use the TimerService SPI instead of referencing specific implementation classes.
-** Adapt existing TimerService implementation to this SPI.
-* Add a CompositeTimerService implementation that delegates to separate persistent and transient TimerService implementations.
-* Implement a distributed implementation of the TimerService SPI.
-** Add a org.wildfly.clustering.ejb.timer SPI to the wildfly-clustering-ejb-spi module to facilitate multiple distributed timer manager implementations.
-** Add a distributed TimerService implementation to the EJB subsystem that delegates to the wildfly-clustering-ejb-spi for timer management and scheduling.
-** Add an Infinispan implementation of the org.wildfly.clustering.ejb.timer API that assigns execution ownership of timers to cluster members based on consistent hashing of the timer identifier.
-*** Leverage the existing Scheduler implementation within wildfly-clustering-ee-cache for timer scheduling
-*** Leverage the existing PrimaryOwnerScheduler decorator that uses consistent hashing of the timer ID to distribute timer execution among cluster members
-* Add a management resource and capability to the distributable-ejb subsystem to configure the Infinispan TimerManagement implementation.
-* Add an persistent-timer-management attribute to the EJB subsystem timer-service resource (as a alternate to the timer-store attribute) that references a TimerManagementProvider capability from the distributable-ejb subsystem.
-** Modify the TimerService DUP to use a CompositeTimerService in the event that the server is configured with a persistent-timer-management, that delegates to the existing TimerServiceImpl for use with transient timers and the DistributedTimerService for use with persistent timers.
+* Create a `TimerService` implementation that delegates to separate persistent vs transient `TimerService` implementations.
+* Implement Infinispan-based `TimerService` for persistent timers that manages timer execution based on the consistent hash of the timer identifier.
+** Make existing ``default-data-store`` attribute of the timer service resource optional.  Use distributable implementation when this attribute is undefined.
+* Add default and named persistent timer management capabilities to distributable-ejb subsystem.
+** Add named Infinispan timer management resource that provides the above capability.
+* Add a distributable-ejb subsystem attribute to determine the server-wide default provider for persistent timer management.
+* Add ability to enable distributed timer management (in lieu of a timer store) within ejb3 subsystem for management of persistent timers.
+* Existing timer service resource operations must continue to work when executed against the distributable `TimerService`.
 
 === Nice-to-Have Requirements
 
-* Add an optional transient-timer-management attribute to the EJB subsystem timer-service resource
-** This would replace the existing TimerService implementation completely, by using a distributed timer service that uses a local, non-persistent cache.
+* Add default and named transient timer management capabilities to distributable-ejb subsystem.
+* Add a distributable-ejb subsystem attribute to determine the server-wide default provider for transient timer management.
+* Add ability to switch between distributed timer management or existing in-memory implementation for management of transient timers
 ** This has the advantage of supporting passivation of timer instances to support applications with a sufficiently large number of transient timers in excess of some configured maximum.
-* Add support to TimerServiceMetaData/TimerServiceMetaDataParser for overriding the persistent timer manager per EJB.
+* Add support to timer service deployment descriptor namespace for overriding persistent/transient timer management per EJB.
 
 === Non-Requirements
 
+* This feature will not change the default timer service behavior or deprecate any existing timer service mechanisms.
+
 == Backwards Compatibility
 
-* Servers started using an old version of the EJB subsystem configuration will continue to behave as normal, using the existing TimerServiceImpl for both transient and persistent timers.
+* Servers started using the default ejb3 subsystem configuration will continue to behave as normal, using the existing in-memory `TimerService` implementation for both transient and persistent timers.
 
 === Default Configuration
 
-The following caches will be added to the default ejb cache-container configuration with the Infinispan subsystem:
+The following caches will be added to the default `ejb` cache-container configuration with the Infinispan subsystem:
 
 For the HA profiles:
 
-	/subsystem=infinispan/cache-container=ejb/distributed-cache=persistent-timers:add()
-	/subsystem=infinispan/cache-container=ejb/distributed-cache=persistent-timers/component=locking:add(isolation=PESSIMISTIC)
-	/subsystem=infinispan/cache-container=ejb/distributed-cache=persistent-timers/component=transaction:add(mode=BATCH)
-	/subsystem=infinispan/cache-container=ejb/distributed-cache=persistent-timers/store=file:add(purge=false)
+	/subsystem=infinispan/cache-container=ejb/distributed-cache=persistent:add()
+	/subsystem=infinispan/cache-container=ejb/distributed-cache=persistent/component=locking:add(isolation=PESSIMISTIC)
+	/subsystem=infinispan/cache-container=ejb/distributed-cache=persistent/component=transaction:add(mode=BATCH)
+	/subsystem=infinispan/cache-container=ejb/distributed-cache=persistent/store=file:add(purge=false)
 
 And for non-HA profiles:
 
-	/subsystem=infinispan/cache-container=ejb/local-cache=persistent-timers:add()
-	/subsystem=infinispan/cache-container=ejb/local-cache=persistent-timers:component=locking:add(isolation=PESSIMISTIC)
-	/subsystem=infinispan/cache-container=ejb/local-cache=persistent-timers:component=transaction:add(mode=BATCH)
-	/subsystem=infinispan/cache-container=ejb/local-cache=persistent-timers/store=file:add(purge=false,passivation=false)
+	/subsystem=infinispan/cache-container=ejb/local-cache=persistent:add()
+	/subsystem=infinispan/cache-container=ejb/local-cache=persistent:component=locking:add(isolation=PESSIMISTIC)
+	/subsystem=infinispan/cache-container=ejb/local-cache=persistent:component=transaction:add(mode=BATCH)
+	/subsystem=infinispan/cache-container=ejb/local-cache=persistent/store=file:add(purge=false, passivation=false)
 
 The following timer management provider will be added to the default distributable-ejb subsystem configuration:
 
-	/subsystem=distributable-ejb/infinispan-timer-management=persistent:add(cache-container=ejb, cache=persistent-timers)
+	/subsystem=distributable-ejb/infinispan-timer-management=persistent:add(cache-container=ejb, cache=persistent)
 
-The default timer-service configuration of the EJB subsystem will be modified as follows:
+To enable use of the distributable timer service, a user will be required to modify the default timer-service configuration of the ejb3 subsystem as follows:
 
 	/subsystem=ejb3/service=timer-service:undefine-attribute(name=default-data-store)
-	/subsystem=ejb3/service=timer-service:write-attribute(name=persistent-timer-management, value=persistent-timers)
 
 === Importing Existing Configuration
 
+N/A
+
 === Deployments
+
+N/A
 
 === Interoperability
 
+N/A
+
 == Implementation Plan
 
-The feature requires WFLY-14953 in order to house the configuration for distributed timer management providers.
+The feature requires the distributable-ejb subsystem (WFLY-14953) in order to house the configuration for distributed timer management providers.
+
+The general implementation plan for this feature is as follows:
+
+* Introduce a proper SPI for a `TimerService` within `org.jboss.as.ejb3.timerservice.spi`, to facilitate multiple implementations.
+** Adapt the existing `TimerResource` runtime attributes/operations to use the `TimerService` SPI instead of referencing specific implementation classes.
+** Adapt existing `TimerService` implementation to this SPI.
+* Create `CompositeTimerService`, a `TimerService` implementation that delegates to separate persistent and transient `TimerService` implementations.
+* Implement an distributed implementation of the `TimerService` SPI.
+** Add a `org.wildfly.clustering.ejb.timer` SPI to the `wildfly-clustering-ejb-spi` module to facilitate multiple distributed timer manager implementations.
+** Create `DistributedTimerService`, a distributed `TimerService` implementation to the ejb3 subsystem that delegates to the `wildfly-clustering-ejb-spi` for timer management and scheduling.
+** Add an Infinispan implementation of the `org.wildfly.clustering.ejb.timer` SPI that assigns execution ownership of timers to cluster members based on consistent hashing of the timer identifier.
+*** Leverage the existing `LocalScheduler` implementation within `wildfly-clustering-ee-cache` for timer scheduling
+*** Leverage the existing `PrimaryOwnerScheduler` decorator that uses consistent hashing of the timer identifier to distribute timer execution among cluster members
+* Add a management resource and capability to the distributable-ejb subsystem to configure the Infinispan timer management implementation.
+* Make the `default-data-store` attribute of the timer service resource optional.
+** Require the default distributable timer management capability when undefined.
+** Modify the `TimerService` DUP to use a `CompositeTimerService` in the event that the server is configured to use persistent timer management, that delegates to the existing `TimerServiceImpl` for use with transient timers and the `DistributableTimerService` for use with persistent timers.
 
 == Security Considerations
 
-////
-Identification if any security implications that may need to be considered with this feature
-or a confirmation that there are no security implications to consider.
-////
+N/A
 
 == Test Plan
 
-* All existing TimerService tests must pass against the new persistent timer implementation, including the EJB TCK.
-* Existing distributed EJB timer tests that use the JDBC timer-store within an EAP cluster may need to be modified since timers may execute on different cluster members than expected by the tests.
-* New tests will be added to the clustering integration testsuite to validate correctness of behavior with a cluster, e.g. failover behavior, etc.
+* All existing `TimerService` tests must pass against the new persistent timer implementation, including the Jakarta Enterprise Beans TCK.
+* New tests will be added to the clustering integration testsuite to validate correctness of behavior of persistent timers with a cluster, including:
+** Validate that auto-timers are only ever created on a single cluster member.
+** Validate normal expiration of auto-timers and manually created single action, interval, and calendar persistent timers within stable cluster.
+** Validate that all timer types continue to trigger expirations without duplication after the cluster member that created them is stopped.
+** Validate that all timer types failback to original cluster member without duplication following restart.
 
 == Community Documentation
-////
-Generally a feature should have documentation as part of the PR to wildfly master, or as a follow up PR if the feature is in wildfly-core. In some cases though the documentation belongs more in a component, or does not need any documentation. Indicate which of these will happen.
-////
+
+* Add section to Jakarta Enterprise Bean documentation that describes the advantages of the distributed Jakarta Enterprise Bean `TimerService` and how to configure the ejb3 subsystem to use it.
+* Add documentation of distributable timer service resources to existing distributable-ejb subsystem documentation.
+
 == Release Note Content
 
 This release adds a new distributed EJB timer implementation that leverages Infinispan to distribute and schedule persistent timers within a cluster, capable of scaling to large clusters.
--- a/clustering/ejb/WFLY-7628_Distributed_EJB_Timers.adoc
+++ b/clustering/ejb/WFLY-7628_Distributed_EJB_Timers.adoc
@@ -65,7 +65,7 @@ Clustering, EJB
 ** Make existing ``default-data-store`` attribute of the timer service resource optional.  Use distributable implementation when this attribute is undefined.
 * Add default and named persistent timer management capabilities to distributable-ejb subsystem.
 ** Add named Infinispan timer management resource that provides the above capability.
-* Add an attribute to the timer service resource within the ejb3 subsystem to determine the server-wide default provider for transient timer management.
+* Add an attribute to the timer service resource within the ejb3 subsystem to determine the server-wide default provider for persistent timer management.
 * Add ability to enable distributed timer management (in lieu of a timer store) within ejb3 subsystem for management of persistent timers.
 * Existing timer service resource operations must continue to work when executed against the distributable `TimerService`.
 
--- a/clustering/ejb/WFLY-7628_Distributed_EJB_Timers.adoc
+++ b/clustering/ejb/WFLY-7628_Distributed_EJB_Timers.adoc
@@ -1,0 +1,142 @@
+= Distributed EJB timers
+:author:            Paul Ferraro
+:email:             paul.m.ferraro@redhat.com
+:toc:               left
+:icons:             font
+:idprefix:
+:idseparator:       -
+
+== Overview
+
+Currently, WildFly supports 2 mechanisms for persistent EJB timers: a file-based implementation and a JDBC-based implementation.
+Of these, only the JDBC-based implementation is suitable for use in a cluster.   However, this implementation does not scale well to large to medium sized clusters and relies on a single point of failure.
+WildFly needs an persistent timer service implementation suitable for use in a cluster is both highly-available and reasonably scalable.
+
+== Issue Metadata
+
+=== Issue
+
+* https://issues.redhat.com/browse/WFLY-7628[WFLY-7628]
+
+=== Related Issues
+
+* https://issues.redhat.com/browse/WFLY-7628[WFLY-7628]
+
+=== Dev Contacts
+
+* mailto:{email}[{author}]
+
+=== QE Contacts
+
+* mailto:msvehla@redhat.com[Martin Svehla]
+
+=== Testing By
+// Put an x in the relevant field to indicate if testing will be done by Engineering or QE. 
+// Discuss with QE during the Kickoff state to decide this
+* [ ] Engineering
+
+* [x] QE
+
+=== Affected Projects or Components
+
+=== Other Interested Projects
+
+=== Relevant Installation Types
+// Remove the x next to the relevant field if the feature in question is not relevant
+// to that kind of WildFly installation
+* [x] Traditional standalone server (unzipped or provisioned by Galleon)
+
+* [x] Managed domain
+
+* [x] OpenShift s2i
+
+* [x] Bootable jar
+
+== Requirements
+
+=== Hard Requirements
+
+* Introduce a proper SPI for a TimerService within org.jboss.as.ejb3.timerservice.spi, to facilitate multiple implementations.
+** Adapt the existing TimerResource runtime attributes/operations to use the TimerService SPI instead of referencing specific implementation classes.
+** Adapt existing TimerService implementation to this SPI.
+* Add a CompositeTimerService implementation that delegates to separate persistent and transient TimerService implementations.
+* Implement a distributed implementation of the TimerService SPI.
+** Add a org.wildfly.clustering.ejb.timer SPI to the wildfly-clustering-ejb-spi module to facilitate multiple distributed timer manager implementations.
+** Add a distributed TimerService implementation to the EJB subsystem that delegates to the wildfly-clustering-ejb-spi for timer management and scheduling.
+** Add an Infinispan implementation of the org.wildfly.clustering.ejb.timer API that assigns execution ownership of timers to cluster members based on consistent hashing of the timer identifier.
+*** Leverage the existing Scheduler implementation within wildfly-clustering-ee-cache for timer scheduling
+*** Leverage the existing PrimaryOwnerScheduler decorator that uses consistent hashing of the timer ID to distribute timer execution among cluster members
+* Add a management resource and capability to the distributable-ejb subsystem to configure the Infinispan TimerManagement implementation.
+* Add an persistent-timer-management attribute to the EJB subsystem timer-service resource (as a alternate to the timer-store attribute) that references a TimerManagementProvider capability from the distributable-ejb subsystem.
+** Modify the TimerService DUP to use a CompositeTimerService in the event that the server is configured with a persistent-timer-management, that delegates to the existing TimerServiceImpl for use with transient timers and the DistributedTimerService for use with persistent timers.
+
+=== Nice-to-Have Requirements
+
+* Add an optional transient-timer-management attribute to the EJB subsystem timer-service resource
+** This would replace the existing TimerService implementation completely, by using a distributed timer service that uses a local, non-persistent cache.
+** This has the advantage of supporting passivation of timer instances to support applications with a sufficiently large number of transient timers in excess of some configured maximum.
+* Add support to TimerServiceMetaData/TimerServiceMetaDataParser for overriding the persistent timer manager per EJB.
+
+=== Non-Requirements
+
+== Backwards Compatibility
+
+* Servers started using an old version of the EJB subsystem configuration will continue to behave as normal, using the existing TimerServiceImpl for both transient and persistent timers.
+
+=== Default Configuration
+
+The following caches will be added to the default ejb cache-container configuration with the Infinispan subsystem:
+
+For the HA profiles:
+
+	/subsystem=infinispan/cache-container=ejb/distributed-cache=persistent-timers:add()
+	/subsystem=infinispan/cache-container=ejb/distributed-cache=persistent-timers/component=locking:add(isolation=PESSIMISTIC)
+	/subsystem=infinispan/cache-container=ejb/distributed-cache=persistent-timers/component=transaction:add(mode=BATCH)
+	/subsystem=infinispan/cache-container=ejb/distributed-cache=persistent-timers/store=file:add(purge=false)
+
+And for non-HA profiles:
+
+	/subsystem=infinispan/cache-container=ejb/local-cache=persistent-timers:add()
+	/subsystem=infinispan/cache-container=ejb/local-cache=persistent-timers:component=locking:add(isolation=PESSIMISTIC)
+	/subsystem=infinispan/cache-container=ejb/local-cache=persistent-timers:component=transaction:add(mode=BATCH)
+	/subsystem=infinispan/cache-container=ejb/local-cache=persistent-timers/store=file:add(purge=false,passivation=false)
+
+The following timer management provider will be added to the default distributable-ejb subsystem configuration:
+
+	/subsystem=distributable-ejb/infinispan-timer-management=persistent:add(cache-container=ejb, cache=persistent-timers)
+
+The default timer-service configuration of the EJB subsystem will be modified as follows:
+
+	/subsystem=ejb3/service=timer-service:undefine-attribute(name=default-data-store)
+	/subsystem=ejb3/service=timer-service:write-attribute(name=persistent-timer-management, value=persistent-timers)
+
+=== Importing Existing Configuration
+
+=== Deployments
+
+=== Interoperability
+
+== Implementation Plan
+
+The feature requires WFLY-14953 in order to house the configuration for distributed timer management providers.
+
+== Security Considerations
+
+////
+Identification if any security implications that may need to be considered with this feature
+or a confirmation that there are no security implications to consider.
+////
+
+== Test Plan
+
+* All existing TimerService tests must pass against the new persistent timer implementation, including the EJB TCK.
+* Existing distributed EJB timer tests that use the JDBC timer-store within an EAP cluster may need to be modified since timers may execute on different cluster members than expected by the tests.
+* New tests will be added to the clustering integration testsuite to validate correctness of behavior with a cluster, e.g. failover behavior, etc.
+
+== Community Documentation
+////
+Generally a feature should have documentation as part of the PR to wildfly master, or as a follow up PR if the feature is in wildfly-core. In some cases though the documentation belongs more in a component, or does not need any documentation. Indicate which of these will happen.
+////
+== Release Note Content
+
+This release adds a new distributed EJB timer implementation that leverages Infinispan to distribute and schedule persistent timers within a cluster, capable of scaling to large clusters.
--- a/clustering/ejb/WFLY-7628_Distributed_EJB_Timers.adoc
+++ b/clustering/ejb/WFLY-7628_Distributed_EJB_Timers.adoc
@@ -91,14 +91,14 @@ The following caches will be added to the default `ejb` cache-container configur
 For the HA profiles:
 
 	/subsystem=infinispan/cache-container=ejb/distributed-cache=persistent:add()
-	/subsystem=infinispan/cache-container=ejb/distributed-cache=persistent/component=locking:add(isolation=PESSIMISTIC)
+	/subsystem=infinispan/cache-container=ejb/distributed-cache=persistent/component=locking:add(isolation=REPEATABLE_READ)
 	/subsystem=infinispan/cache-container=ejb/distributed-cache=persistent/component=transaction:add(mode=BATCH)
 	/subsystem=infinispan/cache-container=ejb/distributed-cache=persistent/store=file:add(purge=false)
 
 And for non-HA profiles:
 
 	/subsystem=infinispan/cache-container=ejb/local-cache=persistent:add()
-	/subsystem=infinispan/cache-container=ejb/local-cache=persistent:component=locking:add(isolation=PESSIMISTIC)
+	/subsystem=infinispan/cache-container=ejb/local-cache=persistent:component=locking:add(isolation=REPEATABLE_READ)
 	/subsystem=infinispan/cache-container=ejb/local-cache=persistent:component=transaction:add(mode=BATCH)
 	/subsystem=infinispan/cache-container=ejb/local-cache=persistent/store=file:add(purge=false, passivation=false)
 
@@ -109,7 +109,7 @@ The following timer management provider will be added to the default distributab
 To enable use of the distributable timer service, a user will be required to modify the default timer-service configuration of the ejb3 subsystem as follows:
 
 	/subsystem=ejb3/service=timer-service:undefine-attribute(name=default-data-store)
-	/subsystem=ejb3/service=timer-service:write-attribute(name=default-persistent-timer-management=persistent)
+	/subsystem=ejb3/service=timer-service:write-attribute(name=default-persistent-timer-management, value=persistent)
 
 === Importing Existing Configuration
 
--- a/clustering/ejb/WFLY-7628_Distributed_EJB_Timers.adoc
+++ b/clustering/ejb/WFLY-7628_Distributed_EJB_Timers.adoc
@@ -23,6 +23,7 @@ This feature will add a new timer service implementation that uses the consisten
 === Related Issues
 
 * https://issues.redhat.com/browse/WFLY-14953[WFLY-14953]
+* https://issues.redhat.com/browse/WFLY-16270[WFLY-16270]
 
 === Dev Contacts
 
--- a/clustering/ejb/WFLY-7628_Distributed_EJB_Timers.adoc
+++ b/clustering/ejb/WFLY-7628_Distributed_EJB_Timers.adoc
@@ -64,14 +64,14 @@ Clustering, EJB
 ** Make existing ``default-data-store`` attribute of the timer service resource optional.  Use distributable implementation when this attribute is undefined.
 * Add default and named persistent timer management capabilities to distributable-ejb subsystem.
 ** Add named Infinispan timer management resource that provides the above capability.
-* Add a distributable-ejb subsystem attribute to determine the server-wide default provider for persistent timer management.
+* Add an attribute to the timer service resource within the ejb3 subsystem to determine the server-wide default provider for transient timer management.
 * Add ability to enable distributed timer management (in lieu of a timer store) within ejb3 subsystem for management of persistent timers.
 * Existing timer service resource operations must continue to work when executed against the distributable `TimerService`.
 
 === Nice-to-Have Requirements
 
 * Add default and named transient timer management capabilities to distributable-ejb subsystem.
-* Add a distributable-ejb subsystem attribute to determine the server-wide default provider for transient timer management.
+* Add an attribute to the timer service resource within the ejb3 subsystem to determine the server-wide default provider for transient timer management.
 * Add ability to switch between distributed timer management or existing in-memory implementation for management of transient timers
 ** This has the advantage of supporting passivation of timer instances to support applications with a sufficiently large number of transient timers in excess of some configured maximum.
 * Add support to timer service deployment descriptor namespace for overriding persistent/transient timer management per EJB.
@@ -109,6 +109,7 @@ The following timer management provider will be added to the default distributab
 To enable use of the distributable timer service, a user will be required to modify the default timer-service configuration of the ejb3 subsystem as follows:
 
 	/subsystem=ejb3/service=timer-service:undefine-attribute(name=default-data-store)
+	/subsystem=ejb3/service=timer-service:write-attribute(name=default-persistent-timer-management=persistent)
 
 === Importing Existing Configuration
 
@@ -140,7 +141,8 @@ The general implementation plan for this feature is as follows:
 *** Leverage the existing `PrimaryOwnerScheduler` decorator that uses consistent hashing of the timer identifier to distribute timer execution among cluster members
 * Add a management resource and capability to the distributable-ejb subsystem to configure the Infinispan timer management implementation.
 * Make the `default-data-store` attribute of the timer service resource optional.
-** Require the default distributable timer management capability when undefined.
+* Add an `default-persistent-timer-management` attribute to the timer service resource, as an alternative to `default-data-store`.
+** Require the named distributable timer management capability when defined.
 ** Modify the `TimerService` DUP to use a `CompositeTimerService` in the event that the server is configured to use persistent timer management, that delegates to the existing `TimerServiceImpl` for use with transient timers and the `DistributableTimerService` for use with persistent timers.
 
 == Security Considerations
--- a/clustering/ejb/WFLY-7628_Distributed_EJB_Timers.adoc
+++ b/clustering/ejb/WFLY-7628_Distributed_EJB_Timers.adoc
@@ -10,7 +10,7 @@
 
 Currently, WildFly supports 2 mechanisms for persistent EJB timers: a file-based implementation and a JDBC-based implementation.
 Of these, only the JDBC-based implementation is suitable for use in a cluster.
-However, this implementation does not scale well in large to medium sized clusters and relies on a single point of failure due to the blocking nature of the mechanism use to determine which cluster member should schedule a given timer expiration.
+However, this implementation does not scale well in large to medium sized clusters and relies on a single point of failure due to the blocking nature of the mechanism employed to determine which cluster member should schedule a given timer expiration.
 WildFly needs an persistent timer service implementation suitable for use in a cluster that is both highly-available and reasonably scalable.
 This feature will add a new timer service implementation that uses the consistent hash of an Infinispan cache to determine which timers a given cluster member should manage.
 


pferraro:
@chengfang Please review.

clustering/ejb/WFLY-7628_Distributed_EJB_Timers.adoc
```diff
@@ -0,0 +1,142 @@
...
+	/subsystem=infinispan/cache-container=ejb/local-cache=persistent-timers/store=file:add(purge=false,passivation=false)
+
+The following timer management provider will be added to the default distributable-ejb subsystem configuration:
+
+	/subsystem=distributable-ejb/infinispan-timer-management=persistent:add(cache-container=ejb, cache=persistent-timers)
+
+The default timer-service configuration of the EJB subsystem will be modified as follows:
+
+	/subsystem=ejb3/service=timer-service:undefine-attribute(name=default-data-store)
+	/subsystem=ejb3/service=timer-service:write-attribute(name=persistent-timer-management, value=persistent-timers)
```
chengfang, line 111: Does that mean timer-service will support both the existing `default-data-store` and the new `persistent-timer-management` attributes, and the infinispan-based `persistent-timer-management` will be the default persistence option?

pferraro: I intended persistent-timer-management to be an alternative attribute to default-data-store, meaning that users can define one or the other, but not both.
When persistent-timer-management is defined, then persistent timers will delegate to a timer manager as defined in the distributable-ejb subsystem.
When a default-data-store is defined instead (meaning that persistent-timer-management must be undefined), the user gets the current behavior.

I don't think we want to change the default configuration within the context of this feature (this can be an independent decision), but if we did, this is what it would look like.

clustering/ejb/WFLY-7628_Distributed_EJB_Timers.adoc
```diff
@@ -0,0 +1,142 @@
...
+** Adapt existing TimerService implementation to this SPI.
+* Add a CompositeTimerService implementation that delegates to separate persistent and transient TimerService implementations.
+* Implement a distributed implementation of the TimerService SPI.
+** Add a org.wildfly.clustering.ejb.timer SPI to the wildfly-clustering-ejb-spi module to facilitate multiple distributed timer manager implementations.
+** Add a distributed TimerService implementation to the EJB subsystem that delegates to the wildfly-clustering-ejb-spi for timer management and scheduling.
+** Add an Infinispan implementation of the org.wildfly.clustering.ejb.timer API that assigns execution ownership of timers to cluster members based on consistent hashing of the timer identifier.
+*** Leverage the existing Scheduler implementation within wildfly-clustering-ee-cache for timer scheduling
+*** Leverage the existing PrimaryOwnerScheduler decorator that uses consistent hashing of the timer ID to distribute timer execution among cluster members
+* Add a management resource and capability to the distributable-ejb subsystem to configure the Infinispan TimerManagement implementation.
```
chengfang, line 69: I had a vague idea to change timer id from string to int (WFLY-15568). Is this doable in infinispan-based distributed timer service impl? The switch will be a long-term consideration, but it will be good to have the flexibility for the potential change of id type. I expect int id will have meaningful performance advantage in both database and infinispan-based timer service impl.

clustering/ejb/WFLY-7628_Distributed_EJB_Timers.adoc
pferraro: All of the timer management services are installed on-demand, so If the timer service is configured to use file or database store in the ejb subsystem, nothing related to the infinispan implementation is started.

clustering/ejb/WFLY-7628_Distributed_EJB_Timers.adoc
pferraro: The infinispan-based timer service currently uses UUIDs, as these need to be unique across cluster members (rather than a counter).  Those UUIDs are then converted to strings for use by the management API.

chengfang (approved):

clustering/ejb/WFLY-7628_Distributed_EJB_Timers.adoc
```diff
@@ -0,0 +1,142 @@
...
+:icons:             font
+:idprefix:
+:idseparator:       -
+
+== Overview
+
+Currently, WildFly supports 2 mechanisms for persistent EJB timers: a file-based implementation and a JDBC-based implementation.
+Of these, only the JDBC-based implementation is suitable for use in a cluster.   However, this implementation does not scale well to large to medium sized clusters and relies on a single point of failure.
+WildFly needs an persistent timer service implementation suitable for use in a cluster is both highly-available and reasonably scalable.
+
```
fabiobrz, line 14: What would mean for an end user to adopt this feature? Both from the developer and the administration perspective? e.g.: how is existing code/developer experience affected? How would an admin move from a current JDBC based solution to the new one?

Of course this cannot be detailed here, but a high level description, examples code or commands would definitely help the reader to have a clear vision of the scope of what would mean to migrate to this feature.

Also, it would be definitely important to have some, even very short and simple, consideration - literally from you - about how this feature could affect perfomance, since there are some issues in the distributed EJB timers portfolio around which happen(ed) to pop up in such context (e.g. high load). This would be of inspiration for more complex testing scenarios.

fabiobrz (changes requested):
Hi @pferraro - Thanks for this proposal. 
I went through the document and added some comments and questions. 
Let me know what you think.

clustering/ejb/WFLY-7628_Distributed_EJB_Timers.adoc
```diff
@@ -0,0 +1,142 @@
...
+== Issue Metadata
+
+=== Issue
+
+* https://issues.redhat.com/browse/WFLY-7628[WFLY-7628]
+
+=== Related Issues
+
+* https://issues.redhat.com/browse/WFLY-7628[WFLY-7628]
```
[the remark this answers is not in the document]
pferraro: This is a public document.  Only public JIRAs are to be linked here.  I will include WFLY-14953.

clustering/ejb/WFLY-7628_Distributed_EJB_Timers.adoc
```diff
@@ -0,0 +1,142 @@
...
+
+=== Dev Contacts
+
+* mailto:{email}[{author}]
+
+=== QE Contacts
+
+* mailto:msvehla@redhat.com[Martin Svehla]
```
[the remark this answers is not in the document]
pferraro: Will do.

clustering/ejb/WFLY-7628_Distributed_EJB_Timers.adoc
```diff
@@ -0,0 +1,142 @@
...
+////
+
+== Test Plan
+
+* All existing TimerService tests must pass against the new persistent timer implementation, including the EJB TCK.
+* Existing distributed EJB timer tests that use the JDBC timer-store within an EAP cluster may need to be modified since timers may execute on different cluster members than expected by the tests.
+* New tests will be added to the clustering integration testsuite to validate correctness of behavior with a cluster, e.g. failover behavior, etc.
+
+== Community Documentation
+////
```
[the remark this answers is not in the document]
pferraro: Will do.

clustering/ejb/WFLY-7628_Distributed_EJB_Timers.adoc
```diff
@@ -0,0 +1,142 @@
...
+== Implementation Plan
+
+The feature requires WFLY-14953 in order to house the configuration for distributed timer management providers.
+
+== Security Considerations
+
+////
+Identification if any security implications that may need to be considered with this feature
+or a confirmation that there are no security implications to consider.
+////
```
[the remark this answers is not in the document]
pferraro: OK.

clustering/ejb/WFLY-7628_Distributed_EJB_Timers.adoc
```diff
@@ -0,0 +1,142 @@
...
+
+=== Importing Existing Configuration
+
+=== Deployments
+
+=== Interoperability
+
+== Implementation Plan
+
+The feature requires WFLY-14953 in order to house the configuration for distributed timer management providers.
```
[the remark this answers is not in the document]
pferraro: It will be added to the related issues section.

clustering/ejb/WFLY-7628_Distributed_EJB_Timers.adoc
```diff
@@ -0,0 +1,142 @@
...
+** Add a org.wildfly.clustering.ejb.timer SPI to the wildfly-clustering-ejb-spi module to facilitate multiple distributed timer manager implementations.
+** Add a distributed TimerService implementation to the EJB subsystem that delegates to the wildfly-clustering-ejb-spi for timer management and scheduling.
+** Add an Infinispan implementation of the org.wildfly.clustering.ejb.timer API that assigns execution ownership of timers to cluster members based on consistent hashing of the timer identifier.
+*** Leverage the existing Scheduler implementation within wildfly-clustering-ee-cache for timer scheduling
+*** Leverage the existing PrimaryOwnerScheduler decorator that uses consistent hashing of the timer ID to distribute timer execution among cluster members
+* Add a management resource and capability to the distributable-ejb subsystem to configure the Infinispan TimerManagement implementation.
+* Add an persistent-timer-management attribute to the EJB subsystem timer-service resource (as a alternate to the timer-store attribute) that references a TimerManagementProvider capability from the distributable-ejb subsystem.
```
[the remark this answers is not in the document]
pferraro: The distributable-ejb subsytem will expose a TimerManagementProvider capability.  `TimerManagement` above should read `TimerManagementProvider`.  Will fix.

clustering/ejb/WFLY-7628_Distributed_EJB_Timers.adoc
```diff
@@ -0,0 +1,142 @@
...
+* [x] Bootable jar
+
+== Requirements
+
+=== Hard Requirements
+
+* Introduce a proper SPI for a TimerService within org.jboss.as.ejb3.timerservice.spi, to facilitate multiple implementations.
+** Adapt the existing TimerResource runtime attributes/operations to use the TimerService SPI instead of referencing specific implementation classes.
+** Adapt existing TimerService implementation to this SPI.
+* Add a CompositeTimerService implementation that delegates to separate persistent and transient TimerService implementations.
```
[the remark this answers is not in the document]
pferraro: None of the existing timer service implementations will be removed as part of this feature - so backwards compatibility is not an issue.

clustering/ejb/WFLY-7628_Distributed_EJB_Timers.adoc
```diff
@@ -0,0 +1,142 @@
...
+* Implement a distributed implementation of the TimerService SPI.
+** Add a org.wildfly.clustering.ejb.timer SPI to the wildfly-clustering-ejb-spi module to facilitate multiple distributed timer manager implementations.
+** Add a distributed TimerService implementation to the EJB subsystem that delegates to the wildfly-clustering-ejb-spi for timer management and scheduling.
```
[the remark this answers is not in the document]
pferraro: 'subsystem` is the correct term here.  An `extension` is a mechanism for adding a `subsystem`.

clustering/ejb/WFLY-7628_Distributed_EJB_Timers.adoc
```diff
@@ -0,0 +1,142 @@
...
+** Add an Infinispan implementation of the org.wildfly.clustering.ejb.timer API that assigns execution ownership of timers to cluster members based on consistent hashing of the timer identifier.
```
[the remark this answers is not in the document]
pferraro: Yes - will fix.

clustering/ejb/WFLY-7628_Distributed_EJB_Timers.adoc
```diff
@@ -0,0 +1,142 @@
...
+*** Leverage the existing Scheduler implementation within wildfly-clustering-ee-cache for timer scheduling
+*** Leverage the existing PrimaryOwnerScheduler decorator that uses consistent hashing of the timer ID to distribute timer execution among cluster members
+* Add a management resource and capability to the distributable-ejb subsystem to configure the Infinispan TimerManagement implementation.
+* Add an persistent-timer-management attribute to the EJB subsystem timer-service resource (as a alternate to the timer-store attribute) that references a TimerManagementProvider capability from the distributable-ejb subsystem.
+** Modify the TimerService DUP to use a CompositeTimerService in the event that the server is configured with a persistent-timer-management, that delegates to the existing TimerServiceImpl for use with transient timers and the DistributedTimerService for use with persistent timers.
```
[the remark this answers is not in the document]
pferraro: `DistributedTimerService` is the core implementation class of this feature (within the ejb subsystem).

clustering/ejb/WFLY-7628_Distributed_EJB_Timers.adoc
```diff
@@ -0,0 +1,142 @@
...
+
+=== Nice-to-Have Requirements
+
+* Add an optional transient-timer-management attribute to the EJB subsystem timer-service resource
```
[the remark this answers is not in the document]
pferraro: Good point.

clustering/ejb/WFLY-7628_Distributed_EJB_Timers.adoc
```diff
@@ -0,0 +1,142 @@
...
+** This would replace the existing TimerService implementation completely, by using a distributed timer service that uses a local, non-persistent cache.
+** This has the advantage of supporting passivation of timer instances to support applications with a sufficiently large number of transient timers in excess of some configured maximum.
+* Add support to TimerServiceMetaData/TimerServiceMetaDataParser for overriding the persistent timer manager per EJB.
+
+=== Non-Requirements
+
+== Backwards Compatibility
+
+* Servers started using an old version of the EJB subsystem configuration will continue to behave as normal, using the existing TimerServiceImpl for both transient and persistent timers.
```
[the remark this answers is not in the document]
pferraro: Yes.

clustering/ejb/WFLY-7628_Distributed_EJB_Timers.adoc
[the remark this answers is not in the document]
pferraro: This document was written with the assumption that WFLY-14953 is complete.

clustering/ejb/WFLY-7628_Distributed_EJB_Timers.adoc
```diff
@@ -0,0 +1,142 @@
...
+
+=== Default Configuration
+
+The following caches will be added to the default ejb cache-container configuration with the Infinispan subsystem:
```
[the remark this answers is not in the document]
pferraro: Will fix.

clustering/ejb/WFLY-7628_Distributed_EJB_Timers.adoc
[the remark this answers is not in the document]
fabiobrz: My question, which is meant to understand the technical details, is whether is it `CompositeTimerService` which will allow for existing application relying on the JDBC based pesistence to continue working transparently.

clustering/ejb/WFLY-7628_Distributed_EJB_Timers.adoc
```diff
@@ -0,0 +1,142 @@
...
+== Security Considerations
+
+////
+Identification if any security implications that may need to be considered with this feature
+or a confirmation that there are no security implications to consider.
+////
+
+== Test Plan
+
+* All existing TimerService tests must pass against the new persistent timer implementation, including the EJB TCK.
```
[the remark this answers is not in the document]
pferraro: The Jakarta EE TCK is very coarse grained (i.e. by spec, not feature), so the EJB TCK (or, rather the Jakarta Enterprise Bean TCK (I will fix references to that)) is relevant module.

clustering/ejb/WFLY-7628_Distributed_EJB_Timers.adoc
```diff
@@ -0,0 +1,142 @@
...
+* Existing distributed EJB timer tests that use the JDBC timer-store within an EAP cluster may need to be modified since timers may execute on different cluster members than expected by the tests.
```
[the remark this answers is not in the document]
pferraro: I did not change the existing test (there is only 1 which only runs against a single server), but rather added a separate integration test to the clustering testsuite that validates timer failover.

clustering/ejb/WFLY-7628_Distributed_EJB_Timers.adoc
```diff
@@ -0,0 +1,142 @@
...
+
+	/subsystem=ejb3/service=timer-service:undefine-attribute(name=default-data-store)
+	/subsystem=ejb3/service=timer-service:write-attribute(name=persistent-timer-management, value=persistent-timers)
+
+=== Importing Existing Configuration
+
+=== Deployments
+
+=== Interoperability
+
```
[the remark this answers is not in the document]
pferraro: I do not foresee any differences.

clustering/ejb/WFLY-7628_Distributed_EJB_Timers.adoc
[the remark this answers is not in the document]
fabiobrz: Ok thanks, then the documentation is a little bit confusing, at least to me.

[1]
https://docs.wildfly.org/26/Admin_Guide.html#General_configuration_concepts

clustering/ejb/WFLY-7628_Distributed_EJB_Timers.adoc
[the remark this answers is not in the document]
fabiobrz: Thanks, then please add a couple of word about that in the document itself, as you did here. Otherwise the reader will keep wondering and try looking for it in the sources in order to understand this technical (and valuable) detail.

clustering/ejb/WFLY-7628_Distributed_EJB_Timers.adoc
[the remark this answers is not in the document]
fabiobrz: Good to know that, thanks.

clustering/ejb/WFLY-7628_Distributed_EJB_Timers.adoc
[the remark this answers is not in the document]
fabiobrz: I see, thanks. But my point is about mapping requirements, and specifically this one, to the related RFE, and this is another valuable technical detail for those who read.

clustering/ejb/WFLY-7628_Distributed_EJB_Timers.adoc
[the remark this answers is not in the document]
fabiobrz: Ok, thanks. Could please add this details to the document itself? Also, the involved testsuite module should be mentioned (better linked) here, since it is not implied - nor required - to look at the PR for understanding what tests will be affected. Indeed, it should be made visible here.

clustering/ejb/WFLY-7628_Distributed_EJB_Timers.adoc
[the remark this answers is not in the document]
fabiobrz: Thanks for the feedback. Could you please add a link to that TCK? I cannot find it in the WildFly codebase and I cannot understand the way these tests are executed.

clustering/ejb/WFLY-7628_Distributed_EJB_Timers.adoc
pferraro: > What would mean for an end user to adopt this feature? Both from the developer and the administration perspective? e.g.: how is existing code/developer experience affected? How would an admin move from a current JDBC based solution to the new one?

The developer experience is identical.
The administrator experience is simpler, since there would no longer be the need to provision and configure a database resource.

> Of course this cannot be detailed here, but a high level description, examples code or commands would definitely help the reader to have a clear vision of the scope of what would mean to migrate to this feature.

This is included in the community documentation.

> Also, it would be definitely important to have some, even very short and simple, consideration - literally from you - about how this feature could affect performance,

The root cause of the performance issues with the existing database timer-store implementation is that it relies on blocking mechanisms. Timer "ownership" is not deterministic, but rather is determined by which EAP instance is able to obtain an exclusive lock on a given table entry.  While the instance that is able to lock the row will return immediately, due to the nature of the JDBC API and relational database locking in general, instances contending for the same row lock will block.  When an instance eventually does obtain the lock, the associated timer will have already scheduled it, and thus each instance will try to lock the next timer, and so on.  For obvious reasons, this does not scale.

We solve this in the new implementation be making timer ownership deterministic, i.e. an intrinsic function of the cluster topology and the timer identifier; in the same manner that EAP manages all other distributed state.  Thus the new implementation does not rely on blocking mechanisms in order to schedule timers.

Memory footprint issues are mitigated through the use of Inifnispan cache passivation.

> since there are some issues in the distributed EJB timers portfolio around which happen(ed) to pop up in such context (e.g. poor performance if high throughput and memory footprint if high load). This would be of inspiration for more complex testing scenarios.

I anticipate that the same QE tests used to reproduce these performance issues can also be used to validate that the new implementation resolves them.

clustering/ejb/WFLY-7628_Distributed_EJB_Timers.adoc
pferraro: >  as you stated, changing the default configuration sounds like a separate feature or it should at least be mentioned and documented (e.g.: release notes).

I am not proposing changing the default configuration as part of this feature proposal.

> As far as this analysis document is concerned, I think that it should be clarified that this is a proposal, since it now reads as an implementation detail.

I'm not sure what you mean.  Can you elaborate?

clustering/ejb/WFLY-7628_Distributed_EJB_Timers.adoc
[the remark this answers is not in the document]
pferraro: I'm talking about the standard Jakarta EE TCK: https://github.com/eclipse-ee4j/jakartaee-tck

I also misspoke, the TCK is not composed of any modules, but rather one can run a set of tests within a given directory (as with any junit testsuite).

clustering/ejb/WFLY-7628_Distributed_EJB_Timers.adoc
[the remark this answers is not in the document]
pferraro: I'm talking about the standard Jakarta EE TCK: https://github.com/eclipse-ee4j/jakartaee-tck

I also misspoke, the TCK is not composed of any modules, but rather one can run a set of tests within a given directory (as with any junit testsuite).

clustering/ejb/WFLY-7628_Distributed_EJB_Timers.adoc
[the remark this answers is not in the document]
pferraro: No existing tests will be affected, but I'll better describe the test scenarios being added to the clustering integration testsuite.

clustering/ejb/WFLY-7628_Distributed_EJB_Timers.adoc
pferraro: > Thanks for sharing these details. I think they add great value and I am wondering if there's any reason for not including the above explanation into the document itself. WDYT?

The justification of this feature has already been discussed in the associated JIRAs.  The purpose of this document is to analyze the feature requirements.

pferraro: > Ok, thanks, then could you please mention this when you'll add the Community docs section contents?

Will do.

clustering/ejb/WFLY-7628_Distributed_EJB_Timers.adoc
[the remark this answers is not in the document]
pferraro: If the application is configured to use a JDBC-based timer store, then the current TimerService implementation is used, not the `CompositeTimerService`, which would only be required when persistent and transient timers are managed separately.

clustering/ejb/WFLY-7628_Distributed_EJB_Timers.adoc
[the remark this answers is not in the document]
pferraro: It will be linked in the "Related Issues" section.

clustering/ejb/WFLY-7628_Distributed_EJB_Timers.adoc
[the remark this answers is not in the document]
pferraro: No problem.

clustering/ejb/WFLY-7628_Distributed_EJB_Timers.adoc
```diff
@@ -0,0 +1,142 @@
...
+* Add a management resource and capability to the distributable-ejb subsystem to configure the Infinispan TimerManagement implementation.
+* Add an persistent-timer-management attribute to the EJB subsystem timer-service resource (as a alternate to the timer-store attribute) that references a TimerManagementProvider capability from the distributable-ejb subsystem.
+** Modify the TimerService DUP to use a CompositeTimerService in the event that the server is configured with a persistent-timer-management, that delegates to the existing TimerServiceImpl for use with transient timers and the DistributedTimerService for use with persistent timers.
+
+=== Nice-to-Have Requirements
+
+* Add an optional transient-timer-management attribute to the EJB subsystem timer-service resource
+** This would replace the existing TimerService implementation completely, by using a distributed timer service that uses a local, non-persistent cache.
+** This has the advantage of supporting passivation of timer instances to support applications with a sufficiently large number of transient timers in excess of some configured maximum.
```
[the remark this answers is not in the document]
pferraro: No, it is a "nice-to-have requirement".

pferraro:
@fabiobrz I've updated this doc with your feedback incorporated.  Please review.

clustering/ejb/WFLY-7628_Distributed_EJB_Timers.adoc
```diff
@@ -56,87 +59,108 @@ WildFly needs an persistent timer service implementation suitable for use in a c
...
-* Add a management resource and capability to the distributable-ejb subsystem to configure the Infinispan TimerManagement implementation.
-* Add an persistent-timer-management attribute to the EJB subsystem timer-service resource (as a alternate to the timer-store attribute) that references a TimerManagementProvider capability from the distributable-ejb subsystem.
-** Modify the TimerService DUP to use a CompositeTimerService in the event that the server is configured with a persistent-timer-management, that delegates to the existing TimerServiceImpl for use with transient timers and the DistributedTimerService for use with persistent timers.
+* Create a `TimerService` implementation that delegates to separate persistent vs transient `TimerService` implementations.
+* Implement Infinispan-based `TimerService` for persistent timers that manages timer execution based on the consistent hash of the timer identifier.
+** Make existing ``default-data-store`` attribute of the timer service resource optional.  Use distributable implementation when this attribute is undefined.
+* Add default and named persistent timer management capabilities to distributable-ejb subsystem.
```
fabiobrz, line 65: s/distributable-ejb/the new distributable-ejb/ ?

clustering/ejb/WFLY-7628_Distributed_EJB_Timers.adoc
```diff
@@ -56,87 +59,108 @@ WildFly needs an persistent timer service implementation suitable for use in a c
...
-* Add a management resource and capability to the distributable-ejb subsystem to configure the Infinispan TimerManagement implementation.
-* Add an persistent-timer-management attribute to the EJB subsystem timer-service resource (as a alternate to the timer-store attribute) that references a TimerManagementProvider capability from the distributable-ejb subsystem.
-** Modify the TimerService DUP to use a CompositeTimerService in the event that the server is configured with a persistent-timer-management, that delegates to the existing TimerServiceImpl for use with transient timers and the DistributedTimerService for use with persistent timers.
+* Create a `TimerService` implementation that delegates to separate persistent vs transient `TimerService` implementations.
+* Implement Infinispan-based `TimerService` for persistent timers that manages timer execution based on the consistent hash of the timer identifier.
+** Make existing ``default-data-store`` attribute of the timer service resource optional.  Use distributable implementation when this attribute is undefined.
```
fabiobrz, line 64: This would mean that a new default server behavior would originate from this new feature in case it is configured to persist EJB timers and `default-data-store` is omitted, thus depending on the Infinispan subsystem configuration, right?

clustering/ejb/WFLY-7628_Distributed_EJB_Timers.adoc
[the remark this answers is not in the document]
fabiobrz: Ok thanks, but will it be a new feature in case the _nice to have_ requirement is implemented?

clustering/ejb/WFLY-7628_Distributed_EJB_Timers.adoc
```diff
@@ -56,87 +59,108 @@ WildFly needs an persistent timer service implementation suitable for use in a c
...
-	/subsystem=infinispan/cache-container=ejb/local-cache=persistent-timers/store=file:add(purge=false,passivation=false)
+	/subsystem=infinispan/cache-container=ejb/local-cache=persistent:add()
+	/subsystem=infinispan/cache-container=ejb/local-cache=persistent:component=locking:add(isolation=PESSIMISTIC)
+	/subsystem=infinispan/cache-container=ejb/local-cache=persistent:component=transaction:add(mode=BATCH)
+	/subsystem=infinispan/cache-container=ejb/local-cache=persistent/store=file:add(purge=false, passivation=false)
 
 The following timer management provider will be added to the default distributable-ejb subsystem configuration:
 
-	/subsystem=distributable-ejb/infinispan-timer-management=persistent:add(cache-container=ejb, cache=persistent-timers)
+	/subsystem=distributable-ejb/infinispan-timer-management=persistent:add(cache-container=ejb, cache=persistent)
```
fabiobrz, line 107: So from what I understand, what above will be added to the server default configuration, while what follows must be set by the user to enable the feature. Is this correct or am I missing something?

clustering/ejb/WFLY-7628_Distributed_EJB_Timers.adoc
```diff
@@ -56,87 +59,108 @@ WildFly needs an persistent timer service implementation suitable for use in a c
...
-or a confirmation that there are no security implications to consider.
-////
+N/A
 
 == Test Plan
 
-* All existing TimerService tests must pass against the new persistent timer implementation, including the EJB TCK.
-* Existing distributed EJB timer tests that use the JDBC timer-store within an EAP cluster may need to be modified since timers may execute on different cluster members than expected by the tests.
-* New tests will be added to the clustering integration testsuite to validate correctness of behavior with a cluster, e.g. failover behavior, etc.
+* All existing `TimerService` tests must pass against the new persistent timer implementation, including the Jakarta Enterprise Beans TCK.
```
fabiobrz, line 152: Is the Jakarta Enterprise Beans TCK execution bound to any WildFly CI execution, like e.g. it happens for the Microprofile TCKs? Basically, in which context/domain will these tests be executed?

clustering/ejb/WFLY-7628_Distributed_EJB_Timers.adoc
```diff
@@ -56,87 +59,108 @@ WildFly needs an persistent timer service implementation suitable for use in a c
...
+* New tests will be added to the clustering integration testsuite to validate correctness of behavior of persistent timers with a cluster, including:
+** Validate that auto-timers are only ever created on a single cluster member.
+** Validate normal expiration of auto-timers and manually created single action, interval, and calendar persistent timers within stable cluster.
+** Validate that all timer types continue to trigger expirations without duplication after the cluster member that created them is stopped.
+** Validate that all timer types failback to original cluster member without duplication following restart.
```
fabiobrz, line 157: The last two bullet should cover failover scenarios, correct?

clustering/ejb/WFLY-7628_Distributed_EJB_Timers.adoc
```diff
@@ -56,87 +59,108 @@ WildFly needs an persistent timer service implementation suitable for use in a c
...
+	/subsystem=infinispan/cache-container=ejb/local-cache=persistent:component=transaction:add(mode=BATCH)
+	/subsystem=infinispan/cache-container=ejb/local-cache=persistent/store=file:add(purge=false, passivation=false)
 
 The following timer management provider will be added to the default distributable-ejb subsystem configuration:
 
-	/subsystem=distributable-ejb/infinispan-timer-management=persistent:add(cache-container=ejb, cache=persistent-timers)
+	/subsystem=distributable-ejb/infinispan-timer-management=persistent:add(cache-container=ejb, cache=persistent)
 
-The default timer-service configuration of the EJB subsystem will be modified as follows:
+To enable use of the distributable timer service, a user will be required to modify the default timer-service configuration of the ejb3 subsystem as follows:
```
fabiobrz, line 109: Looking at the following line below: 

`/subsystem=ejb3/service=timer-service:undefine-attribute(name=default-data-store)`

previously it is stated as a requirement that this attribute should be made optional, so why would the user be requested to unset it?

fabiobrz (changes requested):
Hi @pferraro, and thanks for addressing my comments.
I've dropped some more questions and considerations, and I'd appreciate your feedback.

clustering/ejb/WFLY-7628_Distributed_EJB_Timers.adoc
fabiobrz: BTW - one more question on this. How will be the configuration tested? i.e. attributes and operations on new resources?

clustering/ejb/WFLY-7628_Distributed_EJB_Timers.adoc
pferraro: When `default-data-store` is undefined, the default behavior for persistent timers will be determined via the distributable-ejb subsystem configuration.

clustering/ejb/WFLY-7628_Distributed_EJB_Timers.adoc
pferraro: The word *new* implies that the subsystem is new to this feature.  It is not, hence the omission.

clustering/ejb/WFLY-7628_Distributed_EJB_Timers.adoc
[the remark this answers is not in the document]
pferraro: No - this is all one feature.
_Requirements_ represent the scope of the feature.
_Hard requirements_ are requirements essential to the feature (think "MVP")
_Nice-to-have requirements_ are requirements of the feature that should or could be implemented if possible, but are non-essential.
This particular requirement has a high value add, and a low level of effort, thus will almost certainly be included.

clustering/ejb/WFLY-7628_Distributed_EJB_Timers.adoc
pferraro: Correct.

clustering/ejb/WFLY-7628_Distributed_EJB_Timers.adoc
pferraro: Because this attribute is defined in the default ejb3 subsystem configuration.

clustering/ejb/WFLY-7628_Distributed_EJB_Timers.adoc
pferraro: The execution of the Jakarta EE TCK is completely external to WildFly.  @scottmarlow can answer any specific questions you have about this.

clustering/ejb/WFLY-7628_Distributed_EJB_Timers.adoc
pferraro: > The last two bullet should cover failover scenarios, correct?

Yes.

> How will be the configuration tested? i.e. attributes and operations on new resources?

These are covered by the distributable-ejb subsystem tests.

clustering/ejb/WFLY-7628_Distributed_EJB_Timers.adoc
```diff
@@ -9,8 +9,10 @@
 == Overview
 
 Currently, WildFly supports 2 mechanisms for persistent EJB timers: a file-based implementation and a JDBC-based implementation.
-Of these, only the JDBC-based implementation is suitable for use in a cluster.   However, this implementation does not scale well to large to medium sized clusters and relies on a single point of failure.
-WildFly needs an persistent timer service implementation suitable for use in a cluster is both highly-available and reasonably scalable.
+Of these, only the JDBC-based implementation is suitable for use in a cluster.
+However, this implementation does not scale well in large to medium sized clusters and relies on a single point of failure due to the blocking nature of the mechanism use to determine which cluster member should schedule a given timer expiration.
```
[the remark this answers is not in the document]
pferraro: Fixed.

clustering/ejb/WFLY-7628_Distributed_EJB_Timers.adoc
fabiobrz: Ok, thanks

clustering/ejb/WFLY-7628_Distributed_EJB_Timers.adoc
fabiobrz: Thanks for clarifying

clustering/ejb/WFLY-7628_Distributed_EJB_Timers.adoc
fabiobrz: Thanks @pferraro - Hi @scottmarlow, my point here is to understand which environment/domain/context should be checked in oder to verify the execution of this TCK, when testing the feature.

clustering/ejb/WFLY-7628_Distributed_EJB_Timers.adoc
fabiobrz: > These are covered by the distributable-ejb subsystem tests.

the following one as well?

`/subsystem=ejb3/service=timer-service:undefine-attribute(name=default-data-store)`

It is not about the distributable-ejb subsystem, hence my question. Just double checking.

clustering/ejb/WFLY-7628_Distributed_EJB_Timers.adoc
fabiobrz: Ok, thanks for your feedback.

So, since the implementation of this feature will make the attribute optional in the model, the user will be able to _undefine_ it with the proposed command.
I am thinking at users which will use the feature in future versions of WildFly and who never had any kind of previous experience - neither they would need it - with database based persistence. 
Isn't this approach too _indirect_ in such a use case (and maybe in general)? Do you think there can be another solution for enabling the feature without relying on the semantics of a completely different implementation?

pferraro: It's not ideal - but it was done this way in anticipation of the end result.  However, given that the scope of the distributable timer service implementation encompasses the file and database persistence use cases, the expectation here is that this new implementation will eventually replace the existing ones (meaning, that default-data-store, as well as the data store resources themselves, will be deprecated, and the default configuration will no longer define a default-data-store attribute).  At that point, this will no longer be confusing.

clustering/ejb/WFLY-7628_Distributed_EJB_Timers.adoc
pferraro: Yes.

clustering/ejb/WFLY-7628_Distributed_EJB_Timers.adoc
pferraro: @fabiobrz See the test result attachment in the downstream JIRA.

scottmarlow: > Thanks @pferraro - Hi @scottmarlow, my point here is to understand which environment/domain/context should be checked in oder to verify the execution of this TCK, when testing the feature.

Hi @fabiobrz, the TCK testing is internal to Red Hat currently.

fabiobrz: Thanks for your feedback on this @pferraro and @scottmarlow - I'll check the resources.

clustering/ejb/WFLY-7628_Distributed_EJB_Timers.adoc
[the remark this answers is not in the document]
fabiobrz: Thanks for telling me about requirements, which I am aware of.

Now, to clarify my point: since the nice-to-have we're discussing will be implemented, then there will be a new feature, or new behavior, or whatever we want to call it, and it is about passivation of transient timers, right? 
If the behavior doesn't change, then you're right to resolve this thread. If it does change, then the topic should be discussed further.
 
Specifically In such a case, it has to be clarified explicitly and documented so that no functionality is hidden anywhere.

clustering/ejb/WFLY-7628_Distributed_EJB_Timers.adoc
fabiobrz: In which way it will no longer be confusing? Because the attribute will disappear and the users won't have to enable the distributed EJB timers management by executing this command?

In such a case IMHO it is not ideal, as you stated as well, and should be changed, since the proposal is still being worked on, and there will be users (and customers when it comes to the productized version of this feature) relying on its configuration, which shouldn't be "non-ideal" since the beginning.

Isn't there a more straightforward way to enable this?

clustering/ejb/WFLY-7628_Distributed_EJB_Timers.adoc
fabiobrz: Ok, thanks, then it should be mentioned in the community docs.

fabiobrz:
Hi @pferraro - when looking at linked JIRAs I noticed one discussion about transaction and it came to my mind that there's a section in the docs about a specific use case on k8s, see https://docs.wildfly.org/26/Developer_Guide.html#Jakarta_Enterprise_Beans_on_Kubernetes and the `9.14.2. EJB configuration for Kubernetes` section specifically.

Does the feature implementation relate to his in ant ways?

clustering/ejb/WFLY-7628_Distributed_EJB_Timers.adoc
pferraro: > In which way it will no longer be confusing?

Because timer management will use the distributable implementation by default, i.e. exclusively configured via the distributable-ejb subsystem in our default configuration.

> Because the attribute will disappear

It does not "disappear" - it would be undefined by default.

> Isn't there a more straightforward way to enable the feature?

Such as?  The problem is not just "how to enable the feature", but how to disable the existing configuration.  Ideally, the user performs a single action for both.  Here's a the 3 most obvious options:

1. My first thought was to create a separate distributable-timer-service resource that is mutually exclusive with the existing timer-service resource, which would normally be the cleanest way to introduce a new implementation mean to replace an existing implementation.
     * This prohibits use of different timer service implementations for transient vs persistent timers (not ideal)
     * This suggests requiring a separate distributable-timer-service deployment descriptor namespace used to override distributable timer service behavior per EJB (i.e. more awkward)
     * This complicates any future plans to deprecate features of the existing timer-service, where persistent vs transient timers are not implemented independently.
     * For the above reasons, I discarded this idea
1. The second option is to add separate `default-persistent-timer-management` and `default-transient-timer-management` attributes to the existing timer-service resource, which reference named timer-management providers as defined in the distributable-ejb subsystem.  The `default-persistent-timer-management` attribute would be mutually exclusive with the "default-data-store` attribute.
     * This is more flexible than the proposal 1, since it lets users configure different timer service implementations for transient vs persistent timers
     * From a user perspective, this is more cumbersome than what is described in the current draft of the analysis document.  Enabling this feature now becomes a matter of undefining the `default-data-store` attribute and defining the `default-persistent-timer-management` attribute within a batch (since one of the pair of attributes must be defined).
1. The third option, and what is described in the current draft of the analysis document, is the same as proposal 2, except that the default-*-timer-management attributes reside in the distributable-ejb subsystem instead of the timer-service resource of the ejb3 subsystem.
     * This means that enabling this feature is simply a matter of undefining the `default-data-store` attribute (since the`default-persistent-timer-management` will already be defined in the distributable-ejb subsystem.
     * Additionally, this option was preferred by @rachmatowicz

@fabiobrz What are your thoughts re: proposal 2 vs 3?
@chengfang Any opinion?

pferraro:
> Does the feature implementation relate to his in ant ways?

clustering/ejb/WFLY-7628_Distributed_EJB_Timers.adoc
pferraro: After some follow-up discussions with team members, option 2 is going to be problematic for galleon provisioning, which does not handle well the case where a single attribute requires different capabilities depending on whether or not it is defined.  Consequently, I think option 2 is the way to go.

fabiobrz:
> This should not apply here, since Jakarta Enterprise Bean timers are not invokeable.

Thanks for clarifying @pferraro - About transactions again, I noticed a comment of yours to some related JIRA issue mentioning two use cases involving transactions, i.e.:

* timers created/deleted during a transaction
* timers with transactional timeout methods

and that handling these cases happens outside of the Infinispan context. I am wondering whether the feature implementation would bring changes regarding the above mentioned cases and whether test coverage is needed for that.

pferraro:
> I am wondering whether the feature implementation would bring changes regarding the above mentioned cases and whether test coverage is needed for that.

clustering/ejb/WFLY-7628_Distributed_EJB_Timers.adoc
```diff
@@ -0,0 +1,168 @@
...
+	/subsystem=infinispan/cache-container=ejb/local-cache=persistent/store=file:add(purge=false, passivation=false)
+
+The following timer management provider will be added to the default distributable-ejb subsystem configuration:
+
+	/subsystem=distributable-ejb/infinispan-timer-management=persistent:add(cache-container=ejb, cache=persistent)
+
+To enable use of the distributable timer service, a user will be required to modify the default timer-service configuration of the ejb3 subsystem as follows:
+
+	/subsystem=ejb3/service=timer-service:undefine-attribute(name=default-data-store)
+	/subsystem=ejb3/service=timer-service:write-attribute(name=default-persistent-timer-management=persistent)
```
fabiobrz, line 112: This line looks broken, maybe it should read as follows?

```
/subsystem=ejb3/service=timer-service:write-attribute(name=default-persistent-timer-management, value=persistent)
```

fabiobrz, line 112: This line looks broken, maybe it should read as follows?

```
/subsystem=ejb3/service=timer-service:write-attribute(name=default-persistent-timer-management, value=persistent)
```

clustering/ejb/WFLY-7628_Distributed_EJB_Timers.adoc
```diff
@@ -0,0 +1,168 @@
...
+* Servers started using the default ejb3 subsystem configuration will continue to behave as normal, using the existing in-memory `TimerService` implementation for both transient and persistent timers.
+
+=== Default Configuration
+
+The following caches will be added to the default `ejb` cache-container configuration with the Infinispan subsystem:
+
+For the HA profiles:
+
+	/subsystem=infinispan/cache-container=ejb/distributed-cache=persistent:add()
+	/subsystem=infinispan/cache-container=ejb/distributed-cache=persistent/component=locking:add(isolation=PESSIMISTIC)
```
fabiobrz, line 94: I see that the isolation level is set to `REPEATABLE_READ` in the [remote Infinispan integration test](https://github.com/wildfly/wildfly/blob/b9d1aa85531fa0afa1dd59aed2be72e4d7a8782d/testsuite/integration/clustering/src/test/java/org/jboss/as/test/clustering/cluster/ejb/timer/remote/HotRodPersistentTimerServiceTestCase.java#L64). What is the correct value for this configuration setting?

fabiobrz, line 94: I see that the isolation level is set to `REPEATABLE_READ` in the docs and in the remote Infinispan integration test. What is the correct value for this configuration setting?

clustering/ejb/WFLY-7628_Distributed_EJB_Timers.adoc
fabiobrz: +1, I see this is consolidated now.

fabiobrz: +1, I see this is consolidated now.

fabiobrz (commented):
Hi @pferraro - back to this proposal after some time and I dropped some comments after making some experiments with the feature.

Feel free to let me know what you think.

clustering/ejb/WFLY-7628_Distributed_EJB_Timers.adoc
pferraro: Yeah - that's a typo.

clustering/ejb/WFLY-7628_Distributed_EJB_Timers.adoc
pferraro: Also, a mistype.

clustering/ejb/WFLY-7628_Distributed_EJB_Timers.adoc
```diff
@@ -0,0 +1,168 @@
...
+* Implement Infinispan-based `TimerService` for persistent timers that manages timer execution based on the consistent hash of the timer identifier.
+** Make existing ``default-data-store`` attribute of the timer service resource optional.  Use distributable implementation when this attribute is undefined.
+* Add default and named persistent timer management capabilities to distributable-ejb subsystem.
+** Add named Infinispan timer management resource that provides the above capability.
+* Add an attribute to the timer service resource within the ejb3 subsystem to determine the server-wide default provider for transient timer management.
+* Add ability to enable distributed timer management (in lieu of a timer store) within ejb3 subsystem for management of persistent timers.
+* Existing timer service resource operations must continue to work when executed against the distributable `TimerService`.
+
+=== Nice-to-Have Requirements
+
```
fabiobrz, line 72: All the nice-to-have requirements deal with transient timers, although no tests, between those mentioned in this analysis document, are defined to check such requirement configuration and functionalities, or am I missing something?

clustering/ejb/WFLY-7628_Distributed_EJB_Timers.adoc
fabiobrz: BTW @pferraro - Is the following:

```
Use distributable implementation when this attribute is undefined
```

tested somehow?

fabiobrz (commented):
Hi @pferraro - I went for another round of review on this since some points came to my attention while reviewing the PR changes and building a test plan.

Can you please let me know what you think?

clustering/ejb/WFLY-7628_Distributed_EJB_Timers.adoc
```diff
@@ -0,0 +1,168 @@
...
+** Modify the `TimerService` DUP to use a `CompositeTimerService` in the event that the server is configured to use persistent timer management, that delegates to the existing `TimerServiceImpl` for use with transient timers and the `DistributableTimerService` for use with persistent timers.
+
+== Security Considerations
+
+N/A
+
+== Test Plan
+
+* All existing `TimerService` tests must pass against the new persistent timer implementation, including the Jakarta Enterprise Beans TCK.
+* New tests will be added to the clustering integration testsuite to validate correctness of behavior of persistent timers with a cluster, including:
```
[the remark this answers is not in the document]
pferraro: > Looking at the feature PR it seems that there are no domain tests, correct?

What do you mean by "domain tests"?

> it seems to me that there aren't any tests to ensure that the server can start correctly with the two - mutually exclusive - configuration, except the integration ones, which validate what you detail in the following points.
> Shouldn't such a "smoke" test be part of the PR changes?

In general, it is not feasible to test against every combination of galleon layers - and such testing is truly orthogonal to this feature.
We know that the ejb + ejb-dist-cache layer creates extra distributable-ejb and infinispan cache configuration, but this configuration is not actually mutually exclusive - but it would be a source of confusion for users.  We should definitely address this separately.

clustering/ejb/WFLY-7628_Distributed_EJB_Timers.adoc
```diff
@@ -0,0 +1,168 @@
...
+== Issue Metadata
+
+=== Issue
+
+* https://issues.redhat.com/browse/WFLY-7628[WFLY-7628]
+
+=== Related Issues
+
+* https://issues.redhat.com/browse/WFLY-14953[WFLY-14953]
+
```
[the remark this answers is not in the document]
pferraro: I've created https://issues.redhat.com/browse/WFLY-16270 and will add this link to this section.

clustering/ejb/WFLY-7628_Distributed_EJB_Timers.adoc
pferraro: The same integration test will cover both cases.   I have not yet committed these changes.

fabiobrz: I see, thanks. I'll keep an eye on the feature implementation PR and we can update this very thread accordingly until resolved.

clustering/ejb/WFLY-7628_Distributed_EJB_Timers.adoc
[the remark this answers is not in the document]
fabiobrz: Thanks! Do you agree that we need a correspondent tracker for the product, too?

clustering/ejb/WFLY-7628_Distributed_EJB_Timers.adoc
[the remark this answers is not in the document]
fabiobrz: > What do you mean by "domain tests"?

I mean tests that use the "domain" mode,, not the "standalone" one.

> In general, it is not feasible to test against every combination of galleon layers - and such testing is truly orthogonal to this feature.

I am not asking about Galleon layers, actually. I am asking for a smoke test that validates that the server can start either with the default configuration - which will use the database persistence IIUC - and the new one introduced by the feature. Basically what you describe here:

```
To enable use of the distributable timer service, a user will be required to modify the default timer-service configuration of the ejb3 subsystem as follows:

/subsystem=ejb3/service=timer-service:undefine-attribute(name=default-data-store)
/subsystem=ejb3/service=timer-service:write-attribute(name=default-persistent-timer-management, value=persistent)
```

> We know that the ejb + ejb-dist-cache layer creates extra distributable-ejb and infinispan cache configuration, but this configuration is not actually mutually exclusive - but it would be a source of confusion for users. We should definitely address this separately.

I agree, IMHO there could be room for an issue tracker to improve the documentation. WDYT?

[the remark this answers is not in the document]
pferraro: > I mean tests that use the "domain" mode, not the "standalone" one.

Domain management is a completely separate function - and is tested separately from any functionality of a host controller.

> I am asking for a smoke test that validates that the server can start either with the default configuration - which will use the database persistence IIUC - and the new one introduced by the feature. Basically what you describe here:

Not currently, though now that we have 2 versions of the distributed timer service test (one using a distributed cache and another using a remote infinispan server), I can modify one variant to change the default behavior, while the other overrides the timer service via deployment descriptor.
The default configuration uses file persistence for persistent timers, BTW.

> I agree, IMHO there could be room for an issue tracker to improve the documentation. WDYT?

Better to fix the layers such that this combination does not include the ejb-local-cache layer.

clustering/ejb/WFLY-7628_Distributed_EJB_Timers.adoc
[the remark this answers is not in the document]
pferraro: I do not think that is warranted.

clustering/ejb/WFLY-7628_Distributed_EJB_Timers.adoc
pferraro: The feature PR includes these tests now.

clustering/ejb/WFLY-7628_Distributed_EJB_Timers.adoc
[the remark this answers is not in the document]
pferraro: > I am asking for a smoke test that validates that the server can start either with the default configuration - which will use the database persistence IIUC - and the new one introduced by the feature.

OK - I've updated the test that uses a remote infinispan server to configure the distributed timer service via the ejb3 subsystem, rather than using a deployment descriptor.

clustering/ejb/WFLY-7628_Distributed_EJB_Timers.adoc
pferraro: https://github.com/pferraro/wildfly/blob/WFLY-7628/testsuite/integration/clustering/src/test/java/org/jboss/as/test/clustering/cluster/ejb/timer/remote/HotRodPersistentTimerServiceTestCase.java#L68

fabiobrz: Thanks.

clustering/ejb/WFLY-7628_Distributed_EJB_Timers.adoc
fabiobrz: Thanks.

clustering/ejb/WFLY-7628_Distributed_EJB_Timers.adoc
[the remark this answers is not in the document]
fabiobrz: > Domain management is a completely separate function - and is tested separately from any functionality of a host controller.

IIUC, in the WildFly test suite there is room for "smoke"/"basic" tests to verify that a feature configuration is available, in a managed domain, see https://github.com/wildfly/wildfly/blob/main/testsuite/domain/src/test/java/org/jboss/as/test/integration/domain/suites/DatasourceTestCase.java#L44

I understand that not every feature can be tested this way, but in this very case I am wondering whether it would be an opportunity that we could/should consider, given that - with this new feature - there are quite some combinations for managing timers (persistent/transient, on file, database, embedded or "remote" cache etc.) and maybe such configuration options should be tested to be available from a domain controller.
WDYT?

>> ... I am asking for a smoke test that validates that the server can start either with the default configuration - which will use the database persistence IIUC - and the new one introduced by the feature
> Not currently, though now that we have 2 versions of the distributed timer service test (one using a distributed cache and another using a remote infinispan server), I can modify one variant to change the default behavior, while the other overrides the timer service via deployment descriptor.

Looking at the latest changes that you pushed here. IIUC now the HotRod (i.e. remote Infinispan) test is enabling the feature at the subsystem level, thanks to `TimerServiceSetupTask`. 
In such a case, this _integration_ test is the only "smoke" test that we have at the moment to verify that the server starts with the proposed configuration for enabling the feature. Correct? If this is true then a test execution will have to reach integration testing before assessing a potential failuhre in the configuration, right?

> The default configuration uses file persistence for persistent timers, BTW.

Thanks for clarifying, IMHO the community docs could be improved to list the EJB timer service configuration and managment options, for users to easily map their use cases to them.

> Better to fix the layers such that this combination does not include the ejb-local-cache layer.

I'd like to hear from @jfdenise about this, since I am not sure such an issue exists and whether it would be fixed at all, given the relationships between the layers. Will the ejb + ejb-dist combination fixed in a way that no ejb-local-cache will be included implicitly?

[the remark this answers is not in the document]
pferraro: > IIUC, in the WildFly test suite there is room for "smoke"/"basic" tests to verify that a feature configuration is available in a managed domain

That is not the purpose of the domain testsuite. It is impossible for a feature to *not* be available in a managed domain (unless the management operation explicitly forbids it, which would be very abnormal).

> Looking at the latest changes that you pushed here. IIUC now the HotRod (i.e. remote Infinispan) test is enabling the feature at the subsystem level, thanks to TimerServiceSetupTask.
In such a case, this integration test is the only "smoke" test that we have at the moment to verify that the server starts with the proposed configuration for enabling the feature. Correct? 

Not really.
The ejb3 subsystem unit test validates the use of timer-management references to the distributable-ejb3 subsystem.
The distributable-ejb subsystem unit tests validate the use of cache references to the Infinispan subsystem.
The infinispan subsystem unit tests validate the use hotrod-store references to remote-cache-container resources.

> IMHO the community docs could be improved to list the EJB timer service configuration and managment options, for users to easily map their use cases to them.

Feel free to file a jira to improve the general documentation of the EJB timer service.  This feature is only concerned about documenting the behavior that it adds.

> I'd like to hear from @jfdenise about this, since I am not sure such an issue exists and whether it would be fixed at all, given the relationships between the layers. Will the ejb + ejb-dist combination fixed in a way that no ejb-local-cache will be included implicitly?

I did not get around to filing this bug until now.  Please redirect any discussion of this bug here:  https://issues.redhat.com/browse/WFLY-16326

clustering/ejb/WFLY-7628_Distributed_EJB_Timers.adoc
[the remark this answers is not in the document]
fabiobrz: Ok, thanks @pferraro - can you please add https://issues.redhat.com/browse/WFLY-16270 to the section?

clustering/ejb/WFLY-7628_Distributed_EJB_Timers.adoc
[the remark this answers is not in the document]
jfdenise: @fabiobrz , my understanding of local vs dist is that it is by design. The local cache is the default and must be excluded to use the dist one.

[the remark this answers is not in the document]
fabiobrz: Thanks @jfdenise - I'll report this to https://issues.redhat.com/browse/WFLY-16326, as suggested by @pferraro

fabiobrz (approved):
Hi @pferraro - Thanks for the work you've done in this proposal and for addressing comments and questions with details and context.

I am approving the document, which LGTM.

clustering/ejb/WFLY-7628_Distributed_EJB_Timers.adoc
```diff
@@ -0,0 +1,169 @@
...
+== Requirements
+
+=== Hard Requirements
+
+* Create a `TimerService` implementation that delegates to separate persistent vs transient `TimerService` implementations.
+* Implement Infinispan-based `TimerService` for persistent timers that manages timer execution based on the consistent hash of the timer identifier.
+** Make existing ``default-data-store`` attribute of the timer service resource optional.  Use distributable implementation when this attribute is undefined.
+* Add default and named persistent timer management capabilities to distributable-ejb subsystem.
+** Add named Infinispan timer management resource that provides the above capability.
+* Add an attribute to the timer service resource within the ejb3 subsystem to determine the server-wide default provider for transient timer management.
```
fabiobrz, line 68: Maybe typo? s/transient/persistent/

PR #448, opened 2021-11-24
[WFLY-15746] WildFly Application Development Guide

emmartins:
Issue: https://issues.redhat.com/browse/WFLY-15746


emmartins:
This is just a draft for docs team to review and contribute.

quickstarts/WFLY-15746_WildFly_Application_Development_Guide.adoc
```diff
@@ -0,0 +1,83 @@
...
+=== Other Interested Projects
+
+== Requirements
+
+=== Hard Requirements
+
+* The guide should have 2 major sections, one for first time/beginners app developers, and one for advanced app developers.
+
+* The major section for first time app developers should include the following new content:
+** How to develop a first basic application (helloworld quickstart) from scratch using Maven, a complete user story walkthrough, which should include:
```
darranl, line 44: Do we think this is one quickstart or a small subset of quickstarts?  I don't think it would be a massive number but it feels like there may be a few getting started approaches.

quickstarts/WFLY-15746_WildFly_Application_Development_Guide.adoc
```diff
@@ -0,0 +1,83 @@
...
+*** How to create the application Maven Project, and add Red Hat Maven repositories and BOMs dependencies
+*** How to package, deploy and run the application (war/bootable jar, web console)
+*** How to add logging to the application (basic usage)
+*** How to test the application
```
darranl, line 48: Would it be worth in this section having something like "How to make the first modification" - i.e. the user has moved just beyond deploying an example we provide and made it their own with a small change?

quickstarts/WFLY-15746_WildFly_Application_Development_Guide.adoc
emmartins: This is just the first app from scratch, in my mind just the simplest hello world

quickstarts/WFLY-15746_WildFly_Application_Development_Guide.adoc
emmartins: Please note that there is a quick start guide also, and that should provide just a deploy and run, in this one there is the coding of the hello world, and any further modification is then related with a specific technology (line 53)

quickstarts/WFLY-15746_WildFly_Application_Development_Guide.adoc
jmesnil: This development guide should have some guiding principles that showcases efficient way to develop with WildFly.

While Jakarta EE development does not strictly adhere to https://12factor.net, there are principles that makes developing and operating WildFly much more simpler.

In particular, having the codebase being the source of deployments is a key aspect.
We should recommend & showcase how to provision and configure WildFly using the wildfly-maven-plugin in a reproducible & automated manner.
When we ask the user to copy a XML snippet in `standalone.xml`, we are not providing them a mechanism to keep this change in sync with WildFly releases and put the onus on them to automate this.
Compared to using a CLI scripts to apply this change to a provisioned WildFly server, this is automated, testable without any change to make outside of the codebase.

quickstarts/WFLY-15746_WildFly_Application_Development_Guide.adoc
```diff
@@ -0,0 +1,83 @@
...
+
+== Requirements
+
+=== Hard Requirements
+
+* The guide should have 2 major sections, one for first time/beginners app developers, and one for advanced app developers.
+
+* The major section for first time app developers should include the following new content:
+** How to develop a first basic application (helloworld quickstart) from scratch using Maven, a complete user story walkthrough, which should include:
+*** How to create the application Maven Project, and add Red Hat Maven repositories and BOMs dependencies
```
jmesnil, line 45: Are you planning to leverage https://issues.redhat.com/browse/WFLY-17651 for these instructions?

With https://issues.redhat.com/browse/WFLY-18428, we want to have a user being able to get up and running with WildFly in a few minutes. This is achievable by using a Maven archetypes that will do all this plumbing for them.

quickstarts/WFLY-15746_WildFly_Application_Development_Guide.adoc
```diff
@@ -0,0 +1,83 @@
...
+*** How to package, deploy and run the application (war/bootable jar, web console)
+*** How to add logging to the application (basic usage)
+*** How to test the application
+*** How to use CodeReady Workspaces/Studio (basic usage: add app project, run & debug)
+*** How to build, deploy and run on Openshift (basic usage)
+
+* The major section for first time app developers should also include the following new content:
+** How to develop a first basic application for each specific supported component/technology (e.g. CDI)
```
jmesnil, line 53: Are you planning to modularise that content? How would you sort/present the sections?

quickstarts/WFLY-15746_WildFly_Application_Development_Guide.adoc
```diff
@@ -0,0 +1,83 @@
...
+*** Introduction to the component/Technology
```
jmesnil, line 54: Having a list of prerequisites (in terms of tooling or previous guides)

quickstarts/WFLY-15746_WildFly_Application_Development_Guide.adoc
```diff
@@ -0,0 +1,83 @@
...
+*** Extends the basic application development content (e.g. add needed Maven dependencies related with component/technology)
+*** Any other basic level content which may be useful with respect to the technology
```
jmesnil, line 56: Links to relevant sections in WildFly admin guide and model reference for further information

quickstarts/WFLY-15746_WildFly_Application_Development_Guide.adoc
```diff
@@ -0,0 +1,83 @@
...
+
+* The following content from the existent Developer Guide should be adapted and included in the Advanced app developers section:
+** Modules and Classloading
+** Logging
+** Specific component/technology content (e.g. CDI)
+
+* The specific content/technology content from the existent Quickstart Guides should be adapted and included in the Advanced app developers section, with respect to the Specific component/technology content.
+
```
jmesnil, line 64: Will it cover migration like the current developer guide does? https://docs.wildfly.org/29/Developer_Guide.html#How_do_I_migrate_my_application_from_AS7_to_WildFly

quickstarts/WFLY-15746_WildFly_Application_Development_Guide.adoc
emmartins: IMHO we should include guidance for upstream only features, but consider those as alternatives to the supported features. If an upstream feature is a good enhancement than we should prioritise getting it supported.

quickstarts/WFLY-15746_WildFly_Application_Development_Guide.adoc
emmartins: No, the idea was to start with the standard Maven archetypes, and document the changes needed. The main reason for this is that WildFly archetypes are not supported, but I also thought of aligning with Quarkus, and reuse their enhancements on app development, both the online tool and the cli app project generation by maven plugin.

quickstarts/WFLY-15746_WildFly_Application_Development_Guide.adoc
emmartins: IMHO these would fit nicely as independent modules/guides, which could even be external to the document.

quickstarts/WFLY-15746_WildFly_Application_Development_Guide.adoc
emmartins: I think that we should definitely provide high level migration content, and port the EAP migration guide to upstream.

quickstarts/WFLY-15746_WildFly_Application_Development_Guide.adoc
jmesnil: I'm not sure to follow you.
This guide must focus on the best way to develop with WildFly. If or when this will go in a supported product (I suppose you mean JBoss EAP) should not impact how WildFly is documented.

jmesnil: The comment above is also a good opportunity to split this dev guides in different focused guides that can then be picked and choose when a feature reach a downstream product.

quickstarts/WFLY-15746_WildFly_Application_Development_Guide.adoc
jmesnil: > WildFly archetypes are not supported

We have been providing them and maintaining them. I don't see why we could not use them to bootstrap the necessary Maven project.
I can do a `mvn archetype:generate` and starts writing *my* code in a matter of minutes. Why would I start from scratch and spend 1/2 to 1 hour just to have the pieces in place in the `pom.xm`, `src/main` & `src/test`?

quickstarts/WFLY-15746_WildFly_Application_Development_Guide.adoc
emmartins: I mean this is the upstream for an EAP 8 guide, it should prioritise now on the supported features.

quickstarts/WFLY-15746_WildFly_Application_Development_Guide.adoc
emmartins: > We have been providing them and maintaining them. I don't see why we could not use them to bootstrap the necessary Maven project. 

Because that means we need to make archetypes a product and support it for 10 years, please remember this is the upstream for a new EAP 8 requirement.

> I can do a mvn archetype:generate and starts writing my code in a matter of minutes. Why would I start from scratch and spend 1/2 to 1 hour just to have the pieces in place in the pom.xm, src/main & src/test?

Nobody wants to start from scratch, for now use the standard Maven archetypes, next use whatever is best in the long term, if possible aligned with Quarkus (which AFAIK opted for a Maven Plugin instead of archetypes).

emmartins: Please note that this doesn't mean we can not add WildFly Archetypes in the upstream guide too, we just need to not make the guide work around it since it's not supported. For instance we can document the supported way and then add a section introducing WFLY Archetypes, and how it automates what we first did manually (add plugin and dep management, import boms, etc.)

emmartins: Archetypes existed for ages, and it is not and never was a popular Maven feature among app developers, mostly cause content is not flexible enough and gets deprecated fast, are you sure it's the best direction for us to promote long term?

Anyway let's say it takes 30 min for the user to manually add everything to the Maven Project... isn't that only for the 1st time for the user to learn about our plugins, BOMs, etc? And won't that learning be needed? And would actually those few minutes really mean something on the time it takes to develop a new app? I can understand the need for an as fast as possible "server start", for an as fast as possible "get started", now an as fast as possible "app development"? I can't really see what's wrong with now explaining how to do it manually, starting with the maven archetypes + (upstream only) extend that with the "unstable/unsupported" features like Archetypes.

bstansberry: @ehsavoie @jmesnil This sounds like something that can be handled via doc modularization. I see the value in a content section that spells things out, and also in one that tells you how to use things WildFly provides to do that for you. The guide incorporates both with a bit of preface to tell you about and link to the two options.

The endpoint for both paths would be a consistent maven project.

The "how to use things WildFly provides" cannot be constrained by EAP. WildFly will always be ahead of EAP in this area and can't be constrained because we won't document what we do. If EAP documenters want to reuse WildFly content in its docs, that's great, I'm all for that, but we should organize things so they can reasonably efficiently pick and choose what to use.

The "how to use things WildFly provides" part can also evolve over time; e.g. it's archetypes now because it's what we have, but later it can be a project generator or something.

emmartins: @bstansberry IMHO in WFLY we should have everything we have on EAP + unstable/upstream-only. We need to document the manual options upstream too.

emmartins: @bstansberry @jmesnil btw FYI this JIRA was taken over by DOCs team (this AD is from 2021), and the work is being led by Ashwin, with a minimal version for this to be available on EAP 8.0. Even if this is on docs domain now, we still have a perm status/progress discussion spot on BOMs and QS weekly.

bstansberry: Understood, and we need to consider that downstream work, and I'm fine with orienting the scope of this proposal toward those requirements, as that will certainly be better than what we have now!

But WildFly needs to move beyond that, and quite soon, so it's valuable to consider where we want things to go.

jmesnil: > are you sure it's the best direction for us to promote long term?

No. But this is our best direction for WildFly *short term*. We could have *now* a way for users to bootstrap their Entreprise Java applications in a matter of minutes. This would be a huge productivity boon.
Long term, there can be other solutions (code launcher) but there is no reason not to improve the user experience now with what we have now.
 
> Anyway let's say it takes 30 min for the user to manually add everything to the Maven Project... isn't that only for the 1st time for the user to learn about our plugins, BOMs, etc? And won't that learning be needed? And would actually those few minutes really mean something on the time it takes to develop a new app?

These few initial minutes are very important for the perception of the user and the developer joy. As a developer, do you want to spend 30 to 1 hour reading a doc and doing `mkdir`, `cp` all other the places before you are able to do your actual work? I definitely don't. I want to work on *my* stuff with a project that respects my time.

As a team, we agree that is one of our objective to be able to get a WildFly application up and running in less than 5 minutes (download times not included). This is a valuable objective that drives innovation in terms of user experience and feature capabilities.

> I can understand the need for an as fast as possible "server start", for an as fast as possible "get started", now an as fast as possible "app development"? I can't really see what's wrong with now explaining how to do it manually, starting with the maven archetypes + (upstream only) extend that with the "unstable/unsupported" features like Archetypes.

There is nothing wrong with "explaining how to do it manually". But from a developer experience, a "show (and tell)" user experience beats every time a "tell (and figure out what instructions you missed)" one :) 

> If you truly believe Archetypes is the direction we should go long term then why not an RFE and make them a product with a feature team?

I don't understand your sentence. We can release a new archetype now to improve WildFly user experience without committing to any product.
The important commitment is on the *output* of the tool: a Maven project properly setup to develop with WildFly.

And again, I don't think this is the ideal long term solution but this is something that would benefits our user now and do not prevent us to improve on it down the road...

jmesnil: @emmartins I understand that this proposal is a downstream requirement for EAP 8. As @bstansberry suggested, let's make that a clear requirement and focus on it.

My only remark is then that this proposal should not hamper any progress or changes we need to improve WildFly user experience.

emmartins: > These few initial minutes are very important for the perception of the user and the developer joy. As a developer, do you want to spend 30 to 1 hour reading a doc and doing mkdir, cp all other the places before you are able to do your actual work? I definitely don't. I want to work on my stuff with a project that respects my time.

mkdir, cp, what are you talking about? I was just referring to what we have been using up till now, and what was being worked out by DOCs for this RFE up till now, which is the usage of the standard Maven archetypes. On top of using those you pretty much just add our plugin management, boms import, repos and that's is most cases wrt our stuff... Please note that WFLY Archetypes will never be enough for a real app, there is always user stuff to add to the Maven project (dependencies, repos, versions, etc.) so the time adding our stuff becomes irrelevant. 

> As a team, we agree that is one of our objective to be able to get a WildFly application up and running in less than 5 minutes (download times not included). This is a valuable objective that drives innovation in terms of user experience and feature capabilities.

I think we are too focused on the get started app and forget we are adding a new WFLY/EAP public component that may in the end may just be more noise on our and user's processes.

> I don't understand your sentence. We can release a new archetype now to improve WildFly user experience without committing to any product.

But we need to improve EAP user experience too... I kind of feel that perhaps I am not explaining my concerns properly but in short what is really in my mind is that we should spend some effort NOW figuring out what is the best long term solution, and work on it for both WFLY and EAP. And if we need to extend an OKR to another CY no prob.  In the end we may even end up deciding the archetypes are in fact the best direction!

emmartins: @bstansberry @jmesnil no matter the - very good IMHO - discussion the docs team let me know they will consider the  documentation of both archetypes, using a switch to render differently for WFLY and EAP.

jmesnil: > I was just referring to what we have been using up till now, and what was being worked out by DOCs for this RFE up till now, which is the usage of the standard Maven archetypes.

Could you please link to this doc? I was not able to find it in https://docs.wildfly.org/29/Getting_Started_Developing_Applications_Guide.html

> I think we are too focused on the get started app and forget we are adding a new WFLY/EAP public component that may in the end may just be more noise on our and user's processes.

I never implied or suggested to create archetype for EAP. I'm focused on WildFly.
I'm not focused on the *archetype*. I'm focused on the Maven project setup. The archetype is what we have now (or close to have it) to produce this setup. Is there any other solution we provide to our users to bootstrap their applications *now*?

emmartins: @jmesnil You will find it in the READMEs of several MP Quickstarts, e.g. https://github.com/wildfly/quickstart/blob/a3b0be1576e2cdb93abdb8512074af5006f8e825/microprofile-config/README.adoc?plain=1#L56

jmesnil: thanks (and that's exactly all the `cp` and `mkdir` that I was talking about to fill the general maven archetypes with the specifities of WildFly)

emmartins: Which mkdir or cp you referring to?

jmesnil: all the xml snippets you have to copy to the pom.xml, the directories and files you create to populate `src`, etc.

emmartins: The only non-app-specific there are the pom.xml changes I previously referred, which have little time/effort impact due to the need for app-specific changes there too, e.g. dependencies.

emmartins:
@darranl @jmesnil FYI there is a major rework of the plans for this doc wrt schedule, with a severe reduction of what the doc will cover on first release, targeting EAP 8 GA. Please note that the same content is planned, we are simply moving many parts of it to nice-to-have requirements, which will come post GA, due to schedule limitations. 
Please note also that the new guide will be first developed EAP only, and post GA we will adapt (considering your previous comments and suggestions) and contribute it to WildFly.

zhantemirov:
It's hard to review the AD when it's aiming to EAP Development Guide whereas being located in WFLY proposals.

I've already seen the preview version of this guide, LGTM due to the fact it only contains necessary hard requirements since all previous hard requirements were moved to nice-to-have section.

Having said that, there is no many things to review, but I'll leave the possibility of making the final conclusion to Krystof Stekovic.

emmartins:
@zhantemirov fyi there is now an upstream PR too, did review it today live with Ashwin... so we will follow the standard process and this won't be EAP only any longer. 
I will meet with Ashwin again next thursday to review the changes needed (some also downstream on MR, not flagged in the upstream PR), please let me know if you want to join too.

kstekovi:
Thanks @zhantemirov  for your review.

I don't have more to say. This proposal LGTM.

emmartins:
thanks @zhantemirov @kstekovi for the review

PR #450, opened 2022-01-06
[WFLY-15832] MP Config: Ability to specify root config source directory

kabir:
https://issues.redhat.com/browse/WFLY-15832


microprofile/WFLY-15832-microconfig-config-root-folder.adoc
```diff
@@ -0,0 +1,99 @@
...
+:icons:             font
+:idprefix:
+:idseparator:       -
+
+== Overview
+The MicroProfile Config with SmallRye subsystem currently supports configuration of several types of Config Sources. One of these allows you to specify a directory, wherein each file is used as a config property. The file name is the name of the property, and the contents of the file is the value of the property.
+
+This is useful on OpenShift where ConfigMaps, the ServiceBinding opertator (and probably other things) expose config via this mechanism for use in applications.
+
+This RFE is about enhancing the subsystem to allow specifying a root directory, of which the top-level directories under it will appear as Comfig Sources. The motivation is that each 'thing' on OpenShift installs their config under a root directory. e.g. in the ServiceBinding operator case this is under the `/bindings` directory and ConfigMaps are installed under `/etc/config`.
```
mchoma, line 14: `Comfig Sources` -> `Config Sources`

microprofile/WFLY-15832-microconfig-config-root-folder.adoc
```diff
@@ -0,0 +1,99 @@
...
+
+So if you have two ConfigMaps called `cfg-a` and `cfg-b`, their respective directories containing the config files will be `/etc/config/cfg-a` and `/etc/config/cfg-b`. At present this means we need to add two Config Source entries to the subsystem, and we need to know the exact names of both, in order to pick these up.
+
+While we will keep the existing mechanism, the proposal is to be able to specify a directory as a root for configuration, and that each top level directory becomes a Config Source as with the previous mechanism.
+
+This has the advantage of only needing to specify one Config Source root in order to pick up all ConfigMaps, one to pick up all ServiceBinding configs, and so on. An added bonus is not needing to know the exact names of the configuration directories, which are in some cases autogenerated.
```
mchoma, line 20: Are we going to support 1 level of recursion, right?

mchoma: Ok, answer to my  question is here https://github.com/wildfly/wildfly-proposals/pull/450/files#diff-9f1354c78479d2107750d078d4f846420ac82216e6eda99d34a0ad36ea2d948fR74. Only one level of recursion. +1

microprofile/WFLY-15832-microconfig-config-root-folder.adoc
```diff
@@ -0,0 +1,99 @@
...
+
+=== Testing By
+// Put an x in the relevant field to indicate if testing will be done by Engineering or QE. 
+// Discuss with QE during the Kickoff state to decide this
+* [ ] Engineering
+
+* [ ] QE
+
+=== Affected Projects or Components
+All code changes are in the MicroProfile Config with SmallRye subsystem in WildFly.
```
mchoma, line 48: For me to understand. This feature goes beyond what MicroProfile Specification mandates to be MP compliant?

microprofile/WFLY-15832-microconfig-config-root-folder.adoc
```diff
@@ -0,0 +1,99 @@
...
+
+=== Non-Requirements
+* As mentioned previously, in <<hard-requirements, Hard Requirements>>, only top level directories under a ConfigSource root directory will be considered as ConfigSources. We don't check nested folders.
+* No attempts will be made to avoid having multiple ConfigSources defined in the subsystem referencing the same directories
+
+== Backwards Compatibility
+The feature is backward compatible.
+
+== Security Considerations
+None. The feature doesn't expose anything which would not be possible with what is there today.
```
mchoma, line 90: Is it true? So far EAP administrator had to configure each file explicitely. So administrator guarantees files are correct. With this feature in place, attacker (e.g. other user on OpenShift) can prepare new (unchecked by administrator) files to be processed by EAP. E.g. by putting something into `/bindings` or `/etc/config`. Those files could be malformed (some sort of injection) or a lot of them could be there (DoS attack). Or it can be file which "tricks" ordinal resolving https://github.com/wildfly/wildfly-proposals/pull/450/files#diff-9f1354c78479d2107750d078d4f846420ac82216e6eda99d34a0ad36ea2d948fR77 and value expected by administrator can be overwritten by attacker.

mchoma: Does not need to be attacker. It can happen easily by accident. If `/bindings` and `/etc/config` are common places where OpenShift store ServiceBindings and ConfigMaps from whole cluster. It can happen easily (e.g. on managed cluster like Dedicated or Sandbox instances) key collision will happen and which value will be used will be determined by ordinal resolving (alphabetical order?) https://github.com/wildfly/wildfly-proposals/pull/450/files#diff-9f1354c78479d2107750d078d4f846420ac82216e6eda99d34a0ad36ea2d948fR77. This feature seems very dangerous to me.

microprofile/WFLY-15832-microconfig-config-root-folder.adoc
bstansberry: @mchoma Yes, as does the existing feature it is improving.

The spec only requires us to provide ConfigSources backed by system properties and env vars and any META-INF/ microprofile-config.properties found on the classpath. 

https://download.eclipse.org/microprofile/microprofile-config-3.0.1/microprofile-config-spec-3.0.1.html#default_configsources

Note that the config_ordinal file discussed below is a mechanism to provide the spec behavior from https://download.eclipse.org/microprofile/microprofile-config-3.0.1/microprofile-config-spec-3.0.1.html#_manually_defining_the_ordinal_of_a_built_in_configsource

microprofile/WFLY-15832-microconfig-config-root-folder.adoc
jmesnil: > attacker (e.g. other user on OpenShift)

The ConfigMap must be in the same namespace than the application. In essence, if an OpenShift user can do that, he also has direct access to the application itself.

jmesnil: your 2nd point is correct but that would be a misconfiguration, not a security issue.

We could have 2 service bindings injection config map that would provide the same `USERNAME` property for example.
However there are mechanism to avoid this and have deterministic configuration such as having unambiguous prefix for different service bindings.

mchoma: From wording of [1] and [2] I understood `/bindings` and `/etc/config` we are talking about are OpenShift wide directories. So something from different namespace will be visible in EAP pod. So you are basically saying on OpenShift level it is guaranteed in EAP pod in `/bindings` and `/etc/config` there is mounted only stuff related to Pods namespace? If this is true my concern is not valid. And that would make sense.

But in this case could you make it please in [1], [2] explicit. So any other reader have a chance to misundarstand it like I did. 

[1] https://github.com/wildfly/wildfly-proposals/pull/450/files#diff-9f1354c78479d2107750d078d4f846420ac82216e6eda99d34a0ad36ea2d948fR14
[2] https://github.com/wildfly/wildfly-proposals/pull/450/files#diff-9f1354c78479d2107750d078d4f846420ac82216e6eda99d34a0ad36ea2d948fR16

jmesnil: https://docs.openshift.com/container-platform/4.10/nodes/pods/nodes-pods-configmaps.html

> ConfigMap objects reside in a project.
> They can only be referenced by pods in the same project.

mchoma: I just checked some EAP pod and I really don't see any `/bindings` (I don't have ServiceBinding) and `/etc/config` (I have ConfigMap in namespace) directories available automatically. So I assume someone needs to configure that. 

Especially I am questioning this part how is meant. 

```
ConfigMaps are installed under `/etc/config`.

So if you have two ConfigMaps called `cfg-a` and `cfg-b`, their respective directories containing the config files will be `/etc/config/cfg-a` and `/etc/config/cfg-b`.
```

mchoma (approved):

bstansberry (approved):

PR #455, opened 2022-02-08
[WFLY-16018] RESTEasy should be able to provide trace information about request

liweinan:


jaxrs/WFLY-16018_EAP7-627.adoc
```diff
@@ -0,0 +1,107 @@
...
+=== Hard Requirements
+
+=== Nice-to-Have Requirements
+
+=== Non-Requirements
+
+=== Test Plan
+
+The features are tested in RESTEasy codebase. Successfully running the latest RESTEasy release' testsuite against target WildFly container will be required.
+
```
jamezp, line 102: I think we may need some tests in WildFly too. The reason being is we override the modules in WildFly so we need to ensure that WildFly without modified modules works. We should also test that disabling tracing works as well in WildFly.

jaxrs/WFLY-16018_EAP7-627.adoc
```diff
@@ -0,0 +1,107 @@
...
+* mailto:jperkins@redhat.com[James Perkins]
+
+=== QE Contacts:
+
+TBD
+
+=== Testing By
+
+* [ ] Engineering
+* [ ] QE
```
asoldano, line 75: @liweinan , testing by Eng or QE?

jaxrs/WFLY-16018_EAP7-627.adoc
asoldano: +1

liweinan: I'll start with writing some Arquillian tests.

jaxrs/WFLY-16018_EAP7-627.adoc
liweinan: @asoldano I talked with @jamezp on this and we decide to test by `Engineering` (test by ourselves). I'll update the doc here.

liweinan:
@jamezp I have updated the doc according to your suggestion.

jaxrs/WFLY-16018_EAP7-627.adoc
```diff
@@ -0,0 +1,111 @@
...
+
+=== Hard Requirements
+
+* The tracing feature may expose RESTEasy internal engine information, which may contain security sensitive information, so please use it with caution.
+
+=== Nice-to-Have Requirements
+
+=== Non-Requirements
+
+* The tracing feature can be turned off by configuration.
```
jamezp, line 101: I don't think this should be a non-requirement. It would be fairly simple to have an attribute in the subsystem to control whether or not this should be turned on.

liweinan: okay!

jaxrs/WFLY-16018_EAP7-627.adoc
```diff
@@ -0,0 +1,107 @@
...
+
+* [x] Traditional standalone server (unzipped or provisioned by Galleon)
+* [x] Managed domain
+* [x] OpenShift s2i
+* [x] Bootable jar
+
+== Requirements
+
+=== Hard Requirements
+
```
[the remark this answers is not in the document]
liweinan: I added some requirements.

jaxrs/WFLY-16018_EAP7-627.adoc
liweinan: @jamezp I modified this part a little bit because we have Arquillian test in WildFly now.

bstansberry (changes requested):

jaxrs/WFLY-16018_EAP7-627.adoc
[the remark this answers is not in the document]
liweinan: @bstansberry Here is my reply:

> Can you fill in more details? AIUI there's meant to be some sort of management API support to allow this to be configured. What are the requirements for how that is done? Are there other ways to configure this beyond the management API?

Currently this module doesn't support management API configuration. Now the configuration is like this:

- If `resteasy-tracing` module is added into a project as dependency, then it will be loaded into runtime. By default it's turned off, and users can configure to use it by setting in `web.xml`. The other way to enable it is to use the `ON_DEMAND` mode(The usage is added into RESTEasy doc as QA required: https://github.com/resteasy/resteasy/pull/3235). Which the client side can use headers to control the tracing info.

> Per your note on the next line, this is a security sensitive configuration setting, so how will access to that be controlled via the WFLY management RBAC system?

For the security sensitive issue, it means the tracing info may contain RESTEasy server side internal messages (for example, the message passed into filters, the time information that consumed by each layer of internal RESTEasy classes, etc). Because it's a `tracing` feature, so it may expose a lot of internal tracing information when used.

> What is the required set of information that is exposed? That is, a formal list describing all that's meant to be covered under 'like used interceptors, used exception mappers, used end-points, used providers, etc.'

By default the feature is turned off, and if users want to use it, there are several levels of tracing information can be configured(https://docs.jboss.org/resteasy/docs/6.2.0.Beta1/userguide/html_single/index.html#d4e4526):

- SUMMARY
- TRACE
- INFO

Currently these three levels are hard coded into RESTEasy internal classes, maybe in the future we could provide more fine grained level information levels. Regardless which level, the tracing info is quite internal information.

The full list of the events traced by the module is listed here(like used interceptors, used exception mappers, used end-points, used providers, etc):

- https://docs.jboss.org/resteasy/docs/6.2.0.Beta1/userguide/html_single/index.html#d4e4591


> What's the message format? Is it configurable? What compatibility guarantees do we provide with respect to the format and the message details?

Currently it supports PLAIN-TEXT and JSON format. The example text of the tracing info is shown in the project document: https://docs.jboss.org/resteasy/docs/6.2.0.Beta1/userguide/html_single/index.html#d4e4566

@bstansberry wdyt about the above info? If it's okay guess I can add the above text into Hard Requirement.

jaxrs/WFLY-16018_EAP7-627.adoc
```diff
@@ -0,0 +1,99 @@
...
+* The tracing feature may expose RESTEasy internal engine information, which may contain security sensitive information, so please use it with caution.
+
+=== Nice-to-Have Requirements
+
+=== Non-Requirements
+
+=== Test Plan
+
+The features are tested in RESTEasy codebase and also with the Arquillian test in WildFly. Successfully running the test in WildFly is required.
```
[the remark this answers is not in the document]
liweinan: Currently it doesn't support management API. The the configuration method is tested in WildFly PR: https://github.com/wildfly/wildfly/pull/15198

jaxrs/WFLY-16018_EAP7-627.adoc
```diff
@@ -0,0 +1,99 @@
...
+
+== Community Documentation
+
+The documentation is coming with the release of the RESTEasy.
```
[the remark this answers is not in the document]
liweinan: > The configuration of this sounds like something that should be covered in the WildFly documentation, given the potential security issues.

Maybe the `security issues` is misleading here. I explained the details in above. So I guess the `Hard Requirement` part should be re-rewritten as shown above.

jamezp:
FYI I'm going to clone this PR and force push to your branch @liweinan. I think we need to update this to reflect a couple of notes about how this is not meant for production and there will be a subsystem configuration.

liweinan:
@jamezp Thanks for the great help on this!

jaxrs/WFLY-16018_Treacing_API.adoc
```diff
@@ -0,0 +1,108 @@
...
+* [x] OpenShift s2i
+* [x] Bootable jar
+
+== Requirements
+
+=== Hard Requirements
+
+* Add the `org.jboss.resteasy.resteasy-tracing-api` module.
+* Introduce `resteasy-extensions` as dependency.
+* Add subsystem attributes to control turning this feature on/off. The description *must* note that this feature is not
```
bstansberry, line 85: I think the subsystem should log a WARN if it is enabled.

jamezp: This is a good idea! I'll add that to the proposal as well as update the PR.

jaxrs/WFLY-16018_Treacing_API.adoc
```diff
@@ -0,0 +1,108 @@
...
+  meant for production.
+* Enabling in the subsystem should be constrained via the management model.
+  ** Users can see and read the value. However, in order to write the value they must have permissions to do so.
```
bstansberry, line 88: This should be fleshed out a bit, i.e. that there will be an RBAC sensitivity-classification named 'xxx' added and applied to the attribute.

PR #461, opened 2022-03-16
Proposal for WFLY-14934, WildFly server Maven plugin [adjustments for basic developer workflow]

jfdenise:
* This is an evolution of the initial proposal with some changes that are added to the WildFly Maven plugin allowing for minimal developer workflow.
* In addition, the property ``layers-configuration-file-name`` has been added allowing to integrate the maven plugin in wildfly tests. This update of the proposal documents it.

### DIFF
--- a/maven/WFLY-14934_Server_Plugin.adoc
+++ b/maven/WFLY-14934_Server_Plugin.adoc
@@ -132,20 +132,24 @@ WildFly. This change impacts ```run``` and ```start```goals).
 These configuration items are common to the ```provision``` and ```package``` goals. They are used to configure 
 Galleon to provision a server.
 
-* feature-packs: A list of feature-pack to install.
+* feature-packs: A list of feature-pack to install. This can be configured with the system property ``wildfly.provisioning.feature-packs`` containing 
+a comma separated list of feature-pack locations in the form of Maven coordinates GroupId:ArtifactId:Version or producer reference.
 * log-provisioning-time: Whether to log provisioning time at the end.
 * offline-provisioning: Whether to use offline mode when the plugin resolves an artifact.
 * galleon-options: Arbitrary Galleon options used when provisioning the server.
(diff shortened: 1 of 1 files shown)


maven/WFLY-14934_Server_Plugin.adoc
```diff
@@ -132,18 +132,24 @@ WildFly. This change impacts ```run``` and ```start```goals).
...
 It contains Galleon provisioning information used to provision this server.
-* layers: A set of layers to include when building-up a custom configuration.
-* excluded-layers: A set of layers to exclude when building-up a custom configuration.
+* layers: A set of layers to include when building-up a custom configuration. This can be configured with the system property ``wildfly.provisioning.layers`` containing a comma separated list of layers.
+* excluded-layers: A set of layers to exclude when building-up a custom configuration.  This can be configured with the system property ``wildfly.provisioning.layers.excluded`` containing a comma separated list of layers.
+* layers-configuration-file-name: The name of the configuration file generated from layers. Default value is ``standalone.xml``. 
+If no ``layers`` have been configured, setting this parameter is invalid.
+* overwrite-provisioned-server: By default if the server referenced from the ``provisioning-dir`` exists when provision/package goals are called, the provisioning/packaging is disabled. By setting
+this parameter to false, the provisioned server will get deleted and provisioning/packaging will occur.
```
honza-kasik, line 152: > By setting this parameter to false, the provisioned server will get deleted and provisioning/packaging will occur.

If false is default behaviour, then this has to be set to `true`. Or not?

honza-kasik (commented):
Thank you @jfdenise . Apart from one detail/typo this looks good from QE PoV.

maven/WFLY-14934_Server_Plugin.adoc
jfdenise: This was a typo, I fixed it to be true.

honza-kasik (approved):
Thank you, LGTM from QE point of view.

PR #462, opened 2022-03-18
[ELY-2320] Add integrity support to FilesystemSecurityRealm

Ashpan:


elytron/ELY-2320-integrity-for-FilesystemSecurityRealm.adoc
```diff
@@ -0,0 +1,152 @@
...
+
+```bash
+/subsystem=elytron/key-store=keystore:add(path=keystore, relative-to=jboss.server.config.dir, type=JKS, credential-reference={clear-text=secret})
+/subsystem=elytron/key-store=keystore:generate-key-pair(alias=localhost,algorithm=RSA,key-size=1024,validity=365,distinguished-name="CN=localhost")
+/subsystem=elytron/key-store=keystore:store()
+```
+```bash
+/subsystem=elytron/filesystem-realm=fsRealm:add(path=fs-realm-users, relative-to=jboss.server.config.dir, key-store=keystore, key-store-alias=localhost)
+```
+The signature will be added to the identity xml file at the end before the closing `</identity>` tag.
```
OndrejKotek, line 105: Is an XML standard applied?

elytron/ELY-2320-integrity-for-FilesystemSecurityRealm.adoc
```diff
@@ -0,0 +1,152 @@
...
+[source]
+----
+firstUser.xml:d1b31f991807feb73dba25e075609950
+----
+it would look like this
+[source]
+----
+RUxZAUMQUO9rq4PRc3BB61B5oZSepATonjzjUriqbsy//5fw:RUxZAUMQ41IKUmpazIiXdB5oZSep4wQa67I+y35TLtssweZ
+----
+==== Hard Requirements
```
OndrejKotek, line 73: Is it possible to enable this feature for an existing realm without a need of conversion? In case the encryption feature the `elytron-tool` has to be used.

elytron/ELY-2320-integrity-for-FilesystemSecurityRealm.adoc
```diff
@@ -0,0 +1,152 @@
...
+:author:            Ashpan Raskar
+:email:             araskar@redhat.com
+:toc:               left
+:icons:             font
+:idprefix:
+:idseparator:       -
+
+== Overview
+
+This objective of this proposal is to add the ability to verify integrity for the filesystem realm, this will be done with the use of a Public and Private Key pair, that will create a signature on every identity on the xml identity file.
```
OndrejKotek, line 11: > This objective of this proposal is to add the ability to verify integrity for the filesystem realm

The whole filesystem realm, or just individual identities? I'm able to add a new identity with name of my choice just by copying an existing one. It is also possible to override an existing identity this way.

OndrejKotek, line 11: > This objective of this proposal is to add the ability to verify integrity for the filesystem realm

The whole filesystem realm, or just individual identities? I'm able to add a new identity with name of my choice just by copying an existing one. It is also possible to override an existing identity this way.

elytron/ELY-2320-integrity-for-FilesystemSecurityRealm.adoc
Ashpan: Yes this is doable, will add to Elytron and then update this document

OndrejKotek: The process should be the opposite way - plan first, then implementation. It seems we don't need `elytron-tool` for this functionality as there is the `update-key-pair` CLI operation.

Ashpan: Sorry will update this plan first. 
And that is correct we don't need the elytron tool feature, the CLI operation will work for this.

elytron/ELY-2320-integrity-for-FilesystemSecurityRealm.adoc
Ashpan: Yes this xml schema is used: https://www.w3.org/TR/2002/REC-xmldsig-core-20020212/xmldsig-core-schema.xsd

elytron/ELY-2320-integrity-for-FilesystemSecurityRealm.adoc
Ashpan: I've changed it so now you could copy and existing identity but any other change to the identity (eg changing attributes, name, credentials) will flag as an error next time the identity is being referenced

elytron/ELY-2320-integrity-for-FilesystemSecurityRealm.adoc
```diff
@@ -0,0 +1,149 @@
...
```
fjuma, line 11: s/that will create a signature on every identity on the xml identity file/that will be used to add a signature for each identity in the xml identity files.

elytron/ELY-2320-integrity-for-FilesystemSecurityRealm.adoc
```diff
@@ -0,0 +1,149 @@
...
+
+To store the KeyPair, a key store will be created and the FileSystem realm will utilize the key store to access the PublicKey and PrivateKey to verify filesystem integrity.
+
+For the default out of box configuration for WildFly, the goal is for the FileSystemSecurityRealm to have integrity enabled.
```
fjuma, line 15: Would be good to mention that this would enabled as part of a separate RFE.

elytron/ELY-2320-integrity-for-FilesystemSecurityRealm.adoc
```diff
@@ -0,0 +1,149 @@
...
+=== Affected Projects or Components
+- Widlfly, Wildfly Elytron, Wildfly Core
+
+=== Other Interested Projects
+N/A
+
+== Requirements
+=== Integrity of Filesystem Realm
+
+Filesystem Integrity is a planned upcoming change, but will not be added until after WildFly 26.1 is released
```
fjuma, line 50: This line can be removed now.

elytron/ELY-2320-integrity-for-FilesystemSecurityRealm.adoc
```diff
@@ -0,0 +1,149 @@
...
+
+If the approach of signing a file was chosen, it would require an asymmetric key pair, a public key and private key. This would involve a key store and a method to ensure only the correct party has access to the private key.
+
+It could be made that the public key is available for everyone to use. This means anyone can decrypt the signature to get the hash of the identity, but only people with access to the key store for the realm would have access to the private key that encrypts the hash of each identity. This way an unauthorized user could view the hash via the public key, but cannot change the hash/signature of the identity.
+
+The approach that is taken is to use the key pair file signature approach. To do so, ever time data is to be written or fetched by an identity, first the filesystem realm will verify the signature with identity file. If this signature ever does not match the identity, then the filesystem realm integrity has been compromised.
```
fjuma, line 60: s/by an identity/for an identity

fjuma: s/If this signature ever does not match the identity/If signature verification fails

elytron/ELY-2320-integrity-for-FilesystemSecurityRealm.adoc
```diff
@@ -0,0 +1,149 @@
...
+The identity file with integrity enabled via a Private and Public Key that contains a clear text password along with an attribute would look like the following.
+```XML
+<?xml version="1.0" encoding="UTF-8" standalone="no"?><identity xmlns="urn:elytron:1.0">
+    <credentials>
+        <password algorithm="digest-md5" format="base64">AgR1c2VyCmZpbGVzeXN0ZW3dJf7xqO/siBRfcZNHPL9d</password>
+    </credentials>
+    <attributes>
+        <attribute name="Roles" value="JBossAdmin"/>
+    </attributes><Signature xmlns="http://www.w3.org/2000/09/xmldsig#"><SignedInfo><CanonicalizationMethod Algorithm="http://www.w3.org/TR/2001/REC-xml-c14n-20010315"/><SignatureMethod Algorithm="http://www.w3.org/2001/04/xmldsig-more#rsa-sha256"/><Reference URI=""><Transforms><Transform Algorithm="http://www.w3.org/2000/09/xmldsig#enveloped-signature"/></Transforms><DigestMethod Algorithm="http://www.w3.org/2001/04/xmlenc#sha256"/><DigestValue>KzLR5/cf78iV3iT1L7fgl7z2Wq1XTvoD2FTqk9FXWYs=</DigestValue></Reference></SignedInfo><SignatureValue>IsOfSpixRMUmgdkuCy/gANiETAfEF5/XWIl8yKT50JjpH3ChGE5i/O1mO4YLPC4OwzilJPNSGHTb&#13;
```
fjuma, line 110: Is the signature element here copy/pasted from an identity file? Is it possible to add proper formatting of this XML to the file?

elytron/ELY-2320-integrity-for-FilesystemSecurityRealm.adoc
```diff
@@ -0,0 +1,149 @@
...
+N/A
+
+=== Non-Requirements
+In a follow up RFE the default out of the box configuration will be changed to use the filesystem realm to replace the properties realm.
+Another follow up RFE can also be created to improve the efficiency by using a checksum hash and an index. If that checksum matches from the index, then the signature does not need to be verified.
+
+== Implementation Plan
+=== Filesystem Integrity
+
+Filesystem Integrity is a planned upcoming change, but will not be added until after WildFly 26.1 is released
```
fjuma, line 132: I think this line can be removed.

elytron/ELY-2320-integrity-for-FilesystemSecurityRealm.adoc
Skyllarr, line 60: @Ashpan  Just checking, does the `key pair file signature approach` means computing the checksum of the file and signing this checksum?

Skyllarr: @Ashpan Thank you. Tthe solution mentioned in the last discussion was to compute a checksum, and sign only that checksum. Which reduces the amount of data that is being processed by the signature algorithm. Should this be added to the analysis doc or was it decided to not use a checksum at all?

Ashpan: Oh I see the confusion. For the current time being we decided to not use the checksum at all anymore, and as a follow up task for this RFE, I'll create an issue for using the checksum to increase efficiency.

fjuma (approved):

PR #466, opened 2022-03-22
[WFCORE-5533] Proposal for simplifying a secondary host controller configuration

Skyllarr:
https://issues.redhat.com/browse/WFCORE-5533


elytron/WFCORE-5533-simplify-secondary-host-controller-configuration.adoc
```diff
@@ -0,0 +1,87 @@
...
+
+* https://github.com/wildfly/wildfly-core[WildFly Core project]
+
+== Requirements
+
+=== Hard Requirements
+
+* It must be possible to configure the secondary host's credentials by providing username and password using a credential reference and without having to specify authentication context
+* This username and password will be used to initialize AuthenticationContext for the connection.
+* Backwards compatibility should be kept in mind as well as which credentials should have precedence in case both are defined
```
fjuma, line 49: I think we should make sure that either the `authentication-context` has been specified or a `credential-reference` has been specified. I don't think we should allow both.

elytron/WFCORE-5533-simplify-secondary-host-controller-configuration.adoc
```diff
@@ -0,0 +1,87 @@
...
+
+N/A
+
+=== Non-Requirements
+
+N/A
+
+== Implementation Plan
+
+New elements `username` and `credential-reference` will  be added to remote's configuration.
```
fjuma, line 61: It looks like a `username` attribute already exists. Can that be used here?

PR #469, opened 2022-03-28
[WFCORE-5532] Proposal for realm readiness checking with Elytron

Skyllarr:
https://issues.redhat.com/browse/WFCORE-5532

### DIFF
--- a/elytron/WFCORE-5532-realm-readiness-check.adoc
+++ b/elytron/WFCORE-5532-realm-readiness-check.adoc
@@ -1,0 +1,78 @@
+= WFCORE-5532 Provide an Elytron equivalent for realm readiness checking
+:author:            Diana Krepinska
+:email:             dvilkola@redhat.com
+:toc:               left
+:icons:             font
+:idprefix:
+:idseparator:       -
+
+== Overview
+
+When no management user was added to the management realm and user tried to access the UI console, the wildfly redirected to the error page that advised to add a management user with `ad-user.sh` script. This redirection took effect only when realm could be checked for users, eg. the default properties realm. Legacy LDAP realm and others were always assumed to contain users. This issue seeks to add similar fuctionality to Elytron.
+
+== Issue Metadata
+
+=== Issue
+
+* https://issues.redhat.com/browse/WFCORE-5533[WFCORE-5533]
+
+=== Related Issues
+
+* https://issues.redhat.com/browse/EAP7-1098[EAP7-1098] - EAP issue tracker
+
+=== Dev Contacts
+
+* TODO
+
+=== QE Contacts
+
+* TODO
+
+=== Testing By
+// Put an x in the relevant field to indicate if testing will be done by Engineering or QE.
+// Discuss with QE during the Kickoff state to decide this
+[x] Engineering
+
+[ ] QE
+
+=== Affected Projects or Components
+
+* https://github.com/wildfly/wildfly-core[WildFly Core project]
+* https://github.com/wildfly/wildfly[WildFly project]
+
+== Requirements
+
+=== Hard Requirements
+
+* When realm associated with management interface is empty and user tries to access management console, wildfly will redirect to the error page advising user to use add-user script
+* Above is true only for some types of realms which allow to check if it is empty. Example of such realm is properties realm, identity realm and filesystem realm. LDAP realm, JAAS realm will never redirect to error page because those realms do not provide possibility to check for existence of users
+* When realm is not empty it will provide a challenge
+
+=== Nice-to-Have Requirements
+
+N/A
+
+=== Non-Requirements
+
+N/A
+
+== Implementation Plan
+
+Add new method eg. `isReadyForAuthentication`/`canResultInSuccessfulAuthentication` to SecurityRealm interface, similar to legacy method `isReadyForHttpChallenge()`, that can be used to find out if at least 1 user was added to the realm. New method will be added to security domain that will check its map of realms to find out whether the domain contains users.
+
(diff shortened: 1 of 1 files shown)


elytron/WFCORE-5532-realm-readiness-check.adoc
```diff
@@ -0,0 +1,78 @@
...
+
+== Issue Metadata
+
+=== Issue
+
+* https://issues.redhat.com/browse/WFCORE-5533[WFCORE-5533]
+
+=== Related Issues
+
+* https://issues.redhat.com/browse/EAP7-1823[EAP7-1823] - EAP issue tracker
```
fjuma, line 21: I think this should be EAP7-1098.

elytron/WFCORE-5532-realm-readiness-check.adoc
```diff
@@ -0,0 +1,78 @@
...
+
+LDAP realm, JAAS realm and others that do not provide possibility to check how many users are presentwill default to true, and the error page advising user to add management user will never be shown. But filesystem realm will check if the realm's folder is empty, properties realm will check properties file for users. This method will be used in the `createReadyFunction(Builder builder)` method of `ManagementHttpServer`. The current security domain can be obtained in `createReadyFunction(Builder builder)` method from builder's httpAuthenticationFactory, or the realm will default to "ManagementRealm". The security domain cwill internally check the map of its realms to check for existence of users in management domain.
+
+The error page that prompts to add a user and the redirection is already implemented, so wildfly-elytron changes and updating of the `createReadyFunction(Builder builder)` method of `ManagementHttpServer` in wildfly-core should suffice.
+
```
fjuma, line 66: The legacy realm readiness check also attempted to check if client cert auth was possible. Can we do something similar to that as well?

elytron/WFCORE-5532-realm-readiness-check.adoc
```diff
@@ -0,0 +1,78 @@
...
+=== Non-Requirements
+
+N/A
+
+== Implementation Plan
+
+Add new method to SecurityRealm interface, similar to legacy method `isReadyForHttpChallenge()`, that can be used to find out if at least 1 user was added to the realm. New method will be added to security domain that will check its map of realms to find out whether the domain contains users.
+
+LDAP realm, JAAS realm and others that do not provide possibility to check how many users are presentwill default to true, and the error page advising user to add management user will never be shown. But filesystem realm will check if the realm's folder is empty, properties realm will check properties file for users. This method will be used in the `createReadyFunction(Builder builder)` method of `ManagementHttpServer`. The current security domain can be obtained in `createReadyFunction(Builder builder)` method from builder's httpAuthenticationFactory, or the realm will default to "ManagementRealm". The security domain cwill internally check the map of its realms to check for existence of users in management domain.
+
```
fjuma, line 64: Just something to keep in mind. Currently, the add-user.sh script is used to add users to a properties realm.

When the default out of the box configuration makes use of a filesystem realm instead, a different approach (i.e., add-user 2.0 / CLI commands) would need to be used to add users.

elytron/WFCORE-5532-realm-readiness-check.adoc
```diff
@@ -0,0 +1,78 @@
...
+=== Non-Requirements
+
+N/A
+
+== Implementation Plan
+
+Add new method to SecurityRealm interface, similar to legacy method `isReadyForHttpChallenge()`, that can be used to find out if at least 1 user was added to the realm. New method will be added to security domain that will check its map of realms to find out whether the domain contains users.
+
+LDAP realm, JAAS realm and others that do not provide possibility to check how many users are presentwill default to true, and the error page advising user to add management user will never be shown. But filesystem realm will check if the realm's folder is empty, properties realm will check properties file for users. This method will be used in the `createReadyFunction(Builder builder)` method of `ManagementHttpServer`. The current security domain can be obtained in `createReadyFunction(Builder builder)` method from builder's httpAuthenticationFactory, or the realm will default to "ManagementRealm". The security domain cwill internally check the map of its realms to check for existence of users in management domain.
```
fjuma, line 63: Just a note here. For the filesystem realm, I think it's possible for the realm's folder to contain subdirectories but no files (e.g., try adding a user, then removing it, and checking the resulting directory structure). So when checking if the realm's directory is empty, this is something that needs to be taken into account too.

elytron/WFCORE-5532-realm-readiness-check.adoc
```diff
@@ -0,0 +1,78 @@
...
+
+N/A
+
+=== Non-Requirements
+
+N/A
+
+== Implementation Plan
+
+Add new method to SecurityRealm interface, similar to legacy method `isReadyForHttpChallenge()`, that can be used to find out if at least 1 user was added to the realm. New method will be added to security domain that will check its map of realms to find out whether the domain contains users.
```
fjuma, line 61: I wonder if we should go for a different method name than what was used in the legacy security approach to make it more clear what the method does, e.g., `isNonEmpty`, `containsUsers`, `isReadyForAuthentication` or something like that.

Skyllarr: Yes I agree. The `containUsers` might be misleading for  the CLIENT_CERT mechanism since the realm does not have to contain the users IIUC, only roles? So that leaves `isNonEmpty` and `isReadyForAuthentication` or something like `canResultInSuccessfulAuthentication`.

elytron/WFCORE-5532-realm-readiness-check.adoc
Skyllarr: @fjuma added this info, thanks

elytron/WFCORE-5532-realm-readiness-check.adoc
Skyllarr: Added this info to the proposal to not forget, thanks!

elytron/WFCORE-5532-realm-readiness-check.adoc
Skyllarr: we can check if the mechanism names obtained from the builder's httpAuthenticationFactory contain `CLIENT_CERT`. I have added it to the proposal and it will become more clear during implementation if that works well and is sufficient. I think it should work, thanks!

darranl:
Superseded by https://github.com/wildfly/wildfly-proposals/pull/511

PR #472, opened 2022-03-29
[WFCORE-5864] Proposal to update default configuration to use hardened Elytron configuration.

skyllarr:
https://issues.redhat.com/browse/WFCORE-5864

### DIFF
--- a/elytron/WFCORE-5864-update-default-configuration-to-use-elytron-capabilities-instead-of-legacy-capabilities.adoc
+++ b/elytron/WFCORE-5864-update-default-configuration-to-use-elytron-capabilities-instead-of-legacy-capabilities.adoc
@@ -1,0 +1,81 @@
+= WFCORE-5864 Update default configuration to use Elytron capabilities instead of legacy capabilities.
+:author:            Diana Krepinska
+:email:             dvilkola@redhat.com
+:toc:               left
+:icons:             font
+:idprefix:
+:idseparator:       -
+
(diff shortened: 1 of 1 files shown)


elytron/WFCORE-5864-update-default-configuration-to-use-elytron-capabilities-instead-of-legacy-capabilities.adoc
```diff
@@ -0,0 +1,81 @@
...
+== Overview
+
+The default configuration was already updated to use the Elytron capabilities instead of the legacy security capabilities in WildFly 25.
+
+This RFE seeks to define which Elytron's default security configuration to harden and update default configuration accordingly.
+
+Some of the examples that can be updated include:
+    * usage of the encrypted filesystem realm instead of the legacy properties realm
+    * usage of SCRAM SASL by default instead of DIGEST MD5
+    * disabled older TLS protocols
```
fjuma, line 18: Since the older TLS protocols (like TLSv1.1) are automatically disabled in recent JDK versions, maybe we can update this point to be specifically about enabling TLSv1.3 by default. Currently, that needs to be explicitly enabled by users.

Skyllarr: @fjuma Updated. Thanks!

PR #476, opened 2022-05-26
[WFCORE-5689] Support simplified configuration of the Graceful Shutdo…

yersan:
…wn Timeout and Shutdown Timeout in service scripts

Jira issue: https://issues.redhat.com/browse/WFCORE-5689

### DIFF
--- a/scripts/WFCORE-5689-graceful_shutdown_timeout_and_shutdown_timeout_service_scripts.adoc
+++ b/scripts/WFCORE-5689-graceful_shutdown_timeout_and_shutdown_timeout_service_scripts.adoc
@@ -8,7 +8,7 @@
 
 == Overview
 
-The server can be installed as a Lunix service by using two types of init daemon services:
+The server can be installed as a Linux service by using two types of init daemon services:
 
 * Init, also known as System V init, or SysVinit: The server is shipped with the required files to install it as an Init daemon service at $SERVER_HOME/docs/contrib/scripts/init.d
 * Systemd (System daemon): The server is shipped with the required files to install it as a System daemon service at $SERVER_HOME/docs/contrib/scripts/systemd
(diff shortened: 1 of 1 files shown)


jamezp (approved):
This looks good to me.

scripts/WFCORE-5689-graceful_shutdown_timeout_and_shutdown_timeout_service_scripts.adoc
```diff
@@ -0,0 +1,119 @@
...
+** Systemd services: The user can configure it on $SERVER_HOME/docs/contrib/scripts/systemd/wildfly.service by using the `TimeoutStopSec` directive.
+
+The Shutdown Timeout always wins, if the graceful shutdown timeout is bigger than the shutdown timeout, you could be in the situation where your server is killed in the middle of a suspension.
+
+The initial goal of this RFE was to standardize the configuration of both timeouts on both types of scripts by using environment variables. However, that's not possible for the `TimeoutStopSec` directive, because systemd directives don't allow variable expansions. To accomplish the goal for this directive we would need to stop the server by supplying a new bash script to be executed on the `ExecStop` directive. It is better to use the out-of-the-box directive instead of reimplementing a similar feature by us. In addition, systemd services only would accept command line variables when these variables are configured as global variables. Those global variables are made available to all the systemd services, so changes could affect to any other service installation or have side effects out of its purpose.
```
yersan, line 26: > It is better to use the out-of-the-box directive instead of reimplementing a similar feature by us

I don't want to sound categoric here, so, feel free to add your opinion. I just only found it difficult to justify the implementation by us of this shutdown timeout just because we want to make it available as an environment variable, after all, it is already available on the service configuration file.

The only possible advantage is to allow the user to configure it on wildfly.conf instead of wildfly.service and get them documented there, but in my opinion, that would not justify adding by ourselves something that is already available out of the box. I could be wrong or maybe there is a way to do it I have not found. So, just my opinion.

yersan:
Closed as original issue was closed

PR #489, opened 2022-06-15
WFLY-16345 @Stateful Jakarta Enterprise Bean instance externalization to remote Infinispan cluster

pferraro:
https://issues.redhat.com/browse/WFLY-16345

### DIFF
--- a/clustering/ejb/WFLY-16345_SFSB_externalization_to_remote_Infinispan_server.adoc
+++ b/clustering/ejb/WFLY-16345_SFSB_externalization_to_remote_Infinispan_server.adoc
@@ -1,0 +1,91 @@
+= @Stateful Jakarta Enterprise Bean instance externalization to remote Infinispan cluster
+:author:            Paul Ferraro
+:email:             paul.ferraro@redhat.com
+:toc:               left
+:icons:             font
+:idprefix:
+:idseparator:       -
+
+== Overview
+
+Currently, WildFly only supports the use of embedded Infinispan for persisting/replicating @Stateful Jakarta Enterprise Bean instances.
+This implementation cannot fully support cross-site replication, since the inability to invalidate entries across sites means that site failback can result in loss of state.
+This feature will add the ability to persist @Stateful Jakarta Enterprise Bean instances in a remote Infinispan cluster via the HotRod protocol, where cross-site replication is fully supported.
+
+== Issue Metadata
+
+=== Issue
+
+* https://issues.redhat.com/browse/WFLY-16345[WFLY-16345]
+
+=== Related Issues
+
+* https://issues.redhat.com/browse/WFLY-14953[WFLY-14953]
+
+=== Dev Contacts
+
+* mailto:{email}[{author}]
+
+=== QE Contacts
+
+* TBD
+
+=== Testing By
+// Put an x in the relevant field to indicate if testing will be done by Engineering or QE. 
+// Discuss with QE during the Kickoff state to decide this
+* [X] Engineering
+
+* [ ] QE
+
+=== Affected Projects or Components
+
+=== Other Interested Projects
+
+=== Relevant Installation Types
+// Remove the x next to the relevant field if the feature in question is not relevant
+// to that kind of WildFly installation
+* [x] Traditional standalone server (unzipped or provisioned by Galleon)
+
+* [x] Managed domain
+
+* [x] OpenShift s2i
+
+* [x] Bootable jar
+
+== Requirements
+
+=== Hard Requirements
+
+* Add hotrod-bean-management resource to distributable-ejb subsystem
+** New resource will contain a required attribute referencing a remote-cache-container from the Infinispan subsystem
+* Add hotrod-based bean management provider implementation
+** Implementation will require no strong nor weak affinity, unless near-caching is enabled, where beans should have weak affinity to the local node.
+** Passivation/activation semantics will be the same as the embedded bean management implementation when using a persistent cache configuration
+
+=== Nice-to-Have Requirements
+
+* Support near-caching of SFSB instances, using a configured max-active-beans
+
+=== Non-Requirements
+
+* No support for ProtoStream marshalling of SFSB instances at this time
+
+== Backwards Compatibility
+
+N/A
+
+== Security Considerations
+
+Security of the hotrod connection is already managed by the Infinispan subsystem.
+
+== Test Plan
+
+* Add remote Infinispan variants of SFSB failover and timeout tests.
+
+== Community Documentation
+
+* Update distributable-ejb documentation to include configuring a hotrod-bean-management provider.
+
+== Release Note Content
+
+* WildFly now supports externalization of @Stateful Jakarta Enterprise Bean instances to a remote Infinispan cluster via the HotRod protocol.


clustering/ejb/WFLY-16345_SFSB_externalization_to_remote_Infinispan_server.adoc
```diff
@@ -0,0 +1,91 @@
...
+:author:            Paul Ferraro
+:email:             paul.ferraro@redhat.com
+:toc:               left
+:icons:             font
+:idprefix:
+:idseparator:       -
+
+== Overview
+
+Currently, WildFly only supports the use of embedded Infinispan for persisting/replicating @Stateful Jakarta Enterprise Bean instances.
```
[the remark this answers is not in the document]
pferraro: Yes, the proposed `/subsystem=distributable-ejb/hotrod-bean-management=*` resource is very much analogous to the `/subsystem=distributable-web/hotrod-session-management=*` resource.

Where `/distributable-ejb/infinispan-bean-management=*` configures a distributed bean manager backed by an embedded cache,
`/distributable-ejb/hotrod-bean-management=*` will configure a distributed bean manager backed by a remote cache.

clustering/ejb/WFLY-16345_SFSB_externalization_to_remote_Infinispan_server.adoc
```diff
@@ -0,0 +1,91 @@
...
+=== QE Contacts
+
+* TBD
+
+=== Testing By
+// Put an x in the relevant field to indicate if testing will be done by Engineering or QE. 
+// Discuss with QE during the Kickoff state to decide this
+* [X] Engineering
+
+* [ ] QE
```
tommaso-borgato, line 38: @pferraro is it worth adding at least one trivial test into the clustering TS?

pferraro: There are failover and expiration tests included in the proposed implementation.

clustering/ejb/WFLY-16345_SFSB_externalization_to_remote_Infinispan_server.adoc
```diff
@@ -0,0 +1,91 @@
...
+* [x] Bootable jar
+
+== Requirements
+
+=== Hard Requirements
+
+* Add hotrod-bean-management resource to distributable-ejb subsystem
+** New resource will contain a required attribute referencing a remote-cache-container from the Infinispan subsystem
+* Add hotrod-based bean management provider implementation
+** Implementation will require no strong nor weak affinity, unless near-caching is enabled, where beans should have weak affinity to the local node.
```
tommaso-borgato, line 62: @pferraro having weak affinity to the local node (= node that last handled a request for a given session) when using near-caching, make somewhat sense to me given invalidation messages are involved ... instead the "Implementation will require no strong nor weak affinity" part is unclear to me ... could you try and explain it to me?

tommaso-borgato: IIUC, when we enable the near-cache (which is an option when using a remote Infinispan cluster), we would have the SFSB on the node that handled the request, and then it make sense to have EJB client affinity to that same node (mode = "local")

clustering/ejb/WFLY-16345_SFSB_externalization_to_remote_Infinispan_server.adoc
```diff
@@ -0,0 +1,91 @@
...
+** Passivation/activation semantics will be the same as the embedded bean management implementation when using a persistent cache configuration
```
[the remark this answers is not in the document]
pferraro: The activation/passivation application callbacks are invoked per invocation, in the same manner as they are with the existing persistence use case.
That is to say, on a given SFSB invocation, @PostActivate is invoked prior to the invocation, and @PrePassivate is invoked after the invocation.
The semantics are the same since, in both use cases (i.e. infinispan-server vs embedded cache with shared persistence), deserialization and serialization of SFSB state happens per invocation.

clustering/ejb/WFLY-16345_SFSB_externalization_to_remote_Infinispan_server.adoc
pferraro: Correct.

clustering/ejb/WFLY-16345_SFSB_externalization_to_remote_Infinispan_server.adoc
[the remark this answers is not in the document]
pferraro: Correct.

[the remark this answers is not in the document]
tommaso-borgato: @pferraro in a set of clustered WF nodes, what would be the difference(s) between using the new `hotrod-bean-management` and the existing `invalidation-cache`?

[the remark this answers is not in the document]
pferraro: Advantages of the RemoteCache-based managers (over their embedded invalidation-cache equivalents) include:

- Simpler load balancing requirements, i.e. no need for primary-owner based routing
- Low clustering overhead for EAP servers
  - No need for JGroups discovery, failure detection, merge/join handling, etc.
  - Faster to scale up/down EAP cluster

Disadvantages of the RemoteCache-based managers (over their embedded cache equivalents) include:

- Generally slower individual response times
  - Each request require extra network call (reads always go remote)
  - Near caching should be avoided unless EAP cluster size is small
- Inefficient expiration handling
  - Ensuring \@PreDestroy is invoked on a single cluster member is very expensive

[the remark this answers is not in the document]
tommaso-borgato: @pferraro IIUC the main use case for this feature is using standalone WildFly instances in a cloud environment lile OpenShift; at present, to cache SFSB in a remote Infinispan cluster, we would need to have  WildFly clustered which implies using a [`StafulSet`](https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/) instead of a simple [`Deployment`](https://kubernetes.io/docs/concepts/workloads/controllers/deployment/); this feature would used in cross-site scenarios as well;
If that's correct perhaps we can add a simple test with our usual two nodes cluster connected to a remote Infinispan two nodes cluster and another test running on OpenShift;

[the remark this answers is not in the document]
tommaso-borgato: A more comprehensive example of a configuration we might use without this new feature is the following:
```
embed-server --server-config=standalone-ha.xml

/socket-binding-group=standard-sockets/remote-destination-outbound-socket-binding=remote-jdg-server1:add(host=10.0.145.66, port=11222)
/socket-binding-group=standard-sockets/remote-destination-outbound-socket-binding=remote-jdg-server2:add(host=10.0.145.120, port=11222)

# JDG Nodes
/subsystem=elytron/key-store=wildfly-truststore-containing-ca:add(path=wildfly-truststore.p12, relative-to=jboss.home.dir, credential-reference={clear-text=123PIPPOBAUDO}, type=PKCS12)
/subsystem=elytron/trust-manager=wildfly-trust-manager-containing-ca:add(key-store=wildfly-truststore-containing-ca, algorithm="SunX509")

/subsystem=elytron/client-ssl-context=CLIENT_SSL_CONTEXT:add(trust-manager=wildfly-trust-manager-containing-ca, protocols=["TLSv1.2"])
batch
/subsystem=infinispan/remote-cache-container=web-sessions:add(default-remote-cluster=jdg-server-cluster, statistics-enabled=true, properties={infinispan.client.hotrod.auth_username=admin, infinispan.client.hotrod.auth_password=pass.1234})
/subsystem=infinispan/remote-cache-container=web-sessions/remote-cluster=jdg-server-cluster:add(socket-bindings=[remote-jdg-server1,remote-jdg-server2])
run-batch
/subsystem=infinispan/remote-cache-container=web-sessions/component=security:write-attribute(name=ssl-context,value=CLIENT_SSL_CONTEXT)
if (outcome == success) of /subsystem=infinispan/remote-cache-container=web-sessions:read-attribute(name=marshaller)
/subsystem=infinispan/remote-cache-container=web-sessions:write-attribute(name=marshaller,value=PROTOSTREAM)
end-if
/subsystem=infinispan/cache-container=web/invalidation-cache=offload:add()
/subsystem=infinispan/cache-container=web/invalidation-cache=offload/store=hotrod:add(remote-cache-container=web-sessions, fetch-state=false, preload=false, passivation=false, purge=false, shared=true)
/subsystem=infinispan/cache-container=web/invalidation-cache=offload/component=transaction:add(mode=BATCH)

/subsystem=infinispan/cache-container=web:write-attribute(name=default-cache, value=offload)

/subsystem=infinispan/cache-container=ejb/invalidation-cache=offload:add()
/subsystem=infinispan/cache-container=ejb/invalidation-cache=offload/store=hotrod:add(remote-cache-container=web-sessions, fetch-state=false, preload=false, passivation=false, purge=false, shared=true)
/subsystem=infinispan/cache-container=ejb/invalidation-cache=offload/component=transaction:add(mode=BATCH)

/subsystem=infinispan/cache-container=ejb:write-attribute(name=default-cache, value=offload)
```

clustering/ejb/WFLY-16345_SFSB_externalization_to_remote_Infinispan_server.adoc
```diff
@@ -0,0 +1,91 @@
...
+:email:             paul.ferraro@redhat.com
+:toc:               left
+:icons:             font
+:idprefix:
+:idseparator:       -
+
+== Overview
+
+Currently, WildFly only supports the use of embedded Infinispan for persisting/replicating @Stateful Jakarta Enterprise Bean instances.
+This implementation cannot fully support cross-site replication, since the inability to invalidate entries across sites means that site failback can result in loss of state.
```
tommaso-borgato, line 12: @pferraro it's not clear to me how, with the implementation proposed in this RFE, we can overcome this limitation

clustering/ejb/WFLY-16345_SFSB_externalization_to_remote_Infinispan_server.adoc
tommaso-borgato: @pferraro I am not sure if we agreed to try and add one test on OpenShift where we check that basic replication works (e.g. two not clustered WF PODs connected to some Infinispan Server, we take one down, we verify the other node has the data)

clustering/ejb/WFLY-16345_SFSB_externalization_to_remote_Infinispan_server.adoc
pferraro: An invalidation cache, by definition caches entries.  If this invalidation cache uses a shared remote Infinispan cluster as a cache store, and any entries are update in a remote site, there is no mechanism to invalidate the entries within the invalidation cache.

clustering/ejb/WFLY-16345_SFSB_externalization_to_remote_Infinispan_server.adoc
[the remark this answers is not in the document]
pferraro: This use case is already validated by the test introduced within the feature implementation.

clustering/ejb/WFLY-16345_SFSB_externalization_to_remote_Infinispan_server.adoc
tommaso-borgato, line 11: @pferraro I am not sure I fully understood the following:

> WildFly only supports the use of embedded Infinispan for persisting/replicating @Stateful Jakarta Enterprise Bean instances

That's because I was able to configure WildFly like:

```
embed-server --server-config=standalone-ha.xml

# JDG Nodes
/socket-binding-group=standard-sockets/remote-destination-outbound-socket-binding=remote-jdg-server1:add(host=127.0.0.1, port=11223)
/socket-binding-group=standard-sockets/remote-destination-outbound-socket-binding=remote-jdg-server2:add(host=127.0.0.1, port=11224)

# create the remote-cache-container
batch
/subsystem=infinispan/remote-cache-container=remote-infinispan-cluster:add(default-remote-cluster=jdg-server-cluster, statistics-enabled=true, properties={infinispan.client.hotrod.auth_username=admin, infinispan.client.hotrod.auth_password=pass.1234})
/subsystem=infinispan/remote-cache-container=remote-infinispan-cluster/remote-cluster=jdg-server-cluster:add(socket-bindings=[remote-jdg-server1,remote-jdg-server2])
run-batch

# add hotrod store to existing "distributed-cache=dist"
/subsystem=infinispan/cache-container=ejb/distributed-cache=dist/store=hotrod:add(remote-cache-container=remote-infinispan-cluster, fetch-state=false, preload=false, passivation=false, purge=false, shared=true)

# make "distributed-cache=dist" the default cache for EJB
/subsystem=infinispan/cache-container=ejb:write-attribute(name=default-cache, value=dist)
```

And I could see data being cached on the Infinispan cluster:

image

which looks like the SFSB is actually cached by:

```java
@WebServlet(name = "LocalEjbServlet", urlPatterns = { "/ejbservlet" })
public class LocalEjbServlet extends HttpServlet {

    @Inject
    private LocalStatefulSB bean;

    @Override
    protected void doGet(HttpServletRequest req, HttpServletResponse resp) throws ServletException, IOException {
        resp.setContentType("text/plain");
        int serial = bean.getSerialAndIncrement();
        resp.getWriter().print(serial);
    }

    @Override
    public String getServletInfo() {
        return "Servlet invoking Stateful Session Bean to store serial.";
    }
}
```

tommaso-borgato: SUMMARY: 

The goal of this RFE is to get rid **entirely** of any cache on the WF node that is involved in handling SFSB: we agreed to include this information in the AD ... the WF node would either create the session and the SFSB or retrieve the whole session data, for an existing session, from the remote Infinispan cluster;

Every node, clustered or not, would be somewhat "stateless": every time the node receives a request it would have to store/retrieve session data from the remote Infinispan cluster;

The "embedded Infinispan" refers to any `invalidation-cache`, `distributed-cache`, `replicated-cache`, "whatever" cache on the WF node;

Please note that is doesn't make sense to use a `distributed-cache` backed by a hotrod store (as in the example above) because they would serve the same purpose and we would do it twice! this is why we want to get rid of the `distributed-cache` element in this scenario...

The challenge that we need to solve is handling SFSB expiration: we have several entries involved with handling one (or more) SFSB on a remote Infinispan cluster:
- one entry to handle expiration
- one entry to actually store the serialized SFSB
When expiration happens:
1. the remote Infinispan cluster send the expiration message to **all** the connected WF nodes
2. each WF node leverages the **consistent hashing** mechanism to find out if it is the primary owner and hence in charge for handling that expiration message
3. if it is, it sends a "remove and get" message to the remote Infinispan cluster which actually removes the cache entry and retrieves the serialized SFSB
4. the WF node re-assembles the SFSB and run the code associated with SFSB expiration
5. the nodes that are not in charge for handling the expiration message, receive it regardless, they just have to make sure the primary owner actually processes the message: here is the challenge because this would involve a lot of network traffic

^^^ @pferraro can you validate the above?

PR #371, opened 2021-02-08
[WFCORE-5279] Split Elytron SSL into it's own subsystem

nekdozjam:
https://issues.redhat.com/browse/WFCORE-5279

PR #500, opened 2022-10-21
[WFCORE-5279] Split Elytron SSL into its own subsystem

cam-rod:
- https://issues.redhat.com/browse/EAP7-1564
- https://issues.redhat.com/browse/WFCORE-5279

Continuation from #371

### DIFF
--- a/elytron/WFCORE-5279-elytron-ssl-subsystem.adoc
+++ b/elytron/WFCORE-5279-elytron-ssl-subsystem.adoc
@@ -1,0 +1,172 @@
+= [WFCORE-5279] Split Elytron SSL into it's own subsystem
+:author:            Martin Mazanek
+:email:             mmazanek@redhat.com
+:toc:               left
+:icons:             font
+:idprefix:
+:idseparator:       -
+
+== Overview
+
+
+Working on bootable jar layers it is apparent that SSL requirements for a server can be distinct from the general authentication requirements, also the elytron subsystem takes a number of resources to define a working SSL configuration.
+
+This RFE is to look at providing a new subsystem dedicated to a simplified SSL configuration, the elytron subsystem approach will not be deprecated and will be preserved to leave administrators with complete control but the majority can likely be satisfied with a simplified model.
+
+
+== Issue Metadata
+
+=== Issue
+
+* https://issues.redhat.com/browse/WFCORE[WFCORE-5279]
+
+=== Related Issues
+
+* https://issues.redhat.com/browse/ELY[ELY-2084]
+
+=== Dev Contacts
+
+* mailto:{email}[{author}]
+
+=== QE Contacts
+
+=== Testing By
+// Put an x in the relevant field to indicate if testing will be done by Engineering or QE. 
+// Discuss with QE during the Kickoff state to decide this
+* TBD
+//* [ ] Engineering
+
+//* [ ] QE
+
+=== Affected Projects or Components
+
+* WildFly Elytron
+* WildFly Core
+* WildFly (Community Documentation)
+
+//=== Other Interested Projects
+
+=== Relevant Installation Types
+// Remove the x next to the relevant field if the feature in question is not relevant
+// to that kind of WildFly installation
+* [x] Traditional standalone server (unzipped or provisioned by Galleon)
+
+* [x] Managed domain
+
+* [x] OpenShift s2i
+
+* [x] Bootable jar
+
+== Requirements
+
+=== Hard Requirements
+
+In order to reduce the required configuration, we need to select the appropriate use-case scenarios. In order to make the best of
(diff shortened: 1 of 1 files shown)


elytron/WFCORE-5279-elytron-ssl-subsystem.adoc
```diff
@@ -0,0 +1,196 @@
...
+:idprefix:
+:idseparator:       -
+
+== Overview
+
+Working on bootable jar layers, it is apparent that TLS requirements for a server can be distinct from the general
+authentication requirements, and also that the Elytron subsystem takes a number of resources to define a working TLS
+configuration.
+
+This RFE is to look at providing a new subsystem dedicated to a simplified TLS configuration. The Elytron subsystem
```
OndrejKotek, line 17: Do we plan to replace the default TLS configuration with this new one?

elytron/WFCORE-5279-elytron-ssl-subsystem.adoc
```diff
@@ -0,0 +1,196 @@
...
+not be delivered all in one go to suggest the strategy. If your feature falls 
+into this category, please mention the Release Coordinators on the pull 
+request so they are aware.
+////
+
+
+The Undertow subsystem uses reference to an SSLContext resource provided with the 'org.wildfly.security.ssl-context' capability.
+This makes implementation of a SSL subsystem straight-forward - we need to create a   SSLContext resource with the same capability and we can wire it in by referencing its name.
+
+Unlike the 'elytron' subsystem we want the configuration as simple as possible - ideally one CLI command.
```
OndrejKotek, line 151: One command is the only simplification for users I can see in this RFE now, the rest seems be rather a code refactoring on background. Having two subsystems seems rather confusing. Are there another visible advantages? Is there a way to keep just one subsystem? (For example a Galleon layer with just TLS related configuration in the Elytron subsystem.)

elytron/WFCORE-5279-elytron-ssl-subsystem.adoc
cam-rod: Not at the moment, this is just planned to be a simpler way of setting up a TLS configuration. Elytron's existing code would remain for integrating with security domains, mappers, etc. when needed

elytron/WFCORE-5279-elytron-ssl-subsystem.adoc
cam-rod: Still need to updating this section from #371. Ideally, we'd also include an updated `elytron-tool.sh` for the improvements. I'll keep this in mind as I update the section

cam-rod: One thing I wanted to add from some of the earlier planning discussions is that by splitting this into it's own subsystem, it reduces the number of dependencies needed, like in non-Galleon setups, so it reduces the overall server demand.

cam-rod:
Updated the proposal to clarify some aspects of the scope, planned changes, and features, based on my better understanding of how components interact. Copying the commit messages below:

> **[Clarified scope, updated elytron-tls-tool](https://github.com/wildfly/wildfly-proposals/pull/500/commits/57a0bbed4080cc2f6d838318e149e96dac728995)**
> 
> Main body of the proposal has be updated to better reflect the scope
> of the subsytem (TLS only, no SASL/JASPI). Also details that the
> elytron-tls-tool should support interactive configuration, like
> add-user.

> **[Refined proposal for command and config structure](https://github.com/wildfly/wildfly-proposals/pull/500/commits/63ec10a6c1bbc9cb83688518d7de019d9164733f)**
> 
> The previous proposal discussed using both references and objects to
> store some elements (ex. key stores) within the subsystem. This makes
> it effectively incompatible with Elytron proper and also much harder to
> edit.
> 
> This updated proposal would only use "object attributes" for commands,
> but saved configuration will match that of Elytron as much as possible.

elytron/WFCORE-5279-elytron-ssl-subsystem.adoc
```diff
@@ -0,0 +1,282 @@
...
+
+The subsystem is being developed in the ``wildfly-security-incubator`` org as
+https://github.com/wildfly-security-incubator/elytron-tls[elytron-tls], a standalone project.
+
+=== Hard Requirements
+
+In order to reduce the required configuration, dependencies, and system resources, Elytron TLS targets
+one major collection of use-cases: authentication of a server (and optionally client) using HTTPS. By limiting the
+scope to HTTP over TLS, we can remove most of the alternative authentication and authorization tools available in
+Elytron, including security domains, principal transformers, SASL support, and HTTP authentication mechanisms. This
```
fjuma, line 79: Just a minor comment, I think the SASL part can be removed from here.

PR #501, opened 2022-10-21
[WFLY-16793] Support for identity propagation within an EAR and across EARs with OIDC

fjuma:
https://issues.redhat.com/browse/WFLY-16793

### DIFF
--- a/elytron/WFLY-16793-oidc-identity-propagation.adoc
+++ b/elytron/WFLY-16793-oidc-identity-propagation.adoc
@@ -1,0 +1,199 @@
+= WFLY-16793 Support for Identity Propagation within an EAR and across EARs with OIDC
+:author:            Farah Juma
+:email:             fjuma@redhat.com
+:toc:               left
+:icons:             font
+:idprefix:
+:idseparator:       -
+
+== Overview
+
+For a deployment secured with OpenID Connect using the `elytron-oidc-client` subsystem,
+the subsystem automatically creates and makes use of a virtual security domain. With
+virtual security domains, it's currently possible to propagate a security identity from the
+web layer to an EJB within the same WAR. However, there are some limitations. In particular,
+the following two things are currently missing:
+
+* Ability to propagate an identity from a WAR to an EJB within a *single* EAR
+* Ability to propagate an identity from a WAR within an EAR to an EJB within *another* EAR
+(i.e., across deployments)
+
+This RFE is to add support for these two things.
+
+== Issue Metadata
+
+=== Issue
+
+* https://issues.redhat.com/browse/WFCORE-6190[WFCORE-6190]
+* https://issues.redhat.com/browse/WFLY-16793[WFLY-16793]
+
+=== Related Issues
+
+* https://issues.redhat.com/browse/EAP7-1971[EAP7-1971]
+* https://issues.redhat.com/browse/WFLY-17312[WFLY-17312]
+* https://issues.redhat.com/browse/WFLY-17333[WFLY-17333]
+
+=== Dev Contacts
+
+* mailto:{email}[{author}]
+
+=== QE Contacts
+
+=== Testing By
+// Put an x in the relevant field to indicate if testing will be done by Engineering or QE. 
+// Discuss with QE during the Kickoff state to decide this
+
+TBD
+
+* [ ] Engineering
+
+* [ ] QE
+
+=== Affected Projects or Components
+
+* WildFly Core
+* WildFly
+
+=== Other Interested Projects
+
+=== Relevant Installation Types
+// Remove the x next to the relevant field if the feature in question is not relevant
+// to that kind of WildFly installation
+* [x] Traditional standalone server (unzipped or provisioned by Galleon)
+
+* [x] Managed domain
+
+* [x] OpenShift s2i
+
+* [x] Bootable jar
+
+== Requirements
+
+=== Hard Requirements
+
+* A new `virtual-security-domain` resource will be introduced in the `elytron` subsystem.
+The new resource will make it possible to outflow security identities established by virtual
+security domains to other security domains. The new `virtual-security-domain` resource will have
+a few attributes:
+
+** `name` - This is the runtime name of a deployment associated with a virtual security domain (e.g.,
+`DEPLOYMENT_NAME.ear`, a deployment that has a subdeployment that is secured using OIDC).
+
+** `outflow-security-domains` - This is the list of `security-domains` that security identities from
+the virtual security domain should be automatically outflowed to. This will be an optional attribute
+and must be non-empty if specified.
+
+** `outflow-anonymous` - When outflowing to a security domain, if outflow is not possible, should the
+anonymous identity be used? Outflow to a security domain might not be possible if the domain does
+not trust this domain or if the identity being outflowed to a domain does not exist in that domain.
+Outflowing anonymous has the effect of clearing any identity already established for that domain.
+This attribute defaults to `false`.
+
+** `auth-method` - The authentication mechanism that will be used with the virtual security domain.
+Allowed values: 'OIDC', 'MP-JWT'. The default value is 'OIDC'.
+
+** As an example, a `virtual-security-domain` could be added as follows:
+[source]
+----
+/subsystem=elytron/virtual-security-domain=DEPLOYMENT_NAME.ear:add(outflow-security-domains=...)
+----
+
+* We also need the ability to specify the virtual security domains that should be trusted by a `security-domain`.
+A new `trusted-virtual-security-domains` attribute will be added to the `security-domain` resource.
+** e.g., `trusted-virtual-security-domains=DEPLOYMENT_NAME.ear` would indicate that this `security-domain`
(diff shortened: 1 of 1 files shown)


elytron/WFLY-16793-oidc-identity-propagation.adoc
```diff
@@ -0,0 +1,166 @@
...
+the virtual security domain should be automatically outflowed to. This will be a required attribute
+and must be non-empty.
+
+** As an example, a `virtual-security-domain` could be added as follows:
+[source]
+----
+/subsystem=elytron/virtual-security-domain=DEPLOYMENT_NAME.ear:add(outflow-security-domains=...)
+----
+
+* A new `trusted-virtual-security-domains` attribute will be added to the `security-domain` resource
```
OndrejKotek, line 93: Will the new `virtual-security-domain` resource provide the `org.wildfly.security.security-domain` capability or a new one?

If the existing one, can we leverage `trusted-security-domains` instead of add the new `trusted-virtual-security-domains` attribute?

elytron/WFLY-16793-oidc-identity-propagation.adoc
```diff
@@ -0,0 +1,166 @@
...
+////
+Identification if any security implications that may need to be considered with this feature
+or a confirmation that there are no security implications to consider.
+////
+
+== Test Plan
+
+Elytron subsystem parsing and transformer tests will be added.
+
+Tests will be added to the WildFly testsuite to verify that a security identity
```
OndrejKotek, line 153: Are there any other subsystems/resources establishing virtual domains? If it's the case, are those somehow specific and require more tests?

elytron/WFLY-16793-oidc-identity-propagation.adoc
fjuma: Currently, when a virtual security domain is created, the `SecurityDomain` service that is provided has a name like `DEPLOYMENT_NAME.ear.security-domain.virtual`. So although it provides a `SecurityDomain`, the service name is formatted differently from a normal `security-domain` that's been configured.

We could update this so that the service name for the virtual case is of the usual form `org.wildfly.security.security-domain.DEPLOYMENT_NAME.ear`. That way, a separate `trusted-virtual-security-domains` attribute would not be needed. Instead, the existing `trusted-security-domains` attribute could be used since we'd be able to determine the service name the usual way. WDYT?

elytron/WFLY-16793-oidc-identity-propagation.adoc
```diff
@@ -0,0 +1,166 @@
...
+Note that in general, we are following the same concept of security outflow and trust that
+is used when propagating identities between traditional Elytron `security-domains`. Here, we
+are simply extending this concept for the case where we need to propagate an identity from
+a virtual security domain to a traditional `security-domain`.
+
+=== Nice-to-Have Requirements
+
+=== Non-Requirements
+
+Note that in the future, we might want to consider adding other attributes to the `virtual-security-domain`
```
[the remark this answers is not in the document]
fjuma: Since we've seen recent requests from the community for this, I think we should consider adding an initial set of attributes as part of this RFE. I've updated the Nice-to-Have Requirements section with this list.

elytron/WFLY-16793-oidc-identity-propagation.adoc
fjuma: The only other place where a virtual security domain is used is for MP JWT. We can add some sanity tests for this case as well. I'll update the analysis.

elytron/WFLY-16793-oidc-identity-propagation.adoc
OndrejKotek: I'm not familiar with such internals and possible impacts of such change, but it sounds reasonable and makes sense to me. It would definitively help us eliminate one extra attribute and keep the configuration relatively simple.

fjuma: +1. I've updated the analysis to mention this.

elytron/WFLY-16793-oidc-identity-propagation.adoc
```diff
@@ -0,0 +1,186 @@
...
+
+* outflow-anonymous
+* pre-realm-principal-transformer
+* post-realm-principal-transformer
+* principal-decoder
+* evidence-decoder
+* role-decoder
+* role-mapper
+* permission-mapper
+* security-event-listener
```
OndrejKotek, line 120: Just wondering, are some "identity propagation" events emitted? If it's the case, on which side?

fjuma: For identity propagation, the `ServerAuthenticationContext#importIdentity` method gets called to import the identity into the target domain (see https://github.com/wildfly-security/wildfly-elytron/blob/1.x/auth/server/base/src/main/java/org/wildfly/security/auth/server/ServerAuthenticationContext.java#L1630). If the identity can be propagated successfully, the target domain will emit a `RealmIdentitySuccessfulAuthorizationEvent`. Otherwise, it will emit a `RealmIdentityFailedAuthorizationEvent`.

fjuma: Note that for virtual security domains, the usual security events will also get emitted upon successful / failed authentication.

OndrejKotek: Thank you @fjuma.

elytron/WFLY-16793-oidc-identity-propagation.adoc
fjuma: Looked into this some more and I think we'll need the `trusted-virtual-security-domains` attribute after all. Once a `virtual-security-domain` has been added to the elytron subsystem, the actual `SecurityDomain` service that corresponds to this won't be created until the deployment unit processors get run for the deployment. This means that attempting to reference the corresponding security domain from the `trusted-security-domains` attribute won't be possible because that service won't actually be available yet. Instead, we can reference the `virtual-security-domain` from a new `trusted-virtual-security-domains` attribute. Thus, the complete configuration would look something like:

```
/subsystem=elytron/virtual-security-domain=myapp.ear:add(outflow-security-domains=[MyDomain])
/subsystem=elytron/security-domain=MyDomain:write-attribute(name=trusted-virtual-security-domains,value=[myapp.ear])
```

OndrejKotek: OK, thanks for the clarification.  Do I understand it well that something like

```/subsystem=elytron/virtual-security-domain=myapp.ear:add(outflow-security-domains=[MyDomain])```

will work even without `myapp.ear` being deployed?

Would it be useful (and possible) to check whether all security domains set in `outflow-security-domains` have also corresponding `trusted-virtual-security-domains` set and vice versa? It could help user to identify wrong configuration for example by logging such findings.

claudioweiler:
Hi. I'm following this PR and I see that it will add new configurations to WF security subsystem.

But I have a simple question. This will bring back old behavior back on WF or it will always require additional configurations?

With old behavior, I mean this from original issue: "_propagate an identity from a WAR to an EJB when the WAR and EJB were packaged in an EAR_".

fjuma:
> Hi. I'm following this PR and I see that it will add new configurations to WF security subsystem.
> 
> But I have a simple question. This will bring back old behavior back on WF or it will always require additional configurations?
> 
> With old behavior, I mean this from original issue: "_propagate an identity from a WAR to an EJB when the WAR and EJB were packaged in an EAR_".

Hi @claudioweiler, thanks for the feedback! Just to double check, you're referring to the case where the EJB uses the same security domain as the WAR, right? (So with the old configuration, the case where both the WAR and EJB were using the `KeycloakDomain`?) The `outflow-security-domains` and `trusted-virtual-security-domains` config introduced so far is for the case where the security domains are different. I'm looking further into the case where the security domains are the same and then will add some more thoughts here.

elytron/WFLY-16793-oidc-identity-propagation.adoc
fjuma: Yes, the `virtual-security-domain` resource can be added without the application being deployed (similar to the `secure-deployment` resource in the `elytron-oidc-client` subsystem).

Adding the checks you mentioned is an interesting idea but we don't really have a way to perform such checks, i.e., it wouldn't be possible to do these checks when adding the `virtual-security-domain` resource (because the corresponding `trusted-virtual-security-domains` wouldn't have been configured yet) or when adding a `security-domain` resource. This would need to be something that's analyzed after the whole subsystem has been configured.

fjuma:
Hi @claudioweiler, I've added information about the case where the EJB should be secured using the same virtual security domain as the WAR, please let me know if you have any questions about this:

https://github.com/wildfly/wildfly-proposals/pull/501/files#diff-eea221141c27ec874df64eaab02d81479e6ed12f4287b086b7b51ac3a9fc312bR103-R113

@OndrejKotek FYI ^.

OndrejKotek (approved):

bstansberry:
Thanks @fjuma and @OndrejKotek

PR #502, opened 2022-11-11
[WFLY-17273] Global tracking-mode setting in Undertow

gaol:
https://issues.redhat.com/browse/WFLY-17273


gaol:
@fl4via could you please review? thanks

undertow/WFLY-17273_Global_tracking_mode_setting_in_Undertow.adoc
```diff
@@ -0,0 +1,100 @@
...
+
+* https://issues.redhat.com/browse/WFLY-17273[WFLY-17273]
+
+=== Related Issues
+
+* https://issues.redhat.com/browse/EAP7-1980[EAP7-1980]
+
+=== Dev Contacts
+
+* mailto:{email}[{author}]
```
fl4via, line 37: please, don't forget to add your info here @gaol

undertow/WFLY-17273_Global_tracking_mode_setting_in_Undertow.adoc
```diff
@@ -0,0 +1,100 @@
...
+
+== Requirements
+
+Provide a way to specify global session tracking modes at servlet container level. The configuration takes effect only if there are no session tracking modes specified in the deployment. So the configuration either in `web.xml` or updated in a `ServletContextListener` implementation takes precedence, which keeps consistent behavior as what it is now.
+
+=== Hard Requirements
+
+There will be 2 ways to set up global session tracking modes in the server:
+
+- Use system property with prefix `undertow.session.tracking-modes`. It reads from system property of `undertow.session.tracking-modes.CONTAINER-NAME`, and falls over to `undertow.session.tracking-modes` if it is not defined.
```
fl4via, line 66: I don't see why we need a system property. system property are added usually when it is too late to add something to the model, for example when we need a fix to a bug urgently and the fix cannot wait for the whole RFE process. Still, it is the goal to eventually add those system properties to the model and deprecate those in the code.

undertow/WFLY-17273_Global_tracking_mode_setting_in_Undertow.adoc
```diff
@@ -0,0 +1,100 @@
...
+* The value gets validated on deployment time, the benefit is that users can use `/system-property=*` to update the global session tracking modes after server starts up, the disadvantage is that the wrong configuration can only be noticed on the deployment time.
+* Use system property approach makes it possible to back port to downstream which does not allow model changes.
+
+- Use an attribute in servlet container: `/subsystem=undertow/servlet-container=default:write-attribute(name="session-tracking-modes", value=[COOKIE, URL])`
```
fl4via, line 70: I think that calling it defaullt-tracking-mode makes it clearer the purpose of the new attribute

undertow/WFLY-17273_Global_tracking_mode_setting_in_Undertow.adoc
```diff
@@ -0,0 +1,100 @@
...
+
+WildFly Undertow subsystem
+
+=== Other Interested Projects
+
+N/A
+
+== Requirements
+
+Provide a way to specify global session tracking modes at servlet container level. The configuration takes effect only if there are no session tracking modes specified in the deployment. So the configuration either in `web.xml` or updated in a `ServletContextListener` implementation takes precedence, which keeps consistent behavior as what it is now.
```
fl4via, line 60: Why do we need more than one tracking mode? If I understood the Jira correctly, don't we need a way to define a global tracking mode that is going to apply to all deployed applications that do not specify the session-tracking-mode in their web.xml file?

fl4via (changes requested):

undertow/WFLY-17273_Global_tracking_mode_setting_in_Undertow.adoc
gaol: yes, `defaullt-tracking-mode` makes more sense

undertow/WFLY-17273_Global_tracking_mode_setting_in_Undertow.adoc
gaol: It refers to the vars defined at the top in the doc: `:author:` and `:email:`

undertow/WFLY-17273_Global_tracking_mode_setting_in_Undertow.adoc
gaol: There is a EAPSUP issue linked to https://issues.redhat.com/browse/WFLY-17273, I thought there is a request back porting to EAP 7.4, if that is not the case, I think we can avoid the system property approach.

@bmaxwell would you please comment if we need this back porting to EAP 7.4 ?

undertow/WFLY-17273_Global_tracking_mode_setting_in_Undertow.adoc
gaol: It is possible to have multiple servlet-containers defined in undertow subsystem, so it is possible to specify different `defaullt-tracking-mode` for each servlet-container. For each servlet-container, I think it is possible to define the `defaullt-tracking-mode` to be a combination of `COOKIE`, `URL`, `SSL`(this is exclusive).

But since the default tracking modes are `COOKIE` and `URL` now already, maybe it is better to just have the `defaullt-tracking-mode` to be single choice. thanks

undertow/WFLY-17273_Global_tracking_mode_setting_in_Undertow.adoc
gaol: It is also possible to use system property for EAP 7.4 in case it is needed, and use attribute only upstream. :)

gaol: Thanks, what do you mean by changes in undertow ? I think it can be done in wildfly undertow subsystem directly.  Please take a look at the section of the implementation plan below. maybe there are anything I missed ?

gaol:
@fl4via I updated with the feedback, would you please review again? thanks :)

PR #506, opened 2022-12-22
[WFLY-15260] Support for securing the management console with OIDC

fjuma:
https://issues.redhat.com/browse/WFLY-15260

### DIFF
--- a/elytron/WFLY-15260_oidc_management_console.adoc
+++ b/elytron/WFLY-15260_oidc_management_console.adoc
@@ -1,0 +1,138 @@
+= WFLY-15260 Securing the management console with OIDC
+:author:            Farah Juma
+:email:             fjuma@redhat.com
+:toc:               left
+:icons:             font
+:idprefix:
+:idseparator:       -
+
+== Overview
+
+With the Keycloak OpenID Connect (OIDC) adapter, it was possible to secure the management console
+using OIDC. When accessing the management console, the user would get redirected to the Keycloak
+login page, log in with their credentials, and get redirected back to the management console upon
+successful authentication. It was also possible for the user to log out of the console.
+
+This RFE is to add the ability to secure the management console when using the native support for
+OIDC.
+
+== Issue Metadata
+
+=== Issue
+
+* https://issues.redhat.com/browse/WFLY-15260[WFLY-15260]
+
+=== Related Issues
+
+* https://issues.redhat.com/browse/WFLY-15485[WFLY-15485]
+* https://issues.redhat.com/browse/EAP7-1796[EAP7-1796]
+
+=== Dev Contacts
+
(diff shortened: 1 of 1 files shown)


elytron/WFLY-15260_oidc_management_console.adoc
```diff
@@ -0,0 +1,137 @@
...
+to publish the OIDC configuration for the management console).
+** The previous `keycloak` subsystem published the OIDC configuration to be used for the
+management console via the http://localhost:MANAGEMENT_PORT/keycloak/adapter/wildfly-console/index.html
+endpoint. HAL would access this endpoint and if it found OIDC configuration, HAL would then
+redirect to the Keycloak login page. Similarly, the `elytron-oidc-client` subsystem will publish the
+OIDC configuration to be used for the management console to a new endpoint (e.g.,
+http://localhost:MANAGEMENT_PORT/elytron-oidc-client/wildfly-console/index.html). HAL will need to be
+updated to access this endpoint and if it finds OIDC configuration, HAL would then
+redirect to the OIDC provider login page.
+* It will be possible to log out of the management console after having logged in with OIDC.
```
OndrejKotek, line 85: Just local logout of the console, or the complete OIDC logout?

elytron/WFLY-15260_oidc_management_console.adoc
```diff
@@ -0,0 +1,137 @@
...
+////
+
+== Security Considerations
+
+This is a security RFE.
+
+== Test Plan
+
+Tests will be added to the WildFly testsuite to verify that the management console
+can be secured successfully with OIDC.
```
OndrejKotek, line 128: Simple tests for the logout and for using RBAC would be also useful.

elytron/WFLY-15260_oidc_management_console.adoc
fjuma: I think that with the existing support, logging out resulted in the complete OIDC logout. So the idea would be to do the same here.

elytron/WFLY-15260_oidc_management_console.adoc
fjuma: Will update this section to mention that, thanks @OndrejKotek!

hpehl (approved):
Sorry @fjuma, for the delay. I missed that this still needs my review. 

The RFE looks good to me. No objections from my side.

elytron/WFLY-15260_oidc_management_console.adoc
kstekovi: Will the logout have effect also on an application deployed into EAP secured by OIDC when we are logged in with the same user?

kstekovi:
Looks good also to me.

PR #507, opened 2023-01-18
[WFLY-14735] Management interface FORM authentication,

darranl:
https://issues.redhat.com/browse/WFLY-14735

### DIFF
--- a/elytron/WFLY-14735-Management-FORM-authentication.adoc
+++ b/elytron/WFLY-14735-Management-FORM-authentication.adoc
@@ -1,0 +1,245 @@
+= WFLY-14735 FORM authentication for the management console with MFA
+:author:            Darran Lofthouse
+:email:             darran.lofthouse@redhat.com
+:toc:               left
+:icons:             font
+:idprefix:
+:idseparator:       -
+
+== Overview
+
+The development of this RFE is delivering a subset of functionality from a much larger planned change,
+this RFE will start with two features:
+ * Enabling FORM authentication for admin console access to the application server.
+ * Where custom security realms are in use supporting MFA authentication on the login form to 
+ request additional attributes.
+
+== Issue Metadata
+
+=== Issue
+
+* https://issues.redhat.com/browse/WFLY-14735[WFLY-14735]
+
+=== Related Issues
+
+* https://issues.redhat.com/browse/EAP7-1696[EAP7-1696]
+
+=== Dev Contacts
+
+* mailto:{email}[{author}]
+
+=== QE Contacts
+
+=== Testing By
+// Put an x in the relevant field to indicate if testing will be done by Engineering or QE. 
+// Discuss with QE during the Kickoff state to decide this
+* [ ] Engineering
+
+* [ ] QE
+
+=== Affected Projects or Components
+
+The main development for this RFE will be a new `elytron-authentication` subsystem 
+in the https://github.com/wildfly/wildfly[WildFly] project.
+
+Changes will also be made within https://github.com/wildfly/wildfly-core[WildFly Core] to
+enable and support the activation of FORM authentication.
+
+Updates will be required to the https://github.com/hal/console[Admin Console] to support
+possibly being required to redirect to the login form and to make use of the logout endpoint.
+
+Security components developed for this RFE will be contained within the
+https://github.com/wildfly-security/wildfly-elytron[WildFly Elytron] project.
+
+=== Other Interested Projects
+
+None
+
+=== Relevant Installation Types
+// Remove the x next to the relevant field if the feature in question is not relevant
+// to that kind of WildFly installation
+* [x] Traditional standalone server (unzipped or provisioned by Galleon)
+
+* [x] Managed domain
+
+* [x] OpenShift s2i
+
+* [x] Bootable jar
+
+== Requirements
+
+=== Hard Requirements
+
+==== General Flow
+
+On the HTTP management interface a new context will be made available at the path `/authentication`,
+when a user connects to the root `/` context of the application server they will be redirected to the
+`/authentication` context.  The `/authentication` context will display a login form for the user 
+to authenticate themselves.
+
+The look and feel of the login FORM will make use of `PatternFly` for consistent with the look and
+feel of the web console.
+
+Under the `/authentication` context there will be two sub contexts `/login` and `/logout`, after 
+a user has input their username and password these will be passed to the first of the sub contexts to
+perform the authentication.  Where authentication is successful a cryptographically signed JWT token 
+will be generated and passed back to the client as a cookie, the login form will then redirect the 
+user to the admin console.
+
+The admin console context is unauthenticated so will load as normal, the `/management` enpoint will 
+use a different HTTP authentication mchanism configuration which will use the `BEARER` HTTP 
+authentication mechanism which will be configured to use the JWT token from the incoming request cookie.
+
+When the user has finished managing the application server they can click the usual Logout link in 
+the admin console which will redirect them to `/authentication/logout`, this context will clear 
+the cookie containing the JWT token.
+
+==== Other Flows
+
+If after the username and password is sent to the `/authentication/login` context the authentication 
+fails the user will remain on the login page and a suitably vague error will be delayed to indicate
+that authentication has failed.
+
+If a user has previously been authenticated and their cookie has expired or the user is using the 
+admin console to connect to a different server it is possible the admin console may sent an invocation
+to the `/management` context without an appropriate cookie, in this case the admin console will need
+to be "notified to" and handle a redirect to the `/authentication` context.
+
+==== elytron-authentication
+
+A new subsystem `elytron-authentication` will be developed, the primary role of this subsystem
+will be to make a HTTP context available to be published.  This subsystem will also be responsible for
+the generation of the cryptographically signed JWT token and also provide a configured BEARER HTTP
+authentication mechanism which can validate the tokens produced by this server only.
+
+The generation and signing of the JWT tokens will largely be internal to the subsystem and will
+not be configurable.  The subsystem will manage a dynamically generated key pair for token signing
+and validation, this will live as long as the application server process and will not survive restarts.
+The key pair however will survive reloads as it is common when managing an application server to reload
+the configuration.
+
+Although this subsystem is only being used for the security of the management HTTP interface at this
+time later it will be expanded to be used for deployment security as well.
+
+The HTTP context configuration will cross reference a `security-domain` for the authentication 
+requirements.
+
+This subsystem will also support the definition of arbitrary attributes which can be requested from
+the user at the time of authentication, as an example this could be used to request a single use
+password in addition to the regular password.  A new evidence type will be defined within WildFly Elytron
+which will accept an arbitrary set of attributes during authentication.  The existing security realms
+within WildFly Elytron that handle username/password based authentication will be updated to support
+this new dynamic evidence type.
+
+==== core-management
+
+The number of attributes related to security on the management interfaces has grown with different
+permutations needed for different situations.  To reduce the complexity a new resource `policy=security`
+will be added to the `core-management` subsystem, this will allow references to the sasl or http 
+authentication factories as well as to ssl-contexts to be used.  This new policy resouce will also be
+able to reference the new context made available from the `elytron-authentication` subsystem.
+
+==== Management Interfaces
+
+On the management interfaces a new attribute `use-core-management-security-policy` will be added,
+this will be a boolean and when set to `true` the configuration defined on the `policy=security`
+resource will be used instead of the individual attributes on the management interface.
+
+==== Out Of Scope
+
+The RFE will not be making changes to the SASL authentication for native access, the out of the
+box configuration will still default to supporting the JBOSS_LOCAL_USER and DIGEST SASL mechanisms.
+
+This RFE will also not be changing the getting started experience adding the initial users to the
+server using the add-user utility and this RFE will not be changing the properties files used to
+contain the initial users.
+
+=== Nice-to-Have Requirements
+
+Should the management interfaces have TLS enabled by default, we may be able to also generate a 
+wildfly-config.xml for clients to use poining at a Keystore for the server's certificate.
+
+This could however be high risk breaking various clients.
+
+=== Non-Requirements
+
+The architecture chosen for this RFE will form the basis for future development which will not be covered by
+this RFE.
+
+It is envisaged that the form authentication provided by the `elytron-authentication` subsystem will be 
+enhanced further with capabilities such as:
+
+ * Support for MFA authentication using OTP such as HOTP / TOTP.
+ * Dynamic decisions based on username such as recognising a user is registered for SSO.
+ * Fido2 / WebAuth authentication with or without a username.
+ * Multi factor authentication using push notifications.
+ * Customisation of the look and feel of the form.
+ 
+In addition to using the form for the management interfaces it will be possible to use configurations
+to secure deployments making all of these options available.
+
+These additional features are broadly the reason for placing the initial configuration for the form 
+into a new subsystem.
+
+== Backwards Compatibility
+
+=== Default Configuration
+
+Prior to this change the default configuration for the management interfaces referenced 
+`*-authentication-factory` resources to provide the SASL and HTTP authentication which was 
+DIGEST based, this change updates  the configuration to use a security policy withing the 
+core-management subsystem as well as changing  the default authentication of the HTTP management
+interface to use FORM authentication.
+
+Unless we also enable TLS for the management interfaces clients which connect to the native 
+interface either  directly or via HTTP Upgrade will be unaffected as this will still be using 
+SASL authentication which will be unchanged by this RFE.
+
+=== Importing Existing Configuration
+
+Existing configuration will be imported "as-is" and will continue to use the previously 
+defined behaviour, users who want to add form authentication to the HTTP management interface will 
+need to manually add it themselves.
+
+=== Deployments
+
+This RFE only affects authentication for the management interfaces so does not affect deployments.
+
+=== Interoperability
+
+The main interoperability risk is clients connecting to the HTTP management interface and assuming
+they will be authenticating using DIGEST authentication, modifications will be needed to the new default
+configuration to add additional authentication mechanisms to the HTTP management interface when accessed
+directly using HTTP.
+
+//== Implementation Plan
+////
+Delete if not needed. The intent is if you have a complex feature which can 
+not be delivered all in one go to suggest the strategy. If your feature falls 
+into this category, please mention the Release Coordinators on the pull 
+request so they are aware.
+////
+
+== Security Considerations
+
+////
+Identification if any security implications that may need to be considered with this feature
+or a confirmation that there are no security implications to consider.
+////
+
+== Test Plan
+
+== Community Documentation
+////
+Generally a feature should have documentation as part of the PR to wildfly master, or as a follow up PR if the feature is in wildfly-core. In some cases though the documentation belongs more in a component, or does not need any documentation. Indicate which of these will happen.
+////
+== Release Note Content
+////
+Draft verbiage for up to a few sentences on the feature for inclusion in the
+Release Note blog article for the release that first includes this feature. 
+Example article: http://wildfly.org/news/2018/08/30/WildFly14-Final-Released/.
+This content will be edited, so there is no need to make it perfect or discuss
+what release it appears in.  "See Overview" is acceptable if the overview is
+suitable. For simple features best covered as an item in a bullet-point list 
+of features containing a few words on each, use "Bullet point: <The few words>" 
+////
--- a/elytron/WFLY-14735-Management-FORM-authentication.adoc
+++ b/elytron/WFLY-14735-Management-FORM-authentication.adoc
@@ -10,6 +10,7 @@
 
 The development of this RFE is delivering a subset of functionality from a much larger planned change,
 this RFE will start with two features:
+
  * Enabling FORM authentication for admin console access to the application server.
  * Where custom security realms are in use supporting MFA authentication on the login form to 
  request additional attributes.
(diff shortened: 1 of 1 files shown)


elytron/WFLY-14735-Management-FORM-authentication.adoc
```diff
@@ -0,0 +1,245 @@
...
+:email:             darran.lofthouse@redhat.com
+:toc:               left
+:icons:             font
+:idprefix:
+:idseparator:       -
+
+== Overview
+
+The development of this RFE is delivering a subset of functionality from a much larger planned change,
+this RFE will start with two features:
```
OndrejKotek, line 12: Probably an empty line is required to render the following lines as a list.

elytron/WFLY-14735-Management-FORM-authentication.adoc
```diff
@@ -0,0 +1,245 @@
...
+feel of the web console.
+
+Under the `/authentication` context there will be two sub contexts `/login` and `/logout`, after 
+a user has input their username and password these will be passed to the first of the sub contexts to
+perform the authentication.  Where authentication is successful a cryptographically signed JWT token 
+will be generated and passed back to the client as a cookie, the login form will then redirect the 
+user to the admin console.
+
+The admin console context is unauthenticated so will load as normal, the `/management` enpoint will 
+use a different HTTP authentication mchanism configuration which will use the `BEARER` HTTP
```
OndrejKotek, line 90: mchanism -> mechanism

elytron/WFLY-14735-Management-FORM-authentication.adoc
```diff
@@ -0,0 +1,245 @@
...
+The number of attributes related to security on the management interfaces has grown with different
+permutations needed for different situations.  To reduce the complexity a new resource `policy=security`
+will be added to the `core-management` subsystem, this will allow references to the sasl or http 
+authentication factories as well as to ssl-contexts to be used.  This new policy resouce will also be
+able to reference the new context made available from the `elytron-authentication` subsystem.
+
+==== Management Interfaces
+
+On the management interfaces a new attribute `use-core-management-security-policy` will be added,
+this will be a boolean and when set to `true` the configuration defined on the `policy=security`
```
OndrejKotek, line 145: Do I understand it correctly that the default value will be `true`?

elytron/WFLY-14735-Management-FORM-authentication.adoc
```diff
@@ -0,0 +1,245 @@
...
+This RFE will also not be changing the getting started experience adding the initial users to the
+server using the add-user utility and this RFE will not be changing the properties files used to
+contain the initial users.
+
+=== Nice-to-Have Requirements
+
+Should the management interfaces have TLS enabled by default, we may be able to also generate a 
+wildfly-config.xml for clients to use poining at a Keystore for the server's certificate.
+
+This could however be high risk breaking various clients.
```
OndrejKotek, line 162: Can we mitigate this a bit by allowing to return to the previous behavior in a simple way? I.e. by setting `use-core-management-security-policy` to false?

elytron/WFLY-14735-Management-FORM-authentication.adoc
```diff
@@ -0,0 +1,245 @@
...
+ 
+In addition to using the form for the management interfaces it will be possible to use configurations
+to secure deployments making all of these options available.
+
+These additional features are broadly the reason for placing the initial configuration for the form 
+into a new subsystem.
+
+== Backwards Compatibility
+
+=== Default Configuration
```
OndrejKotek, line 186: Regarding the `security` command, will just `*-http-auth-management` be affected?

elytron/WFLY-14735-Management-FORM-authentication.adoc
```diff
@@ -0,0 +1,245 @@
...
+None
+
+=== Relevant Installation Types
+// Remove the x next to the relevant field if the feature in question is not relevant
+// to that kind of WildFly installation
+* [x] Traditional standalone server (unzipped or provisioned by Galleon)
+
+* [x] Managed domain
+
+* [x] OpenShift s2i
```
OndrejKotek, line 65: I think Webconsole is not supported for OpenShift and bootable jar.

bstansberry: WildFly supports the console in any context.

elytron/WFLY-14735-Management-FORM-authentication.adoc
```diff
@@ -0,0 +1,245 @@
...
+permutations needed for different situations.  To reduce the complexity a new resource `policy=security`
+will be added to the `core-management` subsystem, this will allow references to the sasl or http 
+authentication factories as well as to ssl-contexts to be used.  This new policy resouce will also be
+able to reference the new context made available from the `elytron-authentication` subsystem.
+
+==== Management Interfaces
+
+On the management interfaces a new attribute `use-core-management-security-policy` will be added,
+this will be a boolean and when set to `true` the configuration defined on the `policy=security`
+resource will be used instead of the individual attributes on the management interface.
```
bstansberry, line 146: This attribute design doesn't align well with how attributes work, i.e. x and y are alternatives and the kernel rejects setting both; it doesn't reject setting y if x is 'true'.

elytron/WFLY-14735-Management-FORM-authentication.adoc
darranl: I don't think I can default the attribute to true as we need to handle legacy configuration which does not set this so instead the default configuration will need to set it to true.

elytron/WFLY-14735-Management-FORM-authentication.adoc
darranl: +1 old configurations will continue to work as they always did so users can always set this to false and configure using authentication factory and ssl context references directly.

elytron/WFLY-14735-Management-FORM-authentication.adoc
bstansberry: Also, the server side aspect of this is generic AFAICT. (And if it's not that's a flaw in the requirements.) WildFly supports HTTP based management with the admin console as a particular client.

elytron/WFLY-14735-Management-FORM-authentication.adoc
darranl: We could avoid mutual exclusivity issues if we wanted by saying when set to true we use the items defined in policy=security unless also defined here so local references override central references.

elytron/WFLY-14735-Management-FORM-authentication.adoc
```diff
@@ -0,0 +1,245 @@
...
+that authentication has failed.
+
+If a user has previously been authenticated and their cookie has expired or the user is using the 
+admin console to connect to a different server it is possible the admin console may sent an invocation
+to the `/management` context without an appropriate cookie, in this case the admin console will need
+to be "notified to" and handle a redirect to the `/authentication` context.
+
+==== elytron-authentication
+
+A new subsystem `elytron-authentication` will be developed, the primary role of this subsystem
```
bstansberry, line 110: This name seems quite generic for a subsystem that's targeted at a pretty specific use.

bstansberry: Even if expanded to deployment use the name is kind of generic, and the elytron subsystem itself deals with authentication too.

All this said I know names are hard and I don't have any suggestions, so please just regard this as a generic 'even though you thought about the name 10 times, please think twice more' request.

elytron/WFLY-14735-Management-FORM-authentication.adoc
```diff
@@ -0,0 +1,245 @@
...
+
+Unless we also enable TLS for the management interfaces clients which connect to the native 
+interface either  directly or via HTTP Upgrade will be unaffected as this will still be using 
+SASL authentication which will be unchanged by this RFE.
+
+=== Importing Existing Configuration
+
+Existing configuration will be imported "as-is" and will continue to use the previously 
+defined behaviour, users who want to add form authentication to the HTTP management interface will 
+need to manually add it themselves.
```
bstansberry, line 202: The Hard Requirements section should have some discussion of what the behavior is in this case (or any case where the new behavior is not enabled.)

elytron/WFLY-14735-Management-FORM-authentication.adoc
darranl: +1 this authentication could be used by other clients, the general principal is a user can authenticate to set a cookie and that is sent on other requests so this could be used for other clients after the form login.

elytron/WFLY-14735-Management-FORM-authentication.adoc
darranl: Be careful I will get the thesaurus out - maybe elytron-form-authentication the problem is if we do get round to fully supporting Fido2 WebAuthn the user no longer interacts with the FORM.  Possibly elytron-http-authentication?

elytron/WFLY-14735-Management-FORM-authentication.adoc
OndrejKotek: Thanks, that's what I meant. I'm sorry for expressed myself wrong.

elytron/WFLY-14735-Management-FORM-authentication.adoc
```diff
@@ -0,0 +1,245 @@
...
+:icons:             font
+:idprefix:
+:idseparator:       -
+
+== Overview
+
+The development of this RFE is delivering a subset of functionality from a much larger planned change,
+this RFE will start with two features:
+ * Enabling FORM authentication for admin console access to the application server.
+ * Where custom security realms are in use supporting MFA authentication on the login form to
```
OndrejKotek, line 14: What is the plan/flow/configuration for this part?

elytron/WFLY-14735-Management-FORM-authentication.adoc
bstansberry: That can work if the details are consistent and will remain consistent.

elytron/WFLY-14735-Management-FORM-authentication.adoc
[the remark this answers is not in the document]
bstansberry: Great question!

bstansberry: I wrote this comment before properly digesting the future enhancements section in the 'Non Requirements'. Those don't eliminate the issue but do give a sense of this subsystem over time more clearly justifying the general purpose name.

I don't think elytron-http-authenication is better as when I pretend not to understand things I think "The elytron subsystem handles http authentication; I know because I see http a lot in its config!" So that's just a more verbose variant of the same problem.

darranl: One point to raise regarding subsystem naming is the overall naming scheme for elytron subsystems.  For better or for worse our main subsystem is called "elytron" with no further qualifiers, this subsystem is intended to be a verbose subsystem enabling an administrator to build their own policy from the building blocks we provide and through the use of capabilities potentially provided by others.

There is a new family of subsystems considered that have specialist behaviour such as "elytron-tls" and "elytron-authentication" - these are intended to cover a large percentage of possible requirements whilst minimising the complexity of the configuration.  So there is going to duplication, the elytron-* subsystems simplify specific use cases whilst the elytron subsystem provides the complete verbose set of options.

Continuing the naming for this new subsystem, another possibility could be something like "elytron-interactive", or "elytron-dynamic" to indicate it's purpose is for an interactive authentication process.  possibly "elytron-dynamic-form" to indicate the primary purpose is dynamic form authentication?

OndrejKotek: > There is a new family of subsystems considered that have specialist behaviour such as "elytron-tls" and "elytron-authentication" - these are intended to cover a large percentage of possible requirements whilst minimising the complexity of the configuration. So there is going to duplication, the elytron-* subsystems simplify specific use cases whilst the elytron subsystem provides the complete verbose set of options.

From the analysis for `elytron-tls` the subsystem would not seem to be a simplification as the complexity seems the same, just limited to TLS settings.

What will be the content of `elytron-authentication`? Just a set of something like `http-context` resource referencing `security-domain` and handling a list of dynamic attributes for extra FORM fields? Also there is no requirement to have such resources in the basic `elytron` subsystem.

Resources of `elytron-oidc-client` are also not included in the basic `elytron` subsystem.

Regarding the name, do we expect we will add another (HTTP based) authentication methods (than FORM) to the subsystem in future?

elytron/WFLY-14735-Management-FORM-authentication.adoc
```diff
@@ -0,0 +1,280 @@
...
+password in addition to the regular password.  A new evidence type will be defined within WildFly Elytron
+which will accept an arbitrary set of attributes during authentication.  The existing security realms
+within WildFly Elytron that handle username/password based authentication will be updated to support
+this new dynamic evidence type.
+
+==== core-management
+
+The number of attributes related to security on the management interfaces has grown with different
+permutations needed for different situations.  To reduce the complexity a new resource `policy=security`
+will be added to the `core-management` subsystem, this will allow references to the sasl or http
```
OndrejKotek, line 139: Is adding such resource really a reduction of complexity? One new attribute would be added to management interfaces in all the cases. Do we expect more attributes than the one referencing the `elytron-authentication` subsystem?

Having one more resource with an option to override the attributes defined there seems to be rather increasing the complexity.

elytron/WFLY-14735-Management-FORM-authentication.adoc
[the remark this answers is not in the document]
darranl: As a new subsystem I expect this will be a distinct module, regarding code written outside of WildFly I haven't looked exactly at where it will be added but if possible it should be separate from our existing Elytron modules so can be provisioned independently.

This is a new piece of functionality that is optional i.e. current behaviour can be preserved - for this reason I think it is justified to define a new layer which will include this new subsystem/extension and provide a default definition within the subsystem that depends on the ManagementDomain.  This will however not make it in use by default.

Maybe a follow on discussion for elsewhere, would layers be suitable granularity for some chunks of default security configuration - i.e. break it down further.  A layer for application SSL, one for management SSL, one for existing management domain config, one for new config.  Could that instead be a use for an external feature pack to define chunks of security config which can be selected?

elytron/WFLY-14735-Management-FORM-authentication.adoc
darranl: I think we can review this one, primarily the first part to implement is getting the /authentication context exposed - I still need to work through some planning for token issuance / validation and combining other authentication mechanisms into the mix which is where I was thinking policy=security would start to be the place we add more but we can try without to begin with.

elytron/WFLY-14735-Management-FORM-authentication.adoc
```diff
@@ -0,0 +1,280 @@
...
+
+On the HTTP management interface a new context will be made available at the path `/authentication`,
+when a user connects to the root `/` context of the application server they will be redirected to the
+`/authentication` context.  The `/authentication` context will display a login form for the user 
+to authenticate themselves.
+
+The look and feel of the login FORM will make use of `PatternFly` for consistent with the look and
+feel of the web console.
+
+Under the `/authentication` context there will be two sub contexts `/login` and `/logout`, after
```
darranl, line 84: @jamezp Would you be able to help with some guidance / recommendations here?

Overall this requirement is being delivered in a new subsystem which will be implementing a rest like endpoint. /logout is likely a very simple HTTP request which will result in a cookie being invalidated but /login is the slightly more complex one.

For /login this will receive an authentication request, I think very likely JSON, it will also respons using JSON for an error / additional information state.  The response will also need to be able to set a cookie for the web browser. 

This will not be activate by default to secure the management interface but that is where we are planning to go.  Should we be looking at using RestEasy for these requests or would you recommend something else?

elytron/WFLY-14735-Management-FORM-authentication.adoc
```diff
@@ -0,0 +1,280 @@
...
+to authenticate themselves.
+
+The look and feel of the login FORM will make use of `PatternFly` for consistent with the look and
+feel of the web console.
+
+Under the `/authentication` context there will be two sub contexts `/login` and `/logout`, after 
+a user has input their username and password these will be passed to the first of the sub contexts to
+perform the authentication.  Where authentication is successful a cryptographically signed JWT token 
+will be generated and passed back to the client as a cookie, the login form will then redirect the 
+user to the admin console.
```
darranl, line 88: I need to add here that there will be an additional context like /authentication/status

This will respond to GET requests and provide some information about if FORM based authentication is available.  The web console can then use this to decide if it needs to redirect to the /authentication context.

elytron/WFLY-14735-Management-FORM-authentication.adoc
```diff
@@ -0,0 +1,280 @@
...
+use of the existing `ManagementDomain` for authentication.
+
+The management interfaces however will not use FORM authentication out of the box.  The reason is 
+that when activating FORM authentication TLS should also be enabled for the management interfaces
+at the same time, this way we can document the steps and considerations to take when activating
+this feature.
+
+=== Nice-to-Have Requirements
+
+The CLI will have a new security command adding to activate FORM authentication for the HTTP management
```
darranl, line 182: @jfdenise FYI this RFE may have a nice to have requirement of another command in the CLI to talk about at some point.

elytron/WFLY-14735-Management-FORM-authentication.adoc
darranl: I have added a mention of this in the testing section as to how we will illustrate, do you need the sequence adding to the analysis?

elytron/WFLY-14735-Management-FORM-authentication.adoc
```diff
@@ -0,0 +1,280 @@
...
+in the https://github.com/wildfly/wildfly[WildFly] project.
+
+Changes will also be made within https://github.com/wildfly/wildfly-core[WildFly Core] to
+enable and support the activation of FORM authentication.
+
+Updates will be required to the https://github.com/hal/console[Admin Console] to support
+possibly being required to redirect to the login form and to make use of the logout endpoint.
+
+Security components developed for this RFE will be contained within the
+https://github.com/wildfly-security/wildfly-elytron[WildFly Elytron] project.
```
darranl, line 53: I may add Elytron Web here as that project already has dependencies on Undertow if we need to implement Undertow APIs.

elytron/WFLY-14735-Management-FORM-authentication.adoc
OndrejKotek: Then something like `elytron-dynamic-form`, `elytron-form`,  `elytron-authentication-form` could be more suitable than the general `elytron-authentication`.

I suppose in (now unexpected) case of adding another authn method in the future, we would place it into another new subsystem. Because of layers an so on.

darranl: I like elytron-dynamic-form

It does have a minimal dynamic aspect from the outset, this will then grow in the future.

elytron/WFLY-14735-Management-FORM-authentication.adoc
OndrejKotek: I've spotted the last paragraph of the _elytron-authentication_ section explains this to some extent. I would like to understand whether the form will be dynamically generated just according to the set of attributes defined in the subsystem, or whether we plan for user to be able to set a custom workflow. In case of the former one explaining sentence would be enough.

elytron/WFLY-14735-Management-FORM-authentication.adoc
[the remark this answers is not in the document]
OndrejKotek: I expect the module to be private as we don't need to expose any new API for things like custom realms.

Do I understand it well there is no mechanism to customize layers and we need a layer per specific case? If it's the case is there a recommendation for naming schemas?

elytron/WFLY-14735-Management-FORM-authentication.adoc
```diff
@@ -0,0 +1,280 @@
...
+
+* [x] Bootable jar
+
+== Requirements
+
+=== Hard Requirements
+
+==== General Flow
+
+On the HTTP management interface a new context will be made available at the path `/authentication`,
```
OndrejKotek, line 76: The context name could be the same as the subsystem name, for easier understanding/interlinking. The same is applied in case of `keycloak`/`elytron-oidc-client`, see https://github.com/wildfly/wildfly-proposals/pull/506/files.

elytron/WFLY-14735-Management-FORM-authentication.adoc
darranl: Ok will add a line, in this case it will be dynamically generated to display the list of configured attributes all in one go.

Later we would add more custom workflow ask for X if Y type things.

elytron/WFLY-14735-Management-FORM-authentication.adoc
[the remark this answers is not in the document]
darranl: +1 as an extension / subsystem it will be a private module.

For the layer under this RFE I think the name should be very close to the name of the subsystem.

elytron/WFLY-14735-Management-FORM-authentication.adoc
darranl: I almost agree but in this case the context is advertised by the management endpoint not the subsystem, i.e. the management endpoint will obtain a reference to a capability that can handle the HTTP requests but it is the management endpoint that choose where to register it.

OndrejKotek: OK, then it makes sense to keep it as it is. In (now unexpected) case of adding next authentication related subsystem the endpoint would remain the same and just the underlying implementation would change.

elytron/WFLY-14735-Management-FORM-authentication.adoc
jfdenise: @darranl , should we log another RFE or you plan to have it included in this RFE?

elytron/WFLY-14735-Management-FORM-authentication.adoc
darranl: +1 being a capability reference we can swap the impl

elytron/WFLY-14735-Management-FORM-authentication.adoc
darranl: @jfdenise I think for now as it in the nice to have sesction keep in this RFE - if we don't get to it then split it out to add later.  Otherwise I think it has more issues getting the RFE through.

elytron/WFLY-14735-Management-FORM-authentication.adoc
darranl: Feature team has agreed the name will be elytron-dynamic-form

darranl:
Going to close for now, will reopen once ready to revisit.

Now we have the WildFly feature development process I would like to approach starting with the general behaviour being supported and then follow up with how we can use it for management.

PR #514, opened 2023-02-15
WFLY-14869 Add support for MicroProfile LRA

xstefank:
https://issues.redhat.com/browse/WFLY-14869

### DIFF
--- a/transactions/WFLY-14869-microprofile-lra.adoc
+++ b/transactions/WFLY-14869-microprofile-lra.adoc
@@ -1,0 +1,123 @@
+= Support for MicroProfile LRA
+:author:            Martin Stefanko
+:email:             mstefank@redhat.com
+:toc:               left
+:icons:             font
+:idprefix:
+:idseparator:       -
+:keywords:          transaction, saga, lra, microprofile
+
+== Overview
+
+The goal of this RFE is providing the Galleon layers for WildFly feature pack which integrate the
+https://github.com/jbosstm/narayana/tree/master/rts/lra[Narayana implementation] of the
+https://github.com/eclipse/microprofile-lra[MicroProfile LRA specification].
+
+The MicroProfile LRA is a transaction processing pattern well suited for microservice development based on the saga
+pattern. The motivation of such processing is summarized in the
+https://github.com/eclipse/microprofile-lra/blob/1.0/README.adoc#motivation[MicroProfile Spec readme].
+
+
+== Issue Metadata
+
+=== Issue
+
+* https://issues.redhat.com/browse/WFLY-14869
+
+=== Related Issues
+
+* https://issues.redhat.com/browse/EAP7-1496
+
+=== Dev Contacts
+
+* mailto:mstefank@redhat.com[{author}]
+
+=== QE Contacts
+
+* mailto:fburzigo@redhat.com[Fabio Burzigotti]
+
+=== Testing By
+
+* [x] Engineering
+* [ ] QE
+
+
+=== Affected Projects or Components
+
+* Transactions
+* WildFly
+
+=== Other Interested Projects
+
+* MicroProfile
+* Galleon
+* WildFly Operator
+
+=== Relevant Installation Types
+
+* [x] Traditional standalone server (unzipped or provisioned by Galleon)
+* [x] Managed domain
+* [ ] OpenShift s2i
+* [x] Bootable jar
+
+== Requirements
+
+=== Hard Requirements
+
+The feature will integrate the LRA specification into the WildFly feature pack as two new extensions (one for LRA coordinator and one for LRA participant) together with two new subsystems and Galleon layers accordingly.
+The Galleon layers will be available for provisioning.
+The provisioned layers will provide all functionality defined in LRA specification for users in WildFly.
+
+The integrations will consist of two separate subsystems - microprofile-lra-coordinator and microprofile-lra-participant) and two separate extensions - org.wildfly.extension.microprofile.lra-coordinator and org.wildfly.extension.microprofile.lra-participant.
+
+The first layer provides a capability of the LRA participant - bundling the Narayana client library capable
+to parse the annotations from LRA specification and work with JAX-RS and non-JAX-RS bindings.
+
+The second layer provides the capability of the LRA coordinator - a server which provides HTTP endpoints where
+applications may register as participants which the coordinator then calls back (via HTTP calls) when LRA finishes.
+
+The feature pack implementation will be placed amongst other MicroProfile WildFly subsystems under
+https://github.com/wildfly/wildfly/tree/main/microprofile[the WildFly GitHub repository].
+
+Both subsystems are optional to allow for exclusion during provisioning.
+
+The LRA layers will be dependent on the `transactions` subsystem. The LRA uses the same object store types
+as the JTA/JTS transaction uses, it requires the recovery manager thread to be running and requires transaction
+specifics like unique node-identifier per running application server (WildFly).
+
+As said before the LRA coordinator is based on the Narayana core (`transactions` subsystem within WildFly). The LRA records and JTA records share the same storage and recovery processing.
+
+=== Nice-to-Have Requirements
+
+Create a quickstart on LRA on how to use LRA with WildFly for https://github.com/wildfly/quickstart.
+
+LRA records could be listed under `/subsystem=transactions/log-store=log-store/lras=...`
+and the operations (e.g. a need for probing) will be consistent to what was developed for JTA transaction types
+under `/subsystem=transactions/log-store=log-store/transactions=...`. This is optional, because the LRA Coordinator REST interface already provides this information.
+
+=== Non-Requirements
+
+NONE
+
+== Test Plan
+
+1. Testing LRA functionality within WildFly including TCK and subsystem setup.
++
+The LRA functionality tests are part of the Narayana LRA codebase (https://github.com/jbosstm/narayana/tree/master/rts/lra/test/. Namely basic and crash tests in addition to TCK.
++
+2. The WildFly will be running with LRA TCK module.
++
+https://github.com/wildfly/wildfly/tree/main/testsuite/integration/microprofile-tck
++
+3. Basic integration tests which tests the Galleon layer provisioning.
++
+https://github.com/wildfly/wildfly/tree/main/testsuite/integration/microprofile
+
+
+== Community Documentation
+
+Description of the LRA configuration to be added in the documentation as part of the implementation PR.
+
+== Release Note Content
+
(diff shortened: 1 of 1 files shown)


transactions/WFLY-14869-microprofile-lra.adoc
```diff
@@ -0,0 +1,128 @@
...
+
+== Issue Metadata
+
+=== Issue
+
+* https://issues.redhat.com/browse/WFLY-14869
+
+=== Related Issues
+
+* https://issues.redhat.com/browse/EAP7-1496
```
fabiobrz, line 29: I am not sure it makes sense to have a product tracker here in the community proposal, since it is not public. WDYT?

transactions/WFLY-14869-microprofile-lra.adoc
```diff
@@ -0,0 +1,128 @@
...
+* [x] Traditional standalone server (unzipped or provisioned by Galleon)
+* [ ] Managed domain
+* [ ] OpenShift s2i
+* [x] Bootable jar
+
+== Requirements
+
+=== Hard Requirements
+
+The feature pack will integrate the LRA specification as the Galleon feature pack.
```
fabiobrz, line 67: AFAICT the feature pack will actually integrate the Narayana implementation of the LRA spec into a new Galleon feature pack, right?

transactions/WFLY-14869-microprofile-lra.adoc
```diff
@@ -0,0 +1,128 @@
...
+The Galleon layer will be available for provisioning.
```
fabiobrz, line 68: Maybe s/layer/feature pack/ ?

Since below we state that the feature pack will make two layers available. WDYT?

transactions/WFLY-14869-microprofile-lra.adoc
```diff
@@ -0,0 +1,128 @@
...
+
+The feature pack will be split to two Galleon layers (i.e., two separate subsystems).
+
+The first layer provides a capability of the LRA participant - bundling the Narayana client library capable
+to parse the annotations from LRA specification and work with JAX-RS and non-JAX-RS bindings.
+
+The second layer provides the capability of the LRA coordinator - a server which provides HTTP endpoints where applications may register as participants and then the coordinator
+calls them back (via HTTP calls) when LRA finishes.
+
+The feature pack will be placed amongst other MicroProfile WildFly feature packs
```
fabiobrz, line 79: s/feature pack/feature pack implementation/ maybe?

transactions/WFLY-14869-microprofile-lra.adoc
```diff
@@ -0,0 +1,128 @@
...
+under https://github.com/wildfly/wildfly/tree/main/microprofile[the WildFly GitHub repository].
+
+Each MicroProfile spec is optional to allow for exclusion during provisioning.
+
+The LRA feature packs will be dependent on the `transactions` subsystem. The LRA uses the same object store types
+as the JTA/JTS transaction uses, it requires the recovery manager thread to be running and requires transaction
+specifics like unique node-identifier per running application server (WildFly).
+
+As said before the LRA coordinator is based on the Narayana core (`transactions` subsystem within WildFly)
+the LRA records and JTA records shares the same storage and recovery processing.
```
fabiobrz, line 89: s/shares/share/ ?

transactions/WFLY-14869-microprofile-lra.adoc
```diff
@@ -0,0 +1,128 @@
...
+
+== Requirements
+
+=== Hard Requirements
+
+The feature pack will integrate the LRA specification as the Galleon feature pack.
+The Galleon layer will be available for provisioning.
+The provisioned feature pack will provide all functionality defined in LRA specification for users in WildFly.
+
+The feature pack will be split to two Galleon layers (i.e., two separate subsystems).
```
fabiobrz, line 71: Can we add the new subsystems' names?

transactions/WFLY-14869-microprofile-lra.adoc
```diff
@@ -0,0 +1,128 @@
...
++
+2. The WildFly will be running with LRA TCK module
++
+https://github.com/wildfly/wildfly/tree/main/testsuite/integration/microprofile-tck
++
+3. Basic integration tests which tests the Galleon layer provisioning
++
+https://github.com/wildfly/wildfly/tree/main/testsuite/integration/microprofile
++
+4. OpenShift testing
```
fabiobrz, line 117: I am thinking whether we should remove OpenShift from this community proposal and add such different platform test details to the product feature test plan. WDYT?

fabiobrz (changes requested):
Thanks for this proposal @xstefank - I just dropped some minor comments.
Feel free to let me know what you think.

transactions/WFLY-14869-microprofile-lra.adoc
xstefank: It seems this is aligned with other proposals. This is also for our reference.

transactions/WFLY-14869-microprofile-lra.adoc
fabiobrz: Maybe I didn't make my point. What hit my eye is that the feature pack should be integrating the implementation, rather than the specs. Or at least both, IIUC. But I might be missing something. WDYT?

transactions/WFLY-14869-microprofile-lra.adoc
fabiobrz: Right then, let's leave it there :slightly_smiling_face:

transactions/WFLY-14869-microprofile-lra.adoc
```diff
@@ -0,0 +1,123 @@
...
+:email:             mstefank@redhat.com
+:toc:               left
+:icons:             font
+:idprefix:
+:idseparator:       -
+:keywords:          transaction, saga, lra, microprofile
+
+== Overview
+
+The goal of this RFE is providing the Galleon feature pack for WildFly which integrates the
```
bstansberry, line 12: @xstefank I don't think this RFE is about providing a feature pack. It is about integrating the LRA functionality into the existing 'wildfly' feature pack, by providing two new extensions along with Galleon layers.

There are a lot of wording changes below that are needed to deal with removing the notion that there are new feature packs here.

BTW I think this is just about analysis doc wording. The https://github.com/wildfly/wildfly/pull/16547 PR is what I expect; it's about adding functionality to the existing 'wildfly' feature pack.

transactions/WFLY-14869-microprofile-lra.adoc
bstansberry: 👍

transactions/WFLY-14869-microprofile-lra.adoc
```diff
@@ -0,0 +1,123 @@
...
+=== Other Interested Projects
+
+* MicroProfile
+* Galleon
+* WildFly Operator
+
+=== Relevant Installation Types
+
+* [x] Traditional standalone server (unzipped or provisioned by Galleon)
+* [ ] Managed domain
```
bstansberry, line 59: Why not managed domain?

transactions/WFLY-14869-microprofile-lra.adoc
bstansberry: I don't think there is any new feature pack here. If we're talking a new feature pack, this would stay in WildFly Extras.  This needs to be about adding capabilities to the existing 'wildfly' feature pack.

transactions/WFLY-14869-microprofile-lra.adoc
```diff
@@ -0,0 +1,123 @@
...
+
+== Requirements
+
+=== Hard Requirements
+
+The feature pack will integrate the LRA specification as the Galleon feature pack.
+The Galleon feature pack will be available for provisioning.
+The provisioned feature pack will provide all functionality defined in LRA specification for users in WildFly.
+
+The feature pack will be split to two Galleon layers (i.e., two separate subsystems - microprofile-lra-coordinator and microprofile-lra-participant).
```
bstansberry, line 71: Is not wanting both a common use case? Is there a big downside to including both when you only need one?

IOW should we have just one layer? Should we have three, with the 3rd being 'microprofile-lra' that depends on the other two?

Having one still allows us to respond to user demand by moving to 3 in the future. Starting with 2 also lets us respond to user demand by moving to 3 in the future, but it also commits us to having 3.

transactions/WFLY-14869-microprofile-lra.adoc
```diff
@@ -0,0 +1,123 @@
...
+https://github.com/wildfly/wildfly/tree/main/testsuite/integration/microprofile
+
+
+== Community Documentation
+
+Description of the LRA configuration to be added in the documentation as part of the implementation PR.
+
+== Release Note Content
+
+WildFly introduces a new Galleon layer `microprofile-lra` for provisioning. It provides the MicroProfile LRA specification into WildFly.
```
bstansberry, line 123: This says there's one layer. :)

bstansberry (changes requested):

transactions/WFLY-14869-microprofile-lra.adoc
xstefank: Most common use case would be only the participant. Coordinator is a REST endpoint that can be started separately. Basically similar to message broker. About putting them together I don't know. I will need to see how users will use them. We can always introduce thrid `microprofile-lra` if needed later.

fabiobrz (approved):
Thanks @xstefank (and thanks @bstansberry for your input on the feature pack topic).
The proposal LGTM, so I am approving it.

transactions/WFLY-14869-microprofile-lra.adoc
```diff
@@ -0,0 +1,123 @@
...
+
+The LRA layers will be dependent on the `transactions` subsystem. The LRA uses the same object store types
+as the JTA/JTS transaction uses, it requires the recovery manager thread to be running and requires transaction
+specifics like unique node-identifier per running application server (WildFly).
+
+As said before the LRA coordinator is based on the Narayana core (`transactions` subsystem within WildFly). The LRA records and JTA records share the same storage and recovery processing.
+
+=== Nice-to-Have Requirements
+
+Create a quickstart on LRA on how to use LRA with WildFly for https://github.com/wildfly/quickstart.
```
marcosgopen, line 92: Has this quickstart already been created? Do you have the Jira link to share @xstefank ?

xstefank: no, still on my TODO list

marcosgopen: No problem, please consider opening the Jira issue, I would be glad to help you with that quickstart!

marcosgopen (approved):

PR #521, opened 2023-04-18
[HAL-1827]: Update the server installation using the Web Console

hpehl:
https://issues.redhat.com/browse/HAL-1827


hpehl:
@bstansberry I've finished the first version of the analysis document to integrate the new update features in the console. Please have a look and let me know if you have any questions. I am looking forward to your feedback.

jedla97 (commented):
Except the change of the naming I don't see anything unclear. I like the picture for better visualisation. In my point of view if you don't have these picture save for editing it's OK to leave them with that old naming and just add to new names bracket with old name. 

For note I'll read this definitely at least once after the analysis for EAP7-1917 will be approved and finished. To make sure that there is something which is not align with it. At the moment I didn't found any problem.

console/HAL-1827_Update_server.adoc
```diff
@@ -0,0 +1,213 @@
...
+
+* Promote a Domain Controller to be an artifacts server
+
+== Implementation Plan
+
+This section only describes the new update UI. The old patching UI remains unchanged and is out of the scope of this document.
+
+=== Homepage
+
+In domain mode the homepage has a section about patching. This section is renamed to "Update", and the content is updated to reflect the basic steps for applying an update. This section is now also available in standalone mode.
```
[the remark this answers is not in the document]
hpehl: @jedla97 I've adjusted the naming in the AD and images to reflect the current implementation.

console/HAL-1827_Update_manager.adoc
```diff
@@ -0,0 +1,215 @@
...
+.Updates column
+image::HAL-1827_updates_column.png[]
+
+==== Updates Item
+
+An update item represents one model node of the `result` array returned by `/core-service=installer:history()`. The console uses the following information to render the item:
+
+* Hash: The revision hash
+* Timestamp: The timestamp of the commit
+* Type: The type of the commit: `rollback`, `update` or `install`
```
jedla97, line 150: In analysis of EAP7-1917 there is a 4th type of commit `config_change`. Used when channel is changed. So there should be mention here as well.

console/HAL-1827_Update_manager.adoc
```diff
@@ -0,0 +1,215 @@
...
+
+The item provides the following actions ❶:
+
+* Revert: Revert the update or the rollback. Only available if the type is `update`.
+
+The preview shows the following information:
+
+* Hash: The revision hash
+* Timestamp: The timestamp of the commit
+* Type: The type of the commit: `rollback`, `update` or `install`
```
jedla97, line 161: Same as above with `config_change`

console/HAL-1827_Update_manager.adoc
```diff
@@ -0,0 +1,215 @@
...
+
+==== Channel Item
+
+The console uses the following information to render a subscribed channel:
+
+* Name of the channel
+* Type of the manifest: GAV or URL
+* An icon representing the type of the manifest
+
+The item provides the following actions ❶:
```
jedla97, line 191: Here should be mentioned the update of channel.

jedla97 (changes requested):
I go trough this one again and mention some missing thing.

console/HAL-1827_Update_manager.adoc
hpehl: Thanks for the feedback, I missed that.

PR #532, opened 2023-08-08
[WFLY-17143] Preview - Add the ability to specify that the OIDC Authentication Request should include request and request_uri parameters

PrarthonaPaul:
Issue: https://issues.redhat.com/browse/WFLY-17143
related issue: https://issues.redhat.com/browse/ELY-2584
Relevant github discussions: 

- https://github.com/keycloak/keycloak/discussions/23150
- https://github.com/keycloak/keycloak/discussions/22729
- https://github.com/keycloak/keycloak/discussions/24514

### DIFF
--- a/elytron/WFLY-17143-request-uri-parameters.adoc
+++ b/elytron/WFLY-17143-request-uri-parameters.adoc
@@ -1,0 +1,246 @@
+= [Preview] Add the ability to specify that the OIDC Authentication Request should include request and request_uri parameters
+:author:            Prarthona Paul
+:email:             prpaul@redhat.com
+:toc:               left
+:icons:             font
+:idprefix:
+:idseparator:       -
+
+== Overview
+
+OpenID Connect is an authentication mechanism that builds on OAuth 2.0 and allows a user to login to a web application using credentials established by an OpenID provider. Authentication requests sent using OpenID Connect can be signed and optionally encrypted by encoding query parameters in a JWT. This can be achieved using one of two auth request parameters: `request` and `request_uri`, which are both optional. The current version of `elytron-oidc-client` does not allow the user to specify `request` or `request_uri`. Instead, all requests are sent in using the OAuth 2.0 request syntax. 
+
+This RFE will add an attribute to the `elytron-oidc-client` subsystem called `authentication-request-format`, which can allow the user to specify if they will use either of the request parameters. Additionally, we will be adding supporting attributes called `request-object-signing-algorithm`, `request-object-encryption-alg-value` and `request-object-encryption-enc-value`, which will be used to specify if and how the request parameters will be signed and/or encrypted. All of these attributes will be added to the `secure-server`, `secure-deployment`, `provider` and `realm` resources, which the user can configure using either the `elytron-oidc-client` subsystem, or using a json file in deployment configuration.
+
+We will also be adding some keystore attributes, such as, `request-object-signing-keystore-file`, `request-object-signing-keystore-password`, `request-object-signing-key-password`, `request-object-signing-key-alias` and `request-object-signing-keystore-type` to the resources mentioned above. There are similar attributes that are currently available under the `credential` attribute. However, when they are configured outside of the `credentials` attribute, they will be used to sign Request Objects.
+
+According to the https://openid.net/specs/openid-connect-core-1_0.html#JWTRequests[OpenID docs], requests using `request` and `request_uri` parameters are represented as JWTs, which are respectively passed by value or by reference. Depending on the OpenID Provider, support for these parameters may or may not be available. 
+
+== Issue Metadata
+
+=== Issue
+
+* https://issues.redhat.com/browse/WFLY-17143[WFLY-17143]
+
+=== Related Issues
+
+* https://issues.redhat.com/browse/ELY-2584[ELY-2584]
+
+=== Stability Level
+// Choose the planned stability level for the proposed functionality
+* [ ] Experimental
+
+* [x] Preview
+
+* [ ] Community
+
+* [ ] default
+
+=== Dev Contacts
+
+* mailto:{email}[{author}]
+
+=== QE Contacts
+
+=== Testing By
+* [x] Engineering
+
+* [ ] QE
+
+=== Affected Projects or Components
+
+* WildFly 
+* WildFly Elytron
+
+=== Other Interested Projects
+
+N/A
+
+=== Relevant Installation Types
+* [x] Traditional standalone server (unzipped or provisioned by Galleon)
+
+* [x] Managed domain
+
+* [x] OpenShift s2i
+
+* [x] Bootable jar
+
+== Requirements
+
+=== Hard Requirements
+
+* Some new attributes named `authentication-request-format`, `request-object-signing-algorithm`, `request-object-encryption-alg-value` and `request-object-encryption-enc-value` will be added to the `secure-deployment`, `secure-server`, `realm` and `provider` resources under the `elytron-oidc-client` subsystem. The value for the `authentication-request-format` attribute will be a String with 3 acceptable values: 
+** oauth2: Indicating that the query parameters will be sent as usual following the oauth2 syntax. 
+** request: Indicating that the query will be sent as a JWT by value. 
+** request_uri: Indicating that the query will be sent as a reference to the JWT. 
+
+* The `authentication-request-format` attribute will specify if either `request` or the `request_uri` should be added to the auth request. This will be an optional attribute. The default value for this attribute will be set to `oauth2`, indicating that the request parameters will be sent regularly following the OAuth2 syntax.
+
+** For multiple `secure-server` or `secure-deployment` resources using the same `realm` or `provider`, the `authentication-request-format` can be specified at the higher level resource, such as, `provider` or `realm`. 
+
+** For multiple `secure-server` or `secure-deployment` resources using the same higher level resource, if the Request Object type is not the same for each deployment, then the `authentication-request-format` attribute can be set accordingly for the individual lower level resources. 
+
+* According to the https://openid.net/specs/openid-connect-core-1_0.html#JWTRequests[OIDC specs], only one of the two parameters can be specified at a time. If `request` is added, then `request_uri` MUST NOT be used in the same request. However, the user can choose to specify neither by setting `authentication-request-format` to `oauth2` or not setting it at all. 
+
+* The `request` and `request_uri` parameters MUST NOT be included in Request Objects.
+
+* The Request Object MAY be signed or unsigned (plaintext) and can be specified by value or by reference. The user can specify the algorithm used to sign the JWT request using the `request-object-signing-algorithm` attribute which will also be added to the resources specified above. This algorithm must be one of the Request Object signature algorithms supported by the OpenID provider. 
+
+** The values for `request-object-signing-algorithm` attribute are of type `String`. Default value for this attribute would be `none`, specifying that the request would be sent as a plaintext. In order to sign the jwt using an algorithm other than `none`, the user must specify the KeyPair used. This can be done using the `request-object-signing-keystore-file`, `request-object-signing-keystore-password`, `request-object-signing-key-password`, `request-object-signing-key-alias` and `request-object-signing-keystore-type` attributes. The algorithm for the KeyPair must match the algorithm used to sign the JWT. The keystore should be of type `JKS` or type `PKCS12` and can be generated using `keytool` on the linux terminal. 
+
+** The certificate for the KeyPair used to sign the JWT must be shared with the OpenID provider either by importing it from the keystore, or by uploading a jwksUrl. 
+
+** When configuring the client in deployment or subsystem settings, the `public-client` attribute must be set to `false` and client credentials must be specified as well using the `credentials` attribute. Client keys must be uploaded to the admin console of the OpenID Provider.
+
+* The request object may also be encrypted. To specify that the Request Object will be encrypted, the user needs to specify the relevant algorithms using `request-object-encryption-alg-value` and `request-object-encryption-enc-value`. Encryption is done using https://openid.net/specs/draft-jones-json-web-encryption-02.html#[JWE]. 
+
+** The values for the `request-object-encryption-alg-value` and the `request-object-encryption-enc-value` attributes are of type `String`. The default value for the attributes are undefined, which specifies that the request object will not be encrypted. 
+ 
+** The request object will only be encrypted if both `request-object-encryption-alg-value` and `request-object-encryption-enc-value` are specified. The values for these attributes must be part of the encryption algorithms and content encryption methods supported by the OpenID provider. These values can also be found by sending a GET request to the discovery url or by looking at the client configurations on the admin console of the OpenID provider. An example of what the GET request would return can be seen below (diaplaying attributes relevant to Request Objects only): 
+```
+...
+"request_object_signing_alg_values_supported":["PS384","ES384","RS384","HS256","HS512","ES256","RS256","HS384","ES512","PS256","PS512","RS512","none"],"request_object_encryption_alg_values_supported":["RSA-OAEP","RSA-OAEP-256","RSA1_5"],"request_object_encryption_enc_values_supported":["A256GCM","A192GCM","A128GCM","A128CBC-HS256","A192CBC-HS384","A256CBC-HS512"],
+...
+"request_parameter_supported":true,"request_uri_parameter_supported":true,"require_request_uri_registration":true,
+...
+```
+
+** The JWT must be signed first and then encrypted to make a nested JWT and the key used to encrypt the JWT must be the realm public key shared by the OpenID provider. The key algorithm must be the same as the encrypting algorithm. Signing and encrypting algorithms can be of different types (i. e. ES-256 for signing and RSA-OAEP for encrypting is allowed). The realm key for encrypting the JWT can be obtained by sending a GET request to the `jwksUri` obtained from the `discoveryUrl`(http://localhost:8080/realms/myrealm/.well-known/openid-configuration). For more information about how Keycloak supports Request Object Encryption, please refer to https://issues.redhat.com/browse/KEYCLOAK-18630[this issue] and the following changes made to the Keycloak source code: 
+
+*** https://github.com/keycloak/keycloak/pull/8243[Request object encryption support]
+*** https://github.com/keycloak/keycloak/pull/8261[Configurable constraints for request object encryption]
+
+** An example of the attributes obtained from the `jwksUri` of an OpenID provider can be seen below. Please note that the actual values for some of the fields have been truncated for readability: 
+```
+{"keys":[
+    {"kid":"I0ogHMugKrowsGS5StX8Ihx1qKSaH3QLVHl6FsVSVxE",
+    "kty":"RSA",
+    "alg":"RSA-OAEP",
+    "use":"enc",
+    "n":"vAHUwV28fDBXQpo390nm...",
+    "e":"AQAB",
+    "x5c":["MIICnTCCAYUCBgG..."],
+    "x5t":"EE4X1HpptqoAkDqMokwRwqliJFY","x5t#S256":"bfhIzItRxe9PKE7uPUJMpP4A-AozTWxMVPNwbzUhs2Y"},
+    
+    {"kid":"TSLvxRZEuxTmFgYdrs3ve7-0-DghC-ChN724HB2fGOc",
+    "kty":"RSA",
+    "alg":"RS256",
+    "use":"sig","n":"jL35-5e67PukEk_...",
+    "e":"AQAB",
+    "x5c":["MIICnTCCAYUCBgGL..."],
+    "x5t":"hNGOdZ0Nszq4Oqj1TB-usasLP1I","x5t#S256":"KNsx334iBpN5oPEEYSCf1Q-TlmG69AOUghRRFa9VuWY"}
+]}
+```
+* The authentication request can send the Request Object in one of two ways: 
+
+** By value: The `request` parameter is added to the query and its value is the Request Object itself. 
+
+** The request string can be added to the auth request by adding the whole string: 
+
+```
+    http://localhost:8080/realms/myrealm/protocol/openid-connect/auth?
+    response_type=code&
+    client_id=wildfly&
+    scope=openid&
+    redirect_uri=http%3A%2F%2Flocalhost%3A8090%2Fsimple-webapp-oidc%2Fsecured&
+    request=eyJhbGciOiJSU0EtT0FFUCI...
+```
+** By reference: This is useful when the `request` string is too large and a number of other https://openid.net/specs/openid-connect-core-1_0.html#RequestUriRationale[reasons]. In this case, the Request Object is passed as a reference parameter string using the `request_uri` parameter. The url references the Request Object, which the OpenID provider can use to download the Request Object itself. The entire Request_URI MUST NOT exceed 512 ASCII characters. 
+
+** The `request_uri` value can be obtained by sending POST request to the `pushed_authentication_request_endpoint` of the OpenID provider. This can be also obtained using the discoveryUrl. The content for the POST request contains the value of the signed and/or encrypted Request Object and the content type for this request is "application/x-www-form-urlencoded". 
+
+** Please see the following resources for more information on Keycloak's implemetation of PAR. 
+
+*** https://issues.redhat.com/browse/KEYCLOAK-18353[Jira Issue]
+*** https://github.com/keycloak/keycloak/pull/8144[Implementation in source code]
+*** https://github.com/keycloak/keycloak-community/blob/main/design/pushed-authorization-requests.md[Documentation]
+
+** A successful request will return 201 http code along with a JWT, which contains the value of the `request_uri` and its expiry time.  
+
+** When included in the auth request, the request_uri would be specified as: 
+    
+```
+    http://localhost:8080/realms/myrealm/protocol/openid-connect/auth?
+    response_type=code&
+    client_id=wildfly&
+    scope=openid&
+    request_uri=urn%3Aietf%3Aparams%3Aoauth%3Arequest_uri%3Ad11a6442-d9ca-4db8-a885-f9f43673f859&
+    redirect_uri=http%3A%2F%2Flocalhost%3A8090%2Fsimple-webapp-oidc%2Fsecured
+```
+
+** The base64-encoded SHA-256 hash `23GkurKxf5T0Y-mnPFCHqWOMiZi4VS138cQO_V7PZHAdM` will allow the provider to cache the token for future use. It is a hash of the contents of the Request Object.
+
+* According to the OpenID docs, if the same parameter exists both in the Request Object and the OAuth Authorization Request parameters, the parameter in the Request Object is used.
+
+* The feature must deal with cases where the OpenId Provider does not support request parameters by recognizing `request_not_supported` error and dealing with it accordingly.
+
+* It should be possible to specify that these parameters should be included in the Authentication Request via deployment configuration using the `oidc.json` file inside the `WEB-INF` directory of the web application and `elytron-oidc-client` subsystem configuration.
+
+The `request` attribute is specified through the deployment configuration as follows: 
+```
+{
+    "client-id" : "myclient",
+    "provider-url" : "${env.OIDC_PROVIDER_URL:http://localhost:8080}/realms/myrealm",
+    "public-client" : "false",
+    "authentication-request-format" : "request",
+    "principal-attribute" : "preferred_username",
+    "ssl-required" : "EXTERNAL"
+    "credentials" : {
+        "secret" : "CLIENT_SECRET"
+    }
+}
+``` 
+The `request` attribute is specified through the `elytron-oidc-client` subsystem as follows: 
+```
+    /subsystem=elytron-oidc-client/secure-deployment:add(client_id=myclient, provider-url="http://localhost:8090/", authentication-request-format="request")
+```
+
+The `request_uri` attribute is specified through the deployment configuration as follows: 
+```
+{
+    "client-id" : "myclient",
+    "provider-url" : "${env.OIDC_PROVIDER_URL:http://localhost:8080}/realms/myrealm",
+    "public-client" : "false",
+    "authentication-request-format" : "request_uri",
+    "principal-attribute" : "preferred_username",
+    "ssl-required" : "EXTERNAL"
+    "credentials" : {
+        "secret" : "CLIENT_SECRET"
+    }
+}
+```
+The `request` attribute is specified through the `elytron-oidc-client` subsystem as follows: 
+```
+    /subsystem=elytron-oidc-client/secure-deployment:add(client_id=myclient, provider-url="http://localhost:8090/", authentication-request-format="request_uri")
+``` 
+
+=== Nice-to-Have Requirements
+
+N/A
+
+=== Non-Requirements
+
+N/A
+
+== Backwards Compatibility
+
+N/A
+
+=== Default Configuration
+
+By default, neither `request` or `request_uri` would be specified and the value for `authentication-request-format` would be `oauth2`. 
+
+== Test Plan
+
+* WildFly Elytron Tests: Integration test cases implemented for functionality. 
+* WildFly Testsuite: Test cases will be added to check for subsystem parsing. 
+** Additional integration tests will be added to test the full functionality of the `elytron-oidc-subsystem` when `request` or `request_uri` are configured. 
+** Tests will be performed using signed (using symmetric and asymmetric keys), unsigned, encrypted and plaintext JWT requests, with the request specified both by value and by reference. 
+* Tests will be added for both subsystem and deployment configurations. 
+* Tests may be added to ensure that server configuration fails when the stability level is not specified appropriately. 
+
+== Community Documentation
+Documentation for the new `request` and `request_uri` attributes will be added to https://github.com/wildfly/wildfly/blob/main/docs/src/main/asciidoc/_admin-guide/subsystem-configuration/Elytron_OIDC_Client.adoc[Elytron OpenID Connect Client Subsystem Configuration].
+
+== Release Note Content
+This feature allows authentication requests to be sent as a signed and encrypted request object using request and request_uris. 


elytron/WFLY-17143-request-uri-parameters.adoc
```diff
@@ -0,0 +1,138 @@
...
+:author:            Prarthona Paul
+:email:             prpaul@redhat.com
+:toc:               left
+:icons:             font
+:idprefix:
+:idseparator:       -
+
+== Overview
+
+OpenID Connect is an authentication mechanism that builds on OAuth 2.0 and allows a user to login to a web application using credentials established by an OpenID provider. Authentication requests sent using OpenID Connect can be signed and optionally encrypted. This can be achieved using 2 auth request parameters: `request` and `request_uri`, which are both optional.
```
fjuma, line 11: We should mention in the overview that the `elytron-oidc-client` subsystem doesn't currently include the ability to specify the `request` or `request_uri` in the authentication request.

elytron/WFLY-17143-request-uri-parameters.adoc
```diff
@@ -0,0 +1,138 @@
...
+== Issue Metadata
+
+=== Issue
+
+* https://issues.redhat.com/browse/WFLY-17143[WFLY-17143]
+
+=== Related Issues
+
+* https://issues.redhat.com/browse/EAP7-1974[EAP7-1974]
+* [INSERT ELY ISSUE]
```
fjuma, line 24: Feel free to create the corresponding ELY issue.

elytron/WFLY-17143-request-uri-parameters.adoc
```diff
@@ -0,0 +1,138 @@
...
+
+* [x] OpenShift s2i
+
+* [x] Bootable jar
+
+== Requirements
+
+=== Hard Requirements
+
+* Two new attributes named `request` and `request-uri` will be added to the `secure-deployment` resource under the `elytron-oidc-client` subsystem, which will be used to specify either the `request` or the `request_uri`, which are both JWTs.
```
fjuma, line 59: What will be the type of the new attributes? (i.e., what will their allowed values be, what will they default to, etc.)

elytron/WFLY-17143-request-uri-parameters.adoc
```diff
@@ -0,0 +1,138 @@
...
+
+* [x] Bootable jar
+
+== Requirements
+
+=== Hard Requirements
+
+* Two new attributes named `request` and `request-uri` will be added to the `secure-deployment` resource under the `elytron-oidc-client` subsystem, which will be used to specify either the `request` or the `request_uri`, which are both JWTs.  
+
+* According to the https://issues.redhat.com/browse/EAP7-1974[OIDC specs], only one of the two parameters can be specified at a time. If `request` is added, then `request_uri` MUST NOT be used in the same request. However, the user can choose to specify neither.
```
fjuma, line 61: I think this is the wrong link here.

elytron/WFLY-17143-request-uri-parameters.adoc
```diff
@@ -0,0 +1,138 @@
...
+    %23GkurKxf5T0Y-mnPFCHqWOMiZi4VS138cQO_V7PZHAdM
+    &state=af0ifjsldkj&nonce=n-0S6_WzA2Mj
+    &scope=openid
+```
+
+* According to the OpenID docs, if the same parameter exists both in the Request Object and the OAuth Authorization Request parameters, the parameter in the Request Object is used.
+
+* The feature must deal with cases where the OpenId Provider does not support request parameters by recognizing `request_not_supported` error and dealing with it accordingly.
+
+* It should be possible to specify that these parameters should be included in the Authentication Request via deployment configuration using the `oidc.json` file inside the `WEB-INF` directory of the web application and `elytron-oidc-client` subsystem configuration.
```
fjuma, line 112: What will the new attributes in oidc.json look like?

elytron/WFLY-17143-request-uri-parameters.adoc
```diff
@@ -0,0 +1,138 @@
...
+
+== Test Plan
+
+* WildFly Elytron Tests: Test cases implemented for functionality. 
+* WildFly Testsuite: Test cases will be added to check for subsystem parsing. 
+    * Additional integration tests will be added to test the full functionality of the `elytron-oidc-subsystem` when `request` and `request_uri` are configured. 
+    * 
+
+== Community Documentation
+Documentation for the new scope option will be added to https://github.com/wildfly/wildfly/blob/main/docs/src/main/asciidoc/_admin-guide/subsystem-configuration/Elytron_OIDC_Client.adoc[Elytron OpenID Connect Client Subsystem Configuration].
```
fjuma, line 138: This needs to be updated.

fjuma (changes requested):
Thanks @PrarthonaPaul! I've started adding a few comments here.

elytron/WFLY-17143-request-uri-parameters.adoc
PrarthonaPaul: fixed!

elytron/WFLY-17143-request-uri-parameters.adoc
PrarthonaPaul: added!

elytron/WFLY-17143-request-uri-parameters.adoc
PrarthonaPaul: fixed

elytron/WFLY-17143-request-uri-parameters.adoc
PrarthonaPaul: added

elytron/WFLY-17143-request-uri-parameters.adoc
PrarthonaPaul: Updated! Thank you for the reviews!

elytron/WFLY-17143-request-uri-parameters.adoc
PrarthonaPaul: fixed!

elytron/WFLY-17143-request-uri-parameters.adoc
```diff
@@ -0,0 +1,161 @@
...
+
+* [x] OpenShift s2i
+
+* [x] Bootable jar
+
+== Requirements
+
+=== Hard Requirements
+
+* A new attribute named `request-object-required` will be added to the `secure-deployment` resource under the `elytron-oidc-client` subsystem. The value for this attribute will be a String with 3 acceptable values:
```
OndrejKotek, line 59: > A new attribute named `request-object-required` will be added to the `secure-deployment` resource under the `elytron-oidc-client` subsystem.

Why just to `secure-deployment`? Why not also to `provider`, `realm`, and/or `secure-server`?

PrarthonaPaul: Hi @OndrejKotek, 
Thank you for your review. Adding this attribute to `secure-server` makes sense and I can add that to the analysis doc. However, I am not sure if this attribute relates to `realm` or `provider_url` since request parameters encrypt query parameters. Additionally, you cannot configure `client-id` using `realm` or `provider_url`, which is needed for the auth request. So, they would have to configure `secure-server` or `secure-deployment` regardless. However, if you see other benefits to having `request-object-required` added to `realm` and `provider_url`, I am happy to add it there!

OndrejKotek: IIUIC, `provider` and `realm` resources are just containers for settings that can be referenced in `secure-deployment` or `secure-server` resources. It makes sense to me to allow to define `request-object-required`also in such resources (at higher/more global configuration levels) like the other attributes. WDYT?

elytron/WFLY-17143-request-uri-parameters.adoc
```diff
@@ -0,0 +1,165 @@
...
+=== Hard Requirements
+
+* A new attribute named `request-object-required` will be added to the `secure-deployment`, `secure-server`, `realm` and `provider` resources under the `elytron-oidc-client` subsystem. The value for this attribute will be a String with 3 acceptable values: 
+** none
+** request
+** request_uri
+
+* The `request-object-required` attribute will specify if either `request` or the `request_uri` would be added to the auth request. This will be an optional attribute. The default value for this attribute will be set to `none`, indicating that neither request parameters are used and the query parameters are sent through the url.
+
+** For multiple `secure-server` or `secure-deployment`s using the same `realm` or `provider`, the `request-object-required` can be specified at the higher level resource.
```
OndrejKotek, line 66: > ``` `secure-deployment`s ```

This formatting does not work, there needs to a space before the `s`. I suggest to use something like ``` `secure-deployment` resources ``` instead.

The same in the next point/sentence.

PrarthonaPaul: fixed!

elytron/WFLY-17143-request-uri-parameters.adoc
```diff
@@ -0,0 +1,173 @@
...
+
+* [x] OpenShift s2i
+
+* [x] Bootable jar
+
+== Requirements
+
+=== Hard Requirements
+
+* A new attributes named `request-object-required`, `request-signing-algorithm`, `request-encrypt-algorithm` and `request-encrypt-enc` will be added to the `secure-deployment`, `secure-server`, `realm` and `provider` resources under the `elytron-oidc-client` subsystem. The value for the `request-object-required` attribute will be a String with 3 acceptable values:
```
fjuma, line 59: Instead of `request-object-required`, I'm wondering if `authentication-request-format` might be a better name?

elytron/WFLY-17143-request-uri-parameters.adoc
```diff
@@ -0,0 +1,173 @@
...
+
+* The `request` and `request_uri` parameters MUST NOT be included in Request Objects.
+
+* The Request Object MAY be signed or unsigned (plaintext) and can be specified by value or by reference. The user can specify the algorithm used to sign the JWT request using the `request-signing-algorithm` attribute which will also be added to the resources specified above. 
+
+** The values for `request-signing-algorithm` attribute are of type `String`. Default value for this attribute would be `none`, specifying that the request would be sent as a plaintext. The default value of `request-signing-algorithm` is `none`. The signing algorithm specified must be one of the signing algorithms accepted by the OpenID provider. 
+
+* The request object may also be encrypted. To specify that the request parameter will be encrypted, the user needs to specify the relevant algorithms using `request-encrypt-algorithm` and `request-encrypt-enc`.
```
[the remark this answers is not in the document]
fjuma: Do we know if Keycloak supports request object encryption? (Was just taking a look at the response from http://localhost:8080/auth/realms/WildFly/.well-known/openid-configuration and I do see `request_object_signing_alg_values_supported` but I don't seem to see `request_object_encryption_alg_values_supported` or `request_object_encryption_enc_values_supported`.)

elytron/WFLY-17143-request-uri-parameters.adoc
```diff
@@ -0,0 +1,173 @@
...
+
+* According to the https://openid.net/specs/openid-connect-core-1_0.html#JWTRequests[OIDC specs], only one of the two parameters can be specified at a time. If `request` is added, then `request_uri` MUST NOT be used in the same request. However, the user can choose to specify neither by setting `request-object-required` to `none`. 
+
+* The `request` and `request_uri` parameters MUST NOT be included in Request Objects.
+
+* The Request Object MAY be signed or unsigned (plaintext) and can be specified by value or by reference. The user can specify the algorithm used to sign the JWT request using the `request-signing-algorithm` attribute which will also be added to the resources specified above. 
+
+** The values for `request-signing-algorithm` attribute are of type `String`. Default value for this attribute would be `none`, specifying that the request would be sent as a plaintext. The default value of `request-signing-algorithm` is `none`. The signing algorithm specified must be one of the signing algorithms accepted by the OpenID provider.
```
fjuma, line 76: Maybe we could call this `request-object-signing-algorithm`?

elytron/WFLY-17143-request-uri-parameters.adoc
PrarthonaPaul: That sounds better. I can change it to this. Thanks for the suggestion!

elytron/WFLY-17143-request-uri-parameters.adoc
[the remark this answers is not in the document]
PrarthonaPaul: Thank you for your comment. 
I have added to the discovery code to extract openid information, such as algorithms that are supported for signing and encrypting. There are multiple algorithms for all three supported by keycloak. 
For example here are the algorithms that keycloak supports for encrypting: 
image
This is why I allowed the user to specify the algorithms.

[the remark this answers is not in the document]
PrarthonaPaul: And yes, Keycloak does support request object encryption (please see image above). We can do a check to see if the algorithm specified by the user is one of the ones supported and throw an error/exception otherwise.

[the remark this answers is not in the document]
fjuma: Thanks for clarifying that, if Keycloak provides the ability to configure that, then it's fine to have corresponding attributes in the subsystem / JSON configuration. 

How about renaming to `request-object-encryption-algorithm` and `request-object-content-encryption-algorithm`?

elytron/WFLY-17143-request-uri-parameters.adoc
PrarthonaPaul: Done!

elytron/WFLY-17143-request-uri-parameters.adoc
```diff
@@ -0,0 +1,173 @@
...
+* [x] OpenShift s2i
+
+* [x] Bootable jar
+
+== Requirements
+
+=== Hard Requirements
+
+* A new attributes named `request-object-required`, `request-signing-algorithm`, `request-encrypt-algorithm` and `request-encrypt-enc` will be added to the `secure-deployment`, `secure-server`, `realm` and `provider` resources under the `elytron-oidc-client` subsystem. The value for the `request-object-required` attribute will be a String with 3 acceptable values: 
+** none
```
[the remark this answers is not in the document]
PrarthonaPaul: Done!

elytron/WFLY-17143-request-uri-parameters.adoc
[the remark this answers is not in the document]
PrarthonaPaul: Done!

elytron/WFLY-17143-request-uri-parameters.adoc
PrarthonaPaul: updated! I have also added a point about configuring the keystore information using existing keystore file related attributes. 
Thank you for the comments :)

elytron/WFLY-17143-request-uri-parameters.adoc
```diff
@@ -0,0 +1,177 @@
...
+    "principal-attribute" : "preferred_username",
+    "ssl-required" : "EXTERNAL"
+}
+```
+The `request` attribute is specified through the `elytron-oidc-client` subsystem as follows: 
+```
+    /subsystem=elytron-oidc-client/secure-deployment:add(client_id=myclient, provider-url="http://localhost:8090/", authentication-request-format="request_uri")
+``` 
+
+* For signed and encrypted request objects, we must need to use a private key, whose corresponding public key and certificate must be communicated with the OpenID provider. The public key is communicated to the OpenID provider using the admin console of the OpenID provider (i.e. Keycloak). And the same keypair is added to the deployment settings using existing attributes `client-keystore`, `client-keystore-password`, `client-key-password`, which are available under the `secure-deployment`, `secure-server`, `realm` and provider resources. Additionally, `client-key-alias` and `client-keystore-type` attributes will be added to these resources for additional information about the keystore.
```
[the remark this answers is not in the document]
PrarthonaPaul: Thank you for the comments. I will address these changes. 

Looking at the schema for wildfly, it seems that `elytron-oidc-client.secure-deployment.credential.client-keystore-type` and etc. already exists. This is used to specify the client credentials, such as the client secret using a keystore. But we need to use it for the `secure-deployment`, `secure-server`, `realm` and `provider`. So, we will be adding it one level higher as well.

elytron/WFLY-17143-request-uri-parameters.adoc
```diff
@@ -0,0 +1,177 @@
...
+* [x] Bootable jar
+
+== Requirements
+
+=== Hard Requirements
+
+* A new attributes named `authentication-request-format`, `request-object-signing-algorithm`, `request-object-encryption-algorithm` and `request-object-content-encryption-algorithm` will be added to the `secure-deployment`, `secure-server`, `realm` and `provider` resources under the `elytron-oidc-client` subsystem. The value for the `authentication-request-format` attribute will be a String with 3 acceptable values: 
+** oauth2: Indicating that the query parameters will be sent as usual following the oauth2 syntax. 
+** request: Indicating that the query will be sent as a jwt by value. 
+** request_uri: Indicating that the query will be sent as a reference to the jwt.
```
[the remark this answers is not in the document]
PrarthonaPaul: As discussed, since this is a value and not an attribute name, we will keep it as `request_uri` to be consistent with the specifications.

elytron/WFLY-17143-request-uri-parameters.adoc
```diff
@@ -0,0 +1,177 @@
...
+== Requirements
+
+=== Hard Requirements
+
+* A new attributes named `authentication-request-format`, `request-object-signing-algorithm`, `request-object-encryption-algorithm` and `request-object-content-encryption-algorithm` will be added to the `secure-deployment`, `secure-server`, `realm` and `provider` resources under the `elytron-oidc-client` subsystem. The value for the `authentication-request-format` attribute will be a String with 3 acceptable values: 
+** oauth2: Indicating that the query parameters will be sent as usual following the oauth2 syntax. 
+** request: Indicating that the query will be sent as a jwt by value. 
+** request_uri: Indicating that the query will be sent as a reference to the jwt. 
+
+* The `authentication-request-format` attribute will specify if either `request` or the `request_uri` would be added to the auth request. This will be an optional attribute. The default value for this attribute will be set to `none`, indicating that neither request parameters are used and the query parameters are sent through the url.
```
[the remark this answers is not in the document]
PrarthonaPaul: Fixed!

elytron/WFLY-17143-request-uri-parameters.adoc
```diff
@@ -0,0 +1,177 @@
...
+
+** For multiple `secure-server` or `secure-deployment` resources using the same `realm` or `provider`, the `authentication-request-format` can be specified at the higher level resource. 
+
+** For multiple `secure-server` or `secure-deployment` resources using the same higher level resource, if the Request Object type is not the same, then the authentication-request-format` attribute can be set accordingly for the individual lower level resources.
```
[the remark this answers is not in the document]
PrarthonaPaul: Fixed!

elytron/WFLY-17143-request-uri-parameters.adoc
```diff
@@ -0,0 +1,177 @@
...
+
+* The `authentication-request-format` attribute will specify if either `request` or the `request_uri` would be added to the auth request. This will be an optional attribute. The default value for this attribute will be set to `none`, indicating that neither request parameters are used and the query parameters are sent through the url.
+
+** For multiple `secure-server` or `secure-deployment` resources using the same `realm` or `provider`, the `authentication-request-format` can be specified at the higher level resource. 
+
+** For multiple `secure-server` or `secure-deployment` resources using the same higher level resource, if the Request Object type is not the same, then the authentication-request-format` attribute can be set accordingly for the individual lower level resources. 
+
+* According to the https://openid.net/specs/openid-connect-core-1_0.html#JWTRequests[OIDC specs], only one of the two parameters can be specified at a time. If `request` is added, then `request_uri` MUST NOT be used in the same request. However, the user can choose to specify neither by setting `authentication-request-format` to `none`.
```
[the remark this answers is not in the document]
PrarthonaPaul: Fixed!

elytron/WFLY-17143-request-uri-parameters.adoc
```diff
@@ -0,0 +1,196 @@
...
+    %23GkurKxf5T0Y-mnPFCHqWOMiZi4VS138cQO_V7PZHAdM
+    &state=af0ifjsldkj&nonce=n-0S6_WzA2Mj
+    &scope=openid
+```
+
+** The base64-encoded SHA-256 hash `23GkurKxf5T0Y-mnPFCHqWOMiZi4VS138cQO_V7PZHAdM` will allow the provider to cache the token for future use. It is a hash of the contents of the Request Object.
+
+* According to the OpenID docs, if the same parameter exists both in the Request Object and the OAuth Authorization Request parameters, the parameter in the Request Object is used.
+
+* The feature must deal with cases where the OpenId Provider does not support request parameters by recognizing `request_not_supported` error and dealing with it accordingly.
```
Skyllarr, line 130: @PrarthonaPaul `by recognizing `request_not_supported` error and dealing with it accordingly.` what will be the approach for this situation? Will some error be thrown that says that request parameters are not supported and only oauth2 format is supported?

elytron/WFLY-17143-request-uri-parameters.adoc
```diff
@@ -0,0 +1,196 @@
...
+
+* The `request` and `request_uri` parameters MUST NOT be included in Request Objects.
+
+* The Request Object MAY be signed or unsigned (plaintext) and can be specified by value or by reference. The user can specify the algorithm used to sign the JWT request using the `request-object-signing-algorithm` attribute which will also be added to the resources specified above. This algorithm must be one of the Request object signature algorithms supported by the 
+OpenID provider. 
+
+** The values for `request-object-signing-algorithm` attribute are of type `String`. Default value for this attribute would be `none`, specifying that the request would be sent as a plaintext.In order to sign the jwt using an algorithm other than `none`, the user must specify the KeyPair used. This can be done using the `client-keystore-file`, `client-keystore-password`, `client-key-password`, `client-key-alias` and `client-keystore-type` attributes. The algorithm for the KeyPair must match the algorithm used to sign the JWT. The keystore should be on type `JKS` and can be generated using `keytool` on the linux terminal.
```
Skyllarr, line 81: @PrarthonaPaul Just a minor, s/JKS/PKCS12

elytron/WFLY-17143-request-uri-parameters.adoc
PrarthonaPaul: For now I have coded it so that if either request or request_uri is not supported, then we send it as usual.
But this may not be as secure. So, in that case, if it makes more sense to, I can change it to throw a runtime error saying that request/request_uri is not supported by the OpenID provider
Which one do you think would be more appropriate?

elytron/WFLY-17143-request-uri-parameters.adoc
PrarthonaPaul: Fixed!

elytron/WFLY-17143-request-uri-parameters.adoc
Skyllarr: @PrarthonaPaul I think we should log this and then document that the request and the request_uri methods are used only if available with the OpenID provider, if not available then the warning message will be logged.

elytron/WFLY-17143-request-uri-parameters.adoc
```diff
@@ -0,0 +1,196 @@
...
+ 
+** The request object will only be encrypted if both `request-object-encryption-algorithm` and `request-object-content-encryption-algorithm` are specified. The values for these attributes must be part of the encryption algorithms and content encryption methods supported by the OpenID provider. These values can also be found by sending a GET request to the discovery url or by looking at the client configurations on the admin console of the OpenID provider. 
+
+** The JWT must be signed first and then encrypted to make a nested JWT and the key used to encrypt the JWT must be the realm public key shared by the OpenID provided. The key algorithm must be the same as the encrypting algorithm. Signing and encrypting algorithms can be of different types (i. e. ES-256 for signing and RSA-OAEP for encrypting is allowed). The realm key for encrypting the JWT can be obtained by sending a GET request to the `jwksUri` obtained from the `discoveryUrl`(http://localhost:8080/realms/myrealm/.well-known/openid-configuration).
```
fjuma, line 93: s/OpenID provided/OpenID provider

fjuma: Would be good to add some links to Keycloak issues/PRs/docs where the request object encryption is described, e.g.,

https://issues.redhat.com/browse/KEYCLOAK-18630
https://github.com/keycloak/keycloak/pull/8243
https://github.com/keycloak/keycloak/pull/8261

fjuma: Do you have some sample output that shows the result of sending a GET request to the `jwksUri` endpoint?

elytron/WFLY-17143-request-uri-parameters.adoc
```diff
@@ -0,0 +1,196 @@
...
+    http://localhost:8080/realms/myrealm/protocol/openid-connect/auth?
+    response_type=code&
+    client_id=wildfly&
+    scope=openid&
+    redirect_uri=http%3A%2F%2Flocalhost%3A8090%2Fsimple-webapp-oidc%2Fsecured&
+    request=eyJhbGciOiJSU0EtT0FFUCI...
+```
+** By reference: This is useful when the `request` string is too large and a number of other https://openid.net/specs/openid-connect-core-1_0.html#RequestUriRationale[reasons]. In this case, the Request Object is passed as a reference parameter string using the `request_uri` parameter. The url references the Request Object, which the OpenID provider can use to download the Request Object itself. The entire Request_URI MUST NOT exceed 512 ASCII characters. 
+
+** The `request_uri` value can be obtained by sending POST request to the `pushed_authentication_request_endpoint` of the OpenID provider. This can be also obtained using the discoveryUrl. The content for the POST request contains the value of the signed and/or encrypted Request Object and the content type for this request is "application/x-www-form-urlencoded".
```
fjuma, line 111: Would be good to add some pointers to relevant Keycloak docs/issues/PRs for this, e.g.,

https://issues.redhat.com/browse/KEYCLOAK-18353
https://github.com/keycloak/keycloak/pull/8144
https://github.com/keycloak/keycloak-community/blob/main/design/pushed-authorization-requests.md

elytron/WFLY-17143-request-uri-parameters.adoc
```diff
@@ -0,0 +1,196 @@
...
+
+This RFE will add an attribute to the `elytron-oidc-client` subsystem called `authentication-request-format`, which can allow the user to specify if they will use either of the request parameters. Additionally, we will be adding supporting attributes called `request-object-signing-algorithm`, `request-object-encryption-algorithm` and `request-object-content-encryption-algorithm`, which will be used to specify if and how the request parameters will be signed and/or encrypted. All of these attributes will be added to the `secure-server`, `secure-deployment`, `provider` and `realm` resources, which the user can configure using either the `elytron-oidc-client` subsystem, or using a json file in deployment configuration.
+
+We will also be adding some keystore attributes, such as, `client-keystore-file`, `client-keystore-password`, `client-key-password`, `client-key-alias` and `client-keystore-type` to the resources mentioned above. These attributes are currently available under the `credential` attribute. However, when they are configured outside of the `credentials` attribute, they will be used to sign Request Objects.
```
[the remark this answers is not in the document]
PrarthonaPaul: It could conflict if both OIDC and HTTPS are configured at the same time. 
Would there be a case where the user would want to configure OIDC with HTTPS? 
If so, then I can use two new attributes called `oidc-client-keystore` and `oidc-client-keystore-password`.

elytron/WFLY-17143-request-uri-parameters.adoc
```diff
@@ -0,0 +1,196 @@
...
+
+* The request object may also be encrypted. To specify that the Request Object will be encrypted, the user needs to specify the relevant algorithms using `request-object-encryption-algorithm` and `request-object-content-encryption-algorithm`. 
+
+** The values for the `request-object-encryption-algorithm` and the `request-object-content-encryption-algorithm` attributes are of type `String`. The default value for the attributes are undefined, which specifies that the request object will not be encrypted. 
+ 
+** The request object will only be encrypted if both `request-object-encryption-algorithm` and `request-object-content-encryption-algorithm` are specified. The values for these attributes must be part of the encryption algorithms and content encryption methods supported by the OpenID provider. These values can also be found by sending a GET request to the discovery url or by looking at the client configurations on the admin console of the OpenID provider.
```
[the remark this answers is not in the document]
PrarthonaPaul: I think the output of this would be a bit long. 
image
Do you think it would be better to point to https://github.com/wildfly-security/wildfly-elytron/blob/2.x/http/oidc/src/main/java/org/wildfly/security/http/oidc/OidcClientConfiguration.java#L212-L238 and https://github.com/wildfly-security/wildfly-elytron/blob/2.x/http/oidc/src/main/java/org/wildfly/security/http/oidc/OidcProviderMetadata.java for more information? 
Alternatively, we could give some examples of what metadata is obtained through this GET request and link either the elytron source code links I added above or this [keycloak doc](https://www.keycloak.org/docs/22.0.5/server_admin/#con-basic-settings_server_administration_guide)

elytron/WFLY-17143-request-uri-parameters.adoc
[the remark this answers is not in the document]
fjuma: Yes, it's possible that a user would want their request to be signed and also want to use HTTPS for communication with the OpenID provider. I think it would be good to use attribute names that explain what the attribute is for (e.g., something like `request-object-signing-keystore-file` perhaps).

elytron/WFLY-17143-request-uri-parameters.adoc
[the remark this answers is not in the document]
fjuma: I think we could just mention the specific ones that contain the info we're interested in (I think you had these in a previous version of the proposal), e.g., `request_object_signing_alg_values_supported`, etc.

elytron/WFLY-17143-request-uri-parameters.adoc
PrarthonaPaul: Added!

elytron/WFLY-17143-request-uri-parameters.adoc
PrarthonaPaul: I have added the links and a sample output with some of the longer values truncated for readability.

elytron/WFLY-17143-request-uri-parameters.adoc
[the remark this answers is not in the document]
PrarthonaPaul: Sounds good! 
I have made this change to include the attributes relevant to Request Objects.

elytron/WFLY-17143-request-uri-parameters.adoc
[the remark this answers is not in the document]
PrarthonaPaul: These attribute have been renamed to `request-object-signing-keystore-file`, `request-object-signing-keystore-password`, `request-object-signing-key-password`, `request-object-signing-key-alias` and `request-object-signing-keystore-type`

elytron/WFLY-17143-request-uri-parameters.adoc
```diff
@@ -0,0 +1,244 @@
...
+
+** The certificate for the KeyPair used to sign the JWT must be shared with the OpenID provider either by importing it from the keystore, or by uploading a jwksUrl. 
+
+** In order to be able to upload client keys on the admin console of the OpenID Provider, the `Client Authentication` toggle must be turned on. As a result, when configuring the client in deployment or subsystem settings, the `public-client` attribute must be set to `false` and client credentials must be specified as well using the `credentials` attribute.
```
fjuma, line 94: This is specifically about the Keycloak provider right? (since this mentions the `Client Authentication` toggle).

fjuma (approved):

elytron/WFLY-17143-request-uri-parameters.adoc
```diff
@@ -0,0 +1,244 @@
...
+
+We will also be adding some keystore attributes, such as, `request-object-signing-keystore-file`, `request-object-signing-keystore-password`, `request-object-signing-key-password`, `request-object-signing-key-alias` and `request-object-signing-keystore-type` to the resources mentioned above. There are similar attributes that are currently available under the `credential` attribute. However, when they are configured outside of the `credentials` attribute, they will be used to sign Request Objects.
```
darranl, line 15: @fjuma I see this is re-using a pattern already in use in this subsystem but is there any reason we are not making use of the key-store capability so we can rely on the Elytron subsystem handling all of the different types of KeyStore?  

Or possibly something using a credential reference / credential source type pattern?

Maybe this could be some kind of follow up consideration?

fjuma: Right, this is re-using a pattern that was already present elsewhere in the subsystem.

We could switch to making use of the `key-store` capability instead to simplify things.

We had previously talked about introducing that elsewhere in the subsystem too.

darranl: @fjuma Should we move that to a Jira as a follow up consideration so it does not block this one?

elytron/WFLY-17143-request-uri-parameters.adoc
```diff
@@ -0,0 +1,244 @@
...
+
+== Issue Metadata
+
+=== Issue
+
+* https://issues.redhat.com/browse/WFLY-17143[WFLY-17143]
+
+=== Related Issues
+
+* https://issues.redhat.com/browse/EAP7-1974[EAP7-1974]
```
darranl, line 27: I think we can remove the EAP7 reference as we are going for community stability level.

darranl (approved):
I think we can remove the EAP7 reference and I think we should have a follow up to consider the key-store reference but adding my approval.

elytron/WFLY-17143-request-uri-parameters.adoc
fjuma: Created https://issues.redhat.com/browse/WFLY-19157.

elytron/WFLY-17143-request-uri-parameters.adoc
```diff
@@ -0,0 +1,243 @@
...
+== Test Plan
+
+* WildFly Elytron Tests: Integration test cases implemented for functionality. 
+* WildFly Testsuite: Test cases will be added to check for subsystem parsing. 
+** Additional integration tests will be added to test the full functionality of the `elytron-oidc-subsystem` when `request` or `request_uri` are configured. 
+** Tests will be performed using signed (using symmetric and asymmetric keys), unsigned, encrypted and plaintext JWT requests, with the request specified both by value and by reference. 
+* Tests will be added for both subsystem and deployment configurations. 
+* Tests may be added to ensure that server configuration fails when the stability level is not specified appropriately. 
+
+== Community Documentation
```
fjuma, line 242: Just minor, I think the Release Note Content section is missing here (see https://github.com/wildfly/wildfly-proposals/blob/main/design-doc-template.adoc).

PrarthonaPaul: I have added the section, but not sure what to add.

fjuma: This should be a short summary about what the new feature provides (this is used when creating the blog post about the WildFly release).

elytron/WFLY-17143-request-uri-parameters.adoc
```diff
@@ -0,0 +1,243 @@
...
+
+** The certificate for the KeyPair used to sign the JWT must be shared with the OpenID provider either by importing it from the keystore, or by uploading a jwksUrl. 
+
+** In order to be able to upload client keys on the admin console of the OpenID Provider, the `Client Authentication` toggle must be turned on. As a result, when configuring the client in deployment or subsystem settings, the `public-client` attribute must be set to `false` and client credentials must be specified as well using the `credentials` attribute. 
+
+* The request object may also be encrypted. To specify that the Request Object will be encrypted, the user needs to specify the relevant algorithms using `request-object-encryption-algorithm` and `request-object-content-encryption-algorithm`. Encryption is done using https://openid.net/specs/draft-jones-json-web-encryption-02.html#[JWE].
```
fjuma, line 95: Should this be updated based on the discussion on the Elytron PR?

fjuma:
@PrarthonaPaul Please squash the commits when you get a chance. Thanks!

bstansberry:
Thanks @PrarthonaPaul @OndrejKotek @Skyllarr @fjuma @darranl

PR #534, opened 2023-09-05
[WFLY-18415] Support internal EJB client access to deployments on Ope…

tadamski:
…nShift over HTTP


rachmatowicz:
Also, you mention stateful sets and persistent IP addresses: stateful sets don't provide stable IP addresses, but only stable hostnames (e.g. web-0, web-1, web-2): https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/#stable-network-id

PR #539, opened 2023-09-20
[WFLY-18526] adds proposal to move BOMs to the Server source repository

emmartins:
Issue: https://issues.redhat.com/browse/WFLY-18526

### DIFF
--- a/boms/WFLY-18526_Move_BOMs_to_the_Server_Repository.adoc
+++ b/boms/WFLY-18526_Move_BOMs_to_the_Server_Repository.adoc
@@ -1,0 +1,92 @@
+= Move BOMs to the Server Repository
+:author:            Eduardo Martins
+:email:             emartins@redhat.com
+:toc:               left
+:icons:             font
+:idprefix:
+:idseparator:       -
+
+== Overview
+
+This proposal specifies the move of BOMs to the Server Repository, for both WildFly and JBoss EAP, which should prevent recurrent synchronization issues (e.g. WFLY-18043, WFLY-18155, WFLY-17879), and simplify maintenance and the WildFly/EAP release processes.
+
+== Issue Metadata
+
+=== Issue
+
+* https://issues.redhat.com/browse/WFLY-18526[WFLY-18526]
+
+=== Related Issues
+
+* https://issues.redhat.com/browse/EAP7-2146[EAP7-2146]
+* https://issues.redhat.com/browse/WFLY-18043[WFLY-18043]
+* https://issues.redhat.com/browse/WFLY-18155[WFLY-18155]
+* https://issues.redhat.com/browse/WFLY-17879[WFLY-17879]
+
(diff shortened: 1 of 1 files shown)


boms/WFLY-18526_Move_BOMs_to_the_Server_Repository.adoc
```diff
@@ -0,0 +1,92 @@
...
+// to that kind of WildFly installation
+
+* [] Bootable jar (local and cloud deployment)
+* [] OpenShift S2I
+
+== Requirements
+
+=== Hard Requirements
+
+* The BOMs sources should be moved to the Server repository, and placed in a new /boms/user directory. The BOMs sources repository history should be included in such move.
```
jamezp, line 65: "Server" is used a lot here. By "server repository" do you mean the WildFly repository?

jamezp (commented):
I don't recall when this was done, but we originally pulled the BOM out of the WildFly repo. TBH I don't recall the reason either :) I think it was a chicken and egg type of issue for the Quickstarts.

That said, I was never personally a fan of using the `org.wildfly:wildfly-parent` POM to import the dependencies. I do understand why we did that though.

I'm personally not opposed to moving the BOM back into the main WildFly repository. However, I think we need to take care when we do it. I don't think we really need to preserve the history, but that's just my opinion.

boms/WFLY-18526_Move_BOMs_to_the_Server_Repository.adoc
emmartins: yes, "server" is used an abstraction for WildFly and EAP

emmartins:
> I don't recall when this was done, but we originally pulled the BOM out of the WildFly repo. TBH I don't recall the reason either :) I think it was a chicken and egg type of issue for the Quickstarts.
> 

IIRC it was more of a technical issue but we had a very different solution at that time, with static built BOMs.

> That said, I was never personally a fan of using the `org.wildfly:wildfly-parent` POM to import the dependencies. I do understand why we did that though.

The WildFly BOMs now uses the - fairly recent - internal BOMs to obtain the dep management but it is not much diference TBH, but what we are doing is totally different than in the past, we now have BOM Builders instead, which use the maven plugin taking advantage of the whole dep management "context", to restore the transitive relationships lost with the "excludes".

> I'm personally not opposed to moving the BOM back into the main WildFly repository. However, I think we need to take care when we do it. I don't think we really need to preserve the history, but that's just my opinion.

FYI the feature PR is merged already.

jamezp:
Oh, should this be merged then?

emmartins:
Let's keep it open till feature team approves it, if changes or enhancements are needed we can add another feature PR. Downstream this only targets 8.1 so let's give time everyone some time to digest it.

bstansberry:
Time enough to digest. :) Time to merge.

PR #543, opened 2023-11-22
[WFLY-17180] JCA: add datasource validation timeout

### LINKED ISSUE #800: Connector: add datasource validation timeout parameter
### Description

Validation queries run without any explicit timeout. If the validation query blocks (as it will with most drivers when a connection is invalid/stale), this can block the entire pool in certain cases (e.g. background-validation locks the pool during validation). Even with validate-on-match, the failure to enforce a short timeout can significantly impact time to get a connection if one or more failed connections are found in the pool.

### Issue Contact

tadamski

tadamski:
https://issues.redhat.com/browse/WFLY-17180

### DIFF
--- a/jca/WFLY-17180_Add_queryTimeout_validation_parameter.adoc
+++ b/jca/WFLY-17180_Add_queryTimeout_validation_parameter.adoc
@@ -1,0 +1,120 @@
+---
+categories:
+ - jca
+stability-level: community
+issue: https://github.com/wildfly/wildfly-proposals/issues/669
+feature-team:
+ developer: Tomasz Adamski
+ sme:
+  -
+ outside-perspective: yborgessf
+---
+= [Community] WFLY-17180 Add datasource validation timeout parameter
+:author:            Tomasz Adamski
+:email:             tadamski@redhat.com
+:toc:               left
+:icons:             font
+:keywords:          connector, validation
+:idprefix:
+:idseparator:       -
+:issue-base-url:    https://issues.redhat.com/browse/
+
+== Overview
+
+Validation queries run without any explicit timeout. If the validation query blocks (as it will with most drivers when a connection is invalid/stale), this can block the entire pool in certain cases (e.g. background-validation locks the pool during validation). Even with validate-on-match, the failure to enforce a short timeout can significantly impact time to get a connection if one or more failed connections are found in the pool.
+
+Setting a driver level query timeout is not a suitable workaround because this impacts normal queries that may be expected to run for significant periods. The timeout for normal queries needs to be different from what is tolerated for validation queries (which should respond nearly instantaneously or else the connection is likely broken).
+
+
+== Implementation Plan
+
+=== Proposed solution
+
+A `validation-timeout-seconds` element is added to the `timeout` section of the datasources subsystem. It applies to both regular datasources and XA datasources.
+
+XML:
+
+[source,xml]
+----
+<datasource ...>
+    <timeout>
+        <validation-timeout-seconds>2</validation-timeout-seconds>
+    </timeout>
+</datasource>
+----
+
+CLI:
+
+[source]
+----
+/subsystem=datasources/data-source=ExampleDS:write-attribute(name=validation-timeout-seconds, value=2)
+----
+
+The value is expressed in seconds and defaults to 0 (no timeout). When set, the timeout is passed to the `ValidConnectionChecker.setTimeout(int)` method introduced in IronJacamar (JBJCA-1402).
+
+=== Changes
+
+* *IronJacamar (JBJCA-1402)*: `ValidConnectionChecker` interface gains a default `setTimeout(int)` method. `TimeOut` API and `TimeOutImpl` are extended with `validationTimeoutSeconds`. Built-in checkers (`CheckValidConnectionSQL`, `JDBC4ValidConnectionChecker`, `MySQLReplicationValidConnectionChecker`) apply the timeout; `OracleLegacyValidConnectionChecker` silently ignores it (Oracle's `pingDatabase` does not support a timeout parameter).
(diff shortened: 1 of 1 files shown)


jca/WFLY-17180_Add_queryTimeout_validation_parameter.adoc
```diff
@@ -0,0 +1,72 @@
...
+It might also be useful if a pool property or perhaps a system property could support a configurable timeout period (e.g. to account for potential latency in some systems). The Apache Tomcat pool supports something of this nature with the validationQueryTimeout property.
+
+== Configuration
+
+'query-timeout' attribute would be added to validation tag of datasources subsystem:
+
+{code}
+   <datasource (...)>
+      (...)
+      <validation query-timeout='...'>
```
tadamski, line 28: Maybe we can create this fix will a bit less elaborate configuration?

tadamski (commented):
hi @istudens, I'm trying to bootstart a RFE process for this jira. I'm wondering whether we maybe could make the config just a attribute in validation tag, making the config a little less complex, wdyt?

jca/WFLY-17180_Add_queryTimeout_validation_parameter.adoc
bstansberry: Please primarily discuss configuration in terms of the CLI.

jca/WFLY-17180_Add_queryTimeout_validation_parameter.adoc
```diff
@@ -0,0 +1,72 @@
...
+
+* Engineering
+
+=== Affected Projects or Components:
+
+* IronJacamar, WildFly Connector
+
+== Requirements
+
+* customers should be able to configure validation-timeout parameter
```
bstansberry, line 60: Please add the "Hard Requirements", "Nice-to-Have Requirements" and "Non-Requirements" sections from the template. Then please be more explicit.

The "Overview" talks about things in terms of "Minimally" and "It might also be useful" but in this section you need to say what is actually going to be done, using the 3 requirement categories. Any question in the readers' mind based on the Overview discussion should be removed when reading the requirements section.

This has been around for a while, so do I think you should say "None" under "Nice-to-Have Requirements". Any ideas are either in-scope for this iteration, or are out of scope. Use "Non-Requirements" for things you thought about that you decided are not in scope.

In the "Implementation Plan" section adding a "Future Work" section is sometimes useful, as way of noting things that *might* get done in some future iteration. Besides being good info this helps emphasize that things that might seem reasonable are not part of this iteration.

jca/WFLY-17180_Add_queryTimeout_validation_parameter.adoc
```diff
@@ -0,0 +1,72 @@
...
+
+== Proposed solution
+* modify WildFly Connector subystem datasource configuration to include 'query-timeout' attribute in 'validation' tag
+
+=== Test Plan
+
+* Connector subsystem unit tests will be extended to WildFly integration testsuite. The goal of the test would be to verify that this attribute is parsed and applied correctly.
```
bstansberry, line 67: Can you add this (assuming I'm correct and it's true)?

"The actual validation timeout handling is done in IronJacamar and has been thoroughly tested and available for many releases now. Therefore the test plan for this proposal does not cover such testing."

jca/WFLY-17180_Add_queryTimeout_validation_parameter.adoc
```diff
@@ -0,0 +1,72 @@
...
+
+=== Community docs
+
+* None
```
bstansberry, line 71: Please say something like:

"The new attribute will be documented in wildscribe".

That's the boilerplate info for "add a new attribute" proposals where no further documentation is necessary.

jca/WFLY-17180_Add_queryTimeout_validation_parameter.adoc
```diff
@@ -0,0 +1,72 @@
+= WFLY-17180 Add query-timeout validation parameter
```
bstansberry, line 1: Please add appropriate Jekyll front matter so this ends up correctly categorized on the docs.wildfly.org/wildfly-proposals page.

jca/WFLY-17180_Add_queryTimeout_validation_parameter.adoc
```diff
@@ -0,0 +1,72 @@
...
+== Issue Metadata
+
+=== Issue:
+
+* {issue-base-url}/WFLY-17180[WFLY-17180]
+
+=== Related Issues:
+
+* {issue-base-url}/EAP7-2166[EAP7-2166]
+
```
bstansberry, line 45: Please add the new 'Stability' section as per, with the desired level selected.

https://github.com/wildfly/wildfly-proposals/blob/main/design-doc-template.adoc#stability-level

I suggest 'Community'.

If the selection is other than 'Default', please preface the L1 heading with "[the level"], e.g. "[Community]WFLY-17180 Add query-timeout validation parameter"

bstansberry (changes requested):
@istudens @tadamski 

Is this something we can get done for WF 32?

istudens:
@bstansberry @tadamski I will fix this. It slipped out of my radar.

bstansberry:
@istudens That's understandable. With the new WildFly feature development process things like this that have been stuck for a long time can more forward. I'm not sure but I believe this one is mostly done except for some paperwork stuff, like this doc.

jca/WFLY-17180_Add_queryTimeout_validation_parameter.adoc
tadamski: fixed

jca/WFLY-17180_Add_queryTimeout_validation_parameter.adoc
tadamski: fixed

jca/WFLY-17180_Add_queryTimeout_validation_parameter.adoc
```diff
@@ -0,0 +1,120 @@
+---
+categories:
+ - jca
+stability-level: community
+issue: https://github.com/wildfly/wildfly-proposals/issues/669
+feature-team:
+ developer: Tomasz Adamski
+ sme:
+  -
+ outside-perspective: yborgessf
```
yersan: s/yborgessf/yborgess

PR #545, opened 2023-11-29
[WFLY-14984] Add support for encrypted expressions in the wildfly-config.xml

PrarthonaPaul:
Issue: https://issues.redhat.com/browse/WFLY-14984
Considerations based on preliminary discussions: 
- Multiple attributes that could be candidates for supporting encrypted expressions. May be good to add support for encrypted expression for any attribute for forward compatibility. 
- Proposal considers the support for encrypted expression being its own section which can be specified first and other sections can refer to it. 

todo before implementation: 
-  look into what types of resolution properties are already available for the wildfly client


elytron/WFLY-14984-encrypted-expression-for-client-config.adoc
```diff
@@ -0,0 +1,164 @@
...
+
+Attribute values for client configuration can include sensitive information. And while credential references can be used for some of them, others still need to be defined using plaintext. As a result, those attributes are still exposed and make the application vulnerable. Additionally, the password for the credential store needs to be defined as clear-text, which is not ideal. 
+
+This feature adds support for encrypted expressions for attributes in `wildfly-config.xml` file, which allows users to avoid specifying sensitive information using plaintext. Instead, encrypted expressions can be used to specify these information. A configuration section named `encrypted-expression` would be added, which would reference a `secret-key-credential-store` to extract encrypted expressions. It would also have a resolver, which would be backed by the secret-key-credential-store. The `secret-key-credential-store` can be created either using the `elytron-tool` or using the Elytron subsystem on the WildFly server.
```
fjuma, line 14: This is probably just a copy/paste thing but since wildfly-config.xml is something that's used on the client side, we shouldn't be mentioning the Elytron subsystem here.

elytron/WFLY-14984-encrypted-expression-for-client-config.adoc
```diff
@@ -0,0 +1,164 @@
...
+
+* The ability to specify attribute values as encrypted expressions should be added to wildfly-config.xml. 
+
+* The parsing needs to be able to run separately from the application server, as the server may not be running while the client configuration is being set up. 
+
+* Configurations may not depend on components created for server configuration, such as `resolvers` or `elytron/expression`. The resolver for the encrypted expression should be configured through the client configuration. 
+
+* The `encrypted-expression` configuration can be specified as follows: 
+```
+<encrypted-expression>
```
fjuma, line 66: It would be good to add more details here about the specific elements, attributes, etc. that are being introduced. Then the block below could be an example.

elytron/WFLY-14984-encrypted-expression-for-client-config.adoc
```diff
@@ -0,0 +1,164 @@
...
+
+* The encrypted expression would be created using the `--generate-secret-key` command and can be read using the `--export-secret-key` command. 
+
+=== Nice-to-Have Requirements
+
+* The parsing of the encrypted expression could be added as a separate namespace in the schema and should be added at the top of the xml file. 
+
+** While parsing the xml file, the section for encrypted expression and its resolver would need to be parsed before any other sections that reference it. 
+
+* Encrypted expressions can be added to other subsystems in addition to elytron. As a result, the parsing rules and sequences should be configured to be compatible with those subsystems as well.
```
fjuma, line 102: Since this is client side, the terminology here needs to be updated. It's currently referencing subsystems.

elytron/WFLY-14984-encrypted-expression-for-client-config.adoc
PrarthonaPaul: I was thinking in case the user is doing some server configuration as well like this blog does: https://wildfly-security.github.io/wildfly-elytron/blog/ejb-over-tls/
In these cases, we can also use the management cli to create the credential store as opposed to using the elytron-tool. 
WDYT? Could this be a possible use case?

elytron/WFLY-14984-encrypted-expression-for-client-config.adoc
```diff
@@ -0,0 +1,164 @@
...
+
+== Issue Metadata
+
+=== Issue
+
+* https://issues.redhat.com/browse/WFLY-14984[WFLY-14984]
+
+// === Related Issues
+
+// * https://issues.redhat.com/browse/WFCORE[WFCORE-XXXX]
```
PrarthonaPaul, line 24: Do we need an ELY issue for this?

elytron/WFLY-14984-encrypted-expression-for-client-config.adoc
```diff
@@ -0,0 +1,172 @@
...
+
+Attribute values for client configuration can include sensitive information. And while credential references can be used for some of them, others still need to be defined using plaintext. As a result, those attributes are still exposed and make the application vulnerable. Additionally, the password for the credential store needs to be defined as clear-text, which is not ideal. 
+
+This feature adds support for encrypted expressions for attributes in `wildfly-config.xml` file, which allows users to avoid specifying sensitive information using plaintext. Instead, encrypted expressions can be used to specify these information. A configuration section named `encrypted-expression` would be added, which would reference a `secret-key-credential-store` to extract encrypted expressions. It would also have a resolver, which would be backed by the secret-key-credential-store. The `secret-key-credential-store` can be created either using the `elytron-tool`.
```
fjuma, line 14: "to extract encrypted expressions" should be reworded. The credential store contains the secret key that is used to encrypt/decrypt encrypted expressions.

fjuma: "The `secret-key-credential-store` can be created either using the `elytron-tool`."

This sentence is incomplete I think.

elytron/WFLY-14984-encrypted-expression-for-client-config.adoc
fjuma: Yes. We can reference your newly created issue here.

elytron/WFLY-14984-encrypted-expression-for-client-config.adoc
```diff
@@ -0,0 +1,172 @@
...
+
+* [x] Bootable jar
+
+== Requirements
+
+=== Hard Requirements
+
+* The ability to specify attribute values as encrypted expressions should be added to wildfly-config.xml. 
+
+* The parsing needs to be able to run separately from the application server, as the server may not be running while the client configuration is being set up.
```
fjuma, line 60: I think this could be reword to something like:

Parsing for client configuration is separate from parsing server configuration.

elytron/WFLY-14984-encrypted-expression-for-client-config.adoc
```diff
@@ -0,0 +1,172 @@
...
+
+=== Hard Requirements
+
+* The ability to specify attribute values as encrypted expressions should be added to wildfly-config.xml. 
+
+* The parsing needs to be able to run separately from the application server, as the server may not be running while the client configuration is being set up. 
+
+* Configurations may not depend on components created for server configuration, such as `resolvers` or `elytron/expression`. The resolver for the encrypted expression should be configured through the client configuration. 
+
+* The `encrypted-expression` would be a new configuration block added to the client config under the new schema. The new attributes added include `resolvers`, with attributes called `resolver` under it. Additionally, <credential-stores> attribute would be added to the schema with the <credential-store> attributes added to it. The structures of this attribute would be the same as the one one under the `authentication-client` schema.
```
fjuma, line 64: Just a small thing, resolvers and resolver would be elements (as opposed to attributes).

Same with credential-stores and credential-store.

elytron/WFLY-14984-encrypted-expression-for-client-config.adoc
```diff
@@ -0,0 +1,172 @@
...
+* Configurations may not depend on components created for server configuration, such as `resolvers` or `elytron/expression`. The resolver for the encrypted expression should be configured through the client configuration. 
+
+* The `encrypted-expression` would be a new configuration block added to the client config under the new schema. The new attributes added include `resolvers`, with attributes called `resolver` under it. Additionally, <credential-stores> attribute would be added to the schema with the <credential-store> attributes added to it. The structures of this attribute would be the same as the one one under the `authentication-client` schema. 
+
+** There may be multiple `resolver` attribute listed under <resolvers>, and there may be multiple <credential-store> attributes added under <credential-stores>.
+
+** The configuration for this block can be specified as follows: 
+```
+<encrypted-expression>
+	<credential-store name="credential-store-name" type="PKCS12" provider="provider-name"/>
```
fjuma, line 71: Should there be a <credential-stores> element around this?

elytron/WFLY-14984-encrypted-expression-for-client-config.adoc
```diff
@@ -0,0 +1,172 @@
...
+        <attributes>
+            <attribute name="path" value="path/to/credential-store-name.cs" />
+        </attributes>
+        <protection-parameter-credentials>
+            <clear-password password="password" />
+        </protection-parameter-credentials>
+    </credential-store>
+    <resolvers>
+        <resolver name= “resolver-name” credential-store= “credential-store-name key=“key”/>
```
fjuma, line 80: credential-store-name is missing the closing quote here and below.

elytron/WFLY-14984-encrypted-expression-for-client-config.adoc
```diff
@@ -0,0 +1,172 @@
...
+        <resolver name= “resolver-name-2” credential-store= “credential-store-name key=“example”/>
+    </resolvers>
+    <default-resolver name=resolver-name>
```
fjuma, line 83: Does default-resolver need to be its own element? Can it be an attribute defined on the expression-resolver like in the server side config? e.g., see the xml config here:

https://wildfly-security.github.io/wildfly-elytron/blog/wildfly-encrypted-expressions/

elytron/WFLY-14984-encrypted-expression-for-client-config.adoc
```diff
@@ -0,0 +1,172 @@
...
+<encrypted-expression>
+	<credential-store name="credential-store-name" type="PKCS12" provider="provider-name"/>
+        <attributes>
+            <attribute name="path" value="path/to/credential-store-name.cs" />
+        </attributes>
+        <protection-parameter-credentials>
+            <clear-password password="password" />
+        </protection-parameter-credentials>
+    </credential-store>
+    <resolvers>
```
fjuma, line 79: In the server config, I think we use <expression-resolvers>, right (would be good to double check what's used there)? Any reason to use "resolvers" here instead of "expression-resolvers"?

elytron/WFLY-14984-encrypted-expression-for-client-config.adoc
```diff
@@ -0,0 +1,172 @@
...
+        <resolver name= “resolver-name” credential-store= “credential-store-name key=“key”/>
+        <resolver name= “resolver-name-2” credential-store= “credential-store-name key=“example”/>
+    </resolvers>
+    <default-resolver name=resolver-name>
+</encrypted-expression>
+```
+Note that the initial password for the secret-key-credential-store used for the resolver would need to be specified in clear-text. 
+
+Once parsing is done, the resolver would need to be created programmatically (if not created by the server already) and be connected with the secret-key-credential-store.
```
fjuma, line 88: Can you explain this some more? The client side config is unrelated to the server side config so not quite following the reference to the server side here.

elytron/WFLY-14984-encrypted-expression-for-client-config.adoc
fjuma: "It would also have a resolver, which would be backed by the secret-key-credential-store"

Will the resolver be able to reference a credential-store as well in addition to being able to reference a secret-key-credential-store?

elytron/WFLY-14984-encrypted-expression-for-client-config.adoc
fjuma: Will there be a separate secret-key-credential-store element?

elytron/WFLY-14984-encrypted-expression-for-client-config.adoc
```diff
@@ -0,0 +1,172 @@
...
+    <attributes>
+        <attribute name="path" value="path/to/credential-store-name.cs" />
+    </attributes>
+    <protection-parameter-credentials>
+	    <clear-password password={ENC::resolver-name:encrypted-expression} />
+    </protection-parameter-credentials>
+</credential-store>
+```
+
+* The encrypted expression would be created using the `--generate-secret-key` command and can be read using the `--export-secret-key` command.
```
fjuma, line 102: Might be good to get some examples added here.

elytron/WFLY-14984-encrypted-expression-for-client-config.adoc
PrarthonaPaul: Resolvers are backed by aliases from a [CredentialStore](https://wildfly-security.github.io/wildfly-elytron/documentation/api/eap74/org/wildfly/security/credential/store/package-summary.html) object. Secret-key-credential-store is an instance of that class with type PropertiesCredentialStore. So, yes, we can use either type. 
I will add a note about that. Thanks!

elytron/WFLY-14984-encrypted-expression-for-client-config.adoc
PrarthonaPaul: Nope, we can just use the credential-store element and specify the `type` as keystorecredentialstore (which is the default config) or PropertiesCredentialStore.

elytron/WFLY-14984-encrypted-expression-for-client-config.adoc
fjuma: In the linked blog post, it looks like the credential store is created using elytron-tool.sh. I don't see subsystem commands being used to create it.

The management CLI is used for server config so I think mentioning it just makes things a bit confusing since this RFE is about the client side.

elytron/WFLY-14984-encrypted-expression-for-client-config.adoc
fjuma: Ok thanks. Would be good to update the intro since it currently references a secret-key-credential-store element.

elytron/WFLY-14984-encrypted-expression-for-client-config.adoc
```diff
@@ -0,0 +1,164 @@
...
+```
+<encrypted-expression>
+	<credential-store name="credential-store-name" type="PKCS12" provider="provider-name"/>
+        <attributes>
+            <attribute name="path" value="path/to/credential-store-name.cs" />
+        </attributes>
+        <protection-parameter-credentials>
+            <clear-password password="password" />
+        </protection-parameter-credentials>
+            </credential-store>
```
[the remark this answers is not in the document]
PrarthonaPaul: fixed!

elytron/WFLY-14984-encrypted-expression-for-client-config.adoc
PrarthonaPaul: fixed!

elytron/WFLY-14984-encrypted-expression-for-client-config.adoc
PrarthonaPaul: fixed!

elytron/WFLY-14984-encrypted-expression-for-client-config.adoc
PrarthonaPaul: fixed!

elytron/WFLY-14984-encrypted-expression-for-client-config.adoc
PrarthonaPaul: fixed! 
There are still some mentions of a secret-key-credential, but when mentioning the element itself, I am referring to it as a credential-store. 
Please let me know if I should change it all to be credential-store

elytron/WFLY-14984-encrypted-expression-for-client-config.adoc
PrarthonaPaul: fixed!

elytron/WFLY-14984-encrypted-expression-for-client-config.adoc
PrarthonaPaul: fixed

elytron/WFLY-14984-encrypted-expression-for-client-config.adoc
PrarthonaPaul: if it is not added as a separate element, would we add it as a property for the encrypted expression element? 
so like 
```
<encrypted-expression default-resolver="resolver-name">
INSERT CONFIGURATION
</encrypted expression>
```
If we do go with this, I am not sure if we can ensure that `resolver-name` is one of the resolvers specified under the <expression-resolver> elements. I think the parser currently checks to make sure it is.

elytron/WFLY-14984-encrypted-expression-for-client-config.adoc
PrarthonaPaul: Updated to <expression-resolvers> and <expression-resolver>. I will update the code to reflect this change as well.

elytron/WFLY-14984-encrypted-expression-for-client-config.adoc
PrarthonaPaul: "(if not created by the server already)" should not be there since this is entirely on the client side. 
I have fixed it.

elytron/WFLY-14984-encrypted-expression-for-client-config.adoc
fjuma: Just looking at the example from the blog post:

```
<expression-resolver default-resolver="main-resolver">
    <resolver name="initial-resolver" credential-store="initial" secret-key="key"/>
    <resolver name="main-resolver" credential-store="main" secret-key="main-key"/>
</expression-resolver>
```

Can we make use of this format?

PrarthonaPaul: Yes, I will update the schema and parser to use this. 
Also, would it be better to make the default-resolver a required attribute? or make it optional, but if the user does not specify one, then the first resolver becomes the default resolver?

fjuma: How is this done on the server side?

PrarthonaPaul: The server does neither. 
When adding an expression resource using the following command: 
`/subsystem=elytron/expression=encryption:add(resolvers=[{name=initial-resolver, credential-store=initial, secret-key=key}])` defining a value for the `default-resolver` attribute is optional on the server side. 
Now if someone tries to create an expression without defining a resolver, then the output is as follows: 
```
/subsystem=elytron/expression=encryption:create-expression(clear-text=password)
{
    "outcome" => "failed",
    "failure-description" => "WFLYELY01200: The name of the resolver to use was 
not specified and no default-resolver has been defined.",
    "rolled-back" => true
}
```

On the client, the create-expression part is done using the `--encrypt` option on the wildfly-elytron tool, which would give you an output like this: 
```
Clear text encrypted to token 'RUxZAUMQQ1PVWZuimKZCpt0s6MONrl7cbqLGPhyBBmm83fXdfZA=' using alias 'key'.
```

It would be up to the user to convert it to something like `{ENC::resolver_name:RUxZAUMQQ1PVWZuimKZCpt0s6MONrl7cbqLGPhyBBmm83fXdfZA=}` (where the expression resolver called `resolver-name` would be used to decrypt)
or they can convert it to `{ENC::RUxZAUMQQ1PVWZuimKZCpt0s6MONrl7cbqLGPhyBBmm83fXdfZA=}` (where the default expression resolver would be used to decrypt)

There is code in place already in the resolver class to handle cases where the user goes with the second configuration, but there is no default-expression-resolver defined. So, if we do want to do something similar to what the server does, I think it may be possible to throw an exception when trying to resolve the expression. However, since we are using the wildfly-elytron tool to encrypt, we cannot throw the same exception when creating the expression, which is what the server does.

fjuma: Ok given that, I think it would be good to have the default-resolver attribute be optional here too.

An encrypted expression provided in wildfly-config should either specify the resolver name or rely on the default being used.

If relying on the default and the default-resolver hasn't been defined, then that's invalid configuration and an exception should occur when attempting to resolve it.

elytron/WFLY-14984-encrypted-expression-for-client-config.adoc
```diff
@@ -0,0 +1,171 @@
...
+```
+* In order to encrypt a clear-text expression, `--encrypt` option in the wildfly-elytron tool can be used as follows: 
+```
+$ bin/elytron-tool.sh credential-store --location standalone/configuration/store-one.cs --type PropertiesCredentialStore --encrypt key
+Clear text value:
+Confirm clear text value:
+Clear text encrypted to token 'RUxZAUMQvGzk6Vaadp2cahhZ6rlPhHOZcWyjXALlAthrENvRTvQ=' using alias 'key'.
+```
+
+* expression-resolvers should be backed by an entry from a credential-store of type KeyStoreCredentialStore. However, when using a KeyStoreCredentialStore type credential store, the protection parameter, which is the password, needs to be specified in clear-text inside the <encrypted-expressions> tag.
```
fjuma, line 122: expression-resolvers can reference any type of credential-store right? Here, KeyStoreCredentialStore is mentioned specifically.

elytron/WFLY-14984-encrypted-expression-for-client-config.adoc
```diff
@@ -0,0 +1,171 @@
...
+```
+$ bin/elytron-tool.sh credential-store --location standalone/configuration/store-one.cs --type PropertiesCredentialStore --encrypt key
+Clear text value:
+Confirm clear text value:
+Clear text encrypted to token 'RUxZAUMQvGzk6Vaadp2cahhZ6rlPhHOZcWyjXALlAthrENvRTvQ=' using alias 'key'.
+```
+
+* expression-resolvers should be backed by an entry from a credential-store of type KeyStoreCredentialStore. However, when using a KeyStoreCredentialStore type credential store, the protection parameter, which is the password, needs to be specified in clear-text inside the <encrypted-expressions> tag. 
+
+* Attribute value resolvers inside the wildfly-client-config project should be updated with the ability to parse encrypted expressions and resolve them to extract the clear-text sensitive information. In order to avoid circular dependencies, service loader would be used by the wildfly-client-config project to use an EncryptedExpressionResolver class to decrypt an encrypted expression.
```
fjuma, line 124: Might be good to provide some more details on the Service Loader approach if possible.

elytron/WFLY-14984-encrypted-expression-for-client-config.adoc
```diff
@@ -0,0 +1,171 @@
...
+
+=== Nice-to-Have Requirements
+
+* The parsing of the encrypted expression should be added at the top of the xml file and must be parsed before any other class that uses encrypted expressions.
```
fjuma, line 128: Is this still relevant?

elytron/WFLY-14984-encrypted-expression-for-client-config.adoc
```diff
@@ -0,0 +1,171 @@
...
+
+* The `encrypted-expressions` would be a new schema added to the client config under WildFly client configuration. The new elements added include <expression-resolvers>, with elements called <expression-resolver> under it. Additionally, <credential-stores> attribute would be added to the schema with the <credential-store> elements added to it. The structures of this attribute would be the same as the one one under the `authentication-client` schema. 
+
+** There may be multiple `expression-resolver` attribute listed under <expression-resolvers>, and there may be multiple <credential-store> elements added under <credential-stores>. However, there may be a maximum of 1 default expression resolver. This would be used to decrypt expression when no resolvers are specified. 
+
+** The configuration for this block can be specified as follows: 
+```
+<encrypted-expressions>
```
fjuma, line 70: Just noticed `encrypted-expressions` is used here but `encrypted-expression` is used in the blog post https://github.com/wildfly-security/wildfly-elytron/pull/2088.

elytron/WFLY-14984-encrypted-expression-for-client-config.adoc
```diff
@@ -0,0 +1,171 @@
...
+            <attributes>
+                <attribute name="location" value="path/to/mycred.cs"/>
+            </attributes>
+            <protection-parameter-credentials>
+                <clear-password password="password"/>
+            </protection-parameter-credentials>
+        </credential-store>
+    </credential-stores>
+    <expression-resolvers default-resolver="expression-resolver-name">
+        <expression-resolver name= “expression-resolver-name” credential-store= “credential-store-name" alias=“key”/>
```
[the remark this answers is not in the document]
PrarthonaPaul: No particular reason. I assumed `alias` is more unambiguous than `secret-key`. But it should be easy to change it to `secret-key` if we decide to go with that.

elytron/WFLY-14984-encrypted-expression-for-client-config.adoc
PrarthonaPaul: Changed it to "However, when using a other types of credential stores that are protected by a protection parameter, such as a password, that needs to be specified in clear-text inside the <encrypted-expressions> tag. "

elytron/WFLY-14984-encrypted-expression-for-client-config.adoc
PrarthonaPaul: Added!

elytron/WFLY-14984-encrypted-expression-for-client-config.adoc
PrarthonaPaul: Removed this point since it is no longer relevant.

elytron/WFLY-14984-encrypted-expression-for-client-config.adoc
PrarthonaPaul: fixed it on the blog! 
Thanks, @fjuma for reviewing the proposal!

elytron/WFLY-14984-encrypted-expression-for-client-config.adoc
```diff
@@ -0,0 +1,175 @@
...
+
+This feature adds support for encrypted expressions for elements in `wildfly-config.xml` file, which allows users to avoid specifying sensitive information using plaintext. Instead, encrypted expressions can be used to specify these information. A configuration section named `encrypted-expressions` would be added, which would reference a `secret-key-credential-store` to convert plaintext to encrypted expressions and vice versa. The secret-key-credential-store would be indicated by the `credential-store` tag with the `type` being `PropertiesCredentialStore`. It would also have a expression-resolver, which would be backed by one of the secret key aliases from the the secret-key-credential-store. The `credential-store` can be created using the wildfly-elytron tool using the `credential-store` command and the `--create` option.
```
rsearls, line 14: Change, "these information." to "this information."

It might be helpful to future readers to reference this doc about wildfly-elytron tool,
https://github.com/wildfly/wildfly/blob/main/docs/src/main/asciidoc/_elytron/Credential_Store.adoc

elytron/WFLY-14984-encrypted-expression-for-client-config.adoc
```diff
@@ -0,0 +1,175 @@
...
+```
+* In order to encrypt a clear-text expression, `--encrypt` option in the wildfly-elytron tool can be used as follows: 
+```
+$ bin/elytron-tool.sh credential-store --location standalone/configuration/store-one.cs --type PropertiesCredentialStore --encrypt key
+Clear text value:
+Confirm clear text value:
+Clear text encrypted to token 'RUxZAUMQvGzk6Vaadp2cahhZ6rlPhHOZcWyjXALlAthrENvRTvQ=' using alias 'key'.
+```
+
+* expression-resolvers should be backed by an entry from a credential-store of type KeyStoreCredentialStore. However, when using other types of credential stores that are protected by a protection parameter, such as a password, that needs to be specified in clear-text inside the <encrypted-expressions> tag.
```
rsearls, line 124: Change, "that needs to be specified" to "it needs to be specified".

rsearls (approved):

PR #554, opened 2024-01-31
[WFCORE-6503] Add support for unmanaged deployment to YAML extension

ehsavoie:
Jira: https://issues.redhat.com/browse/WFCORE-6503


rpelisse:
Oh, that's a cool feature! We had an Ansible user asking for that. Right now, we deployed app after the fact, adding some for loop around a JBoss CLI call wrapped by Ansible. It's working, but it's a bit unelegant. Being able to deal with app deployment with template provided to the YAML Config feature would be quite neat.

ehsavoie:
Merged into https://github.com/wildfly/wildfly-proposals/pull/384

PR #555, opened 2024-02-09
[Community] Proposal for WFLY-19021, Stability support in WildFly provisioning

jfdenise:
Proposal for Issue: https://issues.redhat.com/browse/WFLY-19021

### DIFF
--- a/wf-galleon/WFLY-19021-Stability_In_Provisioning.adoc
+++ b/wf-galleon/WFLY-19021-Stability_In_Provisioning.adoc
@@ -1,0 +1,315 @@
+---
+categories:
+  - wf-galleon
+---
+= [Community] WildFly provisioning to support WildFly stability
+:author:            Jean-Francois Denise
+:email:             jdenise@redhat.com
+:toc:               left
+:icons:             font
+:idprefix:
+:idseparator:       -
+
+== Overview
+
+WildFly 31 introduced the notion of stability level. Galleon, WildFly Galleon plugins and provisioning tooling must be evolved to support this notion.
+This proposal describes how stability is handled by the WildFly provisioning. 
+
+== Issue Metadata
+
+=== Issue
+
+* https://issues.redhat.com/browse/WFLY-19021[WFLY-19021] - WildFly provisioning to support WildFly stability
+
+=== Related Issues
+
+* https://issues.redhat.com/browse/WFCORE-6668[WFCORE-6668] - Ensure read-feature-description results contain stability as reported by associated description for use by WildFly Glow
+* https://issues.redhat.com/browse/WFGP-265[WFGP-265] - Support setting a minimum stability level when provisioning
+* https://issues.redhat.com/browse/WFGP-264[WFGP-264] - Support setting a minimum stability level when building a feature pack
+* https://issues.redhat.com/browse/GAL-357[GAL-357] - Support for feature/package stability
+* https://github.com/wildfly-extras/prospero/issues/570[Issue 570] - Support in prospero for stability level
+
+
+=== Stability Level
+// Choose the planned stability level for the proposed functionality
+* [ ] Experimental
+
+* [ ] Preview
+
+* [x] Community
+
+* [ ] default
+
+
+=== Dev Contacts
+
+* mailto:{brian.stansberry@redhat.com}[Brian Stansberry]
+* mailto:{kkhan@redhat.com}[{Kabir Khan}]
+* mailto:{email}[{author}]
+
+=== QE Contacts
+
+* N/A.
+
+=== Testing By
+// Put an x in the relevant field to indicate if testing will be done by Engineering or QE. 
+// Discuss with QE during the Kickoff state to decide this
+* [x] Engineering
+
+* [ ] QE
+
+=== Affected Projects or Components
+
+* https://github.com/wildfly/wildfly[WildFly]
+* https://github.com/wildfly/galleon-plugins[WildFly Galleon plugins]
+* https://github.com/wildfly/galleon[Galleon]
+* https://github.com/wildfly/wildfly-glow[WildFly Glow]
+* https://github.com/wildfly/wildfly-maven-plugin[WildFly Maven Plugin]
+* https://github.com/wildfly-extras/wildfly-jar-maven-plugin[WildFly Bootable JAR]
+* https://github.com/wildfly-extras/prospero[Prospero]
+
+=== Relevant Installation Types
+// Remove the x next to the relevant field if the feature in question is not relevant
+// to that kind of WildFly installation
+* [x] Traditional standalone server (unzipped or provisioned by Galleon)
+
+* [x] Managed domain
+
+* [x] OpenShift s2i
+
+* [x] Bootable jar
+
+== Requirements
+
+Stability becomes a Galleon notion that applies to feature-pack, feature/feature_param and package.
+There is a 1 to 1 mapping with WildFly stability.
+The stability levels are: `default` >`community` > `preview` > `experimental`
+
+Galleon feature-pack descriptor is extended with 2 new attributes:
+
+* `config-stability-level`: The stability to be used when provisioning configurations.
+* `package-stability-level`: The stability to be used when provisioning packages.
+
+These attributes are set when building Galleon feature-packs and used by Galleon runtime at provisioning time.
+
+=== Requirements at feature-pack build time
+ 
+==== feature-pack minimum stability level
+
+A minimum stability level can be set at feature-pack build time. If no stability level is specified, the `default` level is used.
+
+Stability level is set thanks to a new option of the WildFly Galleon Plugins Maven plugin: 
+`<minimum-stability-level>default|community|preview|experimental</minimum-stability-level>`
+
+Feature-specs and packages at a lower stability level than is specified will not be included in the feature pack.
+
+==== feature-pack provisioning stability level
+
+This level is used as the default stability level to be used when provisioning server configurations and packages.
+If no stability level is specified, the `default` level is used for both configs and packages stability levels. 
+It is set thanks to a new option of the WildFly Galleon Plugins Maven plugin: 
+`<stability-level>default|community|preview|experimental</stability-level>`.
+
+[NOTE]
+====
+If one wants to have different values for configs and packages, then `<config-stability-level>` and `<package-stability-level>` are to be used instead of `<stability-level>`.
+The use case for using `config-stability-level` and `package-stability-level` as an alternative to `stability-level` 
+is when the user wishes to generates configurations with features at a given stability level 
+while allowing provisioning of packages at a lower level. 
+The presence of the lower stability level packages allows subsequent update of the configuration to enable lower stability features.
+====
+
+==== feature-pack provisioning config stability level
+
+This level is used as the default stability level to be used when provisioning server configurations.
+If no stability level is specified at feature-pack build time, the `default` level is used.
+Stability level is set thanks to a new option of the WildFly Galleon Plugins Maven plugin: 
+`<config-stability-level>default|community|preview|experimental</config-stability-level>`
+
+This level can be overridden at provisioning time with the Galleon option `<config-stability-level>`
+
+==== feature-pack provisioning package stability level
+
+This level is used as the default stability level to be used when provisioning server packages (e.g.: JBoss Modules modules). 
+If no stability level is specified, the `default` level is used.
+Stability level is set thanks to a new option of the WildFly Galleon Plugins Maven plugin: 
+`<package-stability-level>default|community|preview|experimental</package-stability-level>`
+
+This level can be overridden at provisioning time with the Galleon option `<package-stability-level>`
+
+[NOTE]
+====
+* Configuring `<stability-level>` and `<config|package-stability-level>` options in the same plugin execution is not supported and will lead to a failure.
+* If the specified stability level is not enabled by the level specified by the 'minimum-stability-level' setting, the build of the feature pack will fail.
+* This stability level is not the minimal stability level at which a feature-pack can be provisioned. At provisioning time, a lower stability level 
+for both configs and packages can be specified.
+====
+
+==== JBoss Modules modules stability level
+
+A stability level can be set in `module.xml` thanks to the property `jboss.stability`. 
+Possible value being `default|community|preview|experimental`. If no stability is specified, `default` is assumed.
+
+==== Implementation notes
+
+At build time, the minimum stability level is used in 2 cases:
+
+* To start the embedded server that generates the Galleon features at the correct stability level (allowing the features to be discovered in the server).
+* To filter-out packages at a stability level lower than the minimum-stability level. Such packages are not packaged inside the feature-pack.
+
+==== Stability levels for WildFly feature-packs
+
+Expected minimum stability levels for WildFly feature-packs:
+
+* wildfly-ee-galleon-pack: `experimental`
+* wildfly-galleon-pack: `experimental`
+* wildfly-preview-feature-pack: `experimental`
+* wildfly-core: `experimental`
+
+Expected config stability levels for WildFly feature-packs:
+
+* wildfly-ee-galleon-pack: `community`
+* wildfly-galleon-pack: `community`
+* wildfly-preview-feature-pack: `preview`
+* wildfly-core: `community`
+
+Expected package stability levels for WildFly feature-packs:
+
+* wildfly-ee-galleon-pack: `experimental`
+* wildfly-galleon-pack: `experimental`
+* wildfly-preview-feature-pack: `experimental`
+* wildfly-core: `experimental`
+
+=== Requirements at provisioning time
+
+Provisioning time follows some rules to properly handle stability and avoid features/packages of wrong stability being provisioned.
+
+* The feature-pack contained config stability level is used to constrain the set of provisioned features. Any features at a lower stability level contained in the 
+feature-pack (if any), are not provisioned.
+* The feature-pack contained package stability level is used to constrain the set of provisioned packages. Any packages at a lower stability level contained in the 
+feature-pack (if any), are not provisioned.
+* By default, the config and package stability levels of each feature-pack are used when provisioning its own content. 
+* Each feature-pack present in the provisioning configuration can have a different config stability level. 
+The config stability of each feature-pack applies to its own content and doesn't leak into other feature-packs.
+* Each feature-pack present in the provisioning configuration can have a different package stability level. 
+The package stability of each feature-pack applies to its own content and doesn't leak into other feature-packs.
+* A user can specify a config stability level, by means of the `config-stability-level` Galleon option. NB: This option can't be set when `stability-level` is set.
+This stability level is used to enable the stability level of all features present in configurations generated during provisioning.
+* A user can specify a package stability level, by means of the `package-stability-level` Galleon option. 
+This stability level is used to set the stability level of all provisioned packages. NB: This option can't be set when `stability-level` is set.
+* A user can specify a stability level for both configs and packages, by means of the `stability-level` Galleon option. This option is a convenience 
+allowing to set both the config and package stability level using a single option. It can't be set when `config-stability-level` or `package-stability-level` 
+are in use.
+* At provisioning time, usage of `<stability-level>` and `<config|package-stability-level>` options in the same provisioning execution 
+is not supported and would lead to a failure.
+
+For example, a provisioning configuration containing 3 feature-packs:
+
+* A, config and package stability default, minimum stability default
+* B, config and package stability community, minimum stability community
+* C, config and package stability experimental, minimum stability experimental
+
+If no stability level is specified:
+
+* `default` features/packages of A are provisioned
+* `community` and `default` features/packages of B are provisioned
+* `experimental`, `preview`, `community` and `default` features/packages of C are provisioned
+
+If the config stability level `default` is specified:
+
+* `default` features and packages of A are provisioned
+* `default` features + `community` and `default` packages of B are provisioned
+* `default` features + `experimental` and `preview` and `community` and `default` packages of C are provisioned
+
+If the package stability level `experimental` is specified:
+
+* `default` features/packages of A are provisioned
+* `community` and `default` features/packages of B are provisioned
+* `experimental`, `preview`, `community` and `default` features/packages of C are provisioned
+
+Example of a feature-pack containing packages and features at a lower stability level than its default stability levels:
+
+* The feature-pack A contains features at `default`, `community` and `preview` level.
+* The feature-pack A contains packages at `default`, `community` and `preview` level.
+* The default package stability level is `community`.
+* The default community stability level is `community`.
+* The minimum stability level is set to `preview`.
+
+If no stability level is specified:
+
+* `default` and `community` features/packages of A are provisioned
+
+If the package and config stability level `preview` is specified:
+
+* `default`, `community` and `preview` features/packages of A are provisioned
+
+==== Lowest config stability level needed by WildFly Galleon Plugins
+
+At provisioning time, when WildFly Galleon plugin generates the server configurations, 
+a WildFly embedded server is started. This server is started with a stability level computed in the following way:
+
+* If a config stability level has been specified by the user, this level is used.
+* If no config stability level has been specified, the lowest config stability level of all feature-packs present in the provisioning is used.
+
+=== Provisioning tooling requirement
+
+The following Galleon based tooling should let users specify the `config-stability-level` and `package-stability-level` options:
+
+* Galleon CLI and Maven plugin
+* WildFly Maven Plugin
+* WildFly Bootable JAR Maven Plugin
+* WildFly Glow
+
+* In addition, for WildFly Glow, information about the features/packages that would be not provisioned for a given stability level 
+is displayed to the user.
+
+=== Nice-to-Have Requirements
+
+* Support for stability in Prospero provisioning tooling.
+
+=== Non-Requirements
+
+* Recording and utilizing stability information in Galleon configuration elements 
+at a higher granularity level than a feature-spec or package (i.e. configs, layers, feature-groups).
+
+* If a provisioning plan (i.e. a provisioning.xml or other provisioning tooling configuration that instructs the tooling about what to provision) 
+involves configs, layers of feature-groups that reference features or packages that are not available at the stability level in effect, 
+the result is not specified. That is:
+** Always rejecting provisioning such a server is not required.
+** Always provisioning a working server is not required.
+** Best practice for feature pack authors is to not include features or packages at a lower stability level in configs, 
+layers or feature-groups that are expected to be used in servers provisioned at a higher stability level. 
+However it is not the responsibility of the Galleon tooling to enforce this best practice or to guarantee a particular result 
+if best practice is not followed.
+
+== Future Works
+
+* An attempt could be made to identify, at feature-pack build time, the Galleon content that 
+references packages below the minimal stability level (so not included) and abort.
+
+== Backwards Compatibility
+
+* Galleon must be backward compatible with older WildFly releases not supporting stability.
+
+== Test Plan
+
+* Add new unit tests to Galleon project.
+* Add new tests to the Galleon Plugins project.
+** Define 4 subsystems to cover the 4 known stability levels.
+** Define a set of packages, modules and features at the 4 known stability levels in a common set.
+** Defines 4 feature-packs for a given minimum stability level. Each feature-pack includes resources from the common set, 
+so filtering is applied when building the feature-pack to exclude content at a lower stability level.
+** Test that the feature-pack content must not contain any content at a lower stability level.
+** Test that the provisioned content must be compliant with the provisioning time stability level (if provided, otherwise must be compliant with the feature-pack defined stability level(s)).
+
+== Community Documentation
+
+* Add documentation to Galleon and WildFly Glow.
+
+* Add documentation to WildFly:
+https://docs.wildfly.org/31/Bootable_Guide.html[Bootable Guide] and 
+https://docs.wildfly.org/31/Galleon_Guide.html[Galleon Guide]
+
+* https://docs.wildfly.org/wildfly-maven-plugin/releases/4.2/provision-mojo.html#galleonOptions[WildFly Maven Plugin documentation] 
+should be enhanced with a link to Galleon and WF Galleon Plugin doc that enumerates the options.
+


wf-galleon/WFLY-19021-Stability_In_Provisioning.adoc
```diff
@@ -0,0 +1,165 @@
...
+* XXX
+
+=== Testing By
+// Put an x in the relevant field to indicate if testing will be done by Engineering or QE. 
+// Discuss with QE during the Kickoff state to decide this
+* [x] Engineering
+
+* [ ] QE
+
+=== Affected Projects or Components
```
spyrkob, line 43: Should we expose this option in Prospero as well?

wf-galleon/WFLY-19021-Stability_In_Provisioning.adoc
```diff
@@ -0,0 +1,165 @@
...
+* In addition, for WildFly Glow, information about the features/packages that would be not provisioned for a given stability level 
+is displayed to the user.
+
+=== Nice-to-Have Requirements
+
+* NONE
+
+=== Non-Requirements
+
+* NONE
```
bstansberry, line 153: We need a non-requirement related to stability levels for Galleon constructs that are not directly tied to management model metadata and JBoss Modules modules. Perhaps

* Recording and utilizing stability information in Galleon elements (e.g. configs, layers, feature-groups, features, packages) that are not generated by reading the server management model metadata (i.e. the 'read-feature-description' management operation) or reading 'module.xml' files that are made available during a feature pack build.'

This may be too broad though, as I suspect manually adding stability metadata to a feature-spec.xml or package.xml would work fine. That is I assume when we generate feature-spec.xml and package.xml we are recording stability metadata, so there's no reason humans couldn't add it for non-generated files.

In that case, perhaps:

* Recording and utilizing stability information in Galleon configuration elements at a higher granularity level than a feature-spec (i.e. configs, layers, feature-groups).

wf-galleon/WFLY-19021-Stability_In_Provisioning.adoc
```diff
@@ -0,0 +1,165 @@
...
+
+=== Requirements at feature-pack build time
+
+==== feature-pack minimum stability level
+
+A minimum stability level can be set at feature-pack build time. If no stability level is specified, the `default` level is used.
+
+Stability level is set thanks to a new option of the WildFly Galleon Plugins Maven plugin: 
+`<minimum-stability>default|community|preview|experimental</minimum-stability>`
+
```
bstansberry, line 75: We need something like:

* Feature-specs and packages at a lower stability level than is specified will not be included in the feature pack.

Two questions fall out of that though. (Note: see my comment about 'Non-requirements', which relates to these):

1) What about manually defined (i.e. not generated) feature specs and packages? Will they be excluded? It seems like the answer should be 'yes', and if it is, then what I said above covers it. If it's not 'yes' then we need to clarify the hard requirement, and add a nice-to-have or non-requirement and probably a bit about future work.

2) How will we handle feature-groups, layers, configs that include feature-specs/packages at a lower stability level? Exclude from build? Fail the build? Include and warn? Do nothing and fail at provisioning time?

wf-galleon/WFLY-19021-Stability_In_Provisioning.adoc
```diff
@@ -0,0 +1,165 @@
+= WildFly provisioning to support WildFly stability
```
bstansberry, line 1: Please add Jekyll front matter at the start of this file so it will show up properly categorized in docs.widlfly.org/wildfly-proposals.  See (almost) any merged proposal for format. See _data/wildfly-categories.yaml for the categories.

wf-galleon/WFLY-19021-Stability_In_Provisioning.adoc
```diff
@@ -0,0 +1,165 @@
...
+==== feature-pack minimum stability level
+
+A minimum stability level can be set at feature-pack build time. If no stability level is specified, the `default` level is used.
+
+Stability level is set thanks to a new option of the WildFly Galleon Plugins Maven plugin: 
+`<minimum-stability>default|community|preview|experimental</minimum-stability>`
+
+==== JBoss Modules modules stability level
+
+A stability level can be set in `module.xml` thanks to the property `org.jboss.stability`.
```
bstansberry, line 78: I'm not sure about the property name. :-)

I don't think the 'org' adds much, and we already have 'jboss.api'.

I'm somewhat tempted to use 'wildfly' instead of 'jboss'. This isn't really a downstream concept, and 'wildfly' better emphasizes that. Maybe someday it will become one though.

wf-galleon/WFLY-19021-Stability_In_Provisioning.adoc
```diff
@@ -0,0 +1,165 @@
...
+
+Expected minimum stability levels for WildFly feature-packs:
+
+* wildfly-ee-galleon-pack: `experimental`
+* wildfly-galleon-pack: `experimental`
+* wildfly-preview-feature-pack: `experimental`
+
+=== Requirements at provisioning time
+
+Provisioning time follows some rules to properly handle stability and avoid features/packages of wrong stability to be provisioned.
```
bstansberry, line 98: Nit: s/to be/being/g

wf-galleon/WFLY-19021-Stability_In_Provisioning.adoc
```diff
@@ -0,0 +1,165 @@
...
+
+* The feature-pack contained minimum stability level is used to constrain the set of provisioned features. Any features/packages at a lower stability level contained in the 
+feature-pack (if any), are not provisioned.
+* By default, the minimum stability level of each feature-pack is used when provisioning its own content. 
+* Each feature-pack present in the provisioning configuration can have a different minimum stability level. 
+The minimum stability of each feature-pack applies to its own content and doesn't leak into other feature-packs.
+* A user can specify a stability-level, by the mean of the `stability-level` Galleon option.
```
bstansberry, line 105: Nit: s/the mean/mean/g

wf-galleon/WFLY-19021-Stability_In_Provisioning.adoc
bstansberry: Per the first bullet in the 'Provisioning tooling requirement' hard requirements section, yes. But let's chat about this.

wf-galleon/WFLY-19021-Stability_In_Provisioning.adoc
```diff
@@ -0,0 +1,165 @@
...
+
+At provisioning time, when WildFly Galleon plugin generates the server configurations, 
+a WildFly embedded server is started. This server is started with a stability level computed in the following way:
+
+* If a stability level has been specified by the user, this level is used.
+* If no stability level has been specified, the lowest minimum stability level of all feature-packs present in the provisioning is used.
+
+=== Provisioning tooling requirement
+
+* All Galleon based tooling should let users specify the `stability-level` option.
```
bstansberry, line 143: If we keep this line as is, the 'Affected Projects or Components' is incomplete.

wf-galleon/WFLY-19021-Stability_In_Provisioning.adoc
```diff
@@ -0,0 +1,165 @@
...
+
+* Galleon must be backward compatible with older WildFly releases not supporting stability.
+
+== Test Plan
+
+* Add new tests to Galleon project.
+
+== Community Documentation
+
+* Add documentation to Galleon and WildFly Glow.
```
bstansberry, line 165: and to:

https://docs.wildfly.org/31/Bootable_Guide.html
https://docs.wildfly.org/31/Galleon_Guide.html

Also, the next version of https://docs.wildfly.org/wildfly-maven-plugin/releases/4.2/provision-mojo.html#galleonOptions should be enhanced with a link to some Galleon or WF Galleon Plugin doc that enumerates the options.

wf-galleon/WFLY-19021-Stability_In_Provisioning.adoc
```diff
@@ -0,0 +1,165 @@
...
+
+* NONE
+
+== Backwards Compatibility
+
+* Galleon must be backward compatible with older WildFly releases not supporting stability.
+
+== Test Plan
+
+* Add new tests to Galleon project.
```
bstansberry, line 161: We're going to need to work out what the requirements around this are in the new community process.

wf-galleon/WFLY-19021-Stability_In_Provisioning.adoc
```diff
@@ -0,0 +1,165 @@
...
+
+* https://issues.redhat.com/browse/WFLY-19021[WFLY-19021] - WildFly provisioning to support WildFly stability
+
+=== Related Issues
+
+* https://issues.redhat.com/browse/WFCORE-6668[WFCORE-6668] - Ensure read-feature-description results contain stability as reported by associated description for use by WildFly Glow
+* https://issues.redhat.com/browse/WFGP-265[WFGP-265] - Support setting a minimum stability level when provisioning
+* https://issues.redhat.com/browse/WFGP-264[WFGP-264] - Support setting a minimum stability level when building a feature pack
+* https://issues.redhat.com/browse/GAL-357[GAL-357] - Support for feature/package stability
+
```
bstansberry, line 26: Please add a 'Stability Level' section as illustrated in https://raw.githubusercontent.com/wildfly/wildfly-proposals/main/design-doc-template.adoc.

You need to pick a level, but I'm pretty sure it has to be 'Community'. Definitely not 'Default'. I suppose we could say 'Preview', but it feels a bit odd to use 'Preview' functionality to generate fps and provision servers at a higher level. It's not formally wrong though.

'Community' will come with stricter requirements, e.g. around testing.

bstansberry: Whatever you pick, please preface the line 1 heading and the PR title with it, in braces, e.g. 

[Community] WildFly provisioning to support WildFly stability

Proposals now should all be about doing X at level Y and the top level heading should declare Y in a standard way.

wf-galleon/WFLY-19021-Stability_In_Provisioning.adoc
jfdenise: @bstansberry , for your point 1) it is yes. I will add what you are suggesting. For point 2) at provisioning time, the features/packages are excluded from top level construct (layers, configs, feature-groups). A trace is displayed. So a layer referencing some lower level stability features would still be possibly valid. It could bring no features or just a subset of its features.

wf-galleon/WFLY-19021-Stability_In_Provisioning.adoc
jfdenise: @bstansberry , ok for `jboss.stability`.

wf-galleon/WFLY-19021-Stability_In_Provisioning.adoc
jfdenise: I added prospero and tooling that depend on Galleon.

wf-galleon/WFLY-19021-Stability_In_Provisioning.adoc
jfdenise: @bstansberry , thanks, I added your suggestion in the Non Requirements

wf-galleon/WFLY-19021-Stability_In_Provisioning.adoc
jfdenise: Done.

wf-galleon/WFLY-19021-Stability_In_Provisioning.adoc
jfdenise: Thanks, I added this set of links.

wf-galleon/WFLY-19021-Stability_In_Provisioning.adoc
jfdenise: Done.

wf-galleon/WFLY-19021-Stability_In_Provisioning.adoc
```diff
@@ -0,0 +1,194 @@
...
+=== Requirements at provisioning time
+
+Provisioning time follows some rules to properly handle stability and avoid features/packages of wrong stability being provisioned.
+
+* The feature-pack contained minimum stability level is used to constrain the set of provisioned features. Any features/packages at a lower stability level contained in the 
+feature-pack (if any), are not provisioned.
+* By default, the minimum stability level of each feature-pack is used when provisioning its own content. 
+* Each feature-pack present in the provisioning configuration can have a different minimum stability level. 
+The minimum stability of each feature-pack applies to its own content and doesn't leak into other feature-packs.
+* A user can specify a stability-level, by mean of the `stability-level` Galleon option.
```
bstansberry, line 125: Oops, my previous requested change here had it's own typo!

s/mean/means

wf-galleon/WFLY-19021-Stability_In_Provisioning.adoc
```diff
@@ -0,0 +1,194 @@
...
+
+At provisioning time, when WildFly Galleon plugin generates the server configurations, 
+a WildFly embedded server is started. This server is started with a stability level computed in the following way:
+
+* If a stability level has been specified by the user, this level is used.
+* If no stability level has been specified, the lowest minimum stability level of all feature-packs present in the provisioning is used.
+
+=== Provisioning tooling requirement
+
+* All Galleon based tooling should let users specify the `stability-level` option.
```
bstansberry, line 163: @jfdenise @jamezp @spyrkob is this achievable by March 20, in time for the WF 32 feature freeze?

If it is not, or is doubtful, for some, but not all, tooling we should enumerate the "must have" tooling here and list others in "Nice-to-Have" or "Non-Requirements". If there are Non-Requirements (or later we realize some Nice-to-Have won't happen) then I like the practice of adding a "Future Work" section to give a hint at what may be coming later.

wf-galleon/WFLY-19021-Stability_In_Provisioning.adoc
```diff
@@ -0,0 +1,194 @@
...
+is displayed to the user.
+
+=== Nice-to-Have Requirements
+
+* NONE
+
+=== Non-Requirements
+
+* Recording and utilizing stability information in Galleon configuration elements 
+at a higher granularity level than a feature-spec or package (i.e. configs, layers, feature-groups).
```
bstansberry, line 174: @jfdenise You somewhat cover this with your use of 'utilizing', but, given your comment about 2) at https://github.com/wildfly/wildfly-proposals/pull/555/files#r1502690469 above, I suggest adding:

* If a provisioning plan (i.e. a provisioning.xml or other provisioning tooling configuration that instructs the tooling about what to provision) involves configs, layers of feature-groups that reference features or packages that are not available at the stability level in effect, the result is not specified. That is:
** Always rejecting provisioning such a server is not required.
** Always provisioning a working server is not required.
** Best practice for feature pack authors is to not include features or packages at a lower stability level in configs, layers or feature-groups that are expected to be used in servers provisioned at a higher stability level. However it is not the responsibility of the Galleon tooling to enforce this best practice or to guarantee a particular result if best practice is not followed.

We could say more about how the features/packages that can be provisioned at the given stability level will be provisioned, and then the overall provisioning and the resulting installation may or may not work. But since the end result is it may or may not work, I don't think we should discuss how 'features/packages that can be provisioned at the given stability level will be'. At least not in this iteration.

wf-galleon/WFLY-19021-Stability_In_Provisioning.adoc
bstansberry: @jfdenise This is the key remaining bit.

wf-galleon/WFLY-19021-Stability_In_Provisioning.adoc
jfdenise: Fixed.

wf-galleon/WFLY-19021-Stability_In_Provisioning.adoc
jfdenise: I listed the tools I know that will be ok for WF32. These tools already switched to Galleon 6. So the support for stability is a mater of bumping the Galleon version to the latest 6 release (that also supports upward compat).
Prospero is still using Galleon 5.x, so we should port the feature to Galleoin 5.x or have prospero to move to Galleon 6 (There is a wip branch for it).

wf-galleon/WFLY-19021-Stability_In_Provisioning.adoc
jfdenise: Great phrasing. Included!

wf-galleon/WFLY-19021-Stability_In_Provisioning.adoc
```diff
@@ -0,0 +1,210 @@
...
+=== Issue
+
+* https://issues.redhat.com/browse/WFLY-19021[WFLY-19021] - WildFly provisioning to support WildFly stability
+
+=== Related Issues
+
+* https://issues.redhat.com/browse/WFCORE-6668[WFCORE-6668] - Ensure read-feature-description results contain stability as reported by associated description for use by WildFly Glow
+* https://issues.redhat.com/browse/WFGP-265[WFGP-265] - Support setting a minimum stability level when provisioning
+* https://issues.redhat.com/browse/WFGP-264[WFGP-264] - Support setting a minimum stability level when building a feature pack
+* https://issues.redhat.com/browse/GAL-357[GAL-357] - Support for feature/package stability
```
spyrkob, line 29: Should the Prospero issue be added here as well? (https://github.com/wildfly-extras/prospero/issues/570)

jfdenise: @spyrkob , I added it.

wf-galleon/WFLY-19021-Stability_In_Provisioning.adoc
jfdenise: @bstansberry , I linked the PR that introduces the stability testing in Galleon plugin project.

wf-galleon/WFLY-19021-Stability_In_Provisioning.adoc
```diff
@@ -0,0 +1,252 @@
+---
+categories:
+  - wf-galleon
```
bstansberry, line 3: @jfdenise This needs to be category defined in https://github.com/wildfly/wildfly-proposals/blob/main/_data/wildfly-categories.yaml.

Please have a chat with @jmesnil about what category we should use for this. Unless I missed something, it looks like we need a new one. But if we do, let's pick one that's both reasonably intuitive and can fit with likely future proposals.

jfdenise: @bstansberry , it seems that it already exists: https://github.com/wildfly/wildfly-proposals/blob/main/_data/wildfly-categories.yaml#L136

wf-galleon/WFLY-19021-Stability_In_Provisioning.adoc
```diff
@@ -0,0 +1,252 @@
...
+* The feature-pack contained config stability level is used to constrain the set of provisioned features. Any features at a lower stability level contained in the 
+feature-pack (if any), are not provisioned.
+* The feature-pack contained package stability level is used to constrain the set of provisioned packages. Any packages at a lower stability level contained in the 
+feature-pack (if any), are not provisioned.
+* By default, the config and package stability levels of each feature-pack are used when provisioning its own content. 
+* Each feature-pack present in the provisioning configuration can have a different config stability level. 
+The config stability of each feature-pack applies to its own content and doesn't leak into other feature-packs.
+* Each feature-pack present in the provisioning configuration can have a different package stability level. 
+The package stability of each feature-pack applies to its own content and doesn't leak into other feature-packs.
+* A user can specify a config stability-level, by means of the `config-stability-level` Galleon option.
```
jfdenise, line 164: @spyrkob , FYI, we are changing stability-level by 2 options: config-stability-level and package-stability-level.

spyrkob: Thanks, created https://github.com/wildfly-extras/prospero/issues/600 for this

wf-galleon/WFLY-19021-Stability_In_Provisioning.adoc
```diff
@@ -0,0 +1,252 @@
...
+
+* [x] OpenShift s2i
+
+* [x] Bootable jar
+
+== Requirements
+
+Stability becomes a Galleon notion that applies to feature-pack, feature/feature_param and package.
+There is a 1 to 1 mapping with WildFly stability.
+The stability levels are: `default` < `community` < `preview` < `experimental`
```
kabir, line 84: Is default the highest? In which case I think it should be >?

bstansberry (changes requested):

wf-galleon/WFLY-19021-Stability_In_Provisioning.adoc
bstansberry: oops; sorry for the noise!

wf-galleon/WFLY-19021-Stability_In_Provisioning.adoc
jfdenise: I will add that < means "enables"

wf-galleon/WFLY-19021-Stability_In_Provisioning.adoc
```diff
@@ -0,0 +1,280 @@
...
+`<minimum-stability>default|community|preview|experimental</minimum-stability>`
+
+Feature-specs and packages at a lower stability level than is specified will not be included in the feature pack.
+
+==== feature-pack provisioning stability level
+
+This level is used as the default stability level to be used when provisioning server configurations and packages.
+If no stability level is specified, the `default` level is used for both configs and packages stability levels. 
+It is set thanks to a new option of the WildFly Galleon Plugins Maven plugin: 
+`<stability>default|community|preview|experimental</stability>`.
```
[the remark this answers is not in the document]
spyrkob: nevermind, I see those are different - does it mean there are 3 different provisioning options: `stability` `config-stability` and `package-stability` that need to be supported by tooling?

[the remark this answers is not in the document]
jfdenise: @spyrkob , yes, at provisioning time you have: stability-level, config-stability-level and package-stability-level.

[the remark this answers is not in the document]
bstansberry: @jfdenise The doc says 'stability' but in your last comment you say 'stability-level'.

wf-galleon/WFLY-19021-Stability_In_Provisioning.adoc
```diff
@@ -0,0 +1,280 @@
...
+== Requirements
+
+Stability becomes a Galleon notion that applies to feature-pack, feature/feature_param and package.
+There is a 1 to 1 mapping with WildFly stability.
+The stability levels are (`<` means enables): `default` < `community` < `preview` < `experimental`
+
+Galleon feature-pack descriptor is extended with 2 new attributes:
+
+* `config-stability`: The stability to be used when provisioning configurations.
+* `package-stability`: The stability to be used when provisioning packages.
```
bstansberry, line 91: I think these should have '-level' appended for consistency with the provisioning attributes.

Note that this means you'll need to support both with and without "-level" for a brief period, long enough to update any FPs that have adapted the "without -level" params.

bstansberry (changes requested):

wf-galleon/WFLY-19021-Stability_In_Provisioning.adoc
```diff
@@ -0,0 +1,280 @@
...
+* Add new unit tests to Galleon project.
+* Add new tests to the Galleon Plugins project. Tests https://github.com/wildfly/galleon-plugins/pull/281[Pull Request] 
+
+== Community Documentation
+
+* Add documentation to Galleon and WildFly Glow.
+
+* Add documentation to WildFly:
+https://docs.wildfly.org/31/Bootable_Guide.html[Bootable Guide] and 
+https://docs.wildfly.org/31/Galleon_Guide.html[Galleon Guide]
```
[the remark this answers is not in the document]
jfdenise: @bstansberry I used stability, config-stability, package-stability at feature-pack build time. At provisioning time it is stability-level, config-stability-level and package-stability-level. It seems that I should align to the xxx-level for all, WDYT? I noticed another typo in the proposal related to this naming.

wf-galleon/WFLY-19021-Stability_In_Provisioning.adoc
```diff
@@ -0,0 +1,280 @@
...
+==== feature-pack provisioning stability level
+
+This level is used as the default stability level to be used when provisioning server configurations and packages.
+If no stability level is specified, the `default` level is used for both configs and packages stability levels. 
+It is set thanks to a new option of the WildFly Galleon Plugins Maven plugin: 
+`<stability>default|community|preview|experimental</stability>`.
+
+[NOTE]
+====
+If one wants to have different values for configs and packages, then `<config-stability>` and `<package-stability>` are to be used instead of `<stability>`.
```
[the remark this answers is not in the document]
jfdenise: @bstansberry  I used stability, config-stability, package-stability at feature-pack build time. At provisioning time it is stability-level, config-stability-level and package-stability-level. It seems that I should align to the xxx-level for all, WDYT? I noticed another typo in the proposal related to this naming.

jfdenise:
@bstansberry , thank-you. I incorporated your feedback

wf-galleon/WFLY-19021-Stability_In_Provisioning.adoc
```diff
@@ -0,0 +1,294 @@
...
+
+== Requirements
+
+Stability becomes a Galleon notion that applies to feature-pack, feature/feature_param and package.
+There is a 1 to 1 mapping with WildFly stability.
+The stability levels are (`<` means enables): `default` < `community` < `preview` < `experimental`
+
+Galleon feature-pack descriptor is extended with 2 new attributes:
+
+* `config-stability-level`: The stability to be used when provisioning configurations.
```
jfdenise, line 90: @bstansberry , checking the implementation, it appears that config-stability and package-stability was a typo in the analysis, the XML descriptor already contains config-stability-level and package-stability-level.

wf-galleon/WFLY-19021-Stability_In_Provisioning.adoc
```diff
@@ -0,0 +1,280 @@
...
+
+* [x] OpenShift s2i
+
+* [x] Bootable jar
+
+== Requirements
+
+Stability becomes a Galleon notion that applies to feature-pack, feature/feature_param and package.
+There is a 1 to 1 mapping with WildFly stability.
+The stability levels are (`<` means enables): `default` < `community` < `preview` < `experimental`
```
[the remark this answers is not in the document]
jfdenise: Ok.

wf-galleon/WFLY-19021-Stability_In_Provisioning.adoc
```diff
@@ -0,0 +1,280 @@
...
+* The feature-pack contained package stability level is used to constrain the set of provisioned packages. Any packages at a lower stability level contained in the 
+feature-pack (if any), are not provisioned.
+* By default, the config and package stability levels of each feature-pack are used when provisioning its own content. 
+* Each feature-pack present in the provisioning configuration can have a different config stability level. 
+The config stability of each feature-pack applies to its own content and doesn't leak into other feature-packs.
+* Each feature-pack present in the provisioning configuration can have a different package stability level. 
+The package stability of each feature-pack applies to its own content and doesn't leak into other feature-packs.
+* A user can specify a config stability level, by means of the `config-stability-level` Galleon option. NB: This option can't be set when `stability-level` is set.
+This stability level is used to constrain the stability level of all features present in configurations generated during provisioning. If the specified level is lower than a feature-pack config stability, 
+the feature-pack minimum stability is used.
```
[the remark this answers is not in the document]
jfdenise: Ok

bstansberry (approved):

wf-galleon/WFLY-19021-Stability_In_Provisioning.adoc
```diff
@@ -0,0 +1,294 @@
...
+`<minimum-stability-level>default|community|preview|experimental</minimum-stability-level>`
+
+Feature-specs and packages at a lower stability level than is specified will not be included in the feature pack.
+
+==== feature-pack provisioning stability level
+
+This level is used as the default stability level to be used when provisioning server configurations and packages.
+If no stability level is specified, the `default` level is used for both configs and packages stability levels. 
+It is set thanks to a new option of the WildFly Galleon Plugins Maven plugin: 
+`<stability-level>default|community|preview|experimental</stability-level>`.
```
bstansberry, line 111: Perhaps add a note:

If the specified stability level is not enabled by the level specified by the 'minimum-stability-level' setting, the build of the feature pack will fail.

wf-galleon/WFLY-19021-Stability_In_Provisioning.adoc
```diff
@@ -0,0 +1,294 @@
...
+while allowing provisioning of packages at a lower level. 
+The presence of the lower stability level packages allows subsequent update of the configuration to enable lower stability features.
+====
+
+==== feature-pack provisioning config stability level
+
+This level is used as the default stability level to be used when provisioning server configurations.
+If no stability level is specified at feature-pack build time, the `default` level is used.
+Stability level is set thanks to a new option of the WildFly Galleon Plugins Maven plugin: 
+`<config-stability-level>default|community|preview|experimental</config-stability-level>`
```
bstansberry, line 127: Perhaps add a note:

If the specified stability level is not enabled by the level specified by the 'minimum-stability-level' setting, the build of the feature pack will fail.

wf-galleon/WFLY-19021-Stability_In_Provisioning.adoc
```diff
@@ -0,0 +1,294 @@
...
+
+This level can be overridden at provisioning time with the Galleon option `<config-stability-level>`
+
+==== feature-pack provisioning package stability level
+
+This level is used as the default stability level to be used when provisioning server packages (e.g.: JBoss Modules modules). 
+If no stability level is specified, the `default` level is used.
+Stability level is set thanks to a new option of the WildFly Galleon Plugins Maven plugin: 
+`<package-stability-level>default|community|preview|experimental</package-stability-level>`
```
bstansberry, line 136: Perhaps add a note:

If the specified stability level is not enabled by the level specified by the 'minimum-stability-level' setting, the build of the feature pack will fail.

wf-galleon/WFLY-19021-Stability_In_Provisioning.adoc
bstansberry: Or add a single note after the L140-143 note that covers all the cases I note here and in the next couple sections.

wf-galleon/WFLY-19021-Stability_In_Provisioning.adoc
```diff
@@ -0,0 +1,294 @@
...
+This stability level is used to constrain the stability level of all provisioned packages. NB: This option can't be set when `stability-level` is set. 
+If the specified level is lower than a feature-pack package stability, 
+the feature-pack package stability level is used.
+* A user can specify a stability level for both configs and packages, by means of the `stability-level` Galleon option. This option is a convenience 
+allowing to set both the config and package stability level using a single option. It can't be set when `config-stability-level` or `package-stability-level` 
+are in use.
+* At provisioning time, usage of `<stability-level>` and `<config|package-stability-level>` options in the same provisioning execution 
+is not supported and would lead to a failure.
+
+For example, a provisioning configuration containing 3 feature-packs:
```
bstansberry, line 206: I think this example needs to cover something like the 'provision a preview config from a wildfly-galleon-pack' case. The example right now has no FPs with different values for the 3 stability settings, so the scenario of wanting to provision a lower config stability than the FP's setting is not described. I think it should be supported and that's why I filed https://issues.redhat.com/browse/GAL-361, but reading the spec I see it's not explicitly stated it's supported.  It's implied by statements like 'This level can be overridden at provisioning time with the Galleon option `<config-stability-level>`' but 'overridden' doesn't have to mean it can overridden to any value.

bstansberry (changes requested):

wf-galleon/WFLY-19021-Stability_In_Provisioning.adoc
jfdenise: @bstansberry , I added a clear example and specified that the default are not minimal that constrain provisioning.

wf-galleon/WFLY-19021-Stability_In_Provisioning.adoc
jfdenise: Ok, added a note.

wf-galleon/WFLY-19021-Stability_In_Provisioning.adoc
jfdenise: Ok, added a note.

wf-galleon/WFLY-19021-Stability_In_Provisioning.adoc
jfdenise: Ok, added a note.

wf-galleon/WFLY-19021-Stability_In_Provisioning.adoc
```diff
@@ -0,0 +1,313 @@
...
+
+* Recording and utilizing stability information in Galleon configuration elements 
+at a higher granularity level than a feature-spec or package (i.e. configs, layers, feature-groups).
+
+* If a provisioning plan (i.e. a provisioning.xml or other provisioning tooling configuration that instructs the tooling about what to provision) 
+involves configs, layers of feature-groups that reference features or packages that are not available at the stability level in effect, 
+the result is not specified. That is:
+** Always rejecting provisioning such a server is not required.
+** Always provisioning a working server is not required.
+** Best practice for feature pack authors is to not include features or packages at a lower stability level in configs,
```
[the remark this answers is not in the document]
jfdenise: @yersan , it is currently not supported.  We could make it strict. If we detect that a referenced package/feature is unknown, we could fail at feature-pack build time. @bstansberry would you be ok that we make it strict as a future enhancement of this proposal?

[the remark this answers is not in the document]
bstansberry: @jfdenise Sure. Any 'future work' section is just suggestions anyway, not a commitment, so if you think it's a reasonable thing to do, go for it.

BTW I suggested the language here about not guaranteeing any behavior, but that was because I wanted to limit the immediate scope for this iteration while also being clear about the topic in the analysis (i.e. don't let readers assume anything.) I don't have any particular feelings about what should be done later.

That said, AIUI WFGP-278 isn't really about what I was thinking about for this. My impression there is there's some optional module.xml dependency that is resulting in a provisioning failure, and I'd expect that to just work. But we should discuss the details of that particular case elsewhere.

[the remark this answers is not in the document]
jfdenise: I added a section.

jfdenise:
The community doc is composed of:
* https://github.com/wildfly/galleon/pull/345
* https://github.com/wildfly/wildfly-maven-plugin/pull/508
* https://github.com/wildfly/wildfly/pull/17801

bstansberry (approved):

PR #556, opened 2024-02-28
[WFLY-18749] Deploying a jar with unsatisfied dependencies does not w…

ChristinaDsl:


darranl (changes requested):
FYI you have edited the design doc template directly for your proposal, when working with this repository the intent is you take a copy if this file into one of the other directories and make your changes there,

PR #557, opened 2024-03-12
[WFCORE-6728] Ability to reload tests to a different stability level

kabir:
https://issues.redhat.com/browse/WFCORE-6728

### DIFF
--- a/server/WFCORE-6728-reload-to-stability-level.adoc
+++ b/server/WFCORE-6728-reload-to-stability-level.adoc
@@ -1,10 +1,11 @@
 ---
 categories:
   - core
+  - management
 # Add any category for this proposal
 # if missing, add it to _data/widfly-categories and use its id
 ---
-= Ability to reload a server to a different stability level
+= [Community] Ability to reload a server to a different stability level
 :author:            Kabir Khan
 :email:             kkhan@redhat.com
 :toc:               left
@@ -44,6 +45,9 @@ This proposal attempts to solve that by providing a set of ServerSetupTasks that
 === Dev Contacts
 
 * mailto:{email}[{author}]
+* mailto:brian.stansberry@redhat.com[Brian Stansberry]
+* mailto:yborgess@redhat.com[Yeray Borges Santana]
+* mailto:jperkins@redhat.com[James Perkins]
 
 === QE Contacts
 
@@ -80,7 +84,7 @@ WildFly Core. The main changes will be in the server, testsuite/shared and tests
 - The operation will be registered at stability level 'community' at present
 - The operation is based on the reload operation, and adds a parameter called `stability`. This parameter takes the stability level we want to reload the server to
 - The operation also supports all the parameters supported by the standard reload operation
-- The operation has its own RBAC setting to narrow down who can call it
+- The operation has its own sensitive target access constraint setting to narrow down who can call it. It is called `reload-enhanced` and is set to `false` for accessDefault, readDefault and writeDefault.
 * A set of ServerSetupTasks for the WildFly Core test runner will be introduced. These use the above operation to reload the server to the desired stability level before running the tests, and to reload the server back to the original stability level after the tests have completed.
 ** The ServerSetupTasks perform the following checks, and skip the test using JUnit `Assume.assumeXXX()` if the condition is not met:
 *** The desired stability level is one of the ones supported by the server
@@ -100,7 +104,7 @@ WildFly Core. The main changes will be in the server, testsuite/shared and tests
 === Future Work
 // Use this section to discuss requirements that are not addressed by this proposal
 // but which may be addressed in later proposals.
-This is currently only implemented for standalone servers, so there is no domain mode support. In the future it could be nice to be able to reload the host controllers and possibly the managed servers to different stability levels.
+This is currently only implemented for standalone servers, so there is no domain mode support. In the future it could be nice to be able to reload the host controllers to a different stability level.
 
 Also, as this is currently mainly for the testsuite to be able to reload, there is no special support in the CLI's reload command for the enhanced reload operation. If this is seen usable for wider use, adding handling to the CLI command to expose the stability parameter could be interesting. The CLI command would need to take into account both the stability level of the server and the user's RBAC role in order to determine whether the enhanced reload operation is visible to the user. If the operation is visible, and the user wants to use the stability paramter, the enhanced reload operation will be called instead of the current reload operation.
 
@@ -141,7 +145,13 @@ Tests will be added to the testsuite/standalone module to ensure that the Server
 
 == Community Documentation
 
-The only user facing functionality here is the enhanced reload operation. The current intent is to keep this a bit hidden, so no community documentation will be added for this.
+The only user facing functionality here is the `reload-enhanced` operation. We deliberately don't promote direct use of low level reload operations, as any use of those operations requires the caller to properly handle reconnecting to the reloaded server. Our end-user documentation around reloading should focus on the high level CLI reload command, which is out of scope for this proposal.
+
+Additionally, reloading to a different stability level requires great care to make sure the server configuration is compatible with the target stability level.
+
+So, beyond the wildscribe management API documentation no community documentation will be added for this.
+
+
 
 == Release Note Content
 ////
--- a/server/WFCORE-6728-reload-to-stability-level.adoc
+++ b/server/WFCORE-6728-reload-to-stability-level.adoc
@@ -1,0 +1,155 @@
+---
+categories:
+  - core
+# Add any category for this proposal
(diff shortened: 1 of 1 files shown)


server/WFCORE-6728-reload-to-stability-level.adoc
```diff
@@ -0,0 +1,155 @@
...
+== Requirements
+
+=== Hard Requirements
+
+* The WildFly Core test runner is enhanced to handle JUnit `Assume.assumeXXX()` calls from ServerSetupTasks the same way as in normal test code. Prior to this RFE the resulting `AssumptionViolatedException` would result in a stacktrace and test error, rather than the test being marked as skipped.
+* There will be a new operation based on the current reload operation introduced for standalone servers.
+- The operation will be registered at stability level 'community' at present
+- The operation is based on the reload operation, and adds a parameter called `stability`. This parameter takes the stability level we want to reload the server to
+- The operation also supports all the parameters supported by the standard reload operation
+- The operation has its own RBAC setting to narrow down who can call it
```
bstansberry, line 83: Please specify the constraint name and the OOTB setting for this.

The constraint name is kind of an impl detail, but since you know it, it's helpful to reviewers to write it down.

server/WFCORE-6728-reload-to-stability-level.adoc
```diff
@@ -0,0 +1,155 @@
...
+////
+RBAC constraints will be configured on the enhanced reload operation to narrow down who can call it.
+
+== Test Plan
+
+Tests will be added to the testsuite/standalone module to ensure that the ServerSetupTasks mentioned in the requirements section reload properly to the desired stability levels.
+
+== Community Documentation
+
+The only user facing functionality here is the enhanced reload operation. The current intent is to keep this a bit hidden, so no community documentation will be added for this.
```
bstansberry, line 144: s/so no/so, beyond the wildscribe management API documentation, no/g

bstansberry (commented):
@kabir This looks good. I have a couple minor comments.

server/WFCORE-6728-reload-to-stability-level.adoc
```diff
@@ -0,0 +1,155 @@
+---
+categories:
+  - core
```
bstansberry, line 3: This should also have 'management'

server/WFCORE-6728-reload-to-stability-level.adoc
```diff
@@ -0,0 +1,155 @@
...
+
+* [x] Bootable jar
+
+== Requirements
+
+=== Hard Requirements
+
+* The WildFly Core test runner is enhanced to handle JUnit `Assume.assumeXXX()` calls from ServerSetupTasks the same way as in normal test code. Prior to this RFE the resulting `AssumptionViolatedException` would result in a stacktrace and test error, rather than the test being marked as skipped.
+* There will be a new operation based on the current reload operation introduced for standalone servers.
+- The operation will be registered at stability level 'community' at present
```
yersan, line 80: Does it mean it will never be possible to reload to default, because then there won't be a way to back to community level, or would it be managed behind the scenes?

kabir: @yersan You can reload to default (assuming the server config works), but you will not be able to reload back again :-) If you look at the tests in https://github.com/wildfly/wildfly-core/pull/5895 you will see they avoid reloading to default. Also, a bit of trickery is needed with config snapshots to make sure that once the test is done, we can reload back to the original community level

server/WFCORE-6728-reload-to-stability-level.adoc
bstansberry: I suggest:

"The only user facing functionality here is the reload-enhanced *operation*. We deliberately don't promote direct use of low level reload operations, as any use of those operations requires the caller to properly handle reconnecting to the reloaded server. Our end-user documentation around reloading should focus on the high level CLI reload *command*, which is out of scope for this proposal. So, beyond the wildscribe management API documentation no community documentation will be added for this."

server/WFCORE-6728-reload-to-stability-level.adoc
```diff
@@ -0,0 +1,155 @@
...
+
+
+=== Non-Requirements
+// Use this section to explicitly discuss things that readers might think are required
+// but which are not required.
+
+=== Future Work
+// Use this section to discuss requirements that are not addressed by this proposal
+// but which may be addressed in later proposals.
+This is currently only implemented for standalone servers, so there is no domain mode support. In the future it could be nice to be able to reload the host controllers and possibly the managed servers to different stability levels.
```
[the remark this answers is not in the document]
bstansberry: Yeah, best to drop the managed servers bit, as we don't want to mislead readers re future plans.

server/WFCORE-6728-reload-to-stability-level.adoc
bstansberry: @yersan Correct. Unless this operation is promoted to 'default' stability level it is not possible to use it to reload to 'default'.

I'll add a proposed change to 'Non-Requirements' regarding this.

bstansberry: I'm going to delete my last comment ^^^ as it's about the ServerSetupTask impls that come with this RFE, not the operation itself.

server/WFCORE-6728-reload-to-stability-level.adoc
```diff
@@ -0,0 +1,155 @@
...
+
+* [ ] Preview
+
+* [x] Community
+
+* [ ] default
+
+=== Dev Contacts
+
+* mailto:{email}[{author}]
```
bstansberry, line 47: Please add you, me, Yeray (if he doesn't object), James.

This block (or a slightly tweaked variant) of the doc lists the feature team. Assignee (you) , SME(s) (me and Yeray), outside perspective person (James).

server/WFCORE-6728-reload-to-stability-level.adoc
kabir: @bstansberry  Thanks, I also added "Additionally, reloading to a different stability level requires great care to make sure the server configuration is compatible with the target stability level." 

@yersan this touches on our discussion in your other comment

bstansberry (changes requested):
@kabir Just a last couple things and then this is good to go from my POV.

jamezp (commented):
Overall I think this looks good. 

I unfortunately have not had time to fully read back on all the chat so please forgive me if this was already addressed. I don't really understand why a new `reload-enhanced` operation is needed. For Core, and even WildFly Arquillian, we have control over how a server is started/stopped.

One option would be to simply group tests by a JUnit `@Tag`/`@Category` and run those in a server provisioned at a specific stability level. I assume we'll have to provision test servers at all stability levels enabled.

Another option is we add something to the `ServerController` in core and the `WildFlyContainerController` in WildFly Arquillian, for a `start()` option with takes a stability level argument. We probably need to do this regardless or manual mode tests won't work unless you provision multiple serves and define them in a `arquillian.xml` file.

bstansberry:
@jamezp Using a separate operation always us to meet the requirement that an RBAC constraint be applied to doing this.

Changing the stability level of the process is a significantly different act than just reloading it, so it requires distinct RBAC permissions.

The RBAC mechanism doesn't really allow for enforcing different constraints based of the value of params passed when invoking an operation, e.g. presence/absence of the 'stability' parameter. RBAC constraints are applied at the resource, attribute or entire-operation level. Since we need different constraints for the stability change we need a different operation.

jamezp:
Got it, thank you! I thought it was simply something like `standalone.sh --stability=experimental`, or the effective `java` command of that :)

The RBAC support makes sense to me and could be useful for sure.

server/WFCORE-6728-reload-to-stability-level.adoc
yersan: Ok @kabir , got it. It could be controlled in the code and avoid reloading to a level where this Operation does not exist, but it is unnecessary since this is for internal usage. After reloading to default, you can shutdown(restart=true) and you will come back easily. Thanks for clarifying it.

server/WFCORE-6728-reload-to-stability-level.adoc
bstansberry: @kabir I'm going to approve this, but please put yourself in here. :)

bstansberry (approved):

yersan (approved):

server/WFCORE-6728-reload-to-stability-level.adoc
jamezp: That should be done as it's defined in the attributes.

bstansberry:
Thanks @kabir, @yersan and @jamezp

PR #569, opened 2024-04-29
[WFCORE-6750] Analysis document for the Unstable API Annotation scanning

kabir:
https://issues.redhat.com/browse/WFCORE-6750 (the runtime scanner)
https://issues.redhat.com/browse/WFLY-19152 (creating the indices in the feature pack build)

### DIFF
--- a/server/WFCORE-6750-unstable-api-annotation-scanner.adoc
+++ b/server/WFCORE-6750-unstable-api-annotation-scanner.adoc
@@ -85,7 +85,7 @@ Both the two alternative ideas above would need this conversion to strings, sinc
 === Affected Projects or Components
 
 * WildFly
-** the wildfly-galleon-pack and wildfly-ee-feature pack feature packs will scan their classpaths and provide the indices of API elements annotated with `io.smallrye.common.annotation.Experimental` and `org.hibernate.Incubating`.
+** the 'wildfly' and 'wildfly-ee' feature packs will scan their classpaths and provide the indices of API elements annotated with `io.smallrye.common.annotation.Experimental` and `org.hibernate.Incubating`.
 * WildFly Core
 ** scans the applications when they are deployed, looking for usage of API elements that exist in the index.
 * https://github.com/wildfly/unstable-api-annotation-utils[wildfly/unstable-api-annotation-utils] - contains a library providing:
@@ -115,20 +115,20 @@ Feature packs external to WildFly, as mentioned in the 'Future Work' section.
 * There is a maven plugin that can be used in the feature pack builds.
 ** It will scan for annotations indicating unstable API, and add the following to the `content/` directory of the  `{indexModule}` module (the `content/` directory is defined as a resource root:
 *** a file containing the indices for annotations concerning the feature pack.
-*** The name of the file will be added to a file called `index.txt` in the `content/` folder of the `{indexModule}` module. This `index.txt` is provided as the main lookup mechanism for the files mentioned in the above point.
+**** The file may be either a .txt or a .zip
 ** The maven plugin scans the full classpath of the feature pack looking for dependencies to index.
 ** For this iteration we will scan
-*** the `wildfly-ee-feature-pack` for the `org.hibernate.Incubating` annotation.
-*** the `wildfly-galleon-pack` for the `io.smallrye.common.annotation.Experimental` annotation.
+*** the 'wildfly-ee' feature pack for the `org.hibernate.Incubating` annotation.
+*** the 'wildfly' feature pack for the `io.smallrye.common.annotation.Experimental` annotation.
 *** It also contains a simple configuration mechanism to narrow down which classpath entries to index (e.g when scanning for `io.smallrye.common.annotation.Experimental` we only check the SmallRye libraries on the classpath)
 * In WildFly Core
 ** deployment unit processors will be added to:
-*** Load the indices listed in the `content/index.html` file in the `{indexModule}` module.
+*** Iterate all resources from the `{indexModule}` module. These resources are the .zip or .txt files containing the indices for each feature pack
 *** Use the wildfly/unstable-api-annotation-utils library to scan the bytecode of every class in the user's deployed application and cross-reference against the loaded indices.
 ** The deployment unit processors will use the WildFly classloading rules to make sure that in the case of nested archives, that we only scan each class once. E.g. if a .war contains a `WEB-INF/lib/my-library.jar`, classes in `WEB-INF/lib/my-library.jar` will only be s scanned as part of the `WEB-INF/lib/my-library.jar` resource root scan, and not for the parent .war.
 ** Once scanned, usage of annotated API elements may be reported depending on configuration in the `core-management` subsystem:
 *** A new child resource (currently called `service=unstable-api-annotations`) is added with a `level` attribute to configure the reporting
-**** Since this is a resource/feature with a stability level of preview, if the resource is not present, no reporting is done.
+**** Since this is a resource/feature with a stability level of preview, if the resource is not present, no scanning or reporting is done.
 ****** The resource is added to the `preview` flavour of the current version of the model. At the time of writing it I was told there is no need to bump the version of management model constructs coming in lower than `community` level.
 **** If the attribute is set to `log` (this is the default value for the attribute), the usages will be logged to the server.log. In this case the user application will be deployed.
 **** If the attribute is set to `error`, the deployment will fail with an exception containing the usages.
@@ -209,7 +209,7 @@ WildFly Core:
 ** Thus, a test feature pack will be created offering an API (including its own annotations for indicating unstable API) and a subsystem providing this API on the classpath.
 ** The feature pack will be indexed and the indices will be added to the `{indexModule}` module.
 ** The test will check that
-*** the index file is added to the `${indexModule}` module, and to the list in the module's `index.txt`, as outlined in the requirements
+*** the index file is added to the `${indexModule}` module, as outlined in the requirements
 *** when deploying an application using annotated API elements from the test feature pack API, usage of those is reported in line with the configuration in the `subsystem=core-management/service=unstable-api-annotations` resource (as outlined in the requirements). We will test both with
 **** `level=log` - and inspect that the log messages pick up all the usages of annotated API elements.
 **** `level=error` - and inspect that the application fails to deploy, and that the error message picks up all the usages of annotated API elements.
@@ -218,10 +218,11 @@ WildFly Core:
 WildFly:
 
 * Since we have done the main testing of the mechanism in WildFly Core, we will not test that here
-* A test will be added to `testsuite/integration/basic` which runs in an execution provisioning both `wildfly-ee-feature-pack` and `wildfly-galleon-pack`. It will test
-** that the `{indexModule}` module, and the list in the module's `index.txt`, contain the indices from both feature packs.
-** A number of differently packaged deployments will be tested, to make sure that each class only gets scanned once, in order to test that the annotated class/member usage follows the WildFly classloading rules. The scanner outputs the class count when an undocumented system property is set to `true`.
-* The above test will run at the `preview` stability level
+* The verifier plugin in each feature pack will verify that the `{indexModule}` module in each feature pack contains the expected index files.
+** The 'wildfly-ee' feature pack expects `wildfly-ee-feature-pack.zip`
+** The 'wildfly' feature pack expects `wildfly-ee-feature-pack.zip` and `wildfly-galleon-pack.zip`
+* A test will be added to `testsuite/integration/microprofile` which runs in an execution provisioning both the 'wildfly-ee' and 'wildfly' feature packs. It will test a number of differently packaged deployments, to make sure that each class only gets scanned once, in order to test that the annotated class/member usage follows the WildFly classloading rules. The scanner outputs the class count when an undocumented system property is set to `true`.
+** This test will run at the `preview` stability level
 
 == Community Documentation
 ////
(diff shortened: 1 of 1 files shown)


server/WFCORE-6750-unstable-api-annotation-scanner.adoc
```diff
@@ -0,0 +1,243 @@
...
+*** a file containing the indices for annotations concerning the feature pack.
+*** The name of the file will be added to a file called `index.txt` in the `content/` folder of the `{indexModule}` module. This `index.txt` is provided as the main lookup mechanism for the files mentioned in the above point.
+** The maven plugin scans the full classpath of the feature pack looking for dependencies to index.
+** For this iteration we will scan
+*** the `wildfly-ee-feature-pack` for the `org.hibernate.Incubating` annotation.
+*** the `wildfly-galleon-pack` for the `io.smallrye.common.annotation.Experimental` annotation.
+*** It also contains a simple configuration mechanism to narrow down which classpath entries to index (e.g when scanning for `io.smallrye.common.annotation.Experimental` we only check the SmallRye libraries on the classpath)
+* In WildFly Core
+** deployment unit processors will be added to:
+*** Load the indices listed in the `content/index.html` file in the `{indexModule}` module.
```
bstansberry, line 126: s/index.html/index.txt

server/WFCORE-6750-unstable-api-annotation-scanner.adoc
```diff
@@ -0,0 +1,243 @@
...
+*** Use the wildfly/unstable-api-annotation-utils library to scan the bytecode of every class in the user's deployed application and cross-reference against the loaded indices.
+** The deployment unit processors will use the WildFly classloading rules to make sure that in the case of nested archives, that we only scan each class once. E.g. if a .war contains a `WEB-INF/lib/my-library.jar`, classes in `WEB-INF/lib/my-library.jar` will only be s scanned as part of the `WEB-INF/lib/my-library.jar` resource root scan, and not for the parent .war.
+** Once scanned, usage of annotated API elements may be reported depending on configuration in the `core-management` subsystem:
+*** A new child resource (currently called `service=unstable-api-annotations`) is added with a `level` attribute to configure the reporting
+**** Since this is a resource/feature with a stability level of preview, if the resource is not present, no reporting is done.
```
bstansberry, line 131: s/no reporting is done/no scanning or reporting is done/g

server/WFCORE-6750-unstable-api-annotation-scanner.adoc
```diff
@@ -0,0 +1,243 @@
...
+=== Non-Requirements
+// Use this section to explicitly discuss things that readers might think are required
+// but which are not required.
+* Full indexing of all annotations indicating unstable API
+** See the 'Future Work' section for plans to get full coverage
+* Inspection of user classes not in an application deployment, such as:
+** User provided modules
+** Classes provided by the user via the management model with the typical `module`/`class` attribute pairs.
+* No attempt is made to not add the indices if the server is provisioned a a higher stability level than 'preview'. In other words, if the feature packs contain the indices, they will always end up in the server.
+* Since this is a `preview` feature, and the model version is the same as the original with the added `preview` qualifier, model transformation is not needed.
```
bstansberry, line 153: There should be a layer for this. Not having it now can be a 'Non-Requirement' + 'Future Work'.

server/WFCORE-6750-unstable-api-annotation-scanner.adoc
```diff
@@ -0,0 +1,243 @@
...
+
+=== Future Work
+// Use this section to discuss requirements that are not addressed by this proposal
+// but which may be addressed in later proposals.
+* Promote this feature to Community or higher.
+** This would allow us to enable it out of the box by default, making it more useful.
+* Identify all annotations indicating unstable API used by WildFly itself with the help of component leads.
+** Scan for these annotations during feature pack builds, and make them available in a provisioned server.
```
bstansberry, line 161: What would be the goal of this?

server/WFCORE-6750-unstable-api-annotation-scanner.adoc
```diff
@@ -0,0 +1,243 @@
...
+*** The name of the file will be added to a file called `index.txt` in the `content/` folder of the `{indexModule}` module. This `index.txt` is provided as the main lookup mechanism for the files mentioned in the above point.
+** The maven plugin scans the full classpath of the feature pack looking for dependencies to index.
+** For this iteration we will scan
+*** the `wildfly-ee-feature-pack` for the `org.hibernate.Incubating` annotation.
+*** the `wildfly-galleon-pack` for the `io.smallrye.common.annotation.Experimental` annotation.
+*** It also contains a simple configuration mechanism to narrow down which classpath entries to index (e.g when scanning for `io.smallrye.common.annotation.Experimental` we only check the SmallRye libraries on the classpath)
+* In WildFly Core
+** deployment unit processors will be added to:
+*** Load the indices listed in the `content/index.html` file in the `{indexModule}` module.
+*** Use the wildfly/unstable-api-annotation-utils library to scan the bytecode of every class in the user's deployed application and cross-reference against the loaded indices.
```
bstansberry, line 127: This should be provided by a module that is an optional dependency of whatever uses this. That module has property jboss-stability-level=preview.

That likely means the index files should be in a different module from this lib, one that's at default stability, so the plugin execution that adds them doesn't have to know whether the FP build is configured to allow preview stability modules in the FP.

server/WFCORE-6750-unstable-api-annotation-scanner.adoc
```diff
@@ -0,0 +1,243 @@
...
+** Scan for these annotations during feature pack builds, and make them available in a provisioned server.
+* Engage with owners of feature packs external to WildFly, and identify all annotations indicating unstable API in those feature packs.
+** Provide instructions on how to do the build time scan, and to make the indices available at runtime once their feature pack has been provisioned.
+** Scan for these annotations during feature pack build, and make them available in a provisioned server.
+** I think an appropriate set of feature packs would be the ones WildFly Glow is aware of.
+* Index user provided classes
+** in user modules
+** in classes referenced via the `module`/`class` attribute pairs in the management model
+* Add a deployment marker to force enable/disable the runtime scanning independent of the subsystem setting
+** This could possibly be enhanced with a filter so a user could say to ignore usage of constructs annotated with `org.hibernate.Incubating` (or perhaps an expression to ignore classes matching the expression), while reporting usage of everything else (e.g. constructs annotated with `io.smallrye.common.annotation.Experimental`).
```
bstansberry, line 170: If we're going to allow something like this, which involves the user identifying annotations of interest, then having that config do a mapping to a stability level seems good. Then how we treat the presence of the thing depends on the server's stability level.

If we found something that wasn't configured, that would mean 'not allowed' (so log it or error). IOW, not configured != experimental. Someone running a --stability=experimental server can still reject APIs.

server/WFCORE-6750-unstable-api-annotation-scanner.adoc
```diff
@@ -0,0 +1,243 @@
...
+* [x] Bootable jar
+
+== Requirements
+
+=== Hard Requirements
+* Note this is at preview level only!
+* There is a maven plugin that can be used in the feature pack builds.
+** It will scan for annotations indicating unstable API, and add the following to the `content/` directory of the  `{indexModule}` module (the `content/` directory is defined as a resource root:
+*** a file containing the indices for annotations concerning the feature pack.
+*** The name of the file will be added to a file called `index.txt` in the `content/` folder of the `{indexModule}` module. This `index.txt` is provided as the main lookup mechanism for the files mentioned in the above point.
```
bstansberry, line 118: Yuck. I don't like the idea of FPs coordinating changes to a single file.

Perhaps use ModuleClassLoader.iterateResources() or some other API in MCL to remove the need for this index.txt.

server/WFCORE-6750-unstable-api-annotation-scanner.adoc
```diff
@@ -0,0 +1,243 @@
...
+*** the index file is added to the `${indexModule}` module, and to the list in the module's `index.txt`, as outlined in the requirements
+*** when deploying an application using annotated API elements from the test feature pack API, usage of those is reported in line with the configuration in the `subsystem=core-management/service=unstable-api-annotations` resource (as outlined in the requirements). We will test both with
+**** `level=log` - and inspect that the log messages pick up all the usages of annotated API elements.
+**** `level=error` - and inspect that the application fails to deploy, and that the error message picks up all the usages of annotated API elements.
+** The test will run at the `preview` stability level
+
+WildFly:
+
+* Since we have done the main testing of the mechanism in WildFly Core, we will not test that here
+* A test will be added to `testsuite/integration/basic` which runs in an execution provisioning both `wildfly-ee-feature-pack` and `wildfly-galleon-pack`. It will test
```
bstansberry, line 221: This needs to be in testsuite/integration/microprofile (to be renamed 'expansion').

Use of wildfly-galleon-pack is not allowed in testsuite/integration/basic.

bstansberry: Also s/wildfly-ee-feature-pack/wildfly-ee-galleon-pack

or, my minor preference

"both the 'wildfly-ee' and 'wildfly' feature packs."

server/WFCORE-6750-unstable-api-annotation-scanner.adoc
```diff
@@ -0,0 +1,243 @@
...
+** that the `{indexModule}` module, and the list in the module's `index.txt`, contain the indices from both feature packs.
+** A number of differently packaged deployments will be tested, to make sure that each class only gets scanned once, in order to test that the annotated class/member usage follows the WildFly classloading rules. The scanner outputs the class count when an undocumented system property is set to `true`.
+* The above test will run at the `preview` stability level
+
+== Community Documentation
+////
+Generally a feature should have documentation as part of the PR to wildfly master, or as a follow up PR if the feature is in wildfly-core. In some cases though the documentation belongs more in a component, or does not need any documentation. Indicate which of these will happen.
+////
+
+Community documentations will be added as part of the PR to WildFly full. This PR will also contain the parts of the tests which live in WildFly, as mentioned in the previous section
```
bstansberry, line 231: We need to beef this bit up a lot. I think not implementing everything by the 33 feature freeze is ok, but we should think about things like user journeys.

server/WFCORE-6750-unstable-api-annotation-scanner.adoc
kabir: To index similar annotations from other libraries than the current SmallRye and Hibernate families. E.g. I think JGroups has something. Apache might have something etc. i.e. get better coverage

server/WFCORE-6750-unstable-api-annotation-scanner.adoc
kabir: Good point, it will simplify the test since in basic there were two cases
- the full zip which tests everything
- the galleon provisioned server which don't have the wildfly FP

server/WFCORE-6750-unstable-api-annotation-scanner.adoc
kabir: I consider this more brainstorming, than a "must" have. I've updated the start of the future work section to say we will decide the scope when going to community level, and that these points are ideas

server/WFCORE-6750-unstable-api-annotation-scanner.adoc
bstansberry: My question here is more about why we make them available in the provisioned server. But probably I misinterpreted the parent point.

If your meaning in the parent point is just to include more annotations from other components and check for their use in deployment code, I understand. If you're talking about scanning WF code for our own use of things, I understand why we'd do that, I'm just not clear why we'd want to write the result in the provisioned server.

In any case I'm not asking for any change in the document; this is just a discussion of what you had in mind.

server/WFCORE-6750-unstable-api-annotation-scanner.adoc
bstansberry: I misspoke -- it's jboss.stability=preview, for example

https://github.com/wildfly-extras/mvc-krazo-feature-pack/blob/main/galleon-shared/src/main/resources/modules/system/layers/base/jakarta/mvc/api/main/module.xml

For a brief discussion of this see 'JBoss Modules modules stability level' in https://docs.wildfly.org/wildfly-proposals/wf-galleon/WFLY-19021-Stability_In_Provisioning.html#requirements-at-feature-pack-build-time

server/WFCORE-6750-unstable-api-annotation-scanner.adoc
bstansberry: Kabir implemented this idea.

server/WFCORE-6750-unstable-api-annotation-scanner.adoc
```diff
@@ -0,0 +1,248 @@
...
+
+== Release Note Content
+The core-management subsystem now allows you to enable scanning of your deployments for usage of classes/methods in the SmallRye and Hibernate libraries annotated with `org.hibernate.Incubating` and `io.smallrye.common.annotation.Experimental`. These annotations indicate that those API elements are likely to change at any time
```
bstansberry, line 239: s/usage/the use/g

server/WFCORE-6750-unstable-api-annotation-scanner.adoc
bstansberry: Actually I think what you did is ok for Preview. A blog would be good though, if not now then as part of moving to Community.

ehsavoie (approved):

server/WFCORE-6750-unstable-api-annotation-scanner.adoc
bstansberry: Great; that was my intent. :)

bstansberry (approved):

bstansberry:
Thanks @kabir and @ehsavoie

PR #570, opened 2024-05-13
[WFCORE-6815][Community] Utility to reload a Host Controller to a different stab…

yersan:
…ility level in the testsuite

Jira issue: https://issues.redhat.com/browse/WFCORE-6815

### DIFF
--- a/server/WFCORE-6815-reload-domain-mode-to-stability-level.adoc
+++ b/server/WFCORE-6815-reload-domain-mode-to-stability-level.adoc
@@ -1,0 +1,161 @@
+---
+categories:
+  - core
+  - management
+# Add any category for this proposal
+# if missing, add it to _data/widfly-categories and use its id
+---
+= [Community] Ability to reload a Host Controller to a different stability level
+:author:            Yeray Borges
+:email:             yborgess@redhat.com
+:toc:               left
+:icons:             font
+:idprefix:
+:idseparator:       -
+
+== Overview
+
+WFCORE-6728 introduces the ability to reload a standalone server to a different stability level. This is mainly for the testsuite to be able to reload the server to a different stability level before running the tests, and to reload back to the original stability level after the tests have completed.
+
+This proposal is to extend this functionality to domain mode and provide a method to reload a Host Controller to a different stability level for testing purposes.
+
+== Issue Metadata
+
+=== Issue
+
+* https://issues.redhat.com/browse/WFCORE[WFCORE-6815]
+
+=== Related Issues
+
+//* https://issues.redhat.com/browse/WFLY[WFLY-XXXX]
+
+=== Stability Level
+// Choose the planned stability level for the proposed functionality
+* [ ] Experimental
+
+* [ ] Preview
+
+* [x] Community
+
+* [ ] default
+
+=== Dev Contacts
+
+* mailto:{email}[{author}]
+* mailto:brian.stansberry@redhat.com[Brian Stansberry]
+* mailto:kkhan@redhat.com[Kabir Khan]
+* mailto:jperkins@redhat.com[James Perkins]
+
+=== QE Contacts
+
+=== Testing By
+// Put an x in the relevant field to indicate if testing will be done by Engineering or QE. 
+// Discuss with QE during the Kickoff state to decide this
+* [x] Engineering
+
+* [ ] QE
+
+=== Affected Projects or Components
+
+WildFly Core. The main changes will be in the controller, host-controller and testsuite/shared modules.
+
+=== Other Interested Projects
+
+=== Relevant Installation Types
+// Remove the x next to the relevant field if the feature in question is not relevant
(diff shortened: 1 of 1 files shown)


yersan:
Hello @kabir , any chance you can take a look at this one as SME for a community feature?
@jamezp Would you agree to be the external third-party contributor and review this one from your side?

jamezp:
@yersan Yes, but as an FYI I'm on PTO for the a week starting Wednesday so if this is time sensitive, I might not be that helpful :)

yersan:
Hi @kabir / @jamezp Any chance to look at this proposal? 

The lack of ability to segregate the tests is blocking work for the conversion branches, since we need to properly avoid the execution of tests that are not meant for the default stability level. If possible, I would like to unlock this. Thanks!

server/WFCORE-6815-reload-domain-mode-to-stability-level.adoc
```diff
@@ -0,0 +1,161 @@
...
+
+== Requirements
+
+=== Hard Requirements
+
+* There will be a new operation based on the current reload operation introduced for host controllers.
+- The operation will be registered at stability level 'community' at present.
+- The operation is based on the reload operation, and adds a parameter called `stability`. This parameter takes the stability level we want to reload the server to.
+- The operation also supports all the parameters supported by the standard reload operation, included the restart-servers parameter.
+- Restarting a host controller with a different stability level without restarting the servers to get the value of the new stability level is not an useful operation, for simplicity and consistency with the reload operation, the restart-servers parameter will be allowed.
```
jamezp, line 83: Should this say the "restart-servers parameter will not be allowed"? I'm asking because the first part of the sentence makes it not sound useful.

If the wording is intended, is the host controller meant to act as the minimum stability level for servers?

yersan: @jamezp 

> Should this say the "restart-servers parameter will not be allowed"?

It is currently allowed, but it does not make sense to reload the host to a certain stability level and not restart your servers. If you do that, you could have your host running at one stability level, but your server running at another. I think that could be a problem. Maybe the simplest approach here would be to remove the `restart-servers` attribute from the `reload-enhanced` operation and restart the servers always when this operation is used.

I initially wanted to avoid these differences and assume you should know what you are doing when using the `reload-enhanced`, after all it is for our testsuite, That would bring more flexibility, but maybe that's not a good idea. WDYT?


> is the host controller meant to act as the minimum stability level for servers?

The Host Controller sends its stability level to the managed servers with the managed server boot command, that's all. It dictates what stability level the managed server will run once it gets restarted. So, if you do not restart your servers, you could have the Host Controller running on `preview` and the managed servers still on `community`, I guess that could be problematic somehow, but I'm not sure.

Maybe we should start as restrictive as we can, e.g. remove the restart servers from reload enhanced and always restart them to avoid problems, and if there is a need, open it to allow restarting the server to get more flexibility.

jamezp: @yersan IMO then I think the `restart-servers` attribute should be removed from the new operation. I can understand not wanting the difference, However, attributes on operations that don't exist are ignored so it won't really hurt anything and makes it safer I think.

yersan: @jamezp I've removed the restart-servers attribute from the reload-enhanced operation. Let me know how it looks like now

server/WFCORE-6815-reload-domain-mode-to-stability-level.adoc
```diff
@@ -0,0 +1,161 @@
...
+=== Relevant Installation Types
+// Remove the x next to the relevant field if the feature in question is not relevant
+// to that kind of WildFly installation
+* [] Traditional standalone server (unzipped or provisioned by Galleon)
+
+* [X] Managed domain
+
+* [] OpenShift s2i
+
+* [] Bootable jar
```
jamezp, line 73: Very minor, these should be `[ ]` so they show up as checkboxes. Same with the first one too... ...again minor :)

jamezp (approved):

server/WFCORE-6815-reload-domain-mode-to-stability-level.adoc
yersan: Done, thanks for reviewing this @jamezp

jamezp (approved):

yersan:
@kabir  Let me know if there is anything else to add here, otherwise, approve if you agree, thanks!

yersan:
@kabir Friendly remainder, this one is waiting for your approval to get it merged, please take a look, thanks!

kabir (approved):

bstansberry (approved):

bstansberry:
Thanks @yersan, @kabir and @jamezp

PR #572, opened 2024-05-23
[WFLY-19335] [COMMUNITY] Include instructions to run quickstarts on Kubernetes

kabir:
https://issues.redhat.com/browse/WFLY-19335


quickstarts/WFLY-19335-kubernetes-readmes-for-quickstarts.adoc
```diff
@@ -0,0 +1,166 @@
...
+
+* The quickstarts mentioned in the Overview section must:
+** Have instructions in the README for how to deploy and test the application using Helm on Kubernetes
+** Have tests runnable on GitHub Actions that perform the same steps as the instructions in the README
+* The instructions to deploy on Kubernetes are overall quite similar to how to deploy on OpenShift, except:
+** `kubectl` is used instead of `oc` when interacting with Kubernetes
+*** There is no need to log in to Kubernetes, so there is no `kubectl login` equivalent of `oc login`
+** Some of the quickstarts require installing things to Kubernetes beforehand
+*** When these prerequisites are different in Kubernetes to OpenShift, the README should reflect that
+** Kubernetes does not have s2i, instead we build the image using `mvn wildfly:image`, and tag and deploy that to the Kubernetes registry
```
jmesnil, line 92: This add a requirement to have `podman` or `docker` installed as `mvn wildly:image` will delegate to these tools to build the images

jmesnil, line 92: s/tag and deploy/tag and push

kabir: Updated, thaks!

bstansberry:
@kabir @emmartins What does 'community' stability mean in the context of QS?

Stability levels have meaning in the sense that users can explicitly opt in to allow things of a given level without risking choosing things at a lower level.  This is enforced via things like the --stability param to standalone/domain.sh and parameters passed to galleon-based provisioning tooling.

And, to be blunt, we want to be able to easily make them unavailable downstream.

How does that work for QS?

kabir:
@bstansberry at the moment, the new content should not be rendered in EAP (via Asciidoc if statements around the includes of the k8s content). We already test on Kubernetes but don't advertise it to the users.

emmartins (approved):

emmartins:
@bstansberry what @kabir described is the standard way we ATM disable the rendering of upstream only content, e.g.:

```
ifndef::ProductRelease,EAPXPRelease[]
include::../shared-doc/build-and-run-the-quickstart-with-kubernetes.adoc[leveloffset=+1]
endif::[]
```
 
I guess we could consider to replace those attrs with a stability level, something like:

```
#only defines all stability levels if not product, otherwise only default 
ifndef::ProductRelease[]
:StabilityExperimental:
:StabilityPreview:
:StabilityCommunity:
endif::[]
:StabilityDefault:

#kubernetes feature is community level, once promoted it should use instead  ifdef::StabilityDefault[]
ifdef::StabilityCommunity[]
include::../shared-doc/build-and-run-the-quickstart-with-kubernetes.adoc[leveloffset=+1]
endif::[]
```

kabir:
@bstansberry @emmartins https://issues.redhat.com/browse/WFLY-19335 was merged for WF 33.

Can we merge this issue?

I've updated it to use the new community process headers. I took the liberty of setting Eduardo as the SME and Brian as the outside perspective.

PR #573, opened 2024-05-31
Add README

jmesnil:


jmesnil:
@bstansberry First improvement is to get a README here :) 

It's copied verbatim from https://github.com/openshift/enhancements/blob/master/guidelines/README.md

README.adoc
```diff
@@ -0,0 +1,33 @@
...
+## Get Started
+
+1. Socialize an idea with others.  Make sure others think the work is worth doing, and are willing to review design and code changes required.
+2. Create a Pull Request with your new analysis document based on the link:./design-doc-template.adoc[template].
+
+## FAQs
+
+### Do I have to use the process?
+
+If this is an enhancement or new requirement for WildFly, yes.
```
bstansberry, line 20: Perhaps

"If this introduces new user-controllable behavior to WildFly, yes.  Requirements not involving user controllable behavior that are of sufficient complexity should also use the process."

I want to avoid the word 'enhancement' because we use the JIRA Enhancement type for a different thing -- things that benefit the user but are not user controllable or otherwise require the feature process.

README.adoc
```diff
@@ -0,0 +1,33 @@
...
+
+### Why would I want to use the process?
+
+It provides a mechanism to communicate enhancements and requirements across the WildFly community. It also helps coordinate and collaborate on related requirements.
```
bstansberry, line 24: s/enhancements/features

bstansberry (changes requested):

README.adoc
jmesnil: Thanks for the feedback. I kept "yes" as the short story and further expand it based on your suggestion.

bstansberry (approved):

bstansberry:
Thanks @jmesnil

Issue #568, opened 2024-04-25
Add CI to validate categories


bstansberry:
I have no idea how practical this is, but it would be good to have a github action that validates PRs for the presence of the 'categories' front matter, including at least one category, and that all declared categories map to one of the _data/categories.yaml entries.

I've spent a lot of time today fixing up various analysis docs when preparing the WildFly 32 release announcement.

jmesnil:
@bstansberry I've done something similar on a test project and I'll have a look at it.

I considering also expanding these checks for things required by an analysis document (link to issue tracker, stability level) in a second time.
Depending on the evolution of our process, I'd like to move more metadata to the front matter for further validation (eg reviewer, approved, outside_perspective assignees)

PR #574, opened 2024-06-03
[#568] Add GitHub action to verify the front matter metadata

jmesnil:
The commit adds the GitHub action to validate the `categories` field.

It only checks proposals that are *changed* in the PR. 

The 2nd commit shows what happens if the proposal is not valid and the check fails.


jmesnil:
Before this PR is merged, I'll remove the 2nd commit.

bstansberry (approved):

PR #576, opened 2024-06-04
[#568] Add GitHub action to verify the front matter metadata

### LINKED ISSUE #568: Add CI to validate categories

jmesnil:
This fixes #568

### DIFF
--- a/.github/workflows/check_process.yml
+++ b/.github/workflows/check_process.yml
@@ -1,0 +1,43 @@
+name: Check Proposals
+
+on:
+  pull_request_target:
+
+jobs:
+  check-proposals:
+    runs-on: ubuntu-latest
+
+    permissions:
+      contents: read # to download a workflow artifact
+      pull-requests: write
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+        with:
+          ref: "refs/pull/${{ github.event.number }}/merge"
+      - name: Setup Ruby
+        uses: ruby/setup-ruby@8575951200e472d5f2d95c625da0c7bec8217c42 # v1.161.0
+        with:
+          ruby-version: '3.1' # Not needed with a .ruby-version file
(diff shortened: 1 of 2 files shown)


jmesnil:
@bstansberry @darranl This PR enables us to automatically validate that a proposal as valid `categories` in its metadata.
If that's not the case, it will add a `invalid-categories` label to the PR. Once this is valid, the label would be removed.

I can't provide a test case for this from this PR so I've created a _temporary branch_  at https://github.com/wildfly/wildfly-proposals/tree/github_action_frontmatter. When I opened https://github.com/wildfly/wildfly-proposals/pull/575 against this temp branch, you can see the check is failing (with a clear error message at https://github.com/wildfly/wildfly-proposals/actions/runs/9362939230/job/25772708885?pr=575) and the `invalid-categories` label is added to the PR.

This enhancement would make sure that we have proposals with valid categories before they are merged.
The check only verifies proposals that are modified by the PR.

@darranl With this setup in place, we could expand our metadata in the front matter of the proposal to further validate that a proposal is in a good shape. For example, we could move to the front matter info such as:
* `stability-level`-- validate that the stability level is set, we could use that when we generate the web site to highlight the proposal (without having to have a [EXPERIMENTAL] in the document title)
* `reviewer`, `outside-perspective` -- these fields would point to the right people involved in the proposal
* `issue` -- the issue tracker for the proposal

Having labels for these various fields would also help discoverability and community participations (what are the `experimental` proposals? Which proposals is missing an `outside-perspective` person, etc.)

bstansberry (approved):

bstansberry:
Thanks @jmesnil!

PR #577, opened 2024-06-06
Proposal to update WildFly download distributions using channels

jmesnil:


jmesnil:
@bstansberry @spyrkob This proposal would encompass https://issues.redhat.com/browse/WFLY-19221 but I'd like to get the whole user story in the proposal (including the tooling aspect of it).
We would likely have to update the existing `installation-manager.sh` script to provide this enhancement.

wf-galleon/wildfly_channel_in_zips.adoc
```diff
@@ -0,0 +1,125 @@
...
+However for users that are downloading WildFly from our traditional distributions (zip & tgz archives) they can not take advantage of this mechanism.
+
+If our archives would work out of the box with channels, it would significantely simplify updating an installation.
+
+The workflow would be as simple as:
+
+1. Download WildFly version x.y.z from our download page
+2. Use it as normal
+3. A micro release of WildFly x.y.(z+ 1) is made available
+4. Run a script in the existing WildFly installation to update if from x.y.z to x.y.(z+1)
```
spyrkob, line 30: Does that mean we would include Prospero in the WildFly zip? An alternative could be using standalone prospero tool to perform the update

wf-galleon/wildfly_channel_in_zips.adoc
```diff
@@ -0,0 +1,125 @@
...
+5. Reload the server to run your applications on WildFly x.y.(z+ 1)
+
+If there are multiple available updates for WildFly, the user must be able to decide which one they want to update to.
```
spyrkob, line 33: Just a note - that would be a new requirement for prospero, not something that is currently supported.

wf-galleon/wildfly_channel_in_zips.adoc
```diff
@@ -0,0 +1,125 @@
...
+
+WIP
+
+== Backwards Compatibility
+
+This enhancement does not affect previous versions of WildFly.
+
+=== Default Configuration
+
+Updating an installation could update its default configuration (eg if the update is to a major version).
```
spyrkob, line 93: Currently Galleon (and consequently Prospero) will only update default configurations if there are no user changes in that configuration file. Any changes are recorded in a `.new` file and it's up to the user to merged those. 

Is this the expected behaviour or are we going to try to merge the user changes with upcoming changes?

wf-galleon/wildfly_channel_in_zips.adoc
```diff
@@ -0,0 +1,125 @@
...
+* mailto:{bstansbe@redhat.com}[{Brian Stansberry}]
+* mailto:{bspyrkos@redhat.com}[{Bartosz Spyrko-Smietanko}]
+
+=== Testing By
+
+* [X] Engineering
+
+=== Affected Projects or Components
+
+* WildFly
```
spyrkob, line 63: Prospero would probably need changes as well if we want to enable user to choose the update to apply.

wf-galleon/wildfly_channel_in_zips.adoc
```diff
@@ -0,0 +1,125 @@
...
+
+* [x] Traditional standalone server (unzipped)
+
+== Requirements
+
+* WildFly distributions archives (zip & tgz) must contain the channel metadata that describes their installation state.
+* WildFly distributions archives (zip & tgz) must contain the tooling (as scripts) to let users update their installations
+** Users must be able to list and select the updates to apply to their installations
+** The user MUST specify the updates to apply.
+* Updates must not discard any users changes to an installation (in their configuration files or JBoss modules directory)
```
spyrkob, line 77: Note, currently this does not apply to user changes in JBoss modules marked as `system-path` (modules/system/layers/base). Changes to any modules there will be overwritten if the update requires it.

User modules outside those paths will not be affected.

wf-galleon/wildfly_channel_in_zips.adoc
jmesnil: This is an open question. There is a `bin/installation-manager.sh` that does work (and I've no idea what its intent)

I'm not keen on adding another tool inside WildFly distributions and I wonder if we should make the update mechanism accessible through `./bin/jboss-cli.sh`...

jmesnil: (prospero lib would still be needed as a JBoss module but not as something surfacing to the UI)

wf-galleon/wildfly_channel_in_zips.adoc
spyrkob: Regarding reloading - the changes to the server have to applied while the server is down, so it's a case of starting the server again rather than reloading.

Right now prospero should display a list of such conflicts at the end of update and let the user know they might need to merge those themselves.

wf-galleon/wildfly_channel_in_zips.adoc
jmesnil: as a note, it is possible to decide to which version to update by changing the `manifest.maven.version` in `.installation/installer-channels.yaml` to point to a new manifest

(or use `/prospero update subscribe --version=32.0.1.Final --product=wildfly` seems to do the work too)

bstansberry:
Thanks for this, @jmesnil. This goes well beyond WFLY-19221, which is a good thing. But it goes beyond in ways that seem unlikely to be completed by the June 26 feature freeze for WildFly 33.

So I'm not sure what we (e.g. me, you, @spyrkob) want to do re WF 33. WFLY-19221 itself establishes a kind of new contract for WF, e.g. that in our download zips there will be files in certain locations with certain content. (We somewhat already have contracts in this regard as we publish manifests that tools like wildfly-maven-plugin can use.)

If we know enough about what that contract should be, perhaps we can do something re WFLY-19221 in 33. If not, we shouldn't.

I don't think that right now we know enough, because the use case discussion here may affect things. OTOH, if there's significant value in having WFLY-19221 in 33 (e.g. to provide a base for other tooling development) it might be the case that we'd know enough a couple weeks from now, if we make that a goal of the discussion.

wf-galleon/wildfly_channel_in_zips.adoc
jmesnil: I updated the proposal to reuse the  `installation-manager` script to update WildFly and mention that we would use prospero as lib

jmesnil:
> This goes well beyond WFLY-19221, which is a good thing. But it goes beyond in ways that seem unlikely to be completed by the June 26 feature freeze for WildFly 33.
> So I'm not sure what we (e.g. me, you, @spyrkob) want to do re WF 33. WFLY-19221 itself establishes a kind of new contract for WF, e.g. that in our download zips there will be files in certain locations with certain content. (We somewhat already have contracts in this regard as we publish manifests that tools like wildfly-maven-plugin can use.)

I update the proposal to address some of the questions you raise.
The additional files are in a dotted directory, the feature itself is flagged as preview so we are not tying our hands too tightly :) 

For WildFly 33, I agree that we will not be able to address the tooling aspect of the proposal.
We could limit the scope to WFLY-19221 (as a preview feature) but the full RFE will not be addressed until we have the tooling in place to effectively update WildFly.

The only advantage of having WFLY-19221 done for WildFly 33 is to be able to work incrementally and that it leaves us the full dev cycle of 34 to address the tooling aspect.

That's a weak argument so if @bstansberry  or @spyrkob, you don't see another value for WFLY-19221 in itself, let's move it to WildFly 34 and do the tooling at the same time.

provisioning/wildfly_channel_in_zips.adoc
```diff
@@ -0,0 +1,210 @@
...
+
+With that in mind, we propose to incorporate the channel metadata in the WildFly distributions that describes the exact version of the channel used to provision the server.
+In practice, that means that the installation would be "bound" to a versioned channel (eg `org.wildfly.channels:wildfly:33.0.0.Final`).
+
+Updating to a more recent version is a user decision and would require the user to give as input the version of the WildFly they want to _update to_ (e.g. `33.0.1.Final` or `34.0.0.Final`).
+After the update is successful, the channel metadata would be updated to point to the current version of the installation.
+
+WildFly must contain the tooling to let user update installations without having to download another piece of software.
+
+The tooling should be made available in the `bin` directory of this installation and affects only the parent directory (`JBOSS_HOME` would have no bearing for update).
```
spyrkob, line 61: We should keep JBoss CLI and HAL in sync with those new capabilities.

provisioning/wildfly_channel_in_zips.adoc
```diff
@@ -0,0 +1,210 @@
...
+** Provisioned feature packs remain referenced by their Galleon feature-pack-locations (e.g. `wildfly@maven(org.jboss.universe:community-universe):current#33.0.0.Final`).
+*** Users that use Galleon tool to udpate their WildFly installation will continue to be able to do so.
+*** Galleon universe artifacts (`org.jboss.universe:community-universe` & `org.jboss.universe.producer:wildfly-producers`) will be added to the WildFly channels
+* WildFly distributions archives (zip & tgz) must contain the tooling to let users update their installations
+** tracked in https://github.com/wildfly-extras/prospero/issues/749[prospero #749]
+** Users must be able to list and select the updates to apply to their installations
+** The user MUST specify the updates to apply.
+** As this feature is experimental, the tooling should warn the user that updating their installation is an experimental mechanism
+** these operations will be using Prospero that needs to be integrated as a JBoss module in the WildFly distributions.
+* Updates must not discard any user changes to an installation (in their configuration files or JBoss modules directory)
```
spyrkob, line 132: Note - currently the files under `modules/system` are considers "system-path" and Galleon updates take precedence over user changes

provisioning/wildfly_channel_in_zips.adoc
```diff
@@ -0,0 +1,210 @@
...
+=== Testing By
+
+* [x] Engineering
+
+=== Affected Projects or Components
+
+* https://github.com/wildfly/wildfly[WildFly]
+** WildFly distributions will incorporate channel metadata used to create the distributions and tooling to update the installation to a more recent version.
+* https://github.com/wildfly-extras/prospero[Prospero]
+** Prospero will be used as the library to update WildFly installations but will not surface to the user (the `installation-manager` script is the user entry point)
```
spyrkob, line 109: I'm not sure I understand why we need a new entry point script (`installation-manager`) instead of the script shipped with Prospero?

If we don't want to expose Prospero script to the users, maybe we should use the JBoss CLI/Web Console integration?

provisioning/wildfly_channel_in_zips.adoc
```diff
@@ -0,0 +1,210 @@
...
+* Trimming an existing installation coming from WildFly distributions with Galleon layers is not supported.
+
+=== Future Work
+
+This feature is `experimental`.
+
+To make it a `preview` or `community` feature, we will pay attention to the user experience. In particular, the distributions should be "self-updatable" and should not need
+to rely on the external download of Prospero to be updated. Promotion to `preview` or `community` would involve the integration of Prospero library into WildFly (as a JBoss module or a separate feature pack). The user interface could also be directly the `prospero` cli tool, a streamlined CLI focused on updates, or additional commands provided within `jboss-cli` tool.
+
+=== Implementation Plan
```
spyrkob, line 146: Should the implementation plan mention the new module(s) created to include Prospero libraries?

provisioning/wildfly_channel_in_zips.adoc
```diff
@@ -0,0 +1,210 @@
...
+* https://issues.redhat.com/browse/WFMP-271[WFMP-271] Add a <name> parameter to the <channel> configuration
+
+=== Related Issues
+
+* https://issues.redhat.com/browse/WFLY-19130[WFLY-19130] - [Community] Publish Wildfly channel and manifest during project build
+** http://docs.wildfly.org/wildfly-proposals/build/WFLY-19130_publish_Wildfly_channel_manifest.html[Analysis Document]
+
+=== Stability Level
+
+* [x] Experimental
```
darranl, line 92: I have only had a very fast scan so sorry if I jumped over it, if this is an experimental feature - how does a user turn on this experimental feature / acknowledge that they are using an experimental feature rather than assuming it is a community feature?

provisioning/wildfly_channel_in_zips.adoc
```diff
@@ -0,0 +1,210 @@
...
+** After their confirmation, the update is applied and the installation state changes
+** After the update, the provisioned version of the installation is the more recent version that was applied
+
+Since https://issues.redhat.com/browse/WFCORE-6206[WFCORE-6206], there is `installation-manager` script that is meant for internal usage. This script could be repurposed to provide these operations to the user.
+
+== Issue Metadata
+
+=== Issue
+
+* https://issues.redhat.com/browse/WFLY-19221[WFLY-19221] - [Preview] Incorporate channel metadata in the download zips
```
bstansberry, line 81: This JIRA needs to be retitled.

provisioning/wildfly_channel_in_zips.adoc
```diff
@@ -0,0 +1,210 @@
...
+* https://github.com/wildfly-extras/prospero[Prospero]
+** Prospero will be used as the library to update WildFly installations but will not surface to the user (the `installation-manager` script is the user entry point)
+* https://github.com/wildfly/wildfly-maven-plugin[wildfly-maven-plugin]
+** The WildFly Maven Plug-in provides the functionality to provision WildFly to create the distributions with the expected channel metadata.
+
+=== Other Interested Projects
+
+=== Relevant Installation Types
+
+* [x] Traditional standalone server (unzipped)
```
bstansberry, line 117: This should include domain mode. Unless I missed something it would just work, and given this is Experimental I don't see any benefit to acting like it doesn't work.

provisioning/wildfly_channel_in_zips.adoc
bstansberry: I think that's fine. We don't support users modifying system modules, so if they do they can't use this feature.

provisioning/wildfly_channel_in_zips.adoc
```diff
@@ -0,0 +1,210 @@
...
+
+Testing an installation is out of scope of this enhancement but should be incorporated in the user workflow (e.g. to test the update in a staging environment).
+
+=== Interoperability
+
+This section is not relevant in the context of this enhancement.
+
+== Test Plan
+
+=== Test Plan for WFLY-19221 - [Preview] Incorporate channel metadata in the download zips
```
bstansberry, line 193: s/Preview/Experimental

provisioning/wildfly_channel_in_zips.adoc
```diff
@@ -0,0 +1,210 @@
...
+
+* Verify that WildFly generated distributions (from the `dist`, `ee-dist`, and `preview-dist` Maven Modules) contain the channel metadata files corresponding to their provisioning states.
```
bstansberry, line 195: preview-dist -> preview/dist

My instinct is this should include 'build', 'ee-build' and 'preview/build' as well. The basic concept is dist and build are the same except no jars in the modules and module.xml files include maven coords instead of the paths to the jars. Not adding further differences seems good.

provisioning/wildfly_channel_in_zips.adoc
```diff
@@ -0,0 +1,210 @@
...
+
```
bstansberry, line 196: We have the testsuite/galleon tests which test updating an installation using Galleon tooling. It seems like we could expand this to cover this update mechanism.

provisioning/wildfly_channel_in_zips.adoc
```diff
@@ -0,0 +1,210 @@
...
+** Users must be able to list and select the updates to apply to their installations
+** The user MUST specify the updates to apply.
+** As this feature is experimental, the tooling should warn the user that updating their installation is an experimental mechanism
+** these operations will be using Prospero that needs to be integrated as a JBoss module in the WildFly distributions.
+* Updates must not discard any user changes to an installation (in their configuration files or JBoss modules directory)
+
+=== Non-Requirements
+
+* Changing the type of distributions during an update is not supported (in other words, it is not possible to download the zip for WildFly 33.0.0.Final and update the installation to WildFly Preview)
+* Trimming an existing installation coming from WildFly distributions with Galleon layers is not supported.
```
bstansberry, line 137: This is fine but it gets me thinking about a related scenario. Is the installation-manager.sh script present in a slimmed server? If it is, what happens when the user uses it?

provisioning/wildfly_channel_in_zips.adoc
```diff
@@ -0,0 +1,210 @@
...
+
```
bstansberry, line 138: @spyrkob has mentioned the CLI and the console. If the decision is to not include those, they should be listed as non-requirements as the basic assumption if nothing is said should be consistency across tools.

provisioning/wildfly_channel_in_zips.adoc
```diff
@@ -0,0 +1,210 @@
...
+=== Future Work
+
+This feature is `experimental`.
+
+To make it a `preview` or `community` feature, we will pay attention to the user experience. In particular, the distributions should be "self-updatable" and should not need
+to rely on the external download of Prospero to be updated. Promotion to `preview` or `community` would involve the integration of Prospero library into WildFly (as a JBoss module or a separate feature pack). The user interface could also be directly the `prospero` cli tool, a streamlined CLI focused on updates, or additional commands provided within `jboss-cli` tool.
```
bstansberry, line 144: The first two sentences in this paragraph contradict the requirements.

provisioning/wildfly_channel_in_zips.adoc
```diff
@@ -0,0 +1,210 @@
...
+The WildFly producer is frequently updated to add new major versions of WildFly including the upcoming major version.
+By adding these artifacts to the WildFly channel manifest, Prospero will be able to resolve the feature pack location and ultimately point to the appropriate Maven artifact for the feature pack.
+
+== Backwards Compatibility
+
+This enhancement does not affect previous versions of WildFly.
+
+=== Default Configuration
+
+Updating an installation could update its default configuration (e.g. if the update is to a major version).
```
bstansberry, line 172: or minor

bstansberry (changes requested):

provisioning/wildfly_channel_in_zips.adoc
spyrkob: If the slimming is achieved using Galleon and recorded in the `.galleon/provisioning.xml`, Prospero should only update the components in the slimmed installation. 
Or do you mean a scenario where the slimmed installation does not include Prospero modules?

provisioning/wildfly_channel_in_zips.adoc
darranl: FYI on tooling alignment (especially HAL and JBoss CLI) I have a discussion here about a change I will be proposing against the proposals repo.

Overall based on the stability level of a new feature I am proposing different levels of support in the tooling so features can come it at a lower level maybe without tooling support and then get the tooling support added, then promote the feature to a higher stability level.

A general exception however would be where a feature is dependent on the tooling support entierely, i.e. if you can not use it without the JBoss CLI or HAL then one must support it at the outset - assuming the developers on those projects are available to assist.

provisioning/wildfly_channel_in_zips.adoc
bstansberry: @spyrkob Both I suppose.  TBH I don't recall what I was thinking when I asked this.

spyrkob: I think currently the installation-manager.sh is always provisioned even if the installation.manager package is not present, but I'm not sure if it's possible to exclude that package, maybe @yersan can confirm.

PR #579, opened 2024-06-07
HAL-1981: Console dashboard

hpehl:
https://issues.redhat.com/browse/HAL-1981

### DIFF
--- a/console/HAL-1981_Dashboard.adoc
+++ b/console/HAL-1981_Dashboard.adoc
@@ -1,0 +1,172 @@
+---
+categories:
+- console
+---
+= [Experimental] Console Dashboard
+:author:            Harald Pehl
+:email:             hpehl@redhat.com
+:toc:               left
+:icons:             font
+:idprefix:
+:idseparator:       -
+
+== Overview
+
+A dashboard will replace the homepage of the current console. The dashboard should visualize WildFly's most essential data and follow the guidelines from the PatternFly website: https://www.patternfly.org/patterns/dashboard/design-guidelines.
+
+.Dashboard sample
+image::https://www.patternfly.org/images/dashboard-view-desktop.d9b02461450d9a4e.png[Dashboard sample]
+
+Different cards visualize the data on the dashboard. These cards include data like:
+
+* Name, version, build, and operation mode of WildFly.
+* Runtime information about memory, threads, and other essential resources.
+* MicroProfile health status
+* Information about deployments
+* The last <n> log statements (with errors and warnings highlighted)
+* Links to documentation, help, forums, and information on how to get involved with WildFly.
+
+== Issue Metadata
+
+The feature will be part of the next-gen console (https://github.com/hal/foundation/). It won't be available in the current (3.x) codebase.
+
+=== Issue
+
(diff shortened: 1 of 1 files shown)


hpehl:
This proposal is meant to discuss a dashboard for the next-gen console. In particular, I want to get feedback about:

- the cards on the dashboard
- the information on the cards
- distinction between standalone and domain mode
- whether the cards should be refreshed automatically or only on demand

console/HAL-1981_Dashboard.adoc
```diff
@@ -0,0 +1,172 @@
...
+
+=== Cards
+
+==== Product info
+
+This card shows WildFly's static and configuration data. This includes name, version, operation mode, stability level, OS/CPU, and JVM information.
+
+*Sources*
+
+* `:read-resource(attributes-only,include-runtime)`
```
jmesnil, line 83: thinking out loud but should this card be dynamic to track changes in the running mode of the servers?

console/HAL-1981_Dashboard.adoc
```diff
@@ -0,0 +1,172 @@
...
+
+* [x] Managed domain
+
+* [ ] OpenShift s2i
+
+* [ ] Bootable jar
+
+== Requirements
+
+The dashboard visualizes WildFly's most essential data, including configuration and runtime data and links to valuable resources. Different cards visualize the data on the dashboard. If a card shows dynamic data, it includes a refresh button. In addition, all cards are refreshed automatically (for static cards this is a noop).
```
jmesnil, line 73: > If a card shows dynamic data

Do you infer that from the CLI operations or is it up to the card to define if it's dynamic or not?

console/HAL-1981_Dashboard.adoc
```diff
@@ -0,0 +1,172 @@
...
+
+==== Health
+
+This card is only available in standalone mode. It uses the MicroProfile Health subsystem to show WildFly's healthiness.
+
+* `/subsystem=microprofile-health-smallrye:check-started`
+* `/subsystem=microprofile-health-smallrye:check-ready`
+* `/subsystem=microprofile-health-smallrye:check-live`
+
+If the MicroProfile Health subsystem is not available, this card contains information on how to enable it and a link to https://docs.wildfly.org/32/Admin_Guide.html#MicroProfile_Health_SmallRye.
```
jmesnil, line 135: If the MicroProfile Health subsystem is not present, you could also first check for the base `health` subsystem too (eg if the server only uses the wildfly-ee galleon pack)

console/HAL-1981_Dashboard.adoc
hpehl: That's up to the card and part of the dashboard/card implementation. The cards are divided into static cards, like name, version, OS, and JVM information, which do not change at runtime, and dynamic cards, like deployment, memory, threads, and health information, which change at runtime.

console/HAL-1981_Dashboard.adoc
hpehl: I'd like to keep this static and move the running mode of the servers in the domain resp. runtime card.

PR #584, opened 2024-07-04
Add GitHub action to detect staled PRs

jmesnil:


jmesnil:
@bstansberry We discussed closing staled PRs to keep the queue in check. 
I used 120 days before closing the PR (that can always be reopened) but I'm not sure if it's enough and we want to span 2 WildFly release cycles

bstansberry:
I think 120 days is too short.

Also we need to triage the ones that will be closed, as glancing at https://github.com/wildfly/wildfly-proposals/pulls?q=is%3Apr+updated%3A%3C2024-06-06+is%3Aopen I see a bunch that I know we don't want to close. We'd need to comment on those to keep them alive.

PR #585, opened 2024-07-09
[WFLY-18082] add web-socket specific r/w timeouts

baranowb:
Issue: https://issues.redhat.com/browse/WFLY-13933
Related: https://issues.redhat.com/browse/UNDERTOW-2333

### DIFF
--- a/undertow/WFLY-18082_WebSocket_timeouts.adoc
+++ b/undertow/WFLY-18082_WebSocket_timeouts.adoc
@@ -1,0 +1,144 @@
+---
+categories:
+ - core
+ - management
+---
+= [Preview] Allow to specify WebSocket timeout per listener irrespective of listener general r/w timeouts
+:author:            Bartosz Baranowski
+:email:             bbaranow@redhat.com
+:toc:               left
+:icons:             font
+:idprefix:
+:idseparator:       -
+
+== Overview
+
+== Issue Metadata
+
+=== Issue
+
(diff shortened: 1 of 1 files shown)


fl4via (approved):

TomasHofman:
Reading [WFLY-18082](https://issues.redhat.com/browse/WFLY-18082), it does look like the motivation for this change is to "fix" websocket behavior:

* Originally there was no timeout;
* but that changed after [WFLY-14980](https://issues.redhat.com/browse/WFLY-14980) or other issue which introduced default 60 or 90 secs timeout on all socket listeners;
* and application using websockets would typically expect to have the connection remain open for longer / indefinitely.

(IMO if this is correct, the motivation is valuable info that could be explained in the RFE.)

Currently, the RFE suggest to allow overriding specifically web socket timeout, which would allow users who need it to disable / change websocket timeout. The default server behavior however doesn't change - web sockets, like all other sockets, would still have r/w timeout of 90 secs by default.

The question in my head is: would it be better to introduce the new timeout parameters with the default value of "0" (no timeout), rather than undefined by default; so that web sockets would be by default not limited? I don't know, just offering that for your consideration :).

Finally, it might be worth specifying behavior of special values (i.e. no timeout should be set as 0 or -1? undefined means fallback to generic r/w timeout, that's I think spelled out) and what units are used (I assume ms).

TomasHofman:
Apart from the above suggestions I approve :).

ropalka (approved):

PR #588, opened 2024-07-17
Create a feature request issue template.

jamezp:
Add a new template for feature request issues.

This is currently a draft as I want to make sure we encapsulate everything first.

### DIFF
--- a/.github/ISSUE_TEMPLATE/feature-request.yaml
+++ b/.github/ISSUE_TEMPLATE/feature-request.yaml
@@ -1,0 +1,32 @@
+name: Feature Request
+description: A feature request for WildFly
+title: "[RFE]: "
+labels: ["feature"]
+projects: ["wildfly/7"]
+body:
+  - type: input
+    id: jira
+    attributes:
+      label: JIRA
+      description: A link to the main feature request JIRA
+      placeholder: ex. https://issues.redhat.com/browse/WFLY-XXX
+    validations:
+      required: true
+  - type: dropdown
+    id: stability
+    attributes:
+      label: Stability
+      description: The stability level proposed for the feature
+      options:
+        - community
+        - preview
+        - experimental
+      default: 0
+    validations:
+      required: true
+  - type: textarea
+    id: description
+    attributes:
+      label: Description
+      description: A description of the feature
+


.github/ISSUE_TEMPLATE/feature-request.yaml
```diff
@@ -0,0 +1,32 @@
+name: Feature Request
+description: A feature request for WildFly
+title: "[RFE]: "
+labels: ["feature"]
+projects: ["wildfly/7"]
```
jamezp, line 5: We may want to consider using a workflow to add the project if not everyone create an issue will have write access. See https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/syntax-for-issue-forms.

.github/ISSUE_TEMPLATE/feature-request.yaml
```diff
@@ -0,0 +1,32 @@
+name: Feature Request
```
jamezp, line 1: Do we want to use the term "Feature Request" or is there something better?

jamezp:
Note that outside perspective is required for the template. We could make the other ones required as well, but we don't have to. The question becomes what we want to see in the issue vs what we want on the actual proposal document.

.github/ISSUE_TEMPLATE/feature-request.yaml
```diff
@@ -0,0 +1,53 @@
...
+projects: ["wildfly/7"]
+body:
+  - type: input
+    id: jira
+    attributes:
+      label: JIRA
+      description: A link to the main feature request JIRA.
+      placeholder: ex. https://issues.redhat.com/browse/WFLY-XXX
+    validations:
+      required: true
```
jmesnil, line 14: Is it really required to propose a new feature enhancement?

It's needed for the analysis document but for proposing something that might be too early. It might not be necessarily known beforehand which projects would be impacted by the proposal (WFLY, WFCORE, WFMP, etc.)

.github/ISSUE_TEMPLATE/feature-request.yaml
```diff
@@ -0,0 +1,53 @@
+name: Feature Request
+description: A feature request for WildFly
+title: ""
+labels: ["feature"]
+projects: ["wildfly/7"]
+body:
```
jmesnil, line 6: we could add a [markdown](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/syntax-for-githubs-form-schema#markdown) type to give more contextual information for the user submitting the proposal (without polluting the created issue with this blurb of text).

This would give the user more info about what we are expecting for them when they fill the form.

.github/ISSUE_TEMPLATE/feature-request.yaml
```diff
@@ -0,0 +1,53 @@
...
+    attributes:
+      label: Developer
+      description: The name of the person who is doing the development.
+    validations:
+      required: false
+  - type: input
+    id: outside-perspective
+    attributes:
+      label: Outside Perspective
+      description: The name of the person who is the outside perspective for this feature.
```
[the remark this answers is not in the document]
bstansberry: If we can validate this based on the stability level that would be ideal. It's N/A for 'experimental', otherwise required.

.github/ISSUE_TEMPLATE/feature-request.yaml
bstansberry: Enhancement has a meaning that's not relevant here.  In JIRA we use it for improvements that bring direct user benefit but are not user controllable. I'd prefer not to confuse that.

'Feature' seems reasonable

.github/ISSUE_TEMPLATE/feature-request.yaml
```diff
@@ -0,0 +1,53 @@
+name: Feature Request
+description: A feature request for WildFly
```
bstansberry, line 2: A new feature for WildFly

.github/ISSUE_TEMPLATE/feature-request.yaml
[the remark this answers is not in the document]
bstansberry: I don't get the overall 'validation' pattern. We always need the SME and the developer, while requiring the outside perspective depends on the stability level.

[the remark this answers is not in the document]
bstansberry: TBH I think this feature team stuff should be in the proposal as well, as front matter. We talk about the 'required' feature team, but those are really minimums. There's one person who's the primary assignee (the default person to contact), then there are 1..n SMEs, and then [0|1]..n outsiders. This latter two are modeled as lists, which is doable via adoc front matter.

Recording that here is ok, if we remove it from the proposal and the feature team itself will be able to maintain it in the issue.

Note: I'm really thinking out loud here; I think we can iterate on this and not block too much.

.github/ISSUE_TEMPLATE/feature-request.yaml
jamezp: No, this doesn't really need to be required I suppose. The idea here is we use this for RFE's that are effectively already in progress. This will not be meant to track features we want to have, just features we're actively working on for a release.

.github/ISSUE_TEMPLATE/feature-request.yaml
```diff
@@ -0,0 +1,53 @@
...
+      required: true
+  - type: dropdown
+    id: stability
+    attributes:
+      label: Stability
+      description: The stability level proposed for the feature.
+      options:
+        - community
+        - preview
+        - experimental
```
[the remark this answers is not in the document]
jamezp: We can definitely remove this.

.github/ISSUE_TEMPLATE/feature-request.yaml
```diff
@@ -0,0 +1,53 @@
...
+      default: 0
+    validations:
+      required: true
+  - type: input
+    id: sme
+    attributes:
+      label: Subject Matter Expert
+      description: The name of the person who is the subject matter expert.
```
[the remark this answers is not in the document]
jamezp: We could do that. There isn't really a field type for that.

.github/ISSUE_TEMPLATE/feature-request.yaml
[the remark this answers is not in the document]
jamezp: Sure, we can do whatever really. Making all or none of them required seems reasonable. Maybe for now, just none of them required.

.github/ISSUE_TEMPLATE/feature-request.yaml
jamezp: Sure, we could do that.

.github/ISSUE_TEMPLATE/feature-request.yaml
bstansberry: BTW 'Feature Request' seems ok to me too if we want to follow KISS principles and align with JIRA. But I get how the 'Request' part is odd here as these issues are never requests that someone do something, they are statements of intent to do something.

.github/ISSUE_TEMPLATE/feature-request.yaml
bstansberry: This data is necessary in the proposal, so repeating it here is added overhead.

.github/ISSUE_TEMPLATE/feature-request.yaml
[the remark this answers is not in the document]
bstansberry: Probably the Developer is required (which AIUI is the person we call the 'assignee'). The issue filer may not be that person, particularly early on when those of us working on tooling / process may be doing the paperwork.

jamezp:
@jmesnil @bstansberry @darranl I've updated the template to be more simplistic and not duplicate attributes. You can see an example of what it would look like here https://github.com/jamezp/wildfly-proposals/issues/new?assignees=&labels=feature&projects=wildfly%2F7&template=feature-development.yaml

darranl (approved):

darranl:
FYI for anyone reviewing, we have discussed if we can get a simplified issue template introduced and merged and we will continue to evolve as we experience using it so reviews here are not a last change to make updates etc...

bstansberry (approved):

bstansberry:
Thanks @jamezp @jmesnil @darranl

PR #595, opened 2024-09-06
[Community] [WFCORE-6830] New attributes on the HTTP Management Interface to configure resource constraints.

darranl:


management/WFCORE-6830_Management_Resource_Constraints.adoc
```diff
@@ -0,0 +1,221 @@
...
+
+A change such as this should likely be constrained to:
+
+ * Removing the stability level qualifier from the attribute definitions allowing them to be used at the default stability level.
+ * Moving the XML schema definitions and parsing to an appropriate schema at the default stability level.
+
+== Backwards Compatibility
+
+In considering backwards compatibility if an attribute is defined on the HTTP management interface AND the
+corresponding system property is set then the system property should take precedence.
```
bstansberry, line 169: If this happens we should log a WARN that the system property will be removed. People should move promptly off of system property based configuration once proper attributes are available.

And then remove it in, say, WF 36, assuming this lands in WF 34.

Probably using the same logger category that we use when warning about use of deprecated management API would be good. It's an analogous situation.

hpehl:
The management console page for the management interface is at http://localhost:9900/console/index.html#standalone-server 

Runtime → Select server → View → HTTP Management Interface

CleanShot 2024-09-09 at 13 02 39

The form for the management interface contains all simple configuration attributes. The new attributes should show up automatically.

management/WFCORE-6830_Management_Resource_Constraints.adoc
darranl: I have updated to add the logging:

https://github.com/wildfly/wildfly-proposals/pull/595/files#diff-c4fba9f719d2a6d63183999eb1869081774e7d44fb5d70004133b12de65c8dd6R176

The other deprecation messages all appear to use INFO so I have done the same here.

darranl:
I have added an "Admin Clients" section to the analysis and confirm that the new attributes are automatically handled by both admin clients:
    https://github.com/wildfly/wildfly-proposals/pull/595/files#diff-c4fba9f719d2a6d63183999eb1869081774e7d44fb5d70004133b12de65c8dd6R198

management/WFCORE-6830_Management_Resource_Constraints.adoc
bstansberry: Thanks. LGTM.

management/WFCORE-6830_Management_Resource_Constraints.adoc
```diff
@@ -0,0 +1,251 @@
...
+
+For each of the system properties if defined and if the attribute is supported at the current stability level  a message in the
+following format will be logged to inform the administrator that they should move to using the new attribute definitions.
+
+[source]
+----
+14:58:13,878 INFO  [org.jboss.as.controller.management-deprecated] (Controller Boot Thread) WFLYCTL0515: The system
+property 'org.wildfly.management.backlog' is deprecated and may be removed in a future version, attribute 'backlog' on
```
bstansberry, line 179: s/version, attribute/version. The attribute/g

This is more a comment for the impl PR, so don't feel the need to change this if you wouldn't otherwise.

management/WFCORE-6830_Management_Resource_Constraints.adoc
```diff
@@ -0,0 +1,251 @@
...
+take into account the results of calling :read-resource-description so these new attributes will be automatically
+supported by the CLI when the server is running at the `Community` stability level or lower.
+
+The admin console does contain a screen dedicated to the management of the HTTP management interface available at
+http://localhost:9990/console/index.html#standalone-server[HTTP Management Interface].  This screen is dynamically rendered
+based on the attributes defined on the HTTP management interface resource, this means that when the server is running at
+a stability level that supports these new attributes the admin console will automatically support them - in other cases they
+will not be present in the view.
+
+No follow up tasks are required for either admin client.
```
bstansberry, line 212: OT: We should make this easily cut-and-pasted content; perhaps in a more fleshed out https://github.com/wildfly/wildfly-proposals/blob/main/FEATURE_PROCESS.adoc

bstansberry (approved):

PR #581, opened 2024-06-27
Update template and checks

jmesnil:
* Update the template to move the tracking issue and the stability level to the front matter
* When a PR is open, check that these fields are valid
* Generate a HTML page for the proposal with these fields.

This PR updates the template to be more descriptive about its expected content.

As a demo, I opened a "dummy" proposal in my fork that was missing the `issue` and `stability-level` in the front matter:
* https://github.com/jmesnil/wildfly-proposals/pull/14
  * The `missing-issue` and `invalid-stability-level` labels were automatically added to the PR
* I fixed my PR
  * The labels were removed as the checks passed
  * the asciidoc displays the front matter: https://github.com/jmesnil/wildfly-proposals/blob/main/bootable-jar/dummy_featyre.adoc
* The PR was merged and the metadata were added to the generated proposal page:
  * https://jmesnil.github.io/wildfly-proposals/bootable-jar/dummy_featyre.html 
* If the `stability-level` field is valid in the front matter, add a label to the GitHub PR of the form `stability-level/XXX` (as seen on my fork at https://github.com/jmesnil/wildfly-proposals/pull/15)

**Before that PR is merged, the following labels must be created:**
* `missing-issue` - The analysis document is missing an issue tracker in its metadata
* `invalid-stability-level` - The analysis document is missing a valid stability level in its metadata
* `stability-level/experimental` - Feature is at the "Experimental" stability level
* `stability-level/preview` - Feature is at the "Preview" stability level
* `stability-level/community` - Feature is at the "Community" stability level
* `stability-level/default` - Feature is at the "Default" stability level

PR #600, opened 2024-09-15
Add admin client requirements to the process

darranl:
The updates I am proposing here relate to how in general the admin clients should be considered when new features are added to the application server.

The admin clients are maintained by smaller teams within the overall WildFly team, for this reason it is not going to scale if every feature requires complete admin client integration at every stability level so the general premise behind this proposal is that features can be added at Experimental or Preview level with a reduced set of admin client requirements, the admin clients can then continue to evolve to add these features and then the underlying feature can be promoted to Community / Default.

Note: This is not so much about features being added directly to the admin clients at the stability levels as in that case the admin client developer is already in control of compatibility etc...

### DIFF
--- a/FEATURE_PROCESS.adoc
+++ b/FEATURE_PROCESS.adoc
@@ -75,7 +75,7 @@ All hard requirements in analysis covered
 | *Admin Clients*
 (HAL / JBoss CLI)
 
-| Management model changes must be compatible with the lowe level model
+| Management model changes must be compatible with the lower level model
 manipulation capabilities of the admin clients.
 
 Breakage of high level client functionality is undesirable but accepted.
--- a/FEATURE_PROCESS.adoc
+++ b/FEATURE_PROCESS.adoc
@@ -70,6 +70,32 @@ All hard requirements in analysis covered
 | Encouraged |Encouraged |Encouraged |Required
 //-------
 
+//-------
+
+| *Admin Clients*
+(HAL / JBoss CLI)
+
+| Management model changes must be compatible with the lowe level model
+manipulation capabilities of the admin clients.
+
+Breakage of high level client functionality is undesirable but accepted.
+
+| Experimental plus:
+
+Existing functionality in the admin clients must not be broken by the new
+addition.  Higher level client integration is not required.
+
+| Preview plus:
+
+Existing higher level views and commands in the admin clients must be
+compatible with the new functionality.
+
+New higher level views and command in the admin clients are not mandatory
+unless required for the feature to be considered "complete".
+
+| Same as Community
+
+//-------
 
 //-------
 | *Component Validation*
--- a/design-doc-template.adoc
+++ b/design-doc-template.adoc
@@ -103,6 +103,16 @@ into this category, please mention the Release Coordinators on the pull
 request so they are aware.
 ////
 
+== Admin Clients
+
+////
+Identify the level of compatibility this feature will have with the existing
+admin clients (JBoss CLI and the Admin Console / HAL).
+
+Identify any follow up work that will be required in the clients and link issues
+created to track this work.
+////
+
 == Security Considerations
 
 ////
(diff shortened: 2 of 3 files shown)


design-doc-template.adoc
```diff
@@ -97,12 +97,22 @@ categories:
 
 //== Implementation Plan
 ////
-Delete if not needed. The intent is if you have a complex feature which can 
-not be delivered all in one go to suggest the strategy. If your feature falls 
-into this category, please mention the Release Coordinators on the pull 
+Delete if not needed. The intent is if you have a complex feature which can
+not be delivered all in one go to suggest the strategy. If your feature falls
+into this category, please mention the Release Coordinators on the pull
```
bstansberry, line 102: Please revert this change as it will conflict with other work.

design-doc-template.adoc
```diff
@@ -97,12 +97,22 @@ categories:
...
 
+== Admin Clients
+
+////
+Identify the level of compatibility this feature will have with the existing
+admin clients (JBoss CLI and the Admin Console / HAL).
+
+Identify any follow up work that will be required in the clients and link issues
+created to track this work.
+////
```
bstansberry, line 114: Please use the markdown style illustrated in https://github.com/wildfly/wildfly-proposals/pull/581/files#diff-1e46fd5b24b8df67bec2629d044ee017dc0e62542e582563b1abf9a29db7dfe6.

design-doc-template.adoc
```diff
@@ -119,10 +129,10 @@ Generally a feature should have documentation as part of the PR to wildfly main,
 == Release Note Content
 ////
 Draft verbiage for up to a few sentences on the feature for inclusion in the
-Release Note blog article for the release that first includes this feature. 
+Release Note blog article for the release that first includes this feature.
```
bstansberry, line 132: Please revert this change as it will conflict with other work.

design-doc-template.adoc
```diff
@@ -119,10 +129,10 @@ Generally a feature should have documentation as part of the PR to wildfly main,
...
 Example article: http://wildfly.org/news/2018/08/30/WildFly14-Final-Released/.
 This content will be edited, so there is no need to make it perfect or discuss
 what release it appears in.  "See Overview" is acceptable if the overview is
-suitable. For simple features best covered as an item in a bullet-point list 
-of features containing a few words on each, use "Bullet point: <The few words>" 
+suitable. For simple features best covered as an item in a bullet-point list
+of features containing a few words on each, use "Bullet point: <The few words>"
```
bstansberry, line 137: Please revert this change as it will conflict with other work.

bstansberry (changes requested):

design-doc-template.adoc
darranl: Updated

bstansberry (approved):
Thanks, @darranl. I'll leave merging this to you as you know what other feedback you want.

darranl:
@hpehl / @jfdenise Can you both please have a look at this and check it makes sense for you?

This is more about how the admin clients will be updated based on other changes to the app server rather than features added directly to the admin clients.

FEATURE_PROCESS.adoc
```diff
@@ -70,6 +70,32 @@ All hard requirements in analysis covered
 | Encouraged |Encouraged |Encouraged |Required
 //-------
 
+//-------
+
+| *Admin Clients*
+(HAL / JBoss CLI)
+
+| Management model changes must be compatible with the lowe level model
```
jfdenise, line 78: lowe? you mean lowest?

jfdenise (approved):

hpehl (approved):
LGTM! Thanks @darranl

FEATURE_PROCESS.adoc
darranl: Thank you, I have fixed this.

darranl:
I just sent a message to the dev list, I think will give it one more day then merge.

Being merged does not set it in stone so we can still evolve as feedback comes in / we try and work with these contraints.

PR #616, opened 2024-10-09
[WFLY-19588] Proposal for MicroProfile Platform 7.0

kabir:
https://issues.redhat.com/browse/WFLY-19588


microprofile/WFLY-19588_microprofile_platform_7.adoc
```diff
@@ -0,0 +1,197 @@
+---
+categories:
+- microprofile
+# Add any category for this proposal as a yaml list, e.g.
+# - core
+# - management
+# if missing, add it to _data/widfly-categories and use its id
+---
+= [Experimental|Preview|Community|default]Template <INSERT TITLE HERE>
```
rhusar, line 9: Can you please add the title here?

rhusar (changes requested):

microprofile/WFLY-19588_microprofile_platform_7.adoc
fabiobrz, line 1: It seems the document structure is not aligned with the latest template here: https://github.com/wildfly/wildfly-proposals/blob/main/design-doc-template.adoc

microprofile/WFLY-19588_microprofile_platform_7.adoc
```diff
@@ -0,0 +1,201 @@
...
+= [default] MicroProfile Platform 7.0
+:author:            Kabir Khan
+:email:             kkhan@redhat.com
+:toc:               left
+:icons:             font
+:idprefix:
+:idseparator:       -
+
+== Overview
+This RFE is about updating WildFLy to support MicroProfile Platform 7.0. We currently support MicroProfile Platform 6.1.
```
fabiobrz, line 22: Nitpicking: s/WildFLy/WildFly/

microprofile/WFLY-19588_microprofile_platform_7.adoc
```diff
@@ -0,0 +1,201 @@
...
+
+|Telemetry +
+|1.1
+|1.1
+|2.0
+|y https://issues.redhat.com/browse/WFLY-19589[WFLY-19589]
+
+|===
+
+There is no additional work in this RFE beyond what is done by the individual spec update RFEs, i.e. it is just an umbrella issue.
```
fabiobrz, line 80: I do agree, this is an umbrella issue, then I wonder how it is not being set as an epic, and include child issues, since the platform upgrade can't be certified _unless_ all of the child issues are resolved.

microprofile/WFLY-19588_microprofile_platform_7.adoc
```diff
@@ -0,0 +1,201 @@
...
+|3.0
+|3.0
+|4.0
+|y https://issues.redhat.com/browse/WFLY-19589[WFLY-19589]
+
+|Telemetry +
+|1.1
+|1.1
+|2.0
+|y https://issues.redhat.com/browse/WFLY-19589[WFLY-19589]
```
fabiobrz, line 76: This doesn't look right, as it is the same tracker as the MP REST Client upgrade. It should be https://issues.redhat.com/browse/WFLY-19846 and also https://issues.redhat.com/browse/WFLY-19590.
BTW I don't get how the latter got into `main` without undergoing the usual process.

microprofile/WFLY-19588_microprofile_platform_7.adoc
```diff
@@ -0,0 +1,201 @@
...
+=== Issue
+
+* https://issues.redhat.com/browse/WFLY-19588[WFLY-19588]
+
+=== Related Issues
+
+* https://issues.redhat.com/browse/WFLY-19592[WFLY-19592 - MicroProfile Fault Tolerance 4.1]
+* https://issues.redhat.com/browse/WFLY-19591[WFLY-19591 - MicroProfile OpenAPI 4.0]
+* https://issues.redhat.com/browse/WFLY-19589[WFLY-19589 - MicroProfile REST Client 4.0]
+* https://issues.redhat.com/browse/WFLY-19589[WFLY-19589 - MicroProfile Telemetry 2.0]
```
fabiobrz, line 93: Same as above. This doesn't look right, as it is the same tracker as the MP REST Client upgrade. It should be https://issues.redhat.com/browse/WFLY-19846 and also https://issues.redhat.com/browse/WFLY-19590.

microprofile/WFLY-19588_microprofile_platform_7.adoc
```diff
@@ -0,0 +1,201 @@
...
+=== QE Contacts
+
+* mailto:fburzigo@redhat.com[Fabio Burzigotti]
+
+=== Testing By
+// Put an x in the relevant field to indicate if testing will be done by Engineering or QE. 
+// Discuss with QE during the Kickoff state to decide this
+* [x] Engineering
+
+* [ ] QE
```
fabiobrz, line 118: I do like the fact you added this content (from _Stability_ level down to the contacts) but still, it seems the current proposal template does not include them. Could you please clarify this part?

microprofile/WFLY-19588_microprofile_platform_7.adoc
```diff
@@ -0,0 +1,201 @@
...
+
+=== Affected Projects or Components
+
+WildFly will be updated to pull in the corresponding versions of the Eclipse Microprofile specification jars, as well as the SmallRye jars and dependencies needed to implement each version.
```
fabiobrz, line 122: As far as the product is concerned, if this upgrade will end up in EAP XP 6, then also the MTA is going to be affected, as mentioned during the last call.

That being said, I realize this _might_ not be the actual place where to mention such topic. Could you please check whether or where it should be mentioned in order to be properly tracked?

microprofile/WFLY-19588_microprofile_platform_7.adoc
```diff
@@ -0,0 +1,201 @@
...
+
+
+=== Other Interested Projects
+
+=== Relevant Installation Types
+// Remove the x next to the relevant field if the feature in question is not relevant
+// to that kind of WildFly installation
+* [x] Traditional standalone server (unzipped or provisioned by Galleon)
+
+* [ ] Managed domain
```
fabiobrz, line 132: I am aware MicroProfile is standalone, and should be single-deployment related BTW. My question is, would subsystems be available for activation when the server is in domain mode as well?

microprofile/WFLY-19588_microprofile_platform_7.adoc
```diff
@@ -0,0 +1,201 @@
...
+
+* [x] OpenShift s2i
+
+* [x] Bootable jar
+
+== Requirements
+
+=== Hard Requirements
+* All components will be upgraded to the latest micro versions of the Eclipse Microprofile specification jars, as well as the SmallRye jars and dependencies needed to implement each version, as indicated in the table in the link:#overview[Overview].
```
fabiobrz, line 141: Maybe it's just me, but I am confused, since the table only shows `minor` versions. So do you mean to the last `micro` available for the `minor` version which is documented by the table above?

microprofile/WFLY-19588_microprofile_platform_7.adoc
```diff
@@ -0,0 +1,201 @@
...
+////
+Identification if any security implications that may need to be considered with this feature
+or a confirmation that there are no security implications to consider.
+////
+I don't believe there are any security implications to consider.
+
+== Test Plan
+* TCKs will be updated to the latest corresponding version. Having checked the upgrades to the individual specifications, the respective TCKs tests cover the added/changed functionality, apart from the exceptions listed below.
+* The TCKs will run at default stabiilty level
+* Where one of the specifications being upgraded needs additional testing beyond what is done by the TCK, it should be pointed out in the analysis of the individual specification upgrade.
```
fabiobrz, line 193: I think that here, being this the _umbrella_ proposal, we have a good chance to mention the need for tests that cover integration between two of the specs which are being upgraded.
A blatant example is represented by MP Fault Tolercance and MP Telemetry integration, which is mentioned in the MP 7.0 release documentation, see: https://docs.google.com/presentation/d/1gg67Gv38B8QJ9o0u_pwoiQf3NvigB_MgL7ZgabL_I2U/edit#slide=id.g2f08feab445_0_1

microprofile/WFLY-19588_microprofile_platform_7.adoc
```diff
@@ -0,0 +1,201 @@
...
+
+== Community Documentation
+* Community documentation for the subsystems in question will be updated to reflect the new versions, and any potential differences in behaviour.
```
fabiobrz, line 196: I'd add that this will be tracked by individual spec upgrade WFLY issues, as you've done in previous sections.

microprofile/WFLY-19588_microprofile_platform_7.adoc
```diff
@@ -0,0 +1,201 @@
...
+
+== Release Note Content
+MicroProfile specifications have been updated to the versions that are part of MicroProfile Platform 7.0. (See the table in the link:#overview[Overview] if exact versions are needed)
+
+MicroProfile OpenTracing has been removed from the MicroProfile 6 platform, in favour of MicroProfile Telemetry.
```
fabiobrz, line 201: Don't we know this already? I mean, we removed the integration when we upgraded the specs to 6.1, right? Or am I missing something?

microprofile/WFLY-19588_microprofile_platform_7.adoc
rhusar: There might be some 'common' documentation bit, that just says that "WildFly is MicroProfile 7.0 compliant runtime".

fabiobrz: +1, thanks for pointing this out @rhusar

microprofile/WFLY-19588_microprofile_platform_7.adoc
kabir: AFAIK only Bilge creates Epics in our process

microprofile/WFLY-19588_microprofile_platform_7.adoc
kabir: Yes, they are - AFAIK we just don't 'support' it

microprofile/WFLY-19588_microprofile_platform_7.adoc
kabir: @fabiobrz Not sure, but it is only in main at the Preview stability level (i.e. in WildFly Preview). I believe it was seen as more important to get back in a community release, rather than block merging on the community process for a preview feature

microprofile/WFLY-19588_microprofile_platform_7.adoc
kabir: @fabiobrz @rhusar IMO this belongs in the FT analysis, since it is also covered in the FT specification.

microprofile/WFLY-19588_microprofile_platform_7.adoc
fabiobrz: Thanks for fixing. Feel free to resolve.

microprofile/WFLY-19588_microprofile_platform_7.adoc
fabiobrz: Ok, thanks. Feel free to resolve.

microprofile/WFLY-19588_microprofile_platform_7.adoc
fabiobrz: Ok, thanks for clarifying. Feel free to resolve.

microprofile/WFLY-19588_microprofile_platform_7.adoc
fabiobrz: I see you updated the proposal in order to follow the latest template version, thanks. 
Feel free to resolve.

microprofile/WFLY-19588_microprofile_platform_7.adoc
fabiobrz: Ok, not a blocking comment, indeed. Feel free to resolve.

microprofile/WFLY-19588_microprofile_platform_7.adoc
fabiobrz: Thanks for fixing. Feel free to resolve.

fabiobrz (approved):
Thanks for this proposal @kabir - changes LGTM, approving from QA perspective.

darranl:
@rhusar you previously requested changes, is it possible to change your review to approved or anything outstanding?

rhusar (approved):
LGTM now (@darranl all feedback has been addressed already)

jamezp (approved):

PR #617, opened 2024-10-09
Update template and checks

bstansberry:
This adds a commit to #581 that:

1) Fixes some rendering issues (mostly avoiding writing an empty [] for existing proposals with no stability data)
2) Converts merged proposals at < default stability to use the generated stability stuff instead of hard coding it in the title


core/WFCORE-6960_Simple_config_export_for_a_server_as_an_attachment_for_standalone_or_domain_mode.adoc
```diff
@@ -2,8 +2,9 @@
 categories:
  - core
  - management
+stability-level: default
 ---
-= [Community] Simple config export for a server as an attachment for standalone or domain mode
```
bstansberry, line 6: Note that the '[Community]' here was a mistake, probably cut and paste from the earlier proposal for the community stability feature.

Issue #619, opened 2024-10-11
[MP Platform 7] MicroProfile Telemetry 2.0


jasondlee:
### Description

MicroProfile Telemetry 2.0 is targeted for EAP XP 6 and part of its MicroProfile 7 support so this must be promoted.

### Issue Contact

jasondlee

PR #624, opened 2024-10-14
[WFLY-19846] Migrate MP Tel 2.0 from PREVIEW to DEFAULT

jasondlee:
Add analysis document

Resolves https://github.com/wildfly/wildfly-proposals/issues/619

### DIFF
--- a/observability/WFLY-19846-promote-mp-tel-2-to-default.adoc
+++ b/observability/WFLY-19846-promote-mp-tel-2-to-default.adoc
@@ -1,0 +1,124 @@
+---
+categories:
+- microprofile
+- observability
+stability-level: default
+issue: https://github.com/wildfly/wildfly-proposals/issues/619
+feature-team:
+ developer: jasondlee
+ sme:
+  -
+ outside-perspective:
+  - Marek Kopecky
+  - Rahuul Chettri
+promotes: https://issues.redhat.com/browse/WFLY-19590
+promoted-by: https://github.com/wildfly/wildfly-proposals/issues/619
+---
+= [Default] Promote MicroProfile Telemetry 2.0 to WildFly Standard
+:author:            Jason Lee
+:email:             jasondlee@redhat.com
+:toc:               left
+:icons:             font
+:idprefix:
+:idseparator:       -
+
+== Overview
+
+MicroProfile Telemetry 2.0 shipped with WildFly Preview 34. To meet WildFly and product roadmap goals, this feature needs to be migrated to the WildFly feature pack, and its stability level promoted to `DEFAULT`.
+
+=== User Stories
+
+User wants to deploy an application using MicroProfile Telemetry 2.0, which adds support for OpenTelemetry Metrics and OpenTelemetry Logging, as well as updating underlying OpenTelemetry library versions, which brings in bug fixes and miscellaneous features.
+
+== Issue Metadata
+
+- https://issues.redhat.com/browse/WFLY-19846[]
+
+=== Related Issues
+
+- https://issues.redhat.com/browse/WFLY-19835[WFLY-19835 - Integrate OpenTelemetry with the MP Reactive Messaging Kafka connector]
+- https://issues.redhat.com/browse/WFLY-19836[WFLY-19836 - Integrate OpenTelemetry with the MP Reactive Messaging AMQP connector]
+
+=== Affected Projects or Components
+
+- MicroProfile Fault Tolerance
+- AMQP
+- Kafka
+
+=== Other Interested Projects
+
+=== Relevant Installation Types
+
+* Traditional standalone server (unzipped or provisioned by Galleon)
+* Managed domain
+* OpenShift Source-to-Image (S2I)
+* Bootable jar
+
+== Requirements
+
+* WildFly must fully support MicroProfile Telemetry 2.0:
+** Tracing - Existing support
+** Metrics - New feature
+** Logs - New feature
+* OpenTelemetry versions must be updated to match those required by MicroProfile Telemetry 2.0
+
+=== Changed requirements
+
+N/A
+
+=== Non-Requirements
+
+N/A
+
+=== Future Work
+
+* Add a quickstart covering the new spec
+
+== Backwards Compatibility
+
+* There are no known breaking API changes. The underlying update to OpenTelemetry 1.39.0, though, does update the semantic conventions used in describing the exported spans. Users updating their applications and environments to MicroProfile Telemetry 2.0 should expect some breakage in their dashboards, as the span attributes may have changed as a result of the upgrade. See https://github.com/open-telemetry/semantic-conventions/blob/main/docs/non-normative/http-migration.md[]
+* One other note: If a deployment makes use of the environment variable `OTEL_SDK_DISABLED`, the specified behavior has changed: https://microprofile.io/specifications/telemetry/2-0/[]
+
+=== Default Configuration
+
+* The default configuration will be unchanged and continue to work as is.
+
+=== Deployments
+
+Deployment behavior will be unaffected.
+
+=== Interoperability
+
+* There are no Interoperability issues, per se, but this update does introduce functionality (OpenTelemetry Metrics) that overlaps functionality provided by both the WildFly Metrics and Micrometer subsystems. Having multiple metrics libraries enabled simultaneous is likely unwanted due to -- if nothing else -- the performance implications of duplicate metrics gathering. This overlap and steps for mitigation should be discussed in the docs. 
+
+== Implementation Plan
+
+N/A. Will be delivered in one release.
+
+== Admin Clients
+
+N/A
+
+== Security Considerations
+
+This change will have no effect on security. As stated in the MicroProfile Telemetry 1.0, there is no baked in notion of security with OpenTelemetry, so systems administrators will need to make sure that the endpoint for the push publication is properly configured, and that the receiving system -- whatever it may, e.g. and OpenTelemetry Collector -- is secured as desired and appropriate for that system. Such a configuration is outside the scope of the work here.
+
+[[test_plan]]
+== Test Plan
+
+There are a number of tests already in WildFly and various downstream repositories. The WildFly test suite will be modified to add test coverage for the newly added observability signals, metrics and logging. Those tests reside in `testsuite/integration/microprofile`.
+
+Additionally, the MicroProfile Telemetry TCK, located in `testsuite/integration/microprofile-tck/telemetry`, will be updated to run the new MicroProfile Telemetry 2.0 TCK.
+
+== Community Documentation
+
+MicroProfile Telemetry already has some documentation in the community guides covering the tracing aspect. This documentation will be updated to cover metrics and logging as well.
+
+For more information on getting started with these new features, see the MicroProfile Telemetry specification:
+
+* https://download.eclipse.org/microprofile/microprofile-telemetry-2.0/microprofile-telemetry-spec-2.0.html#metrics[Metrics]
+* https://download.eclipse.org/microprofile/microprofile-telemetry-2.0/microprofile-telemetry-spec-2.0.html#logs[Logs]
+
+== Release Note Content
+
+"MicroProfile Telemetry support in WildFly has been updated to version 2.0. This new release brings bug fixes and updates to OpenTelemetry tracing support, as well as adding support for metrics and logging."


observability/WFLY-19846-promote-mp-tel-2-to-default.adoc
```diff
@@ -0,0 +1,96 @@
...
+
+Additionally, the MicroProfile Telemetry TCK, located in `testsuite/integration/microprofile-tck/telemetry`, will be updated to run the new MicroProfile Telemetry 2.0 TCK.
+
+== Community Documentation
+
+MicroProfile Telemetry already has some documentation in the community guides covering the tracing aspect. This documentation will be updated to cover metrics and logging as well.
+
+== Release Note Content
+
+"MicroProfile Telemetry support in WildFly has been update to version 2.0. This new release brings bug fixes and updates to OpenTelemetry tracing support, as well as adding support for metrics and logging."
```
rhusar, line 96: s/update/updated

rhusar (changes requested):

observability/WFLY-19846-promote-mp-tel-2-to-default.adoc
```diff
@@ -0,0 +1,96 @@
...
+stability-level: default
+issue: https://github.com/wildfly/wildfly-proposals/issues/619
+feature-team:
+ developer: jasondlee
+ sme:
+  - jasondlee
+ outside-perspective:
+  -
+promotes: https://issues.redhat.com/browse/WFLY-19590
+promoted-by: https://issues.redhat.com/browse/WFLY-19846
```
jamezp, line 14: These should be the link to the GitHub issue. The idea will be to have tooling to create a link between the two generated documents.

That said, I know the old issue, likely doesn't have a GitHub Issue associated with it. I'm not sure what we do in that case. However, of the `promoted-by` that should be a link to the GitHub Issue.

observability/WFLY-19846-promote-mp-tel-2-to-default.adoc
```diff
@@ -0,0 +1,96 @@
...
+:idseparator:       -
+
+== Overview
+
+MicroProfile Telemetry 2.0 shipped with WildFly Preview 34. To meet WildFly and product roadmap goals, this feature needs to be migrated to the WildFly feature pack, and its stability level promoted to `DEFAULT`.
+
+=== User Stories
+
+__<Provide one or more brief user stories that illustrate the intended users of
+ the feature and the goal they will seek to achieve by using the feature.>__
```
jamezp, line 31: We should remove this if it's not going to be used.

observability/WFLY-19846-promote-mp-tel-2-to-default.adoc
```diff
@@ -0,0 +1,96 @@
...
+
+== Issue Metadata
+
+- https://issues.redhat.com/browse/WFLY-19846[]
+
+=== Related Issues
+
+- https://issues.redhat.com/browse/EAP7-2227[]
```
jamezp, line 39: I can't recall if there was a decision to stop adding these links in the analysis doc. @darranl, @bstansberry or @jmesnil do you recall?

rhusar: Good question, these are useless for community. They are just convenience for us so we don't have to open the main issue and then click the linked EAP7 issue.

observability/WFLY-19846-promote-mp-tel-2-to-default.adoc
rhusar: How about:

> User wants to deploy an application using MicroProfile Telemetry 2.0 from MicroProfile Platform 7.0.

At least that's what I am putting in my proposal.

observability/WFLY-19846-promote-mp-tel-2-to-default.adoc
jasondlee: Updated.

observability/WFLY-19846-promote-mp-tel-2-to-default.adoc
rhusar: Excellent!

observability/WFLY-19846-promote-mp-tel-2-to-default.adoc
```diff
@@ -0,0 +1,115 @@
...
+
+=== User Stories
+
+User wants to deploy an application using MicroProfile Telemetry 2.0 from MicroProfile Platform 7.0.
+
+== Issue Metadata
+
+- https://issues.redhat.com/browse/WFLY-19846[]
+
+=== Related Issues
```
marekkopecky, line 36: I noticed that telemetry PR adds grpc dependency [here](https://github.com/wildfly/wildfly/pull/18305/files#diff-0309302dfed110d75a04cb208108f436ddebfb77aed51fc8bca7d4155c7fdb4aR31). Is this telemetry RFE blocked by grpc RFE? (https://issues.redhat.com/browse/EAP7-1819 / https://issues.redhat.com/browse/WFLY-15495 ) If yes, I believe that it should be stressed out here.

rhusar: @marekkopecky No, it doesn't, this is just an internal dependency.

observability/WFLY-19846-promote-mp-tel-2-to-default.adoc
```diff
@@ -0,0 +1,115 @@
...
+:idprefix:
+:idseparator:       -
+
+== Overview
+
+MicroProfile Telemetry 2.0 shipped with WildFly Preview 34. To meet WildFly and product roadmap goals, this feature needs to be migrated to the WildFly feature pack, and its stability level promoted to `DEFAULT`.
+
+=== User Stories
+
+User wants to deploy an application using MicroProfile Telemetry 2.0 from MicroProfile Platform 7.0.
```
marekkopecky, line 30: Can you please elaborate how users can benefit from new features brought by MP Telemetry 2.0? https://microprofile.io/specifications/telemetry/2-0/ describes new features against MP Telemetry 1.1 that we have earlier

observability/WFLY-19846-promote-mp-tel-2-to-default.adoc
```diff
@@ -0,0 +1,115 @@
...
+- https://issues.redhat.com/browse/WFLY-19846[]
+
+=== Related Issues
+
+- https://issues.redhat.com/browse/EAP7-2227[]
+
+=== Affected Projects or Components
+
+- MicroProfile Fault Tolerance
+- AMQP
```
marekkopecky, line 43: How this WF feature affects AMQP or Kafka?

observability/WFLY-19846-promote-mp-tel-2-to-default.adoc
```diff
@@ -0,0 +1,115 @@
...
+
+Currently, neither OpenTelemetry nor MicroProfile Telemetry are supported in domain mode. While this is now technically feasible, given the push nature of the export process, support for this will have to be delivered under a different (to-be-file) feature request.
+
+=== Future Work
+
+N/A
+
+== Backwards Compatibility
+
+* There are no known breaking changes. The upstream OpenTelemetry project made a number of breaking changes to the metrics semantic conventions (which could affect user dashboards), but metrics support will be new with this release, so those changes will not affect existing installations.
```
marekkopecky, line 76: MP Telemetry main page explicitelly mentioned two backward incompatible changes against MP telemetry 1.1 that we used earlier. Can you please double-check this?

https://microprofile.io/specifications/telemetry/2-0/

observability/WFLY-19846-promote-mp-tel-2-to-default.adoc
```diff
@@ -0,0 +1,115 @@
...
+
+=== Default Configuration
+
+* The default configuration will be unchanged and continue to work as is.
+
+=== Deployments
+
+Deployment behavior will be unaffected.
```
marekkopecky, line 84: It seems that any work/checks for some env variables needs to be changed, see "Incompatible changes" section in https://microprofile.io/specifications/telemetry/2-0/

observability/WFLY-19846-promote-mp-tel-2-to-default.adoc
```diff
@@ -0,0 +1,115 @@
...
+N/A
+
+== Security Considerations
+
+This change will have no effect on security. As stated in the MicroProfile Telemetry 1.0, there is no baked in notion of security with OpenTelemetry, so systems administrators will need to make sure that the endpoint for the push publication is properly configured, and that the receiving system -- whatever it may, e.g. and OpenTelemetry Collector -- is secured as desired and appropriate for that system. Such a configuration is outside the scope of the work here.
+
+[[test_plan]]
+== Test Plan
+
+There are a number of tests already in WildFly and various downstream repositories. The WildFly test suite will be modified to add test coverage for the newly added observability signals, metrics and logging. Those tests reside in `testsuite/integration/microprofile`.
```
marekkopecky, line 105: Are you going to add tests for observability signals and logging to https://github.com/wildfly/wildfly/pull/18305/files ? Or is there some another tracker for this?

observability/WFLY-19846-promote-mp-tel-2-to-default.adoc
```diff
@@ -0,0 +1,115 @@
...
+== Admin Clients
+
+N/A
+
+== Security Considerations
+
+This change will have no effect on security. As stated in the MicroProfile Telemetry 1.0, there is no baked in notion of security with OpenTelemetry, so systems administrators will need to make sure that the endpoint for the push publication is properly configured, and that the receiving system -- whatever it may, e.g. and OpenTelemetry Collector -- is secured as desired and appropriate for that system. Such a configuration is outside the scope of the work here.
+
+[[test_plan]]
+== Test Plan
```
marekkopecky, line 103: Can you please add "Manual tests", "Miscellaneous checks" and "Compatibility tests" sections here according to the template? https://github.com/wildfly/wildfly-proposals/blob/790b15b89999850f151e25190266a44d8e16ab35/design-doc-template.adoc?plain=1#L145

marekkopecky:
@jasondlee Thank you for this analysis! I left a few comments here and there, I would appreciate if you can provide some feedback on them.

observability/WFLY-19846-promote-mp-tel-2-to-default.adoc
marekkopecky: Thank you for your feedback! I'm not sure whether I understand. Is this dependency in module.xml not used in module.xml in built WF?

jasondlee: @marekkopecky That is an internal module dependency that is being added independently of the proposed gRPC subsystem. At some point, we'll have to work out the overlapping module definitions between the two, but there is no relationship at the moment.

observability/WFLY-19846-promote-mp-tel-2-to-default.adoc
kabir: @marekkopecky I think he is referring to https://issues.redhat.com/browse/WFLY-19835 and https://issues.redhat.com/browse/WFLY-19836 which is about integrating the MP RM AMQP and Kafka connectors with Otel. That integration requires dependencies brought in via MP Telemetry 2.0, and will not work with the existing MP Telemetry 1.x deps

observability/WFLY-19846-promote-mp-tel-2-to-default.adoc
jasondlee: Ah, right. I was thinking API changes, but that semantic convention change is worth noting. I don't think the env var is of too much interest to WF directly. I'll update the language here.

observability/WFLY-19846-promote-mp-tel-2-to-default.adoc
jasondlee: We don't have any checks for the env var in the WF code base. The SR test suites might test that explicitly, but we do not.

observability/WFLY-19846-promote-mp-tel-2-to-default.adoc
jasondlee: I did add a test for metrics here: https://github.com/wildfly/wildfly/pull/18305/files#diff-d5cdf7a6dbc86418242e90a663cc9c18317efa2ee3eb12f2757749dccd049013

Logging functionality seems pretty limited in the otel spec itself, but it is tested via the MP Telemetry TCK. I don't see any value we might be able to add on top of the logging tests there.

observability/WFLY-19846-promote-mp-tel-2-to-default.adoc
jasondlee: If you have something specific you'd like to see add, please describe that for me. I removed the section as I didn't see anything relevant. Perhaps different eyes will, of course, but these sections were deemed superfluous and removed.

observability/WFLY-19846-promote-mp-tel-2-to-default.adoc
jasondlee: Removed and replaced with related WFLY links

observability/WFLY-19846-promote-mp-tel-2-to-default.adoc
marekkopecky: @jasondlee Thank you for the update. What about integration between Telemetry and Fault Tolerance? My understanding is that WF users can benefit from it as well?

I would like to explain my motivation in this discussion, because I understand that this could be considered as nit-picking. I'm not sure about original reason why "User stories" section has been added to the template for these analysis. But I can guess - I would say that identification of important user stories is useful for later parts of process, specifically for documentation. I really don't want to push on process/analysis without no reasons, but I actually believe that early identification of user-stories can improve documentation, which will be useful of users at the end.

marekkopecky: AFAIU server is going to implement micrometer metrics and telemetry metrics. Telemetry seems to be newer and also is a part of spec. Possible user story could be also to migrate micrimeter metrics to telemetry metrics?

marekkopecky:
Can you please elaborate why it's useful to keep micrometer basic metrics, micrometer metrics and telemetry metrics? Could all these metrics in server be confusing for users?

jasondlee:
> Can you please elaborate why it's useful to keep micrometer basic metrics, micrometer metrics and telemetry metrics? Could all these metrics in server be confusing for users?

Potentially, yes, but both libraries have been offered for some time before they began overlapping in functionality as they do now. There (slow-moving) discussions on how to handle that, but, for now, the decision is to offer both and allow the users to choose.

observability/WFLY-19846-promote-mp-tel-2-to-default.adoc
jasondlee: > @jasondlee Thank you for the update. What about integration between Telemetry and Fault Tolerance? My understanding is that WF users can benefit from it as well? cc @fabiobrz

MP Fault Tolerance is listed in the `Affected Projects or Components` section. I don't see the need to list out every feature of the technology.

marekkopecky: Thank you for your feedback! I tried to explain the motivation behind this above. If that's not a reason for you, we can consider this topic as resolved (in scope of analysis) and handle documentation separately in later RFE phases.

observability/WFLY-19846-promote-mp-tel-2-to-default.adoc
marekkopecky: Thank you for your feedback! AFAIU user can use default WF Metrics, MP Telemetry metrics, MicroMeter metrics. But AFAIU it's not recommended to use more then one metrics together. What about to add manual check whether the server prints proper warning if more metrics systems are use together? My understanding is that such situation is not recommended based on [WF admin guide](https://docs.wildfly.org/34/Admin_Guide.html#Micrometer_Metrics): 

> Note that this an alternative to the existing WildFly Metrics extension. While they may be run concurrently, it is not advisable, as this will likely have an impact on server performance due to the duplicated metrics collection.

marekkopecky:
What is the plan about default behaviour? Are both WF metrics and MP Telemetry Metrics enabled by default?

observability/WFLY-19846-promote-mp-tel-2-to-default.adoc
jasondlee: > AFAIU server is going to implement micrometer metrics and telemetry metrics. Telemetry seems to be newer and also is a part of spec. Possible user story could be also to migrate micrimeter metrics to telemetry metrics?

I think migration only makes sense if we're encouraging or requiring users to move, which we are not at this point.

jasondlee:
> So if users want to use micrometer now, they need to disable MP telemetry metrics, right? If yes, this affects migration from previous WF versions to latest WF version and needs to be elaborate in the analysis document itself.

Yes. I've added some language to the AD to have this addressed in the community docs.

jasondlee:
> What is the plan about default behaviour? Are both WF metrics and MP Telemetry Metrics enabled by default?

Yes, if the admin starts with server using any of the MicroProfile configurations. See note above about updating the community docs.

jasondlee:
@marekkopecky I can't seem to add you to the `wildfly` PR, so here's a link to the docs update: https://github.com/wildfly/wildfly/pull/18357

observability/WFLY-19846-promote-mp-tel-2-to-default.adoc
marekkopecky: +1, thank you for your feedback!

observability/WFLY-19846-promote-mp-tel-2-to-default.adoc
jasondlee: @marekkopecky We currently document that running multiple is not wise, and I think that's sufficient for now. We can certainly add a warning, but that's not just a single line change here or there, so, if we want to do that, that will have to be an RFE in a future release.

observability/WFLY-19846-promote-mp-tel-2-to-default.adoc
```diff
@@ -0,0 +1,122 @@
...
+- AMQP
+- Kafka
+
+=== Other Interested Projects
+
+=== Relevant Installation Types
+
+* Traditional standalone server (unzipped or provisioned by Galleon)
+* OpenShift Source-to-Image (S2I)
+* Bootable jar
```
rhusar, line 54: I think we decided to keep managed domain in the list.

rhusar (approved):

observability/WFLY-19846-promote-mp-tel-2-to-default.adoc
marekkopecky: Thank you for your feedback! I understand your PoV, but I should mainly take care about the quality, so I created an issue for this, so it can be triaged separately.

jasondlee: That's fine. As I noted, it's not a simple, one line change, and it will affect multiple modules, so that change should take some more time to make sure it's done correctly, and definitely should done under another JIRA. Thanks for filing that.

observability/WFLY-19846-promote-mp-tel-2-to-default.adoc
fabiobrz: @marekkopecky we've been discussing this settings in the community proposal lately. It is controversial, because MP apps would naturally be executed by a standalone server, specifically on K8s/OpenShift - were HA is not just about WildFly/EAP. But we are aware that nothing would technically block a user from enabling the MP extensions on a domain mode server. 

This is why we started to add the domain mode to the list of `Relevant Installation Types` - it is indeed relevant since the feature can be executed, although it might not be supported.

observability/WFLY-19846-promote-mp-tel-2-to-default.adoc
bstansberry: @jasondlee What do you mean by RFE? A Feature Request for a new feature at default stability? Or do you mean an Enhancement, not involving any user controllable behavior?

jasondlee: > @jasondlee What do you mean by RFE? A Feature Request for a new feature at default stability? Or do you mean an Enhancement, not involving any user controllable behavior?

@bstansberry Sorry. I meant an Enhancement.

observability/WFLY-19846-promote-mp-tel-2-to-default.adoc
```diff
@@ -0,0 +1,124 @@
...
+** Logs - New feature
+* OpenTelemetry versions must be updated to match those required by MicroProfile Telemetry 2.0
+
+=== Changed requirements
+
+N/A
+
+=== Non-Requirements
+
+Currently, neither OpenTelemetry nor MicroProfile Telemetry are supported in domain mode. While this is now technically feasible, given the push nature of the export process, support for this will have to be delivered under a different (to-be-file) feature request.
```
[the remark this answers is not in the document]
jasondlee: You're right. I can't remember when I wrote this, but I know we've talked about domain mode and metrics in one context or another, and I think I'm way off here. I'll just delete this block, as there's no need for it at all.

observability/WFLY-19846-promote-mp-tel-2-to-default.adoc
```diff
@@ -0,0 +1,124 @@
...
+* There are no known breaking API changes. The underlying update to OpenTelemetry 1.39.0, though, does update the semantic conventions used in describing the exported spans. Users updating their applications and environments to MicroProfile Telemetry 2.0 should expect some breakage in their dashboards, as the span attributes may have changed as a result of the upgrade. See https://github.com/open-telemetry/semantic-conventions/blob/main/docs/non-normative/http-migration.md[]
+* One other note: If a deployment makes use of the environment variable `OTEL_SDK_DISABLED`, the specified behavior has changed: https://microprofile.io/specifications/telemetry/2-0/[]
+
+=== Default Configuration
+
+* The default configuration will be unchanged and continue to work as is.
+
+=== Deployments
+
+Deployment behavior will be unaffected.
```
marekkopecky, line 88: AFAIU (cc @emmartins )  work for https://github.com/konveyor/rulesets should be mentioned in this section of the analysis

Issue #631, opened 2024-10-17
Add support for webservice throttling feature


jimma:
### Description

CXF provides the throttling feature to limit the request traffic to the webservice endpoint. This is to add support to integrate and enable this throttling feature in WFLY webservice subsystem.

### Issue Contact

jimma

PR #632, opened 2024-10-17
[WFLY-19469]:Add support for webservice throttling feature

jimma:
Analysis document for : https://issues.redhat.com/browse/WFLY-19469
Resolves : https://github.com/wildfly/wildfly-proposals/issues/631

### DIFF
--- a/webservices/WFLY-19469-Webservice-throttling-feature.adoc
+++ b/webservices/WFLY-19469-Webservice-throttling-feature.adoc
@@ -1,0 +1,144 @@
+---
+categories:
+- webservices
+stability-level: community
+issue: https://github.com/wildfly/wildfly-proposals/issues/631
+feature-team:
+ developer: ema
+ sme:
+  - ema
+ outside-perspective:
+  -
+promotes:
+promoted-by:
+---
+= Add support for webservice throttling feature
+:author:            Jim Ma
+:email:             ema@redhat.com
+:toc:               left
+:icons:             font
+:idprefix:
+:idseparator:       -
+
+== Overview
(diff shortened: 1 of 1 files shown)


webservices/WFLY-19469-Webservice-throttling-feature.adoc
```diff
@@ -0,0 +1,144 @@
...
+== Requirements
+* WildFly webservice subsystem is installed/provisioned and enabled
+* This feature is enabled and configured with jaxws-endpoint-config.xml packaged in user's deployment
+[source,xml]
+ <?xml version="1.0" encoding="UTF-8"?>
+<jaxws-config xmlns="urn:jboss:jbossws-jaxws-config:4.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns:javaee="http://java.sun.com/xml/ns/javaee"
+  xsi:schemaLocation="urn:jboss:jbossws-jaxws-config:4.0 schema/jbossws-jaxws-config_4_0.xsd">
+  <endpoint-config>
+    <config-name>org.jboss.test.ws.jaxws.cxf.throttling.HelloImpl</config-name>
+    <property>
```
darranl, line 60: Where do these properties get loaded / interpreted?

If this feature is targeting the community stability level my assumption would be that they are disabled at the default stability level - will that be the case?

jimma: > Where do these properties get loaded / interpreted?

This is loaded and parsed when deploy this deployment. 

> If this feature is targeting the community stability level my assumption would be that they are disabled at the default stability level - will that be the case?

This feature actually is disabled if user doesn't include this deployment descriptor, so I set it to community level.
Should it be default level ?

PR #653, opened 2024-10-24
Feature process template doc: add quickstarts as typical test to include in manual testing.

rhusar:
@bstansberry @darranl After lengthy discussions with @fabiobrz and @marekkopecky about manual testing, I have realized that part of the manual verification we do for every MicroProfile related RFE and that cannot be automated, is testing a corresponding existing quick start. Here's a suggestion to add this to the template so that people are reminded to do that and add that to the list (case in point, I forgot about it). Note this doesn't only pertain to MP, we have recently used a quickstart to validate clustering RFEs.


asoldano (approved):

rhusar:
Thanks @bstansberry !

Issue #657, opened 2024-10-29
Verify signatures of WildFly components during provisioning and updates.


spyrkob:
### Description

The wildfly-channels and prospero allow to provision and update WildFly server using components downloaded from Maven repositories.

This proposal aims to add support for validating detached GPG signatures of those components to ensure the origin of downloaded artifacts and protect against malicious artifacts.

### Issue Contact

spyrkob

PR #658, opened 2024-10-30
[#657] Enable signature validation of components during provisioning

spyrkob:
Fixes: #657


spyrkob:
@jmesnil can I put you as SME on this RFE?

PR #665, opened 2024-11-20
[WFLY-19954] Add the vertx extension/subsystem from wildfly-vertx-feature-pack to WildFly Preview Feature Pack

### LINKED ISSUE #662: Add the vertx extension/subsystem from wildfly-vertx-feature-pack to WildFly Preview Feature Pack
### Description

There is a discussion in [smallrye-reactive-messaging to configure vertx options](https://github.com/smallrye/smallrye-reactive-messaging/discussions/2725) to ask the capability to configure the vertx options used in the microprofile-reactive-messaging subsystem, and there is a [wildfly-vertx-feature-pack](https://github.com/wildfly-extras/wildfly-vertx-feature-pack) ready to fulfill the requirement, integrate it to WildFly should satisfy the requirement.

More discussions can be found at:
* https://wildfly.zulipchat.com/#narrow/channel/174184-wildfly-developers/topic/vertx.20subsystem.20integration
* https://lists.jboss.org/archives/list/wildfly-dev@lists.jboss.org/thread/CHUJ5GZSHXUSVZNXS3X32H7UQ4M64NKC/

### Issue Contact

gaol

gaol:
Fixes #662

### DIFF
--- a/microprofile/WFLY-19954_Preview_Support_vertx_feature_pack.adoc
+++ b/microprofile/WFLY-19954_Preview_Support_vertx_feature_pack.adoc
@@ -1,0 +1,232 @@
+---
+categories:
+  - microprofile
+stability-level: preview
+issue: https://github.com/wildfly/wildfly-proposals/issues/662
+feature-team:
+ developer: gaol
+ sme:
+  - kabir
+  - fl4via
+  - jasondlee
+ outside-perspective:
+  -
+promotes:
+promoted-by:
+---
+
+= Add the vertx extension/subsystem from wildfly-vertx-feature-pack to WildFly Preview Feature Pack
+:author:            Lin Gao
+:email:             lgao@redhat.com
+:toc:               left
+:icons:             font
+:idprefix:
+:idseparator:       -
+:wildfly_vertx_feature_pack_with_link: https://github.com/wildfly-extras/wildfly-vertx-feature-pack[wildfly-vertx-feature-pack]
+:vertx_options_with_link: https://github.com/eclipse-vertx/vert.x/blob/4.5.10/src/main/java/io/vertx/core/VertxOptions.java[VertxOptions]
+:filesystem_options_with_link: https://github.com/eclipse-vertx/vert.x/blob/4.5.10/src/main/java/io/vertx/core/file/FileSystemOptions.java[FileSystemOptions]
+:address_resolver_options_with_link: https://github.com/eclipse-vertx/vert.x/blob/4.5.10/src/main/java/io/vertx/core/dns/AddressResolverOptions.java[AddressResolverOptions]
+== Overview
+
+The demand comes from https://github.com/smallrye/smallrye-reactive-messaging/discussions/2725[a discussion in smallrye-reactive-messaging] to ask the capability to configure the vertx options used in the microprofile-reactive-messaging subsystem, and there is a {wildfly_vertx_feature_pack_with_link} ready to fulfill, integrate it to WildFly should satisfy the requirement.
+
+The goal of the {wildfly_vertx_feature_pack_with_link} is to be the central place to provide the Vertx instance within WildFly server for other subsystems which may need it, example subsystems are https://github.com/wildfly/wildfly/tree/main/microprofile/telemetry-smallrye[microprofile-telemetry] and https://github.com/wildfly/wildfly/tree/main/microprofile/reactive-messaging-smallrye[microprofile-reactive-messaging]. Currently, they create their own vertx instances (microprofile-reactive-messaging creates one per deployment). Advice from the VertX team is to have only one shared instance for everything.
+
+=== User Stories
+
+The Vertx instance used by microprofile-reactive-messaging subsystem is constructed by: `Vertx.vertx()`, which uses the default {vertx_options_with_link}. Some users may want to customize the options used to construct the Vertx instance for their own reasons. Like updating `maxEventLoopExecuteTime` to control how long Vertx will log a warning message when it detects that the event loop threads are blocked; or updating `eventLoopPoolSize` to control how many event loop threads will be created; or updating `preferNativeTransport` on if native transport is preferred on any Vertx I/O operations, and more.
+
+== Issue Metadata
+
+=== Issue
+* https://issues.redhat.com/browse/WFLY-19954 - Add the vertx extension/subsystem from wildfly-vertx-feature-pack to WildFly Preview Feature Pack
+
+=== Related Issues
+* https://github.com/smallrye/smallrye-reactive-messaging/discussions/2725
+* https://github.com/smallrye/smallrye-reactive-messaging/issues/2771
+* https://github.com/smallrye/smallrye-opentelemetry/issues/380
+
+=== Stability Level
+* [ ] Experimental
+* [X] Preview
+* [ ] Community
+* [ ] default
+
+=== Dev Contacts
+* mailto:{email}[{author}]
+
+=== QE Contacts
+* TBD
+
+=== Testing By
+* [x] Engineering
+* [ ] QE
+
+=== Affected Projects or Components
+* https://github.com/wildfly/wildfly
+* https://github.com/wildfly-extras/wildfly-vertx-feature-pack
+
+=== Other Interested Projects
+* https://github.com/smallrye/smallrye-reactive-messaging
+* https://github.com/smallrye/smallrye-opentelemetry
+
+=== Relevant Installation Types
+// Remove the x next to the relevant field if the feature in question is not relevant to that kind of WildFly installation
+* [x] Traditional standalone server (unzipped or provisioned by Galleon)
+* [x] Managed domain
+* [x] OpenShift s2i
+* [x] Bootable jar
+
+== Requirements
+
+=== Hard Requirements
+As normally it is sufficient to have one Vertx instance for everything, the vertx subsystem provides either nothing or one Vertx instance according to the configuration. Users can set up a Vertx instance using WildFly management model, and it is exposed to the CDI container, other subsystems that want to consume it can opt in.
+
+The vertx extension and subsystem will be integrated to `preview/feature-pack` in `Preview` stability level.
+
+==== Configure the {vertx_options_with_link} in vertx subsystem
+The {vertx_options_with_link} is used to customize the options to construct a Vertx instance. The following fields in primitive types are selected to configure as attributes in the vertx subsystem with address `/subsystem=vertx/vertx-option=xx`:
+
+.Attributes in address: `/subsystem=vertx/vertx-option=xxx`
+[frame=all, grid=all]
+|===
+^| Field Name ^| Data Type ^| Source Class ^| Resource Attribute Name ^| Note
+
+| eventLoopPoolSize | int | {vertx_options_with_link} | event-loop-pool-size | number of event loop threads to be used
+| workerPoolSize | int | {vertx_options_with_link} | worker-pool-size | number of threads in the worker pool
+| internalBlockingPoolSize | int | {vertx_options_with_link} | internal-blocking-pool-size | number of threads in the internal blocking pool
+| preferNativeTransport | boolean | {vertx_options_with_link} | prefer-native-transport | If native transport is preferred or not
+| blockedThreadCheckInterval | long | {vertx_options_with_link} | blocked-thread-check-interval | blocked thread check interval
+| blockedThreadCheckIntervalUnit | TimeUnit | {vertx_options_with_link} | blocked-thread-check-interval-unit | blocked thread check interval time unit
+| maxEventLoopExecuteTime | long | {vertx_options_with_link} | max-eventloop-execute-time | max event loop thread execution time
+| maxEventLoopExecuteTimeUnit | TimeUnit | {vertx_options_with_link} | max-eventloop-execute-time-unit | max event loop thread execution time unit
+| maxWorkerExecuteTime | long | {vertx_options_with_link} | max-worker-execute-time | max worker thread execution time
+| maxWorkerExecuteTimeUnit | TimeUnit | {vertx_options_with_link} | max-worker-execute-time-unit | max worker thread execution time unit
+| warningExceptionTime | long | {vertx_options_with_link} | warning-exception-time | max time a thread can be blocked before stack traces get logged
+| warningExceptionTimeUnit | TimeUnit | {vertx_options_with_link} | warning-exception-time-unit | the time unit for warningExceptionTime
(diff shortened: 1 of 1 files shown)


gaol:
@kabir @jasondlee @fl4via may I ask you to review ? thanks :)

microprofile/WFLY-19954_Preview_Support_vertx_feature_pack.adoc
```diff
@@ -0,0 +1,234 @@
...
+N/A
+
+=== Deployments
+N/A
+
+=== Interoperability
+N/A
+
+== Implementation Plan
+The {wildfly_vertx_feature_pack_with_link} repository is hosted under wildfly-extras github organization. It contains the source codes for all subsystem functionalities, unit tests and integration tests, including the tests on the vertx subsystem management model operations.
```
jasondlee, line 214: Is the goal of this effort to move the sources from the external FP into the WF source tree?

gaol: No, the source code will stay in the external FP.

microprofile/WFLY-19954_Preview_Support_vertx_feature_pack.adoc
```diff
@@ -0,0 +1,234 @@
...
+  :reload
+else
+  echo INFO: vertx subsystem already in configuration, subsystem not added.
+end-if
+----
+
+There is no Vertx instance set up by default, users need to do it explicitly like:
+[source,shell]
+----
+/subsystem=vertx/vertx=vertx:add()
```
gaol, line 201: @jasondlee @kabir Do you see any options that we need to stay using the default ones ?  

I am not sure if the following options should be updated by users or not:
* `disableTCCL` - defaults to `false`, so a context will be associated to the TCCL when it was created.
* `useDaemonThread` - defaults to `false`, so all Vertx threads are not daemons.

Also do you see some system properties that we need to set before constructing the Vertx instance in microprofile-opentelemetry and microprofile-reactive-messaging subsystem ?

One example I see coming from opentelemetry subsystem[1]:
```java
        // We need to disable vertx's DNS resolver as it causes failures under k8s
        if (System.getProperty(VERTX_DISABLE_DNS_RESOLVER) == null) {
            System.setProperty(VERTX_DISABLE_DNS_RESOLVER, "true");
        }
```

[1] https://github.com/wildfly/wildfly/blob/34.0.0.Final/observability/opentelemetry/src/main/java/org/wildfly/extension/opentelemetry/OpenTelemetrySubsystemRegistrar.java#L184

microprofile/WFLY-19954_Preview_Support_vertx_feature_pack.adoc
```diff
@@ -0,0 +1,234 @@
...
+| maxWorkerExecuteTimeUnit | TimeUnit | {vertx_options_with_link} | max-worker-execute-time-unit | max worker thread execution time unit
+| warningExceptionTime | long | {vertx_options_with_link} | warning-exception-time | max time a thread can be blocked before stack traces get logged
+| warningExceptionTimeUnit | TimeUnit | {vertx_options_with_link} | warning-exception-time-unit | the time unit for warningExceptionTime
+| disableTCCL | boolean | {vertx_options_with_link} | disable-tccl | whether Vert.x sets the Context classloader as the thread context classloader on actions executed on that Context.
+| useDaemonThread | int | {vertx_options_with_link} | use-daemon-thread | whether we want to use daemon vertx thread
+| classPathResolvingEnabled | boolean | {filesystem_options_with_link} | classpath-resolving-enabled | whether classpath resolving is enabled
+| fileCachingEnabled | boolean | {filesystem_options_with_link} | file-cache-enabled | whether file caching is enabled for class path resolving
+| fileCacheDir | String | {filesystem_options_with_link} | file-cache-dir | file cache directory
```
gaol, line 110: The default value is `System.getProperty("java.io.tmpdir", ".") + "/vertx-cache"`, shall we update the default cache directory to `${jboss.server.temp.dir}/vertx-cache` ?

microprofile/WFLY-19954_Preview_Support_vertx_feature_pack.adoc
```diff
@@ -0,0 +1,234 @@
...
+:filesystem_options_with_link: https://github.com/eclipse-vertx/vert.x/blob/4.5.10/src/main/java/io/vertx/core/file/FileSystemOptions.java[FileSystemOptions]
+:address_resolver_options_with_link: https://github.com/eclipse-vertx/vert.x/blob/4.5.10/src/main/java/io/vertx/core/dns/AddressResolverOptions.java[AddressResolverOptions]
+== Overview
+
+The demand comes from https://github.com/smallrye/smallrye-reactive-messaging/discussions/2725[a discussion in smallrye-reactive-messaging] to ask the capability to configure the vertx options used in the microprofile-reactive-messaging subsystem, and there is a {wildfly_vertx_feature_pack_with_link} ready to fulfill, integrate it to WildFly should satisfy the requirement.
+
+The goal of the {wildfly_vertx_feature_pack_with_link} is to be the central place to provide the Vertx instance within WildFly server for other subsystems which may need it, example subsystems are https://github.com/wildfly/wildfly/tree/main/microprofile/telemetry-smallrye[microprofile-telemetry] and https://github.com/wildfly/wildfly/tree/main/microprofile/reactive-messaging-smallrye[microprofile-reactive-messaging].
```
kabir, line 33: Might be worth pointing out that these currently create their own vertx instances (and I think it is actually worse, since at least RM will create one per deployment). Advice from the VertX team is to have only one shared instance

microprofile/WFLY-19954_Preview_Support_vertx_feature_pack.adoc
kabir: At least for reactive messaging I'm not setting any properties in WildFly. I don't know if any are set internally in SmallRye RM. I guess not since when running in Quarkus they use the Vert.X instance provided by Quarkus

microprofile/WFLY-19954_Preview_Support_vertx_feature_pack.adoc
```diff
@@ -0,0 +1,234 @@
...
+|===
+
+Options in {address_resolver_options_with_link} are grouped to the address `/subsystem=vertx/address-resolver-option=xxx`, this can also be merged into `/subsystem=vertx/vertx-option=xxx` too if that is more user-friendly.
+
+.Attributes for AddressResolverOptions in address: `/subsystem=vertx/address-resolver-option=xxx`
+[frame=all, grid=all]
+|===
+^| Field Name ^| Data Type ^| Source Class ^| Resource Attribute Name ^| Note
+
+| hostsPath | String | {address_resolver_options_with_link} | hosts-path | the path to the alternate hosts configuration file
```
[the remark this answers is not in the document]
gaol: The content is something like what is inside `/etc/hosts` in Linux. The `hostsValue` is the alternative which contains the content in String. So maybe we can just support `hostsValue`.

microprofile/WFLY-19954_Preview_Support_vertx_feature_pack.adoc
jasondlee: > I am not sure if the following options should be updated by users or not:

Those sound like properties that I'm not sure we'd want users managing inside an application server, but I could be wrong.

> One example I see coming from opentelemetry subsystem[1]:

That's kind of an ugly hack that we had to put in because vertx did not (and may still not. I haven't looked) provide an API to set. I don't know if setting that property in `opentelemetry` will affect the vertx instance from the FP. I'd be surprised if it's set in time, given the module system lifecycle, etc., so we might either need to make that a configurable option, or just add that to the FP.

microprofile/WFLY-19954_Preview_Support_vertx_feature_pack.adoc
jasondlee: Defaulting to the server default directory seems wise to me.

microprofile/WFLY-19954_Preview_Support_vertx_feature_pack.adoc
gaol: I added a consideration of moving it to WildFly codebase in the future when it becomes mature enough and has a  higher stability level.

microprofile/WFLY-19954_Preview_Support_vertx_feature_pack.adoc
gaol: Maybe we should keep the 2 options(`disableTCCL`, `useDaemonThread`) untouched by users, and open only when there are explicit demands to configure them.

gaol: > because vertx did not (and may still not. I haven't looked) provide an API to set ...
> so we might either need to make that a configurable option, or just add that to the FP.

I don't see any API either to set, so probably be good to add that to the FP as well before a Vertx instance is constructed so that it takes effect.

gaol:
Thank you for the feedback, main updates were done:

* Mention that we will consider moving the code to WildFly code base in the future
* Make `disableTCCL` and `useDaemonThread` not configurable via vertx subsystem (users still can do via system property though)
* Add the setting of `vertx.disableDnsResolver` system property in the vertx subsystem
* Make the default value of `fileCacheDir` to `${jboss.server.temp.dir}/vertx-cache`
* Do not configure `hostsPath`, use only the alternative `hostsValue` for the AddressResolverOptions
* Remove the function of configuring the VertxOptions from a JSON file, as it may lead to some options that are not listed above get configured which may be confusing.

microprofile/WFLY-19954_Preview_Support_vertx_feature_pack.adoc
jasondlee: For those curious and unfamiliar with this issue, the code in question is here: https://github.com/eclipse-vertx/vert.x/blob/4.x/src/main/java/io/vertx/core/spi/resolver/ResolverProvider.java#L35-L50

jasondlee (approved):

microprofile/WFLY-19954_Preview_Support_vertx_feature_pack.adoc
```diff
@@ -0,0 +1,231 @@
...
+  - kabir
+  - fl4via
+  - jasondlee
+ outside-perspective:
+  -
+promotes:
+promoted-by:
+---
+
+= Integrate wildfly-vertx-feature-pack to WildFly Preview
```
darranl, line 18: Just a small comment about terminology here, WildFly Preview is itself a feature pack - is this proposal more about adding the Vert.x subsystem / extension to the WildFly Preview Feature Pack?

gaol: Yes, it is.  I thought WildFly Preview is a  different distribution, so `Adding the vertx subsystem / extension to the WildFly Preview Feature Pack` would be more accurate, will update it. :)

microprofile/WFLY-19954_Preview_Support_vertx_feature_pack.adoc
fl4via: I vote for keeping those options untouched for now. We can always add later

fl4via (approved):

kabir (changes requested):
Some small suggestions but apart from that I think it looks good. 
Mention/pong me when fixed since somehow I don’t get any notifications in this repository

gaol:
Thank you @fl4via and @kabir for the reviews and feedback, I updated accordingly. :)

kabir (approved):

microprofile/WFLY-19954_Preview_Support_vertx_feature_pack.adoc
```diff
@@ -0,0 +1,232 @@
...
+| maxWorkerExecuteTime | long | {vertx_options_with_link} | max-worker-execute-time | max worker thread execution time
+| maxWorkerExecuteTimeUnit | TimeUnit | {vertx_options_with_link} | max-worker-execute-time-unit | max worker thread execution time unit
+| warningExceptionTime | long | {vertx_options_with_link} | warning-exception-time | max time a thread can be blocked before stack traces get logged
+| warningExceptionTimeUnit | TimeUnit | {vertx_options_with_link} | warning-exception-time-unit | the time unit for warningExceptionTime
+| classPathResolvingEnabled | boolean | {filesystem_options_with_link} | classpath-resolving-enabled | whether classpath resolving is enabled
+| fileCachingEnabled | boolean | {filesystem_options_with_link} | file-cache-enabled | whether file caching is enabled for class path resolving
+| fileCacheDir | String | {filesystem_options_with_link} | file-cache-dir | file cache directory, defaults to `${jboss.server.temp.dir}/vertx-cache`
+| addressResolverOptions | AddressResolverOptions | {vertx_options_with_link} | address-resolver-option | reference to the AddressResolverOptions name
```
bstansberry, line 110: I don't think should block integration at 'preview' stability, but for 'community' we should consider changing all these attributes to follow the WildFly style.

s/eventLoopSize/event-loop-size

etc

AFAIK there are no deviations from this style in subsystem's that ship in WildFly.

microprofile/WFLY-19954_Preview_Support_vertx_feature_pack.adoc
```diff
@@ -0,0 +1,232 @@
...
+| cacheMaxTimeToLive | int | {address_resolver_options_with_link} | cache-max-time-to-live | the cache max TTL in seconds
+| cacheNegativeTimeToLive | int | {address_resolver_options_with_link} | cache-negative-time-to-live | the cache negative TTL in seconds
+| queryTimeout | long | {address_resolver_options_with_link} | query-time-out | the query timeout in milliseconds
+| maxQueries | int | {address_resolver_options_with_link} | max-queries | the maximum number of queries to be sent during a resolution
+| rdFlag | boolean | {address_resolver_options_with_link} | rd-flag | the DNS queries Recursion Desired flag value
+| searchDomains | List<String> | {address_resolver_options_with_link} | search-domains | the list of search domains
+| ndots | int | {address_resolver_options_with_link} | n-dots | the ndots value
+| rotateServers | boolean | {address_resolver_options_with_link} | rotate-servers | if the dns server selection uses round-robin
+| roundRobinInetAddress | boolean | {address_resolver_options_with_link} | round-robin-inet-address | if the inet address selection uses round-robin
```
bstansberry, line 134: Same comment as above re the attribute names in a 'community' stability variant:

s/hostsValue/hosts-value

microprofile/WFLY-19954_Preview_Support_vertx_feature_pack.adoc
kabir: Aha, good point. But I agree on changing for the next version

Issue #608, opened 2024-09-25
Jakarta Data support in standard WildFly


bstansberry:
### Description

In WildFly Preview 34 we added `preview` stability support for Jakarta Data. But that is only available in WildFly Preview. It relies on the Hibernate ORM 6.6 implementation present in standard WildFly, so we can bring it into standard WildFly as well.

### Issue Contact

bstansberry

PR #666, opened 2024-11-22
[Issue_608] [Preview] Jakarta Data support in standard WildFly

bstansberry:
For #608


ee/WFLY-19776_Jakarta_Data_in_standard_WildFly.adoc
```diff
@@ -0,0 +1,132 @@
...
+
+=== User Stories
+
+I'm an application developer and I'd like to be less concerned with the mechanics of manipulating persistent data. I'd like to define my entities and then write an interface that describes the CRUD and data access operations I need. I code against that interface and let WildFly take care of generating the interface implementation that deals with the DB.
+
+I'm an application developer and am familiar with the Repository pattern from use of Spring Data. I'd like to use that pattern in WildFly in non-Spring applications.
+
+I'm an application developer who's used Deltaspike Data, which uses a repository pattern. But it doesn't support EE > 8, so I'm looking for a similar technology to run on the latest WildFly releases.
```
yrodiere, line 42: You might want to mention that users need to be fine with having less flexibility, in particular:

* Simpler query language.
* No lazy loading -- or only explicit lazy loading by using Hibernate ORM native APIs.
* More limited support for dynamic queries; I think there is some support for dynamic predicates, but it far from the Criteria API.
* And so on.

Which is fine, but it does mean that it's not just about what developers "like", it's also about the complexity of their use case: Jakarta Data is great in many cases, but may be too limited in others.

Note that it's perfectly possible to use standard JPA, native Hibernate ORM APIs, and Jakarta Data in the same application. Depending on how the stateless session is injected into Jakarta Data, it's probably even possible to use Hibernate ORM APIs (`StatelessSession`) and Jakarta Data in the same transaction.

yrodiere (commented):
Looks great, I just have a comment about the user stories. No sure it requires any change, but at least we'll be on the same page.

ee/WFLY-19776_Jakarta_Data_in_standard_WildFly.adoc
bstansberry: Thanks @yrodiere. I incorporated a lot of your info above in the first user story.

yrodiere (approved):

ee/WFLY-19776_Jakarta_Data_in_standard_WildFly.adoc
```diff
@@ -0,0 +1,137 @@
...
+
+== Requirements
+
+* The JBoss Modules modules required for the `jakarta-data` subsystem and the Jakarta Data API will be made available through `wildfly-ee` feature pack.
+** These modules will be shared with the `wildfly-preview` feature pack. (This is the norm for `wildfly-ee` modules).
+* The `jakarta-data` Galleon layer will be made available through the `wildfly-ee` feature pack.
+** This layer will be shared with the `wildfly-preview` feature pack. (This is the norm for `wildfly-ee` layers).
+* The Jakarta Data API artifact *must not* be part of the standard WildFly user boms. We limit the user boms to `community` stability artifacts.
+* The Jakarta Data API artifact must contintue to be part of the WildFly Preview user boms.
+* The wildfly-tck-runners Jakarta Data runner will be modified to test the `wildfly` feature pack by default, while allowing testing of `wildfly-preview`.
```
bstansberry, line 79: Note: I'll do this after the 35 Beta release as the runner shouldn't depend on a SNAPSHOT.

jasondlee (approved):

bstansberry:
Thanks @yrodiere and @jasondlee

Issue #670, opened 2024-12-03
[WFLY-18581] Add access logs to default configurations


ropalka:
### Description

This will be very helpful for troubleshooting. It would be helpful to include "%S %I %T" in the pattern and have record-request-start-time attribute configured on the listeners so that we have some info on the session activity and also response times and thread details to correlate to thread dumps as needed for performance issue investigations. 

### Issue Contact

ropalka

PR #671, opened 2024-12-03
[WFLY-18581] Enable access logs in default configs

ropalka:
https://issues.redhat.com/browse/WFLY-18581

Add analysis document

Resolves https://github.com/wildfly/wildfly-proposals/issues/670


jamezp:
What is the performance impact of this?

ropalka:
The performance impact should be really minimal. For this particular proposal 'use-server-log' attribute defaults to 'false'  [1] and that means AccessLogService [2] will use DefaultAccessLogReceiver [3] . DefaultAccessLogReceiver uses Executor framework for writing logs so the penalty is just queuing the log message in logging queue (plus later in different thread async write of the log message to the file system). The message queuing is the most important question here because it has direct impact on the HTTP request-response latency. Considering networking latency (HTTP message exchanges) vs. queuing log message to the logging queue, the impact should be IMHO unmeasurable.

[1] [org/wildfly/extension/undertow/AccessLogService.java](https://github.com/wildfly/wildfly/blob/main/undertow/src/main/java/org/wildfly/extension/undertow/AccessLogService.java#L85)
[2] https://github.com/wildfly/wildfly/blob/main/undertow/src/main/resources/schema/wildfly-undertow_14_0.xsd#L435
[3] https://github.com/undertow-io/undertow/blob/2.3.x/core/src/main/java/io/undertow/server/handlers/accesslog/DefaultAccessLogReceiver.java#L55
[4] https://github.com/undertow-io/undertow/blob/2.3.x/core/src/main/java/io/undertow/server/handlers/accesslog/JBossLoggingAccessLogReceiver.java

TomasHofman:
I have two concerns. I think none of them is critical, so just for your consideration...

**Documentation**

Looking through the docs living directly in the Wildfly repo, I didn't find anything that would specifically describe this functionality (I found a section on "Console Access Logging", which is related, but not exactly what is being enabled by this PR.

Googling, the most relevant page that describes this is the MasterTheBoss website (https://www.mastertheboss.com/web/jboss-web-server/how-do-you-configure-jboss-to-enable-http-logging/), and the WildScribe docs on wildfly.org (https://docs.wildfly.org/35/wildscribe/subsystem/undertow/server/host/setting/access-log/index.html).

So, considering this is quite basic and useful functionality, and as well that we are changing the default behavior, would it be worth it to add docs section specifically on "Access Logging", where we could at least mention that the access log is enabled by default and describe how to disable it and cover the basics?

**Log rotation / retention time**

Two questions in my head, not covered by the proposal doc, are:

* "how large can the log file get?"
* and "how long do the log data survive?"

That boils down to whether there is any log rotation strategy by default. Looking at Wildscribe it looks the logs are rotated daily (that's nice IMO), and are never removed (that may be less OK).

Could this cause a problem for people running busy Wildfly instances (running out of space?). Would it be reasonable to limit the amount of data logged in the default configuration?

The access-log resource doesn't seem to have capabilities to do any more complex rotation strategies, the mastertheboss page suggests to delegate access logging to the logging subsystem if this is needed.

If we don't change this aspect maybe just clarify the behavior in the proposal..?

Issue #672, opened 2024-12-09
Add default async context timeout undertow option to management model


ropalka:
### Description

At present the default asynchronous context timeout for Undertow asynchronous operations is hard coded.
There is a high demand from users to make it configurable. To achieve this we need to
add  default asynchronous context timeout option to Undertow subsystem.

### Issue Contact

ropalka

PR #673, opened 2024-12-09
[WFLY-14748] Allow configuration of default async context timeout in configs

ropalka:
https://issues.redhat.com/browse/WFLY-14748

Add analysis document

Resolves https://github.com/wildfly/wildfly-proposals/issues/672

### DIFF
--- a/undertow/WFLY-14748_Add_Async_Context_Timeout_Option.adoc
+++ b/undertow/WFLY-14748_Add_Async_Context_Timeout_Option.adoc
@@ -1,0 +1,174 @@
+---
+categories:
+ - undertow
+stability-level: preview
+issue: https://github.com/wildfly/wildfly-proposals/issues/672
+---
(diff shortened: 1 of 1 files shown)


undertow/WFLY-14748_Add_Async_Context_Timeout_Option.adoc
```diff
@@ -0,0 +1,174 @@
...
+:author:            Richard Opalka
+:email:             ropalka@redhat.com
+:toc:               left
+:icons:             font
+:idprefix:
+:idseparator:       -
+
+== Overview
+
+At present the default asynchronous context timeout for Undertow asynchronous operations is hard coded.
```
darranl, line 17: When you say "the default" is there a second mechanism they can use to set the timeout?

TomasHofman (approved):

fl4via (changes requested):
@ropalka it would be great if we could update this analysis document to reflect the latest model.
I also think we should add to the overview how the user can set the timeout today without the RFE

PR #682, opened 2025-01-27
WFLY-20296 - Add support for more OpenTelemetry Config options

jasondlee:
https://issues.redhat.com/browse/WFLY-20296

Add analysis document


observability/WFLY-20296-add-support-for-more-opentelemetry-config-options.adoc
```diff
@@ -0,0 +1,140 @@
...
+*** Export interval (`otel.metric.export.interval`)
+** Logs
+*** Schedule delay (`otel.blrp.schedule.delay`)
+*** Max queue size (`otel.blrp.max.queue.size`)
+*** Max batch size (`otel.blrp.max.export.batch.size`)
+*** Export timeout (`otel.blrp.export.timeout`)
+* Changes to the management model should be split out based on signal types
+** For ease of use by users, the structure of the updated XML should reflect this
+** Existing trace-specific configuration options should be moved to the trace child element
+** The shape of the management model need not share the nested nature unless that is a more natural structure in the Java code
```
bstansberry, line 76: Please make the basic structure of the management model part of the requirements.

The basic point I want to see covered is any new resources (or a statement there will not be), and any intent to use complex attributes vs a flat list of simple attributes.

In general, new resources should not be added unless they result in necessary new underlying MSC services.

Complex attributes simplify (shorten) the look of the resource (i.e. the read-resource output) but they are more complex to configure via the CLI. They can also be 'undefined', which may or may not make conceptual sense.  Their introduction to encapsulate any existing simple attributes is also a breaking change requiring migration. So their use (or not) requires more thought.

Any attribute can have an attribute-group applied to help drive categorization in the console or in XML.

observability/WFLY-20296-add-support-for-more-opentelemetry-config-options.adoc
```diff
@@ -0,0 +1,140 @@
...
+* Managed domain
+* OpenShift Source-to-Image (S2I)
+* Bootable jar
+
+== Requirements
+
+* Additional configuration options should include:
+** General
+*** TLS server certificate (`otel.exporter.otlp.certificate`)
+*** TLS client certificate (`otel.exporter.otlp.client.certificate`)
```
bstansberry, line 63: Please discuss with the security team how this kind of thing should be represented; i.e. if we should provide deeper integration with existing Elytron services.

observability/WFLY-20296-add-support-for-more-opentelemetry-config-options.adoc
```diff
@@ -0,0 +1,140 @@
...
+* Changes to the management model should be split out based on signal types
+** For ease of use by users, the structure of the updated XML should reflect this
+** Existing trace-specific configuration options should be moved to the trace child element
+** The shape of the management model need not share the nested nature unless that is a more natural structure in the Java code
+** Shared properties (e.g., the OpenTelemetry Collector endpoint) should be exposed on the root element.
+*** Signal-specific overrides need not be supported
+* All new properties added will have, as defaults, values described in the relevant OpenTelemetry documentation.
+* Support should not depend on, e.g., MicroProfile Telemetry or MicroProfile Config
+* Any configured properties should still be able to be overridden via the existing MicroProfile support
+* While it is acceptable to relate WildFly management model attributes to their respective system property in the OpenTelemetry documentation to help explain their intent, the system property is an implementation detail that should not necessarily be leaked in the model to avoid as much coupling as is possible.
```
bstansberry, line 82: As work on this progresses, please tighten this language. It's fine now for initial exploration but by the end it should be stated as a requirement, i.e. what must/must not be done.

observability/WFLY-20296-add-support-for-more-opentelemetry-config-options.adoc
```diff
@@ -0,0 +1,140 @@
...
+
+N/A
+
+== Admin Clients
+
+N/A
+
+== Security Considerations
+
+In general, this should have no impact on security. However, one of the proposed changes includes adding support for configuring TLS certificates for communicating with the OpenTelemetry Collector. The use of these TLS certificates will allow the server to both secure communications with the Collector, and verify that the collector is, indeed, the correct receiver for the signals. This will have a positive impact on the data security for the subsystem and user applications.
```
bstansberry, line 126: See my comment above about discussing with the security team.

observability/WFLY-20296-add-support-for-more-opentelemetry-config-options.adoc
```diff
@@ -0,0 +1,140 @@
...
+
+[[test_plan]]
+== Test Plan
+
+The existing tests under `testsuite/integration/microprofile` will be extended to cover the relevant newly added properties. Not all properties (such as those affecting publish rates or batch sizes) need extensive integration testing. Others, though, such as the TLS properties, will need thorough testing, both positive and negative.
+
+== Community Documentation
+
+The existing documentation in the Administration Guide will be augmented to cover the new options
```
bstansberry, line 135: Please be sure to require coverage of the relationships between new attributes, system props and MP config options, particularly the latter as the intent above is those will be supported.

observability/WFLY-20296-add-support-for-more-opentelemetry-config-options.adoc
```diff
@@ -0,0 +1,140 @@
...
+:idprefix:
+:idseparator:       -
+
+== Overview
+
+While every aspect of OpenTelemetry is technically configurable via MP Telemetry and MP Config, there have been several requests for more direct support for the various options in the management model. The RFE will identify and implement the options most likely to be of general interest to WildFly users.
+
+=== User Stories
+
+- As an application developer, I would like greater control in how OpenTelemetry is configured in the server without having to add an external library (e.g., MicroProfile Config) to my deployment(s).
```
bstansberry, line 33: I'm not sure how to phrase this better, but a user would not add MicroProfile Config to their deployments as a library; i.e. by packaging a jar.

observability/WFLY-20296-add-support-for-more-opentelemetry-config-options.adoc
darranl: What type of data would attributes for these take?

In general we are trying to centralise configuration related to SSLContexts into the Elytron subsystem so users and administrators don't need to learn how to confiure TLS related resources independently on each separate subsystem,

jasondlee: The value passed to the underlying library is the full path to the certificates (and keys) used by the OpenTelemetry Collector to configure TLS on its end. I guess it's possible that we switch the perspective here: rather than pointing WF to what the Collector is using to configure itself, we point the Collector to what WF is using to configure itself?

jasondlee: Yes. It's a process that runs in another JVM (or even data center). In my head, the plan is currently to point the OpenTelemetry subsystem at the certificate and key files known to and use by the OpenTelemetry Collector. What I mean by the inversion is that rather than point from inside the app servers config (and FS area) to files owned/managed by the Collector, that could instead point the Collector to cert/key files owned and managed by WildFly. Perhaps I was being too cute. :P I was just thinking "out loud" about how a WildFly SSL context might used to configure an external service. Ultimately, it's going to come down to each system sharing one or more files, regardless of who "owns" those files.

observability/WFLY-20296-add-support-for-more-opentelemetry-config-options.adoc
jasondlee: I have update the text a bit. I hope it's moving in the direction you were hoping to see. If it is (or still not explicit enough), please let me know.

observability/WFLY-20296-add-support-for-more-opentelemetry-config-options.adoc
jasondlee: Is that better?

observability/WFLY-20296-add-support-for-more-opentelemetry-config-options.adoc
```diff
@@ -0,0 +1,154 @@
...
+:idprefix:
+:idseparator:       -
+
+== Overview
+
+While every aspect of OpenTelemetry is technically configurable via MP Telemetry and MP Config, there have been several requests for more direct support for the various options in the management model. The RFE will identify and implement the options most likely to be of general interest to WildFly users.
+
+=== User Stories
+
+- As an application developer, I would like greater control in how OpenTelemetry is configured in the server without having to depend on an external library (e.g., MicroProfile Config).
```
TomasHofman, line 33: Out of curiosity, it's not clear to me how exactly it affects the developer to rely on MP Config to configure opentelemetry.

My understanding is that the MP Config way simply means that I would e.g. define a system property via `-Dotel.service.name=myname` (tried and it works), or I could create "permanent" system property in standalone.xml. Perhaps I can also package an application.properties file into my deployment (haven't tried that).

Nothing else is needed on the application side (no special dependency, etc). Am I correct? The "external library" is a detail not visible to the app developer.

TomasHofman:
You probably have reasons not to do it that way, but I will ask :):

Was it considered to introduce a "properties" section in the opentelemetry config model, which would allow users to specify otel properties directly as they are in the otel spec (same keys & values), rather than introducing custom named attributes?

The advantages would be:

* the administrator would not need to investigate how Wildfly attributes map to OTel properties,
* it would (maybe?) allow to configure full set of OTel properties, without extensive changes to the management model,
* it might be more accommodating to potential new attributes added in future.

observability/WFLY-20296-add-support-for-more-opentelemetry-config-options.adoc
jasondlee: I more had in mind the server administrator, as the MP Config extension/subsystem would have to be enabled to make it work. I'll try to clean up that wording there.

TomasHofman: OK, I get what you meant now.

jasondlee:
> You probably have reasons not to do it that way, but I will ask :):
> 
> Was it considered to introduce a "properties" section in the opentelemetry config model, which would allow users to specify otel properties directly as they are in the otel spec (same keys & values), rather than introducing custom named attributes?

I have not thought about that, to be honest, but, having done so a bit now, I'm not sure I'm real excited about that. You're probably right the current approach puts some onus on the admin to map attributes to properties, but I *hope* the attribute names make it clear enough what they do so the admin doesn't need to know about the mapping. I think more importantly, depending too heavily on an assortment of properties defined elsewhere makes the management interface vague and opaque. Using clearly defined management model attributes for clarity coupled with the properties-based approach via MP Config (or sysprops) gives us a nice middle ground, albeit at the expense of some complexity (like this PR) in adding new, explicit attribute support. I can be persuaded otherwise, but I think that's my take. :)

observability/WFLY-20296-add-support-for-more-opentelemetry-config-options.adoc
```diff
@@ -0,0 +1,154 @@
...
+||Schedule delay|`otel.blrp.schedule.delay`|The interval, in milliseconds, between two consecutive exports.|`logs-export-interval`|1000|
+|===
+* Changes to the management model should be split out based on signal types
+** For ease of use by users, the structure of the updated XML should reflect this grouping of signal-specific attributes via `setAttributeGroup(String)`
+** The shape of the management model should be flat, using a list of attributes, rather than nested resources
+*** Each attribute will be declared on the existing `OpenTelemetrySubsystemRegistrar`. No new child resources should be created.
+*** Attribute groups will be used to group properties by signal type to make the XML file and admin console more user-friendly
+** Shared properties (e.g., the OpenTelemetry Collector endpoint) should be exposed on the root element.
+*** Signal-specific overrides need not be supported
+* All new properties added will have, as defaults, values described in the relevant OpenTelemetry documentation (see the official docs https://opentelemetry.io/docs/languages/java/configuration/[here])
```
bstansberry, line 84: s/as defaults/as initial defaults/g

Once we set a default, it's *our* default. It's no longer determined based on what some outside party decides. Changing it is a semi-breaking or breaking change, depending on what the attribute configures. (We can change accept that and change the default but we should not promise that we will always do what some outside party said.)

observability/WFLY-20296-add-support-for-more-opentelemetry-config-options.adoc
```diff
@@ -0,0 +1,154 @@
...
+||TLS client certificate|`otel.exporter.otlp.client. certificate`|The path to the client certificate on disk|TBD||
+||Client key|`otel.exporter.otlp.client. key`|The path to the key on disk|TBD||
+||Compression|`otel.exporter.otlp. compression`|The compression type to use on OTLP trace, metric, and log requests.|`compression`|null|Only `gzip` supported
+|Traces|Enabled|`otel.traces.exporter`|Are traces enabled?|`traces-enabled`|true|If true, the value is set to `otlp`. If not, the value is set to `none`.
+|Metrics|Enabled|`otel.metrics.exporter`|Are metrics enabled?|`metrics-enabled`|true|If true, the value is set to `otlp`. If not, the value is set to `none`.
+||Export interval|`otel.metrics.export.interval`|The interval, in milliseconds, between the start of two export attempts.|`metrics-export-interval`|60000|
+||Exemplar filter|`otel.metrics.exemplar.filter`|The filter for exemplar sampling.|`exemplar-filter`|`trace_based`|Options are `always_off`, `always_on` or `trace_based`.
+||Metrics cardinality limit|`otel.experimental.metrics. cardinality.limit`|If set, configure cardinality limit; the maximum number of distinct points per metric.|cardinality-limit|2000|
```
bstansberry, line 71: Sounds like a -1 from me.

If it's 'experimental' to OTel it should not be in the WF management API, except in it's own low-stability-variant of the management API.

observability/WFLY-20296-add-support-for-more-opentelemetry-config-options.adoc
```diff
@@ -0,0 +1,154 @@
...
+* Managed domain
+* OpenShift Source-to-Image (S2I)
+* Bootable jar
+
+== Requirements
+
+* Additional configuration options should include:
+|===
+|Signal type|Name|OpenTelemetry Property|Description|Attribute Name|Default|Notes
+|General|TLS server certificate|`otel.exporter.otlp. certificate`|The path to the server certificate on disk|TBD||_TLS props are pending a discussion with Elytron_
```
bstansberry, line 63: Re 'pending discussion with Elytron' -- yes, for sure. A long term strategy for security configuration is a blocker for having this at community stability.

To me one of the value adds of having a subsytem is it does proper cross-subsystem integration for key concerns like managing TLS.

jasondlee: I am working on this implementation in consultation with Elytron at this very moment.

observability/WFLY-20296-add-support-for-more-opentelemetry-config-options.adoc
bstansberry: +1 to focusing on an administrator.

Beyond not 'depending on a library' (or setting system properties) in terms of a programming API, someone using this is *depending on us* for a robust, long-term-compatible API. They are depending on us to insulate them from underlying changes in OTel. If OTel renames one of their props, or whatever, in a future release, that is our problem, not the user's problem. They count on us to shield them from such things. That is what the WildFly management API does for people who use the many libraries we integrate.

So, think very hard about what you expose, as you'll be living with it for a long time. :-)

observability/WFLY-20296-add-support-for-more-opentelemetry-config-options.adoc
```diff
@@ -0,0 +1,154 @@
...
+||TLS client certificate|`otel.exporter.otlp.client. certificate`|The path to the client certificate on disk|TBD||
+||Client key|`otel.exporter.otlp.client. key`|The path to the key on disk|TBD||
+||Compression|`otel.exporter.otlp. compression`|The compression type to use on OTLP trace, metric, and log requests.|`compression`|null|Only `gzip` supported
+|Traces|Enabled|`otel.traces.exporter`|Are traces enabled?|`traces-enabled`|true|If true, the value is set to `otlp`. If not, the value is set to `none`.
```
bstansberry, line 67: What's the plan for when we decide to support other exporters?

I can imagine one; I just want to make sure you have too, as you'll be living with this for many years. :-)

Same applies to Metrics|Enabled

observability/WFLY-20296-add-support-for-more-opentelemetry-config-options.adoc
```diff
@@ -0,0 +1,154 @@
...
+* Changes to the management model should be split out based on signal types
+** For ease of use by users, the structure of the updated XML should reflect this grouping of signal-specific attributes via `setAttributeGroup(String)`
+** The shape of the management model should be flat, using a list of attributes, rather than nested resources
+*** Each attribute will be declared on the existing `OpenTelemetrySubsystemRegistrar`. No new child resources should be created.
+*** Attribute groups will be used to group properties by signal type to make the XML file and admin console more user-friendly
+** Shared properties (e.g., the OpenTelemetry Collector endpoint) should be exposed on the root element.
+*** Signal-specific overrides need not be supported
+* All new properties added will have, as defaults, values described in the relevant OpenTelemetry documentation (see the official docs https://opentelemetry.io/docs/languages/java/configuration/[here])
+* Support should not depend on, e.g., MicroProfile Telemetry or MicroProfile Config
+* Any configured properties should still be able to be overridden via the existing MicroProfile support (this is the current default behavior)
```
bstansberry, line 86: The textual description of any such attribute must explicitly declare that if MicroProfile Telemetry is used the config property 'foo' takes precedence.

I hope if MP Tel is not used that system properties are not preferred to what is set via the management API. If that's not the case let's have a zulip or dev list discussion.

observability/WFLY-20296-add-support-for-more-opentelemetry-config-options.adoc
```diff
@@ -0,0 +1,154 @@
...
+Open question: Some of the values specified by OpenTelemetry are rather large and unwieldy (e.g., `base2_exponential_bucket_histogram`). Should we provide shorthand mappings, use the exact strings from otel docs, or both?
+=====
+
+=== Changed requirements
+
+N/A
+
+=== Non-Requirements
+
+* The OpenTelemetry documentation lists several signal-specific parameters to configure each signal explicitly. The following parameters have the same defaults as those already supported (i.e., for traces), so, in the name of simplicity, these will be configured implicitly by the system using a single configuration attribute:
```
bstansberry, line 99: I'm not clear on the 'non-requirement' nature of this item.

observability/WFLY-20296-add-support-for-more-opentelemetry-config-options.adoc
```diff
@@ -0,0 +1,154 @@
...
+
+No interoperability concerns are foreseen.
+
+== Implementation Plan
+
+N/A
+
+== Admin Clients
+
+N/A
```
bstansberry, line 136: For sure it's applicable. Needs a statement, even if it's just that the CLI and HAL will just work with no changes on their side.

PR #684, opened 2025-02-18
[WFLY-19915] Make MEMORY_STORAGE_THRESHOLD configurable

xjusko:
https://issues.redhat.com/browse/WFLY-19915

### DIFF
--- a/undertow/WFLY-19915_Make MEMORY_STORAGE_THRESHOLD configurable.adoc
+++ b/undertow/WFLY-19915_Make MEMORY_STORAGE_THRESHOLD configurable.adoc
@@ -1,0 +1,111 @@
+---
+categories:
+ - undertow
+stability-level: preview
+---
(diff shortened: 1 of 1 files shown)


jmesnil:
@xjusko The check is failing because your asciidoc file has space in its name. could you fix it please?

xjusko:
@jmesnil Thanks. I fixed that, but now it says that the issue field is missing. I am not sure why, as it is there.

spyrkob:
@xjusko I think you're missing https://github.com/wildfly/wildfly-proposals/blob/main/design-doc-template.adoc?plain=1#L14

bstansberry:
@xjusko Yes, please use the current template as the base for a proposals.

Please also start with the steps in https://docs.wildfly.org/wildfly-proposals/FEATURE_PROCESS.html. Thanks!

PR #686, opened 2025-02-25
Add \n between metadata table and TOC syntax start

baranowb:
Just new line to separate metadata and TOC. Some editors are tad more strict or too strict and will swallow first paragraph of TOC.


bstansberry:
Thanks @baranowb

Issue #676, opened 2024-12-20
Add OIDC Logout to Elytron and corresponding subsystem


rsearls:
### Description

Add support for the three types of OIDC logout mechanisms when securing apps deployed to WildFly with OIDC, RP-Initiated Logout, Front-Channel Logout, and Back-channel Logout.
Three logout configuration attributes will be added to the elytron-oidc-client subsystem and to the oidc.json configuration file.

### Issue Contact

rsearls

PR #687, opened 2025-02-25
[Preview][WFLY-19314] Logout support for OIDC

rsearls:
Analysis for https://github.com/wildfly/wildfly-proposals/issues/676
related to https://issues.redhat.com/browse/ELY-2534
PR: https://github.com/wildfly-security/wildfly-elytron/pull/2249

### DIFF
--- a/elytron/WFLY-19314-oidc-logout-support.adoc
+++ b/elytron/WFLY-19314-oidc-logout-support.adoc
@@ -1,0 +1,195 @@
+---
+categories:
+  - elytron
+  - security
+---
+= [Preview] WFLY-19314 Logout Support for OIDC
+:author:            Farah Juma
+:email:             fjuma@redhat.com
+:toc:               left
+:icons:             font
+:idprefix:
+:idseparator:       -
+
+== Overview
+
+It is currently possible to secure applications deployed to WildFly using OpenID Connect (OIDC),
+as specified in the https://openid.net/specs/openid-connect-core-1_0.html[OpenID Connect Core 1.0 spec].
+
+There are additional optional OIDC specifications that may be used in addition to the Core spec to
+provide support for logging out:
+
+* https://openid.net/specs/openid-connect-rpinitiated-1_0.html[RP-Initiated Logout] is an optional
+OIDC specification that defines how a relying party (RP) can request that an OpenID provider logs
+out the end-user.
+* https://openid.net/specs/openid-connect-frontchannel-1_0.html[Front-Channel Logout] is an optional
+OIDC specification that defines a front-channel logout mechanism that does not use an OP iframe on RP
+pages.
+* https://openid.net/specs/openid-connect-core-1_0.html#OpenID.BackChannel[Back-channel Logout] is an
+optional OIDC specification that defines a logout mechanism that uses direct back-channel communication
+between the OP and RPs being logged out.
+
+This RFE is to add support for these three types of logout mechanisms when securing apps deployed to
+WildFly with OIDC.
+
+== Issue Metadata
+
+=== Issue
+
+* https://issues.redhat.com/browse/WFLY-19314
+
+=== Related Issues
+
+* https://issues.redhat.com/browse/ELY-2534
+
+=== Stability Level
+// Choose the planned stability level for the proposed functionality
+* [ ] Experimental
+
+* [x] Preview
+
+* [ ] Community
+
+* [ ] default
+
+=== Dev Contacts
+
+* mailto:{email}[{author}]
+
+=== QE Contacts
+
+=== Testing By
+// Put an x in the relevant field to indicate if testing will be done by Engineering or QE. 
+// Discuss with QE during the Kickoff state to decide this
+* [x] Engineering
+
+* [ ] QE
+
+=== Affected Projects or Components
+
+* WildFly Elytron
+* WildFly Core
+* WildFly
+
+=== Other Interested Projects
+
+=== Relevant Installation Types
+// Remove the x next to the relevant field if the feature in question is not relevant
+// to that kind of WildFly installation
+* [x] Traditional standalone server (unzipped or provisioned by Galleon)
+
+* [x] Managed domain
+
+* [x] OpenShift s2i
+
+* [x] Bootable jar
+
+== Requirements
+
+=== Hard Requirements
+
+* It will be possible to log out of an application secured with OIDC using RP-Initiated Logout, Front-Channel Logout,
+and Back-Channel Logout.
+* The new configuration options described below will be added for both the `elytron-oidc-client` subsystem configuration
+and deployment configuration.
+* This feature will be available when starting WildFly with the Preview Stability Level.
+
+==== RP-Initiated Logout
+
+For RP-Initiated Logout, we'll make use of the `id_token_hint` in the logout request to the logout endpoint.
+The use of `id_token_hint` is recommended by the spec.
+
+We'll introduce the following new configuration options:
+
+* `logout-uri`: Specifies a relative URI referencing the endpoint that will start a logout (RP-Initiated).
+This will default to `/logout`.
+* `post-logout-redirect-uri`: Specifies a relative URI referencing the endpoint that the OpenID provider
+should redirect users after logging out sessions. This will be undefined by default.
+
+==== Front-Channel Logout and Back-Channel Logout
+
+For Front-Channel Logout and Back-Channel Logout, we'll introduce the following new configuration options:
+
+* `logout-callback-uri`: A relative URI referencing the endpoint that will handle logout requests
+from the OpenID provider. This will default to `/logout/callback`.
+* `logout-session-required`: A boolean value that indicates whether `elytron-oidc-client` should enforce
+both `sid` and `iss` parameters when processing front-channel logout requests and should enforce the `sid`
+parameter when processing back-channel requests. If set to false, for front-channel logout, there will
+be no check and any authenticated request is going to invalidate the local user session. To make use of back-channel
+logout, this must be set to `true`. (This is because the `sid` will be marked for invalidation using a bounded map
+so that subsequent requests from the user will force the session to be invalidated.) This will default to `true`.
+
+
+
+==== Back-Channel Logout
+
+We'll introduce the following new configuration option for Back-Channel Logout:
+
+* `logout-session-invalidation-limit`: The maximum size for a bounded map when marking sessions for
+invalidation during back-channel logout. Once this maximum size has been reached,
+
+
+
+=== Nice-to-Have Requirements
+// Requirements in this section do not have to be met to merge the proposed functionality.
+// Note: Nice-to-have requirements that don't end up being implemented as part of
+// the work covered by this proposal should be moved to the 'Future Work' section.
+
+
+=== Non-Requirements
+// Use this section to explicitly discuss things that readers might think are required
+// but which are not required.
+The ability to specify the `logout_hint` and `client_id` for RP-Initiated Logout is a
+non-requirement. Neither or these are required since we'll be passing the `id_token_hint`
+in the logout request. The `logout_hint` isn't supported by all OpenID providers.
+
+
+=== Future Work
+// Use this section to discuss requirements that are not addressed by this proposal
+// but which may be addressed in later proposals.
+
+== Backwards Compatibility
+
+// Does this enhancement affect backwards compatibility with previously released
+// versions of WildFly?
+// Can the identified incompatibility be avoided?
+
+=== Default Configuration
+
+=== Importing Existing Configuration
+
+=== Deployments
+
+=== Interoperability
+
+//== Implementation Plan
+////
+Delete if not needed. The intent is if you have a complex feature which can 
+not be delivered all in one go to suggest the strategy. If your feature falls 
+into this category, please mention the Release Coordinators on the pull 
+request so they are aware.
+////
+
+== Security Considerations
+
+////
+Identification if any security implications that may need to be considered with this feature
+or a confirmation that there are no security implications to consider.
+////
+
+== Test Plan
+
+== Community Documentation
+////
+Generally a feature should have documentation as part of the PR to wildfly master, or as a follow up PR if the feature is in wildfly-core. In some cases though the documentation belongs more in a component, or does not need any documentation. Indicate which of these will happen.
+////
+== Release Note Content
+////
+Draft verbiage for up to a few sentences on the feature for inclusion in the
+Release Note blog article for the release that first includes this feature. 
+Example article: http://wildfly.org/news/2018/08/30/WildFly14-Final-Released/.
+This content will be edited, so there is no need to make it perfect or discuss
+what release it appears in.  "See Overview" is acceptable if the overview is
+suitable. For simple features best covered as an item in a bullet-point list 
+of features containing a few words on each, use "Bullet point: <The few words>" 
+////
--- a/elytron/WFLY-19314-oidc-logout-support.adoc
+++ b/elytron/WFLY-19314-oidc-logout-support.adoc
@@ -101,29 +101,22 @@ The use of `id_token_hint` is recommended by the spec.
 
 We'll introduce the following new configuration options:
 
-* `logout-path`: This option designates the path component of a URL.  It is the path component of the RP's URL to trigger a logout of the user.  Elytron uses it to identify a logout request and take the appropriate action internally.  The default values is `/logout`.
+* `logout-path`: This option designates the path component of a URL.  It is the path component to trigger a logout of the user.  Elytron uses it to identify a logout request and take the appropriate action internally.  The default values is `/logout`.
 - The option can not be declared as an empty string or "/" only
 - It is expected that the text start with a forward slash (e.g. "/").
 - It can be a sequence of path segments (e.g. "/call/back/home")
-- It cannot be identical to the text declared in `logout-callback-path`
 
+==== RP-Initiated Logout
+For RP-Initiated Logout, we'll introduce the following new configuration option:
 
-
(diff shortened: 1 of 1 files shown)


elytron/WFLY-19314-oidc-logout-support.adoc
```diff
@@ -0,0 +1,202 @@
...
+
+=== Testing Configuration option values.
+Tests will be created to check the restrictions to the configuration option values.  An exception will be thrown when the value is invalid.
+
+* `logout-path`
+- The value can not be the empty string
+- The value can not be "/" only
+- It must start with the "/" character
+* `post-logout-uri` and `logout-callback-path`
+-  Both must start with `http`.
```
pedro-hos, line 138: Does the T`post-logout-uri` and `logout-callback-path` will only acept `http` or `https` as well?

rsearls: It will accept both http and https

fjuma:
@rsearls This is looking good! Would you be able to update the PR title to something like "[Preview][WFLY-19314] Logout support for OIDC" and move the PR from Draft to ready to review?

The last four commits could also be squashed together so all your changes are in a single commit.

Thanks!

elytron/WFLY-19314-oidc-logout-support.adoc
```diff
@@ -101,34 +101,50 @@ The use of `id_token_hint` is recommended by the spec.
...
 logout, this must be set to `true`. (This is because the `sid` will be marked for invalidation using a bounded map
 so that subsequent requests from the user will force the session to be invalidated.) This will default to `true`.
 
-
-
-==== Back-Channel Logout
-
-We'll introduce the following new configuration option for Back-Channel Logout:
-
-* `logout-session-invalidation-limit`: The maximum size for a bounded map when marking sessions for
```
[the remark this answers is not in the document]
rsearls: I don't remember why I removed it.  Most likely I did not find it in the specs.  

Do you know why it is listed in the "Back-Channel Logout" section and not as general use for Back-Channel and Front-Channel?

I will return it to the doc.  This addition will entail touching elytron code and wildfly/elytron-oidc-client as well.  
Should I also make the name change to post-logout-uri since I will be touching all that code anyway?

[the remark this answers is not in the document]
fjuma: Yes I think we should do the name change for post-logout-uri.

Please check with Pedro from the Keycloak team to see if logout-session-invalidation-limit should be configurable or not.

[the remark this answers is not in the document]
rsearls: I put this back

rsearls:
Add back-channel-logout-session-invalidation-limit.
Revised some property descriptions.

Ready for review.

elytron/WFLY-19314-oidc-logout-support.adoc
```diff
@@ -0,0 +1,221 @@
...
+
+	The applications URL is https://localhost:8080/mySecureApp. If the callback address is to be https://www.wildfly.org/, the user would need to declare `logout-callback-path` to be `https://www.wildfly.org/`
+```
+
+
+* `logout-session-required`: A boolean value. The default value is true
+   - For Front-Channel Logout the boolean value specifies whether the request requires the iss (issuer) and sid (session ID) query parameters be included to identify the session with the OpenId Provider.
+   - For Back-Channel Logout the boolean value specifies whether the request requires the sid (session ID) Claim be included in the Logout Token to identify the session with the OpenId Provider.  This value must be `true` for back-channel logout to be executed.  If the value is `false` the `sid` will be marked for invalidation using a bounded map so that subsequent requests from the user will force the session to be invalidated.
+
+* `back-channel-logout-session-invalidation-limit`: The maximum size for a bounded map when marking sessions for invalidation during back-channel logout. Once this maximum size has been reached, the eldest entry will be removed.  The initial size is 16.
```
fjuma, line 145: s/initial/default

elytron/WFLY-19314-oidc-logout-support.adoc
```diff
@@ -0,0 +1,221 @@
...
+
+=== Testing Configuration option values.
+Tests will be created to check the restrictions to the configuration option values.  An exception will be thrown when the value is invalid.
+
+* `logout-path`
+- The value can not be the empty string
+- The value can not be "/" only
+- It must start with the "/" character
+* `post-logout-uri`
```
fjuma, line 154: This should now be updated to reference the new name.

back-channel-logout-session-invalidation-limit should be mentioned here too.

It would also be good to mention doing some manual tests to verify that logout can be used with an OpenID provider other than Keycloak.

elytron/WFLY-19314-oidc-logout-support.adoc
rsearls: done

rsearls:
Updated pushded

fjuma (approved):

elytron/WFLY-19314-oidc-logout-support.adoc
```diff
@@ -0,0 +1,191 @@
...
+
+* [x] Bootable jar
+
+== Requirements
+
+* It will be possible to log out of an application secured with OIDC using RP-Initiated Logout, Front-Channel Logout,
+and Back-Channel Logout.
+* The new configuration options described below will be added for both the `elytron-oidc-client` subsystem configuration
+and deployment configuration.
+* This feature will be available when starting WildFly with the Preview Stability Level.
```
OndrejKotek: Will the feature be enabled by default? If yes, will it be possible to turn it off?

elytron/WFLY-19314-oidc-logout-support.adoc
```diff
@@ -0,0 +1,191 @@
...
+
+	The applications URL is https://localhost:8080/mySecureApp. If the callback address is to be https://www.wildfly.org/, the user would need to declare `logout-callback-path` to be `https://www.wildfly.org/`
+```
+
+
+* `logout-session-required`: A boolean value. The default value is true
+   - For Front-Channel Logout the boolean value specifies whether the request requires the iss (issuer) and sid (session ID) query parameters be included to identify the session with the OpenId Provider.
+   - For Back-Channel Logout the boolean value specifies whether the request requires the sid (session ID) Claim be included in the Logout Token to identify the session with the OpenId Provider.  This value must be `true` for back-channel logout to be executed.  If the value is `false` the `sid` will be marked for invalidation using a bounded map so that subsequent requests from the user will force the session to be invalidated.
+
+* `back-channel-logout-session-invalidation-limit`: The maximum size for a bounded map when marking sessions for invalidation during back-channel logout. Once this maximum size has been reached, the eldest entry will be removed.  The default size is 16.
```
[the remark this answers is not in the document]
rsearls: I inherited this RFE.  This is pre-existing information.  I think it is prudent to keep the default.  Customers can change this value through this attribute.

elytron/WFLY-19314-oidc-logout-support.adoc
```diff
@@ -0,0 +1,191 @@
...
+There are additional optional OIDC specifications that may be used in addition to the Core spec to
+provide support for logging out:
+
+* https://openid.net/specs/openid-connect-rpinitiated-1_0.html[RP-Initiated Logout] is an optional
+OIDC specification that defines how a relying party (RP) can request that an OpenID provider logs
+out the end-user.
+* https://openid.net/specs/openid-connect-frontchannel-1_0.html[Front-Channel Logout] is an optional
+OIDC specification that defines a front-channel logout mechanism that does not use an OP iframe on RP
+pages.
+* https://openid.net/specs/openid-connect-core-1_0.html#OpenID.BackChannel[Back-channel Logout] is an
```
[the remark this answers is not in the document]
rsearls: url ref changed

elytron/WFLY-19314-oidc-logout-support.adoc
```diff
@@ -0,0 +1,191 @@
...
+The use of `id_token_hint` is recommended by the spec.
+
+We'll introduce the following new configuration options:
+
+* `logout-path`: This option designates the path component of a URL.  It is the path component to trigger a logout of the user.  Elytron uses it to identify a logout request and take the appropriate action internally.  The default values is `/logout`.
+- The option can not be declared as an empty string or "/" only
+- It is expected that the text start with a forward slash (e.g. "/").
+- It can be a sequence of path segments (e.g. "/call/back/home")
+
+==== RP-Initiated Logout
```
OndrejKotek: This heading seems extra as the previous one is the same.

elytron/WFLY-19314-oidc-logout-support.adoc
rsearls: Description added

elytron/WFLY-19314-oidc-logout-support.adoc
```diff
@@ -0,0 +1,194 @@
...
+
+* `back-channel-logout-session-invalidation-limit`: The maximum size for a bounded map when marking sessions for invalidation during back-channel logout. Once this maximum size has been reached, the eldest entry will be removed.  The default size is 16.
+
+* `provider-jwt-claims-typ`: The logout specification defines the default (JWT) Logout Token "typ"
+       string to be "logout+jwt".  If the OpenID Provider uses a different
+       string, it can be specified with this element.
+       - This is an optional parameter.  If defined, it can not be the empty string or null.
+       - Elytron checks for Keycloak's "typ", "Logout" and "logout+jwt" by default.
+
+Note Keycloak does not support configuring both back-channel and front-channel logout for the same application.
```
OndrejKotek: This should not be about Keycloak but about WF Elytron OIDC client.

elytron/WFLY-19314-oidc-logout-support.adoc
```diff
@@ -0,0 +1,195 @@
...
+Example
+	The applications URL is https://localhost:8080/mySecureApp.  Using the default `logout-callback-path` value after logout the user would be sent to address of https://localhost:8080/mySecureApp/logout/callback .
+
+	The applications URL is https://localhost:8080/mySecureApp. If the callback address is to be https://www.wildfly.org/, the user would need to declare `logout-callback-path` to be `https://www.wildfly.org/`
+```
+
+
+* `logout-session-required`: A boolean value. The default value is true
+   - For Front-Channel Logout the boolean value specifies whether the request requires the iss (issuer) and sid (session ID) query parameters be included to identify the session with the OpenId Provider.
+   - For Back-Channel Logout the boolean value specifies whether the request requires the sid (session ID) Claim be included in the Logout Token to identify the session with the OpenId Provider.  This value must be `true` for back-channel logout to be executed.  If the value is `false` the `sid` will be marked for invalidation using a bounded map so that subsequent requests from the user will force the session to be invalidated.
```
OndrejKotek: What is the case for back-channel logout when someone wants logout-session-required set to false?

rsearls: In Keycloak, Backchannel Logout Session Required is a configuration setting 
within an OpenID Connect (OIDC) client that determines whether Keycloak must 
include a session identifier in the logout request sent to that client.

When True, Keycloak will only send logout requests that include a valid session ID. 
This is typically used by applications that manage multiple concurrent sessions 
for the same user and need to know exactly which one to clear.

When FALSE, The client may receive a logout token without a specific session ID, 
potentially signaling a "global" logout for that user across the application.

OndrejKotek: I understand these theoretical possibilities. However there is stated in this analysis that `This value must be true for back-channel logout to be executed.`, and IIUIC the implementation returns BAD REQUEST when the option is set to false. The description of the related subsystem attributes is also not helpful. I'm just trying to understand the behavior we want to implement and document.

elytron/WFLY-19314-oidc-logout-support.adoc
```diff
@@ -0,0 +1,195 @@
...
+- It can be a sequence of path segments (e.g. "/call/back/home")
+* `post-logout-redirect-uri`: This attribute must be an absolute URI.  It must be registered with the OpenId Provider.
+This URI will be called after logout is executed.
+
+
+==== Front-Channel Logout and Back-Channel Logout
+
+For Front-Channel and Back-Channel Logout, we'll introduce the following new configuration options:
+
+* `logout-callback-path`: This is an optional property. It designates the path component of a URL
```
OndrejKotek, line 111: This needs to be updated according to https://github.com/wildfly/wildfly/pull/18781#discussion_r3072844169

elytron/WFLY-19314-oidc-logout-support.adoc
rsearls: I have not been able to find a de finitive answer.  
I suggest creating a  jira to follow up on this question.
As of the jira upgrade I don't have creation permission.
Here is what I intended to post.

Title:  Review proper behavior for OIDC Logout's logout-session-required=false

Description:
It is unclear what the proper (required) behavior is for BackChannel logout
when logout-session-required=false.

The spec,  https://openid.net/specs/openid-connect-backchannel-1_0.html
states,
" backchannel_logout_session_required
    OPTIONAL. Boolean value specifying whether the RP requires that a sid (session ID) 
    Claim be included in the Logout Token to identify the RP session with the OP when 
    the backchannel_logout_uri is used. If omitted, the default value is false.
"

It needs to be determined what Keycloak returns in this case and if any
special action is required of Elytron.

elytron/WFLY-19314-oidc-logout-support.adoc
rsearls: Proposal was updated

elytron/WFLY-19314-oidc-logout-support.adoc
OndrejKotek: Well, the proposed Jira issue looks OK, however we should define our current expectations (so that it's clear what to test) and document the implementation accordingly so that it's clear to users how to use the attribute.

elytron/WFLY-19314-oidc-logout-support.adoc
OndrejKotek: Do I understand it correctly that any OIDC client configuration will have the new logout functionality enabled by default? The main attribute `logout-path` has a value by default (`/logout`), hence there is no option to turn it off.

elytron/WFLY-19314-oidc-logout-support.adoc
OndrejKotek: > It designates the path component of a URL which is called after logout has completed. The value of this property must be an absolute URL.

These two sentences seem to contradict each other. Also, the name of the attribute is misleading in case it should be a absolute URL. Looking closer to the implementation the attribute really seems to be a path, even having default value of `"/logout/callback"`.

elytron/WFLY-19314-oidc-logout-support.adoc
```diff
@@ -0,0 +1,188 @@
...
+
+==== RP-Initiated Logout
+
+For RP-Initiated Logout, we'll make use of the `id_token_hint` in the logout request to the logout endpoint.
+The use of `id_token_hint` is recommended by the spec.
+
+We'll introduce the following new configuration options:
+
+* `logout-path`: This option designates the path component of a URL.  It is the path component to trigger a logout
+of the user.  Elytron uses it to identify a logout request and take the appropriate action internally.  The default
```
OndrejKotek:
```suggestion
of the user from OP.  Elytron uses it to identify a logout request and take the appropriate action internally (redirecting to OP's Logout Endpoint with `id_token_hint` and `post_logout_redirect_uri` parameters if available).  The default
```

elytron/WFLY-19314-oidc-logout-support.adoc
```diff
@@ -0,0 +1,188 @@
...
+values is `/logout`.
+- The option can not be declared as an empty string or "/" only
+- It is expected that the text start with a forward slash (e.g. "/").
+- It can be a sequence of path segments (e.g. "/call/back/home")
+* `post-logout-redirect-uri`: This attribute must be an absolute URI.  It must be registered with the OpenId Provider.
+This URI will be called after logout is executed.
+
+
```
OndrejKotek, line 106:
```suggestion
The local log out of the End-User is not performed before redirecting the User Agent to the OP's Logout Endpoint. We rely on a logout message from the OP triggering the Front-Channel Logout or Back-Channel Logout mechanism.

```

elytron/WFLY-19314-oidc-logout-support.adoc
rsearls: If the user has a URL that ends with "/logout" which they don't want used in this way they can disable that by redefining attribute logout-path to something else such as "/NEVER-logout" or "/lkjadsfjldj"

elytron/WFLY-19314-oidc-logout-support.adoc
OndrejKotek:
```suggestion
   - The value must be `true` for Back-Channel Logout to be executed.
```

elytron/WFLY-19314-oidc-logout-support.adoc
rsearls: I suggest creating a jira for this.  A broader discussion and evaluation is needed for such a change

elytron/WFLY-19314-oidc-logout-support.adoc
rsearls: Thank you.   I have updated the text accordingly.

elytron/WFLY-19314-oidc-logout-support.adoc
rsearls: Change implemented

elytron/WFLY-19314-oidc-logout-support.adoc
rsearls: Change implemented

elytron/WFLY-19314-oidc-logout-support.adoc
OndrejKotek: I don't think so. The original text here corresponds to the implementation and the documentation should be updated accordingly.

Issue #711, opened 2025-04-03
[WFLY-20384] Restore the original behaviour of RoleMapping login module in the mapped role mapper


pedro-hos:
### Description

**BACKGROUND**

Some people migrating from previous versions of Wildlfy to the latest Wildfly version are struggling with the migration of the RoleMapping login module.
The equivalent of the RoleMapping login module is the "mapped-role-mapper" where the same "behavior" could be reached.
The main pain of these customers comes from the fact they used a properties file to do the mapping between REALM roles to APPLICATION roles and with "mapped-role-mapper" this seems to not be possible anymore.

**ENGINEERING REQUEST**

Is it possible to integrate [1] mapper code in the Wildlfly? I have developed this for a customer who wanted an example of how to use the "mapped-role-mapper" to work as a RoleMapping login module.

I believe that [1] would help the ones who are migrating to the latest Wildfly versions, at least for the ones migrating from the legacy RoleMapping login module to Elytron

[1] https://github.com/gabrielpadilh4/custom-properties-mapped-role-mapper
[2] https://issues.redhat.com/browse/WFLY-20384

### Issue Contact

pedro-hos

PR #712, opened 2025-04-04
[WFLY-20384] - Restore the original behaviour of RoleMapping login module in the mapped role mapper

pedro-hos:
Issue: #711 
Jira: [WFLY-20384](https://issues.redhat.com/browse/WFLY-20384)


elytron/WFLY-20384-properties-role-mapper.adoc
```diff
@@ -0,0 +1,219 @@
...
+
+EnterpriseRole=role-admin,role-123
+```
+
+Based on that file, the new `PropertiesMappedRoleMapper.java` will map the roles.
+
+=== Changed requirements
+Update the schema n order to add the new `properties-role-mapper` configuration.
+
+=== Non-Requirements
```
OndrejKotek, line 88: IIUIC, the original login module is able to load also property files from deployments: https://issues.redhat.com/browse/WFCORE-7203 I assume such functionality is not planned.

elytron/WFLY-20384-properties-role-mapper.adoc
```diff
@@ -0,0 +1,219 @@
...
+N/A
+
+=== Future Work
+N/A
+
+== Backwards Compatibility
```
OndrejKotek, line 94: Is it possible to use a file for the original login module directly without changes in the new mapper?

elytron/WFLY-20384-properties-role-mapper.adoc
darranl: No managed components should be loading from deployed content.

OndrejKotek: Also, the original module has a `replaceRole` property. Also out of the scope?

elytron/WFLY-20384-properties-role-mapper.adoc
pedro-hos: I am sorry, I didn't got your question.

OndrejKotek: I think the Backward Compatibility section should discuss how migrate from the original login module to the new Elytron based/integrated solution. The configuration is completely different and there is not much to be compatible (`replaceRole` option?), the data/properties file used by the configuration should be ideally usable as is without modifications. My question/suggestion is to write here the assurance that the properties file can be used without any changes when migrating from the original login module to Elytron.

PR #713, opened 2025-04-09
WFLY-20532-Support client side and server side version handling 

rachmatowicz:
The analysis document for the EAP7 RFE: https://issues.redhat.com/browse/EAP7-2298


rachmatowicz:
Placed this PR in draft mode as it is not fully complete.

PR #721, opened 2025-04-30
[WFCORE-7192] Add brute force mitigation to the Elytron security realms.

darranl:
https://issues.redhat.com/browse/WFCORE-7192

### DIFF
--- a/elytron/WFCORE-7192-brute-force-mitigation.adoc
+++ b/elytron/WFCORE-7192-brute-force-mitigation.adoc
@@ -1,0 +1,234 @@
+---
+categories:
+- elytron
+- security
+stability-level: default
+issue: https://github.com/wildfly/wildfly-proposals/issues/719
+feature-team:
+ developer:
+  - darranl
+ sme:
+ outside-perspective:
+---
+= [WFCORE-7192] Brute Force Mitigation
+:author:            Darran Lofthouse
+:email:             darran.lofthouse@jboss.com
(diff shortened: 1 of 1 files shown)


elytron/WFCORE-7192-brute-force-mitigation.adoc
```diff
@@ -0,0 +1,234 @@
...
+  * SASL Digest
+  * SASL Scram
+  * SASL Plain
+  * HTTP Basic
+  * HTTP Digest
+  * HTTP Form
+  * HTTP Programmatic
+
+The wrapper will be applied automatically by the `elytron` subsystem, it will be possible to override the
+default behaviour by setting the following realm specific system properties:
```
OndrejKotek, line 91: Would it be useful to have also global settings for all security realms to be configurable at once? (E.G. just by omitting `[REALM NAME]`.)

darranl: @OndrejKotek I have been thinking about this and IMO we should not add a global option.

The system property approach we are taking here is a short term configuration option until we are ready to add management attributes to configure this - once we add those attributes they will be on a realm by realm basis so we will not have a global equivalent.

We could consider a global management option but we don't really do that for other resources so I think I would prefer to keep per realm unless we receive end user requests for a global config option.

OndrejKotek (approved):

PR #724, opened 2025-05-09
[WFCORE-5744] Improve key-store certificate health checks

baranowb:
Issues:
https://issues.redhat.com/browse/WFCORE-5744
https://issues.redhat.com/browse/EAP7-1863


elytron/WFCORE-5744-ssl-certificate-expiration-warning.adoc
```diff
@@ -0,0 +1,218 @@
...
+
+Long runing server, with periodic information on certificates will allow admins to not only setup tools to warn them about potential security risk, but also schedule maintanence window based on this.
+
+== Issue Metadata
+
+
+
+=== Related Issues
+
+* https://issues.redhat.com/browse/EAP7-1863[EAP7-1863]
```
darranl, line 58: Please remove the EAP7 references, at this time this is being proposed as a community feature.

elytron/WFCORE-5744-ssl-certificate-expiration-warning.adoc
```diff
@@ -0,0 +1,218 @@
...
+ * read-alias command will have additional KVP "Validity=[valid, EXPIRED, about to expire, not yet]"
+ ** Validity semantics:
+ *** not yet - ceritifacete is still too young
+ *** EXPIRED - current date is after expiration
+ *** about to expire - certificate is still valid. If certificate lifespan is week or smaller, this rougly corresponds to day to EOL, otherwise 7 days to EOL
+ *** valid - more than above boundries
+ * add config options to handle:
+ ** frequency of warning
+ *** 0 - turn off - one time check on startup as in previous version
+ *** n+ - millisecconds between periodic checks
```
darranl, line 86: IMO ms is too fine a granularity for a check, I can't imagine a situation where some users will want their keystores hitting every 1,000 ms whilst others would want it every 500ms.

elytron/WFCORE-5744-ssl-certificate-expiration-warning.adoc
```diff
@@ -0,0 +1,218 @@
...
+
+
```
darranl, line 88: If we are going to do some periodic checks I think validity intervals may need to be configurable, I can imaging different users having different notification requirements.

elytron/WFCORE-5744-ssl-certificate-expiration-warning.adoc
```diff
@@ -0,0 +1,218 @@
...
+NOTE: above is subject to change based on feedback( default to 0, turn-off periodic so its essentially like this change wasnt introduced)
+
+=== Default Configuration
+
+__<Does the proposed work change the default value of any current configuration attributes? Does it change the configuration generated by any current Galleon layers?>__
+
+=== Importing Existing Configuration
+
+Should default to predefined values.
```
darranl, line 114: I don't think we have the default values specified in this analysis?

elytron/WFCORE-5744-ssl-certificate-expiration-warning.adoc
```diff
@@ -0,0 +1,218 @@
...
+        }]
+    }
+}
+----
+
+* Global config:
+
+== Community Documentation
+
+Model/XSD description should be enough as change is not deep and only introudce simple config parameter and additional output KVP in existing command.
```
darranl, line 212: I think there should be a little more in the community documentation, if a user is activating (assuming it is them activating it) a periodic check I think we should have a small paragraph to let them know that that check entails.

elytron/WFCORE-5744-ssl-certificate-expiration-warning.adoc
```diff
@@ -0,0 +1,218 @@
...
+:idprefix:
+:idseparator:       -
+
+__<The entire document should be one to two pages long. We will write each analysis document as if it is a conversation with a future developer. This requires a good writing style, with full sentences organized into paragraphs. Bullets are acceptable only for visual style, not as an excuse for writing sentence fragments.>__
+
+== Overview
+
+Goal of this RFE is two fold, to formalize and improve ways to get feedback on state of SSL certificates in server and to add periodic check, based on configuration.
+
+Periodic check is more or less self explanatory. Currently upon startup elytron subsystem will will print out information about certificates that expired. There are few problems with this. Firstly its exception driven and lacks a bit of information( ie keystore). Secondly its one-off on startup, this operation potentially should be performed periodically.
```
darranl, line 42: I do think we should describe how we will handle this periodic check as this could have performance / testing implications.

But I would also consider for the different KeyStore types is a periodic check really required? i.e. once we have iterated once for many KeyStores the certificates are not going to change unless we reload the store so we know the expiration dates from the set of certificates.

One risk I also see is KeyStores could contain keys that are not used in the installation, these risk creating noise in the logs.

elytron/WFCORE-5744-ssl-certificate-expiration-warning.adoc
```diff
@@ -0,0 +1,218 @@
...
+CLI should not be affected. However, HAL will most likely require follow up, since this feature will introduce new model params/context. Given defaults, its not a breaking change.
+
+== Security Considerations
+
+Slightly increase security risk compared to current state. If attacker has access to log/events it will provide view of current health of server, rather than possible peek on startup. Convenient, but not something that could not be circumvented without this feature.
+
+[[test_plan]]
+== Test Plan
+
+Integration tests should be fairly acceptable.
```
darranl, line 140: We should expand on this section, we have a number of supported KeyStore types and some of these even have sub types so we should specify the tests and the permutations we will run.

elytron/WFCORE-5744-ssl-certificate-expiration-warning.adoc
```diff
@@ -0,0 +1,218 @@
...
+__<Is this feature impacting interoperability?>__
+No.
+
+== Implementation Plan
+
+N/A
+
+== Admin Clients
+
+CLI should not be affected. However, HAL will most likely require follow up, since this feature will introduce new model params/context. Given defaults, its not a breaking change.
```
darranl, line 131: We should be able to work with @hpehl on this one once the configuration model is agreed to check if any specifics are needed for HAL, if they are they will need to be addressed before we can merge at the community stability level.

elytron/WFCORE-5744-ssl-certificate-expiration-warning.adoc
```diff
@@ -0,0 +1,218 @@
...
+* Bootable jar
+
+== Requirements
+
+ * create output format that is not driven by exception, so possibly it can be easily digested via log munching, mail notification or any other mechanism users see fit to employ over log handler.
+ * read-alias command will have additional KVP "Validity=[valid, EXPIRED, about to expire, not yet]"
+ ** Validity semantics:
+ *** not yet - ceritifacete is still too young
+ *** EXPIRED - current date is after expiration
+ *** about to expire - certificate is still valid. If certificate lifespan is week or smaller, this rougly corresponds to day to EOL, otherwise 7 days to EOL
```
darranl, line 81: As I mention above, I think any validity tolerance may need to be configurable.

elytron/WFCORE-5744-ssl-certificate-expiration-warning.adoc
```diff
@@ -0,0 +1,218 @@
...
+# Specify the stability level of the feature.
+# Values can be one of: experimental preview community default
+stability-level:
+ - Community
+issue:
+ - https://github.com/wildfly/wildfly-proposals/issues/685
+feature-team:
+ developer: bbaranow@redhat.com
+ sme:
+  - dlofthou@redhat.com
```
darranl, line 15: Have you asked me to be SME?

elytron/WFCORE-5744-ssl-certificate-expiration-warning.adoc
```diff
@@ -0,0 +1,218 @@
...
+N/A
+
+
+=== Non-Requirements
+
+N/A
+
+=== Future Work
+
+N/A
```
darranl, line 100: This should not block this specific implementation but it does bring up a feature again I want to explore I have been calling "Administrator Encouragement" ideally it would be a service registered more centrally maybe in the management subsystem but we could do the equivalent local to the Elytron subsystem.

The idea is there are a lot of things we discover could be improved in the configuration of the server at runtime, possibly some could be detected by static analysis of config but many are runtime only and this certificate expiration is a good example.

So once the server is running the idea is we have this capability that individual subsystems can depend on and as runtime problems are discovered they can be registered with this service.

As the service is central our admin clients can then use management ops on this service to query the current known issues, so HAL could detect we have warnings about certificate expiration.  The admin clients could detect some common issues and even provide guidance to resolve them,  In the certificate space another example is self signed certificates.

Anyway that should not block this solution to log to the server logs but it could be a future feature we can consider.  IMO if we did consider it I would suggest that features starts at Preview so we can start to add subsuystems to use it and then promote after we have demonstrated some stability.

darranl (changes requested):
I have added some comments inline.

I think the main things to begin with are we need to lose the references to the EAP7 as this is Community stability level and begin the WildFly Feature development process and call for volunteers to participate in the feature team.

elytron/WFCORE-5744-ssl-certificate-expiration-warning.adoc
baranowb: not yet, consider this a grace period.

elytron/WFCORE-5744-ssl-certificate-expiration-warning.adoc
baranowb: Ok. Initially I was leaning towards having cron like regex, but ATM WFLY does not have this in codebase/deps.

elytron/WFCORE-5744-ssl-certificate-expiration-warning.adoc
```diff
@@ -0,0 +1,218 @@
...
+=== Other Interested Projects
+
+=== Relevant Installation Types
+
+* Traditional standalone server (unzipped or provisioned by Galleon)
+* Managed domain
+* OpenShift Source-to-Image (S2I)
+* Bootable jar
+
+== Requirements
```
bstansberry, line 74: There should be a requirement(s) around administrative RBAC controls related to this.

This is something we've probably missed in a lot of analyses, so I'm being a bit unfair asking for it here, but we should start making it a practice to cover this and thus hopefully ensure the code does too.

@darranl Do you think this is something to mention in the template comment in 'Security Considerations'? The info would naturally be Requirement items, but the suggestion comment more naturally fits in security considerations.

elytron/WFCORE-5744-ssl-certificate-expiration-warning.adoc
baranowb: Done. However I wonder if default in case of periodic be  0 - ie, its turned off and server works as without change.

elytron/WFCORE-5744-ssl-certificate-expiration-warning.adoc
darranl: For this proposal I think it should be an ACK that the Elytron subsystem has RBAC defined at the root - but +1 I think the relationship to RBAC would be good to call out and test in the default template.

bstansberry: https://github.com/wildfly/wildfly-proposals/pull/726

baranowb: @darranl so no need to include this in proposal?

skyllarr (approved):

elytron/WFCORE-5744-ssl-certificate-expiration-warning.adoc
baranowb: HAL picks up those automatically.

darranl:
FYI I think we are missing a WFLY Jira issuse to update the documentation and the documentation PR.

PR #727, opened 2025-05-15
Template update

jasondlee:
- Rework wording around promotes to promoted-by for clarity 
- General document clean up (normalizing line lengths, etc)

### DIFF
--- a/design-doc-template.adoc
+++ b/design-doc-template.adoc
@@ -1,19 +1,19 @@
 ---
-# Add any category for this proposal as a yaml list, e.g.
+# Add any category for this proposal as a YAML list, e.g.
 # - core
 # - management
-# if missing, add it to _data/wildfly-categories and use its id
+# If missing, add it to _data/wildfly-categories and use its id
 categories:
 # Specify the stability level of the feature.
-# Values can be one of: experimental preview community default
+# Values can be one of: experimental, preview, community, or default
 stability-level:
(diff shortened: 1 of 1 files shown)


design-doc-template.adoc
```diff
@@ -1,33 +1,38 @@
...
 feature-team:
  developer:
  sme:
   -
  outside-perspective:
   -
-# If this issue tracks the promotion to a higher stability level of a previously
-# completed feature, provide the URL of the https://github.com/wildfly/wildfly-proposals/issues
-# issue that was used to track the previous feature.
+# During initial development of a feature, this field should be left blank. If this issue tracks the promotion of a
```
bstansberry, line 24: I think 'During initial development of a feature, this field should be left blank.' should be removed. I think you mean leave it blank in the very first, non-promotion feature analysis, but 'initial development' is unclear. I'd interpret it as meaning 'early work on the promotion that this document covers' and then wonder why I'm to leave it blank. 'If this issue tracks the promotion' covers the 'leave it blank' case.

bstansberry (commented):
Thanks. The one point I commented on aside, LGTM.

design-doc-template.adoc
jasondlee: Updated.

bstansberry:
Thanks @jasondlee

PR #733, opened 2025-06-05
Make comments real AsciiDoc comments

jasondlee:
Convert the comments to real AsciiDoc comments so that they can remain in the document to help guide future edits without actually appearing in the rendered document itself.


jamezp (approved):
This makes sense to me.

bstansberry:
@jmesnil IIRC it was this way before but you changed it. Any inputs here?

jamezp (approved):
Having these as comments makes sense to me. This allows us to leave comments in for sections we want without the placeholder being rendered.

rhusar (changes requested):
👎🏼 

I got burned by this too but I believe this is intentional.

Also, all of these SHOULD be removed when somebody opens a proposal. I don't want to waste time looking through text that I should ignore. That's not a valid use of our time.

design-doc-template.adoc
```diff
@@ -44,36 +44,48 @@ promoted-by:
 :idprefix:
 :idseparator:       -
 
-__<The entire document should be one to two pages long. We will write each analysis document as if it is a conversation
+////
+The entire document should be one to two pages long. We will write each analysis document as if it is a conversation
 with a future developer. This requires a good writing style, with full sentences organized into paragraphs. Bullets are
-acceptable only for visual style, not as an excuse for writing sentence fragments.>__
+acceptable only for visual style, not as an excuse for writing sentence fragments.
+////
```
bstansberry: Keep as a visible comment and force the writer to delete so reader doesn't have to.

design-doc-template.adoc
```diff
@@ -44,36 +44,48 @@ promoted-by:
...
 
 === User Stories
 
-__<Provide one or more brief user stories that illustrate the intended users of the feature and the goal they will seek
-to achieve by using the feature.>__
+////
+Provide one or more brief user stories that illustrate the intended users of the feature and the goal they will seek
+to achieve by using the feature.
+////
```
bstansberry: This is good as an adoc comment as it reminds the reviewer what a user story should be like.

design-doc-template.adoc
```diff
@@ -44,36 +44,48 @@ promoted-by:
...
 
 == Issue Metadata
 
 === Related Issues
 
-__<List the issues related to this feature>__
+////
+List the issues related to this feature
+////
```
bstansberry: Keep as a visible comment and force the writer to delete so reader doesn't have to.

design-doc-template.adoc
```diff
@@ -44,36 +44,48 @@ promoted-by:
...
 
 === Affected Projects or Components
 
-__<List the projects or components that are affected by the feature. List them using their Git repositories.>__
+////
+List the projects or components that are affected by the feature. List them using their Git repositories.
+////
```
bstansberry: Keep as a visible comment and force the writer to delete so reviewer doesn't have to process it.

In theory the "List them using their Git repositories" could be adoc to driver reviewers to enforce that. I don't know how much we actually care about enforcing it though.

design-doc-template.adoc
```diff
@@ -44,36 +44,48 @@ promoted-by:
...
 
 === Other Interested Projects
 
 === Relevant Installation Types
 
-__<List the installation types that are relevant for the features and remove any that are not relevant>__.
+////
+List the installation types that are relevant for the features and remove any that are not relevant.
+////
```
bstansberry: Keep as a visible comment and force the writer to delete so reviewer doesn't have to process it.

design-doc-template.adoc
```diff
@@ -82,73 +94,98 @@ __<List the installation types that are relevant for the features and remove any
...
+Describe the requirements that must be fulfilled by this feature.
 
-__<For analyses of a promotion of an existing feature to 'preview' or 'community' stability, only list new requirements;
+For analyses of a promotion of an existing feature to 'preview' or 'community' stability, only list new requirements;
 existing requirements from the feature being promoted are assumed to continue unless otherwise noted in the 'Changed
 requirements' section. Other analyses, including those for promotion to the 'default' stability level, must list all
-requirements.>__
+requirements.
+////
```
bstansberry: OK as adoc comment as it reminds the reviewer of how things are meant to be done.

design-doc-template.adoc
```diff
@@ -82,73 +94,98 @@ __<List the installation types that are relevant for the features and remove any
...
+Use this section to explicitly discuss things that readers might think are required but which are not required.
+////
 
 === Future Work
 
-__<Use this section to discuss requirements that are not addressed by this proposal but which may be addressed in later proposals.>__
+////
+Use this section to discuss requirements that are not addressed by this proposal but which may be addressed in later proposals.
+////
```
bstansberry: Keep as a visible comment and force the writer to delete so reader doesn't have to process it.

design-doc-template.adoc
```diff
@@ -82,73 +94,98 @@ __<List the installation types that are relevant for the features and remove any
...
 
 == Backwards Compatibility
 
-__<Does this enhancement affect backwards compatibility with previously released versions of WildFly? Can the identified
-incompatibility be avoided?>__
+////
+Does this enhancement affect backwards compatibility with previously released versions of WildFly? Can the identified
+incompatibility be avoided?
```
bstansberry:
```suggestion
incompatibility be avoided? Discussion should go in the various subsections.
```

design-doc-template.adoc
```diff
@@ -82,73 +94,98 @@ __<List the installation types that are relevant for the features and remove any
...
+////
```
bstansberry: Good adoc comment. Reviewers need reminders of what this area is meant to cover.

design-doc-template.adoc
```diff
@@ -82,73 +94,98 @@ __<List the installation types that are relevant for the features and remove any
...
+Is this feature impacting interoperability?
+////
 
 == Implementation Plan
 
-__<This section is optional. If you have a complex feature which cannot be delivered all in one go, suggest the strategy.>__
+////
+This section is optional. If you have a complex feature which cannot be delivered all in one go, suggest the strategy.
+////
```
bstansberry: Keep as a visible comment and force the writer to delete so reader doesn't have to process it.

design-doc-template.adoc
```diff
@@ -82,73 +94,98 @@ __<List the installation types that are relevant for the features and remove any
...
-__<Is this feature impacting interoperability?>__
+////
+Is this feature impacting interoperability?
+////
 
 == Implementation Plan
 
-__<This section is optional. If you have a complex feature which cannot be delivered all in one go, suggest the strategy.>__
+////
+This section is optional. If you have a complex feature which cannot be delivered all in one go, suggest the strategy.
```
bstansberry:
```suggestion
This section is optional. If you have a complex feature which cannot be delivered all in one go, suggest the strategy. Otherwise, delete this section.
```

design-doc-template.adoc
```diff
@@ -174,13 +211,19 @@ of what the test is doing.
...
-__Example article: https://www.wildfly.org/news/2024/01/25/WildFly31-Released/.__
-__This content will be edited, so there is no need to make it perfect or discuss what release it appears in.>__
+////
+Draft verbiage for up to a few sentences on the feature for inclusion in the Release Note blog article for the release
+that first includes this feature.
+
+Example article: https://www.wildfly.org/news/2024/01/25/WildFly31-Released/
+
+This content will be edited, so there is no need to make it perfect or discuss what release it appears in.
+////
```
bstansberry: Good adoc comment. Reviewers need reminders of what this area is meant to cover.

design-doc-template.adoc
```diff
@@ -82,73 +94,98 @@ __<List the installation types that are relevant for the features and remove any
...
 == Test Plan
 
-__<Depending on the selected stability level, the appropriate section below should be completed, including a brief
-description of how testing is to be performed in accordance with the selected stability level. The non-relevant sections
-may be removed as needed.>__
 ////
+Depending on the selected stability level, the appropriate section below should be completed, including a brief
+description of how testing is to be performed in accordance with the selected stability level. The non-relevant sections
+may be removed as needed.
+
```
bstansberry: This feels more like something that should be deleted, but since it's a preface to a whole block of adoc commentary I think having it be adoc as well is fine.

design-doc-template.adoc
```diff
@@ -44,36 +44,48 @@ promoted-by:
 :idprefix:
 :idseparator:       -
 
-__<The entire document should be one to two pages long. We will write each analysis document as if it is a conversation
+////
+The entire document should be one to two pages long. We will write each analysis document as if it is a conversation
 with a future developer. This requires a good writing style, with full sentences organized into paragraphs. Bullets are
-acceptable only for visual style, not as an excuse for writing sentence fragments.>__
+acceptable only for visual style, not as an excuse for writing sentence fragments.
```
bstansberry:
```suggestion
acceptable only for visual style, not as an excuse for writing sentence fragments.

NOTE: this template includes instruction comments like this one that will visibly render and must be removed. 
It also includes instruction comments in asciidoc format that will not render and can be left in place or removed, at the discretion of the analysis author.
```

bstansberry (commented):
I've made a bunch of comments based on my previous "middle ground" comment. Maybe that's a bad idea; I don't know. It makes things more conceptually complex. But if it seems ok to folks, the comments below are a suggestion for how to implement.  (I probably mixed in a few other possibly OT suggested improvements too.)

PR #739, opened 2025-06-20
[WFGP-292] Generate documentation for Galleon packs

### LINKED ISSUE #738: Generate documentation for feature packs
### Description

WildFly ecosystem provides multiple feature packs but they are lacking some documentation to make them on par with WildFly (model reference, metadata information about the feature pack, its layers, etc.)

When a feature pack is built, we should also generate a companion doc archive that describes the feature pack. This would be very similar to javadoc for Java libraries.

### Issue Contact

jmesnil

jmesnil:
This fixes #738.


provisioning/WFGP-292_doc_galleon-pack.adoc
```diff
@@ -0,0 +1,159 @@
...
+
+== Implementation Plan
+
+=== Galleon
+
+* We might need to expose additional metadata for feature pack through the Galleon API (to be investigated).
+
+=== Wildscribe
+
+* Modify wildscribe's site-generator so that it can read the DMR output of a `:read-resource-description` from a JSON file (in addition to read it from a DMR file).
```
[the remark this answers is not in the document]
jamezp: Writing something more modern to replace wildscribe makes sense to me. It uses velocity which isn't bad, but something more modern might make more sense.

provisioning/WFGP-292_doc_galleon-pack.adoc
```diff
@@ -0,0 +1,161 @@
...
+*** the Model reference of the resources provided by the feature pack (in a format similar to the existing https://docs.wildfly.org/36/wildscribe/[WildFly model reference])
+*** the log messages that can be displayed by the feature pack (in a format similar to the existing https://docs.wildfly.org/36/wildscribe/log-message-reference.html[WildFly log messages])
+** the model reference of the resources provided by the feature pack (stored in a DMR representation in a JSON file) - generated during the `build-feature-pack` goal and stored in a `META-INF/model.json` file
+** the metadata of the feature pack (stored in a JSON file) - generated during the `build-feature-pack` goal and stored in a `META-INF/metadata.json` file
+* This companion doc.zip is attached to the feature pack's Maven model with the coordinates:
+** `groupId` - same as the feature pack
+** `artifactId` - same as the feature pack
+** `version` - same as the feature pack
+** `classifier` - `doc`
+** `extension` - `zip`
```
jamezp, line 96: One thing we'll want to double check with this is that attached artifacts get signed with the gpg plugin before being published. This might already happen, but just something to look out for as we will need to start signing artifacts.

provisioning/WFGP-292_doc_galleon-pack.adoc
```diff
@@ -0,0 +1,161 @@
...
+* Other tools that wants to leverage this documentation artifacts must not expect they are always availavle alongside the feature pack. It will only exist when feature packs project are upgraded to the Galleon Maven plugin that provides this feature. 
+
+=== Non-Requirements
+
+* the Look and Feel of the generated documenation is not configurable (it is based on wildscribe L&F).
+
+=== Future Work
+
+* This documentation archive can be leveraged by other tools to provide an extensive and exhaustive documentation for all feature packs provided by the WildFly community
+* We might consider adding the documentation of the feature packs that are used to provision a WildFly installation *inside* the installation itself.
```
jamezp, line 107: One nice thing to have too, which we don't need right away, would be to have a generated searchable index for searching the docs. It's a bit out of date now and didn't quite work as I'd expect, but I had done this for Wildscribe https://github.com/wildfly/wildscribe/compare/main...jamezp:wildscribe:attribute-search. Having a real index instead of creating one on the fly is probably better though.

provisioning/WFGP-292_doc_galleon-pack.adoc
```diff
@@ -0,0 +1,161 @@
...
+** `artifactId` - same as the feature pack
+** `version` - same as the feature pack
+** `classifier` - `doc`
+** `extension` - `zip`
+* When the feature pack is installed or published, this companion doc.zip is also installed and published
+* Other tools that wants to leverage this documentation artifacts must not expect they are always availavle alongside the feature pack. It will only exist when feature packs project are upgraded to the Galleon Maven plugin that provides this feature. 
+
+=== Non-Requirements
+
+* the Look and Feel of the generated documenation is not configurable (it is based on wildscribe L&F).
```
bstansberry, line 102: This should specify that coverage of Galleon constructs besides the feature pack itself is not in scope.

* Config
* *Layer
* Feature
* Feature group
* Package

If any of those are in scope, the requirements section should describe the details.

provisioning/WFGP-292_doc_galleon-pack.adoc
```diff
@@ -0,0 +1,161 @@
...
+
+== Requirements
+
+
+* When the `doc` goal of the `org.wildfly.galleon-plugins:wildfly-galleon-maven-plugin` is invoked (i.e. when the feature pack is built), generate a `doc.zip` archive that contains:
+** human-readable HTML documention of the feature pack stored in a `doc` directory. This documentation includes:
+*** the Model reference of the resources provided by the feature pack (in a format similar to the existing https://docs.wildfly.org/36/wildscribe/[WildFly model reference])
+*** the log messages that can be displayed by the feature pack (in a format similar to the existing https://docs.wildfly.org/36/wildscribe/log-message-reference.html[WildFly log messages])
+** the model reference of the resources provided by the feature pack (stored in a DMR representation in a JSON file) - generated during the `build-feature-pack` goal and stored in a `META-INF/model.json` file
+** the metadata of the feature pack (stored in a JSON file) - generated during the `build-feature-pack` goal and stored in a `META-INF/metadata.json` file
```
bstansberry, line 90: What metadata is intended?

provisioning/WFGP-292_doc_galleon-pack.adoc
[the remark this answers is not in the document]
jmesnil: In my prototype, I moved away from WildScribe and uses Qute as the template engine to generate the doc.
Wildscribe would become obsolete and be archived.

provisioning/WFGP-292_doc_galleon-pack.adoc
jmesnil: yes, I've updated the requirements section.

provisioning/WFGP-292_doc_galleon-pack.adoc
jmesnil: I updated the requirements section

provisioning/WFGP-292_doc_galleon-pack.adoc
```diff
@@ -0,0 +1,164 @@
...
+* https://issues.redhat.com/browse/WFGP-292[WFGP-292 - Generate documentation for Feature Pack]
+
+=== Affected Projects or Components
+
+* WildFly Galleon Plugins - https://github.com/wildfly/galleon-plugins
+** Generate documentation when the `build-feature-pack` goal is invoked. It generates a companion doc archive next to the feature pack itself
+* Galleon - https://github.com/wildfly/galleon
+** Might be impacted if we are missing any metadata from the feature pack to generate its documentation
+* Wildscribe - https://github.com/wildfly/wildscribe
+** This RFE would be a full replacement of Wildscribe that would be archived.
```
bstansberry, line 69: I think archiving Wildscribe should be out of scope. It is used downstream and we shouldn't mix downstream-impacting changes into this, as doing so just adds unnecessary friction.

I suggest replacing "that would be archived" with a sub-bullet saying "The https://github.com/wildfly/wildscribe/blob/main/README.md will be updated to state that the project is in maintenance mode and is no longer used by the WildFly application server itself."

provisioning/WFGP-292_doc_galleon-pack.adoc
```diff
@@ -0,0 +1,164 @@
...
+
+*_"Javadoc" for feature packs_*
+
+The goal of that proposal is to provide documentation of the capabilities and resources that are provided by WildFly and its ecosystem of Galleon feature packs.
+This documentation is self-contained and automatically generated when the feature packs are built.
+
+== Overview
+
+At its core, WildFly is composed of extensions that provide management resources (grouped in subsystems and deployment resources).
+WildFly is providing a model reference for these resources at https://docs.wildfly.org/36/wildscribe/.
```
bstansberry, line 32: WDYT about starting to call this "management API reference" not "model reference"?

We've used the term "model" around this general area for years but I couldn't tell you precisely what we mean by it. And I suspect a user has no clue.

provisioning/WFGP-292_doc_galleon-pack.adoc
```diff
@@ -0,0 +1,164 @@
...
+
+This feature will be documented in the Galleon Maven Plugin project and is meant to be read by feature pack developers.
+
+However once, this feature is delivered, it will have a high impact on community documentation as any of the feature pack provided by the WildFly community will be able to display this documentation (either in a standalone website, ususally on github.io, or aggregated in docs.wildfly.org).
+
+In particular, once this feature is delivered, it is recommended that feature pack developers review and check their documentation (model reference, metadata, etc.) to ensure that the information are correct, accurate and helpful.
+
+== Release Note Content
+
+Feature packs that are used to provision WildFly provide now generated documentation that describes the feature pack, its layers and the model references of the subsystems provided by the feature pack.
```
bstansberry, line 164: s/provide now/now provide/g

Perhaps

"describes the feature pack and its layers, along with the management API reference for the management resources provided by the feature pack."

^^^ is partly about readability, partly about my point from the overview about the unclear term "model reference" and partly about the fact that feature packs provide server kernel resources, not just subsystems.

provisioning/WFGP-292_doc_galleon-pack.adoc
```diff
@@ -0,0 +1,164 @@
...
+
+Traditionally, this model reference was exhaustive and provided the full management capabilities of WildFly.
+It relied on its standalone-full-ha.xml configuration to provide it.
+
+However, over time, this approach has shown some cracks:
+
+* Some new subsystems are provided by other configurations (e.g. standalone-microprofile.xml). This was worked around by adding additional extensions when the model referenced is generated (in WildFly's `docs` module) but is out of sync.
+* WildFly is now composed of 2 Galleon packs (wildfly-ee-galleon-pack & wildfly-galleon-pack). Each Galleon pack provides different extensions and resources but this is not clear from the management model (e.g. which Galleon pack provides the `microprofile-config-smallrye` subsystem?)
+* Additional capabilities for WildFly are provided by other feature packs (mvc-krazo, grpc, wildfly-preview, etc.) that do not provide their model references and makes it more difficult for the users to understand which resources, attributes and operations are available to configure their installations. They have to rely on jboss-cli to get the information and while it is available, it does not make it very user-friendly.
```
bstansberry, line 41: The mvc-krazo FP isn't a good example as the production code jar and galleon content that GitHub project provides are incorporated in the 'wildfly' feature pack. That project still builds its own FP but it's not really used.

BTW we do the same with vertx, so don't switch your example to that. ;)

provisioning/WFGP-292_doc_galleon-pack.adoc
```diff
@@ -0,0 +1,164 @@
...
+** This RFE would be a full replacement of Wildscribe that would be archived.
+* Any project publishing feature packs (wildfly, grpc-feature-pack, ai-feature-pack, etc.)
+** They will be updated to bump the version of the Galleon maven plugin
+** *No other changes should be required*
+* WildFly - https://github.com/wildfly/
+** Its model references will be generated when its feature packs are built.
+*** For WildFly, we could generate a single documentation entry points that would document the wildfly-galleon-pack and also the capabilites coming from the wildfly-ee-galleon-pack.
+** The dependency to wildscribe from its docs module will be removed and replaced by unarchiving the documentation when the  wildfly-galleon-pack feature pack is built.
+** There would be no changes to the WildFly release process.
+
```
bstansberry, line 78: Suggested detail requirement

* Documentation for feature packs that don't provide layers or don't provide their own management resources will be appropriately formatted.

I just intend this as a reminder to check this, or at least consciously handle it in the implementation so we don't end up with unprofessional output in these scenarios.

For example, wildfly-datasources-galleon-pack provides no new management resource API. https://github.com/wildfly-extras/deployment-transformer-feature-pack provides neither layers nor any new management API

provisioning/WFGP-292_doc_galleon-pack.adoc
jmesnil: I agree, this statement was too aggressive. I've used your suggestion.

Regarding downstream uses, this RFE could also be interesting to apply to downstream feature packs that are lacking this kind of documentation at the moment too. But that's out of scope of what we want to achieve in that proposal

provisioning/WFGP-292_doc_galleon-pack.adoc
```diff
@@ -0,0 +1,164 @@
...
+* WildFly Galleon Plugins - https://github.com/wildfly/galleon-plugins
+** Generate documentation when the `build-feature-pack` goal is invoked. It generates a companion doc archive next to the feature pack itself
+* Galleon - https://github.com/wildfly/galleon
+** Might be impacted if we are missing any metadata from the feature pack to generate its documentation
+* Wildscribe - https://github.com/wildfly/wildscribe
+** This RFE would be a full replacement of Wildscribe that would be archived.
+* Any project publishing feature packs (wildfly, grpc-feature-pack, ai-feature-pack, etc.)
+** They will be updated to bump the version of the Galleon maven plugin
+** *No other changes should be required*
```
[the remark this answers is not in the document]
jmesnil: Actually, I was wrong, we don't need a new Galleon release, @jfdenise is using existing layers-spec's properties to provide this additional info.
And of course, he already did the work for WildFly :) 
He added `description` & `note` properties to the WildFly layer-spec.xml (I cherry-picked his [commit](https://github.com/wildfly/wildfly/commit/10f6b01fb2316d9bcbaaf52c873cfd3d87d3a556)).

So for example, the [jpa layer](https://github.com/wildfly/wildfly/commit/10f6b01fb2316d9bcbaaf52c873cfd3d87d3a556#diff-4e417096d8c25028fa9580f2433cf24b844d6845faa62997dd5832d5bc903ce6) is now rendered as https://jmesnil.github.io/wildfly.github.io/37/feature-pack/doc/#layer-jpa.

It on par with https://docs.wildfly.org/37/Galleon_Guide.html#gal.jpa.

Issue #747, opened 2025-07-30
WildFly Maven Plugin, Bootable JAR packaging for the cloud


jfdenise:
### Description

Description
We plan to introduce a new feature-pack allowing for bootable JAR packaging on the cloud

Issue Contact
jfdenise

### Issue Contact

jfdenise

PR #748, opened 2025-07-30
[WFLY-20810] WildFly Maven Plugin, bootable JAR packaging for the cloud

jfdenise:
Issue: https://github.com/wildfly/wildfly-proposals/issues/747

### DIFF
--- a/maven/WFLY-20810_maven_plugin_bootable_jar_cloud.adoc
+++ b/maven/WFLY-20810_maven_plugin_bootable_jar_cloud.adoc
@@ -1,0 +1,269 @@
+---
+categories:
+  - maven
+  - bootable-jar
+  - cloud
+  - wf-galleon
+stability-level: community
+issue: https://github.com/wildfly/wildfly-proposals/issues/747
+feature-team:
+ developer: jfdenise
+ sme:
+  - jmesnil
+ outside-perspective:
+  - ehsavoie
+---
+= WildFly Maven Plugin, bootable JAR packaging on the cloud
+:author:            Jean-Francois Denise
+:email:             jdenise@redhat.com
+:toc:               left
+:icons:             font
+:idprefix:
+:idseparator:       -
+
+== Overview
+
+This proposal defines the changes on the https://github.com/wildfly/wildfly-cloud-galleon-pack[WildFly cloud feature pack] and the 
+https://github.com/wildfly/wildfly-maven-plugin[WildFly Maven Plugin] in order to package and run a WildFly Bootable JAR for the cloud.
+
+== Issue Metadata
+
+=== Issue
+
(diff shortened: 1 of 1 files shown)


bstansberry (commented):
A general comment is I think the Overview can be much simpler and the Hard Requirements should be more expansive. Don't use the Overview to detail the feature. Most of what's in 'WildFly and the cloud galleon feature-pack' and below in the Overview can be reduced to a few sentences that say that a feature pack to adapt a bootable jar to the cloud environment is needed, but the existing `org.wildfly.cloud:wildfly-cloud-galleon-pack` is tightly coupled to the s2i use case and isn't appropriate. So a new feature pack will be added.

Formally list hard requirements so those reviewing the feature can clearly see what will and won't be done. Don't make them parse the Overview etc to get that.

maven/WFLY-20810_maven_plugin_bootable_jar_cloud.adoc
```diff
@@ -0,0 +1,192 @@
...
+
+No impact.
+
+== Security Considerations
+
+No impact.
+
+== Test Plan
+
+* Some tests are added to the new project that defines the feature-pack.
```
bstansberry, line 184: That cover what?

fabiobrz:
Thanks @jfdenise - I am off this week, but will review as soon as I will be back.

maven/WFLY-20810_maven_plugin_bootable_jar_cloud.adoc
```diff
@@ -0,0 +1,269 @@
...
+* Galleon layers adjustment to take into account cloud execution context. No changes, existing adaptation is re-used.
+* Custom logic run at bootable jar boot time to handle:
+** `jboss.node.name` and `jboss.tx.node.id` correct value handling.
+** HA, clustering password handling using the env var `JGROUPS_CLUSTER_PASSWORD`. Set the auth password in the correct subsystem configuration.
+* Advertise in the well known file `/tmp/wildfly-bootable-jar/install-dir` where the server is installed (required by the WildFly Operator). This implies a change in the 
+WildFly Operator to change the well known directory to be `/tmp/wildfly-bootable-jar/install-dir`.
+
+==== Excluding logic executed at boot time
+
+* When provisioning a server installation, the package `org.wildfly.cloud.bootable.runtime` can be excluded.
```
mchoma: What does `can be excluded` mean ? Why user may or may not explicitly remove it?

When server is provisioned, I would expect `org.wildfly.cloud.bootable.runtime` would be excluded automatically, because it can do some problems.

bstansberry:
@jfdenise https://redhat.atlassian.net/browse/WFLY-20810 was resolved for WF 41 so I am merging this.

PR #749, opened 2025-08-05
Prospero update operation to a specific version

### LINKED ISSUE #705: Add ability to update WildFly to a specific version of a manifest using wildfly-channels
### Description

Currently, Prospero update operation always picks the latest available manifest(s) version(s) to update to. Some users might prefer to perform an update to a concrete version that they can pre-test.

The goal of this proposal is to give a user an option to perform an update to any available manifest(s) version(s).

### Issue Contact

spyrkob

pedro-hos:
Resolves: #705

### DIFF
--- a/provisioning/Prospero_update_to_version.adoc
+++ b/provisioning/Prospero_update_to_version.adoc
@@ -1,0 +1,182 @@
+---
+categories:
+ - provisioning
+stability-level: default
+# Specify the Feature Development tracker issue for the feature.
+# This must be an issue tracked in https://github.com/orgs/wildfly/projects/7/views/1.
+# To create a Feature Development tracker issue, go to  https://github.com/wildfly/wildfly-proposals/issues/new/choose
+# and select 'Feature Development'
+issue: https://github.com/wildfly/wildfly-proposals/issues/705
+# Provide the github ids of the members of the feature team, organized by role.
+# Provide a single id for the 'assignee' role. Use a yaml list for the 'sme' and
+# 'outside-perspective' roles, even if there is only one person in a role.
+feature-team:
+ developer: Pedro Henrique de Oliveira Silva
+ sme:
+  - Jeff Mesnil
+ outside-perspective:
+  - Tomas Hofman
+# This should be blank during initial development of a feature. It may be used
+# after the feature is completed if a subsequent issue is field to track promotion
+# of this feature to a higher stability level
+promoted-by:
+---
+= Add ability to update WildFly to a specific version of a manifest using wildfly-channels.
+:author:            Bartosz Spyrko-Smietanko
+:email:             bspyrkos@redhat.com
+:toc:               left
+:icons:             font
+:idprefix:
+:idseparator:       -
+
+== Overview
+
+Currently, Prospero update operation always picks the latest available manifest(s) version(s) to update to. Some users might prefer to perform an update to a concrete version that they can pre-test.
+
+This proposal adds a new parameter to the `update perform` and `update prepare` operations that will accept a version of a manifest version.
+
+=== User Stories
+
+ - As an admin, I want to use a physical manifest version to update the server to a specific version. For example, use `--version=36.0.0.Beta1` to move to this specific version.
+ - As an admin, I want to be able to perform a downgrade if a specified version is lower than a current version. Such operation should result in a warning that the user has to accept.
+ - As an admin, I want to be able to perform a downgrade in a non-interactive way by specifying a command-line flag. For example, use `--version=...` on the `update perform` and `update prepare` in order to downgrade.
+ - As an admin, I want to be prevented from downgrading a manifest version unless the new version is specified explicitly.
+
+
+==== Nice-to-have:
+ - As an admin customer I want to be able to see a list of potential manifest versions I can update (or downgrade to).
+
+== Issue Metadata
+
+=== Related Issues
+
+ - https://github.com/wildfly/prospero/issues/972
+ - https://github.com/wildfly/prospero/issues/577
+ - https://issues.redhat.com/browse/EAP7-2291
+
+=== Affected Projects or Components
+
+ - https://github.com/wildfly-extras/prospero[Prospero]
+ - https://github.com/wildfly-extras/wildfly-installation-manager-api[Installation Manager API]
+ - https://github.com/wildfly/wildfly-core[WildFly Core]
+ - https://github.com/hal/console[HAL Console]
+
+=== Other Interested Projects
+
+=== Relevant Installation Types
+
+* Traditional standalone server (unzipped or provisioned by Galleon)
+* Managed domain
+
+== Requirements
+
+==== Hard requirements:
+- Add a `--version` parameter to the update operation with values in format `<CHANNEL_NAME>::<VERSION>`
+- If the `--version` parameter is present, the `update perform` and `update prepare` operations use a manifest with this version to provision the server.
+- If the `--version` provided is lower than the current version, the user should be **warned** and asked to confirm the operation.
+- If the `--version` is used in conjunction with `--yes` flag, the downgrade confirmation is skipped.
+- If a server is subscribed to multiple channels, the customer has to provide versions of all the involved manifests.
+- If the `--version` is not specified and the `update` operation would result in a manifest downgrade, the operation should fail as an error.
+- Above changes need to be available in JBoss CLI and HAL console.
+
+==== Nice-to-have requirements:
+- A new `update list-versions` operation should present a list of possible manifest updates and downgrades.
+- If the channel uses a URL manifest, the user should be able to provide the URL of the new manifest. Note, providing a version for a URL-based channel should be considered an error.
+- If the URL manifest provided is a GAV-based manifest, it's expected an error
+- Improve the `prospero history --dir wildfly-ee/` history command output to provide more detailed info - both maven version and logical versions should be displayed. Currently only the *logical version* is displayed:
+
+```
+$ ./prospero history --dir wildfly-ee/
+[0b685213] 2025-10-27T14:44:57Z - update [WildFly EE 38.0.0.Final]
+[f967d959] 2025-10-27T14:29:05Z - install [WildFly EE 37.0.0.Final]
+```
+
+==== Non-requirements:
+- When the server is subscribed to multiple channels, update operation does not verify the compatibility of selected versions.
+- "Fixing" a server to a manifest version - i.e. the update operation *does not* change the channel configuration. If before the update server was subscribed to a `org.wildfly.channels:wildfly`, the version will not be added after the update.
+
+=== Future Work
+
+- Supporting rollback to a physical manifest.
+- Possibly listing available versions in `update list` operation.
+
+== Backwards Compatibility
+
+This change would prevent a downgrade without an explicit `--version` argument. Some users may have used this option, but this is not a recommended use case.
+
+=== Default Configuration
+
+n/a
+
+=== Importing Existing Configuration
+
+n/a
+
+=== Deployments
+
+n/a
+
+=== Interoperability
+
+n/a
+
+== Implementation Plan
+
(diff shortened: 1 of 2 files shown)


provisioning/Prospero_update_to_version.adoc
```diff
@@ -0,0 +1,178 @@
...
+---
+= Add ability to update WildFly to a specific version of a manifest using wildfly-channels.
+:author:            Bartosz Spyrko-Smietanko
+:email:             bspyrkos@redhat.com
+:toc:               left
+:icons:             font
+:idprefix:
+:idseparator:       -
+
+__<The entire document should be one to two pages long. We will write each analysis document as if it is a conversation with a future developer. This requires a good writing style, with full sentences organized into paragraphs. Bullets are acceptable only for visual style, not as an excuse for writing sentence fragments.>__
```
jmesnil, line 33: you can remove this section.

provisioning/Prospero_update_to_version.adoc
```diff
@@ -0,0 +1,178 @@
...
+
+
+==== Nice-to-have:
+ - As an admin customer I want to be able to see a list of potential manifest versions I can update (or downgrade to).
+
+== Issue Metadata
+
+=== Related Issues
+
+ - https://github.com/wildfly-extras/prospero/issues/754
```
jmesnil, line 56: Could you also add a link to https://github.com/wildfly/wildfly-proposals/pull/577 please?

provisioning/Prospero_update_to_version.adoc
```diff
@@ -0,0 +1,178 @@
...
+
+__<List the installation types that are relevant for the features and remove any that are not relevant>__.
+
+* Traditional standalone server (unzipped or provisioned by Galleon)
+* Managed domain
+
+== Requirements
+
+==== Hard requirements:
+- Add a `--version` parameter to the update operation with values in format `<CHANNEL_NAME>::<VERSION>`
```
jmesnil, line 78: I'm not sure about the format.  In the simple case (only 1 manifest was used to provision the installation), only the `version` from the manifest's Maven GAV coordinates needs to be specified.
For more complex cases (multiple manifests were used), the full Maven coordinates would needs to be specified for the manifest(s) to update to.

`--version` could be either `<version>` or a comma-separated lists of `<groupId>:<artifactId>:<version>`.

wdyt?

provisioning/Prospero_update_to_version.adoc
```diff
@@ -0,0 +1,178 @@
...
+- If the channel uses a URL manifest, the user should be able to provide the URL of the new manifest. Note, providing a version for a URL-based channel should be considered an error.
+
+==== Non-requirements:
+- When the server is subscribed to multiple channels, update operation does not verify the compatibility of selected versions.
+- "Fixing" a server to a manifest version - i.e. the update operation *does not* change the channel configuration. If before the update server was subscribed to a `org.wildfly.channels:wildfly`, the version will not be added after the update.
+
+=== Future Work
+
+- Supporting rollback to a logical/physical manifest.
+- __Possibly listing available versions in `update list` operation.__
```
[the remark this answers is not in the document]
pedro-hos: That's the same from `==== Nice-to-have requirements:`. I guess that we can use only one. wdyt?

provisioning/Prospero_update_to_version.adoc
```diff
@@ -0,0 +1,178 @@
...
+- If the `--version` is not specified and the `update` operation would result in a manifest downgrade, the operation should fail as an error.
+- Above changes need to be available in JBoss CLI and HAL console.
+
+==== Nice-to-have requirements:
+- A new `update list-versions` operation should present a list of possible manifest updates and downgrades.
+- If the channel uses a URL manifest, the user should be able to provide the URL of the new manifest. Note, providing a version for a URL-based channel should be considered an error.
+
+==== Non-requirements:
+- When the server is subscribed to multiple channels, update operation does not verify the compatibility of selected versions.
+- "Fixing" a server to a manifest version - i.e. the update operation *does not* change the channel configuration. If before the update server was subscribed to a `org.wildfly.channels:wildfly`, the version will not be added after the update.
```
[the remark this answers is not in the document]
pedro-hos: The first one is related to multiple channels. I don't know if you agree, but the ideia is not verify the compatibility of the versions present on `--version` parameter when using multiple channels.

TBH I didn't understand the second as well. I don't know what Bartosz was thinking on this section. Do you think that we can remove it @jmesnil?

provisioning/Prospero_update_to_version.adoc
TomasHofman: Following two points are relevant for this discussion (reference via channel name or via manifest coordinate):

> - If a server is subscribed to multiple channels, the customer has to provide versions of all the involved manifests.

This point is actually not a problem, as I'm checking that the wildfly-channel lib doesn't allow to initialize a session with multiple channels that have identical manifest. That means that manifest coordinate also uniquely represents a channel. So we could use both - channel name or manifest coordinate.

> - If the channel uses a URL manifest, the user should be able to provide the URL of the new manifest. Note, providing a version for a URL-based channel should be considered an error.

Here the manifest coordinate I think becomes bit problematic. We need to identify a channel we want to change, and at the same time provide a new manifest coordinate.

With channel name it could be as Bartek suggested `--version channel-0::https:/...`. 

With manifest coordinates, it would have to be something like `--version oldurl::newurl`? Single colon doesn't look like sufficient delimiter.

provisioning/Prospero_update_to_version.adoc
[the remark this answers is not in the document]
TomasHofman: It looks reasonable behavior to me, I wouldn't remove it. :)

provisioning/Prospero_update_to_version.adoc
```diff
@@ -0,0 +1,175 @@
...
+
+== Overview
+
+Currently, Prospero update operation always picks the latest available manifest(s) version(s) to update to. Some users might prefer to perform an update to a concrete version that they can pre-test.
+
+This proposal adds a new parameter to the `update perform` and `update prepare` operations that will accept a version of a manifest version.
+
+=== User Stories
+
+ - As an admin I want to use a logical or physical manifest version to update the server to a specific version.
```
TomasHofman, line 41: Note that allowing the logical version would probably mean downloading all versions of manifest files form maven to be able to create mapping between logical version and real maven version. I kinda see that as problematic.

We would also need to be able to distinguish if given string is a logical or a real maven version, which I'm not sure if can be done reliably.

(The logical version is metadata field inside the manifest YAML, see https://download.eng.brq.redhat.com/brewroot/repos/jb-eap-8.1-maven-build/latest/maven/org/jboss/eap/channels/eap-8.1/1.0.0.GA-redhat-00012/eap-8.1-1.0.0.GA-redhat-00012-manifest.yaml)

TomasHofman:
@jmesnil as I'm looking at https://github.com/wildfly-extras/prospero/issues/754 / https://github.com/wildfly/wildfly-proposals/issues/705 linked as related to this proposal, this proposal doesn't in fact enable that. Am I right and is that OK?

In this proposal the --version parameter only affects single invocation of given update command. The https://github.com/wildfly/wildfly-proposals/issues/705 requires permanent change of the manifest version in `.installation/installer-channels.yaml`.

provisioning/Prospero_update_to_version.adoc
pedro-hos: I agree, w/ the name seems the correct way to make it work. Like you said: `--version channel-0::https:/...`. With manifest coordinates I don't see other way as  `--version oldurl::newurl` 

Should I add those examples on each bullet point? I think that it could be useful.

jmesnil:
@TomasHofman That seems right, #705 proposal actually correspond to the non-requirement ("Fixing" a server to a manifest version) from this #749 proposal.

provisioning/Prospero_update_to_version.adoc
```diff
@@ -0,0 +1,175 @@
...
+- If the `--version` parameter is present, the `update perform` and `update prepare` operations use a manifest with this version to provision the server.
+- If the `--version` provided is lower than the current version, the user should be warned and asked to confirm the operation.
+- If the `--version` is used in conjunction with `--yes` flag, the downgrade confirmation is skipped.
+- If a server is subscribed to multiple channels, the customer has to provide versions of all the involved manifests.
+- If the `--version` is not specified and the `update` operation would result in a manifest downgrade, the operation should fail as an error.
+- Above changes need to be available in JBoss CLI and HAL console.
+
+==== Nice-to-have requirements:
+- A new `update list-versions` operation should present a list of possible manifest updates and downgrades.
+- If the channel uses a URL manifest, the user should be able to provide the URL of the new manifest. Note, providing a version for a URL-based channel should be considered an error.
```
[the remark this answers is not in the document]
pedro-hos: I'll add the following line, what do you think?

`- If the URL manifest provided is a GAV-based manifest, it's expected an error`

provisioning/Prospero_update_to_version.adoc
pedro-hos: Do you think that we could only remove the `logical version` for example: `logical-version: "EAP 8.1 GA"` and keep only the `physical manifest version`, like: `org.jboss.eap.channels:eap-8.1` from the text?

TomasHofman:
I'm considering whether we are going to provide enough context for the users to figure this out.

The scenario:
* User updates the installation and finds an issue -> needs to rollback to the previous manifest. But, what was the previous manifest?
* User runs history command:
```
$ ./prospero history --dir wildfly-ee/ 
[0b685213] 2025-10-27T14:44:57Z - update [WildFly EE 38.0.0.Final]
[f967d959] 2025-10-27T14:29:05Z - install [WildFly EE 37.0.0.Final]
```
-> only the logical version is displayed, which is not usable unless we enable logical version as input parameter.
* User can run list-channels to see how logical version map to maven versions:
```
$ ./prospero update list-channels --dir wildfly-ee/ --all
Checking available updates for /home/thofman/Projects/prospero/wildfly-ee

Found new versions of channel manifests available:
 - channel name: channel-0
   current version: 38.0.0.Final
   available versions:
   ...
   - 36.0.0.Beta1 (WildFly EE 36.0.0.Beta1)
   - 36.0.0.Final (WildFly EE 36.0.0.Final)
   - 36.0.1.Final (WildFly EE 36.0.1.Final)
   - 37.0.0.Beta1 (WildFly EE 37.0.0.Beta1)
   - 37.0.0.Final (WildFly EE 37.0.0.Final)
   - 37.0.1.Final (WildFly EE 37.0.1.Final)
   - 38.0.0.Beta1 (WildFly EE 38.0.0.Beta1)
   - 38.0.0.Final (WildFly EE 38.0.0.Final)
```
So the maven version is 37.0.0.Final, but it's not very good experience, and quite hard to explain.

I think we have two options:
* Improve the history command output to provide more detailed info - both maven version and logical versions should be displayed. I would like this change to be there.
* Keep the logical version as an option for --version parameter input. We can now add it as optional requirement and see what it takes to implement it.

WDYT?

TomasHofman (approved):

TomasHofman (approved):
Looks good, thanks @pedro-hos !

provisioning/Prospero_update_to_version.adoc
```diff
@@ -0,0 +1,182 @@
...
+issue: https://github.com/wildfly/wildfly-proposals/issues/705
+# Provide the github ids of the members of the feature team, organized by role.
+# Provide a single id for the 'assignee' role. Use a yaml list for the 'sme' and
+# 'outside-perspective' roles, even if there is only one person in a role.
+feature-team:
+ developer: Pedro Henrique de Oliveira Silva
+ sme:
+  - Jeff Mesnil
+ outside-perspective:
+  - Tomas Hofman
```
bstansberry: github id please

jmesnil (approved):

Issue #760, opened 2025-10-09
[WFLY-14559] Add new attributes for global configuration for RESTEasy deployments in the jaxrs subsystem


jamezp:
### Description

There are at least 3 new attributes that should be added to the `jaxrs` subsystem. These attributes can currently be configured via a system property or as a servlet context parameter. Adding them as attributes to the `jaxrs` subsystem allows them to be configured at a global level with better validation for the values.

### Issue Contact

jamezp

PR #761, opened 2025-10-10
[WFLY-14559] Add new attributes to the jaxrs subsystem.

jamezp:
JIRA: https://issues.redhat.com/browse/WFLY-14559

Issue: #760

### DIFF
--- a/jaxrs/WFLY-14559-new-attributes.adoc
+++ b/jaxrs/WFLY-14559-new-attributes.adoc
@@ -1,0 +1,130 @@
+---
+categories:
+ - jaxrs
+ - management
+stability-level: default
+issue: https://github.com/wildfly/wildfly-proposals/issues/760
+feature-team:
+ developer: jamezp
+ sme:
+  - tommaso-borgato
+ outside-perspective:
+  - Raul Chettri
+---
+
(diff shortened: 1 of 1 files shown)


jaxrs/WFLY-14559-new-attributes.adoc
```diff
@@ -0,0 +1,141 @@
...
+
+=== Interoperability
+
+N/A
+
+== Implementation Plan
+
+The implementation will be to simply add the new attributes to the subsystem. In a `DeploymentUnitProcessor`, which
+already exists, the attributes will be converted into context properties set in the servlet context parameters. The
+values from the subsystem will not overwrite values defined in the `web.xml` of the deployment being processed.
```
tommaso-borgato: @jamezp if one of the new attributes is not set in the model, would or wouldn't it be added to the deployment context parameters ... I mean, would it or wouldn't it influence whatever default is already set in resteasy?

I am asking this because so far, we had:

- resteasy default values

NExt we'll have:

- resteasy default values 
- wf default values 

in case the 2 default values diverge any time in some future, we might have "wf default != resteasy default" resulting in a dis-alignment ... I guess I am just asking how we deal with such things with existing attributes

bstansberry (approved):

PR #769, opened 2025-12-03
[WFCORE-7247][WFLY-21238] Filtering mechanism for management audit logging and configuration change history

### LINKED ISSUE #768: Filtering mechanism for management audit logging and configuration change history
### Description

Add a mechanism that allows the enablement of redaction of susceptible attributes when auditing management operations or saving the history of configuration changes 

### Issue Contact

yersan

yersan:
Jira issues:
* https://issues.redhat.com/browse/WFCORE-7247
* https://issues.redhat.com/browse/WFLY-21238

Fixes https://github.com/wildfly/wildfly-proposals/issues/768

### DIFF
--- a/management/WFCORE-7247_Filtering_mechanism_for_management_audit_logging_and_configuration_change_history.adoc
+++ b/management/WFCORE-7247_Filtering_mechanism_for_management_audit_logging_and_configuration_change_history.adoc
@@ -1,0 +1,152 @@
+---
+categories:
+- management
+- security
+stability-level: community
+# Specify the Feature Development tracker issue for the feature.
+# This must be an issue tracked in https://github.com/orgs/wildfly/projects/7/views/1.
+# To create a Feature Development tracker issue, go to https://github.com/wildfly/wildfly-proposals/issues/new/choose
+# and select 'Feature Development'
+issue: https://github.com/wildfly/wildfly-proposals/issues/768
+# Provide the GitHub ids of the feature team members, organized by role.
+# Provide a single id for the 'assignee' role. Use a YAML list for the 'sme' and
+# 'outside-perspective' roles, even if there is only one person in a role.
+feature-team:
+ developer:
+  - yersan
+ sme:
+  - darranl
+ outside-perspective:
+  - jamezp
+# If this issue tracks the promotion of a previously completed feature to a higher stability level, provide the URL of
+# the original https://github.com/wildfly/wildfly-proposals/issues issue that was used to track the implementation of
+# that feature.
+promotes:
+# During initial development of a feature, this field should be left blank. If after the feature is completed and its
+# stability level is promoted via a later issue, the developer promoting the issue should return to this document ("this"
+# being the original analysis document for the feature). The field should be updated to point to the
+# https://github.com/wildfly/wildfly-proposals/issues issue that promotes it.
+# For example,
+#              | Implementation Issue (A) | Promotion Issue (B)
+# promotes:    |                          | https://github.com/.../A
+# promoted-by: | https://github.com/.../B |
+promoted-by:
+---
+
+= Filtering mechanism for management audit logging and configuration change history
+:author:            Yeray Borges
+:email:             yborgess@redhat.com
+:toc:               left
+:icons:             font
+:idprefix:
+:idseparator:       -
+
+== Overview
+
+Management operations often involve sensitive data, such as clear-text passwords. If management audit logging is enabled, this data is written to the audit log. Additionally, if the configuration change history limit is greater than zero, these operations are stored in memory by the server and can be listed later by an authenticated user.
+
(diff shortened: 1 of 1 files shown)


bstansberry:
A general point on this is adding this facility on its own does not protect anything. The entire management API would need to be audited with the flag set where appropriate.

I don't think we could properly report this feature to our users until that is done. We could report interim steps or something, but the end user-feature is having stuff not reported in the audit log. Just having the mechanism in place with  little or nothing actually redacted is not useful.

Implementing it without a full audit also introduces an incompatible change risk -- i.e. a user decides to allow redaction in version X based on what we'd marked as redactable, and then in version Y we add more things that they'd not want to be redacted. (I fully admit this is a corner case. It's likely not relevant unless different subsystems are making inconsistent decisions about what should be redactable.)

The audit would need to check for any cases where a distinct AttributeDefinition object is used for operations (e.g. add) vs what’s used in the resource. (This is unusual; I’m just writing it here so it is recorded.)

To be formally correct, any subsystem API that added the flag should have an API version bump. (Particularly since the read-operation-description output should report the value; I'll look and see if the analysis requires that and if not I'll comment to add it.)

Note that the initial use case that led to this RFE was operation parameter values being logged that didn't end up being persisted in the configuration, i.e. from ops not related to the persistent config. That is a much smaller thing. It's also a conceptually distinct category from stuff that ends up being in the config file where it can be read by anyone with file read access. I can imagine users wanting separate treatment of those use cases (e.g. use an enum for the attribute that controls this, not a boolean.) Generally I would expect access to the management audit log to be restricted to the same or fewer user organizational roles as ones that can read the config file. If we hide things in the log that are visible in the config I can imagine auditors objecting to that. Particularly if we end up hiding 'safe' values, e.g. credential expressions. (Checking whether a redactable expression is 'safe' is a whole other aspect.)

management/WFCORE-7247_Filtering_mechanism_for_management_audit_logging_and_configuration_change_history.adoc
```diff
@@ -0,0 +1,152 @@
...
+* Provide a new attribute named `redacted` at `/core-service=management/access=audit/logger=audit-log` so users can enable or disable the redaction of sensitive data in the management audit log.
+** It defaults to `true`.
+** When enabled, any attribute flagged as `redactable` will be redacted when the management operation that includes the attribute is logged in the management audit log.
+* Provide a new attribute named `redacted` at `/subsystem=core-management/service=configuration-changes` so users can enable or disable the redaction of sensitive data in the configuration change history.
+** It defaults to `true`.
+** When enabled, any attribute flagged as `redactable` will be redacted when the management operation that includes the attribute is recorded in the configuration change history.
+* WildFly and WildFly Core subsystem attributes will be reviewed to find the existent attributes that will be flagged as `redactable`, for example, any attribute that contains a password or clear-text.
+* The method for the redaction would be a hash function using SHA-256 cryptography algorithm.
```
bstansberry, line 91: * If set to `true` the `redactable` value for any operation parameter will be reported in the `read-operation-description` output.
* If set to `true` the `redactable` value for any resource attribute will be reported in the `read-operation-description` output.

management/WFCORE-7247_Filtering_mechanism_for_management_audit_logging_and_configuration_change_history.adoc
```diff
@@ -0,0 +1,152 @@
...
+=== Relevant Installation Types
+
+* Traditional standalone server (unzipped or provisioned by Galleon)
+* Managed domain
+* OpenShift Source-to-Image (S2I)
+* Bootable jar
+
+== Requirements
+
+* Provide a new Attribute Flag named `redactable` to allow WildFly authors to classify an attribute as **sensitive** data that is eligible for redaction when included as a parameter in a management operation.
```
bstansberry, line 83: A vague general point on all this is I think we need to be more precise about 'attribute' and 'parameter' throughout this document. This kind of deep kernel thing should involve a really hard core highly precise spec.

What is logged are operation parameter values.

The fact that for those we use object's whose class name includes the word 'Attribute' doesn't make them resource attributes.  Even if a particular instance of that class is also used for a resource attributes.

So don't use the word attribute unless you are describing very specific cases involving resource attributes. Which this requirement set should explicitly do. :) The very specific cases being operations that take a generic parameter (e.g. 'value') and write that parameter's value to a resource attribute. For example, write-attribute, undefine-attribute, any others I'm not thinking of. There the operation parameter ('value') is not redactable but the definition of the target attribute tells us the value should be redacted.

bstansberry: The sentence I commented on mentions 'Attribute Flag'.  That should be `Attribute.Flag`.  

My comment above about not using 'attribute'  is about the general common noun 'attribute'. Where using a proper noun like `Attribute.Flag` or `AttributeDefinition` is necessary to precisely communicate, like `Attribute.Flag` is in this sentence I'm commenting on, that's fine.

Issue #784, opened 2026-01-28
[HAL-1842] Support MicroProfile LRA in the console


hpehl:
### Description

Support the configuration of the MicroProfile LRA spec in the console. Console part of [WFLY-14869](https://issues.redhat.com/browse/WFLY-14869). 

### Issue Contact

hpehl

PR #785, opened 2026-01-28
[HAL-1842] Support MicroProfile LRA in the console

hpehl:
JIRA: https://issues.redhat.com/browse/HAL-1842
Issue: #784


rhusar:
> I'm fine with adding the missing sections from the template. I'm just not sure how to apply the template to new analysis documents:
> 
> * [ ]  Leave out sections that don't apply to the RFE
> * [ ]  Include all sections and mark those that do not apply with something like 'n/a'
> 
> Related to that, there used to be sections like "Dev Contacts", "QE Contacts", and "Testing By" that are no longer in the template. Are they still required?

Yes still required, here https://github.com/wildfly/wildfly-proposals/blob/main/design-doc-template.adoc?plain=1#L15-L23

PR #797, opened 2026-02-05
[WFLY-15495] Add gRPC to WildFly via a new subsystem.

### LINKED ISSUE #796: Add gPRC support and subsystem to WildFly
### Description

Add a `grpc` subsystem to WildFly. This will not be as simple as adding the [wildfly-grpc-feature-pack](https://github.com/wildfly-extras/wildfly-grpc-feature-pack). We will create a new subsystem that does not use Netty for the server.

### Issue Contact

jamezp

jamezp:
https://issues.redhat.com/browse/WFLY-15495

resolves #796

### DIFF
--- a/grpc/WFLY-15495-add-grpc-support.adoc
+++ b/grpc/WFLY-15495-add-grpc-support.adoc
@@ -1,0 +1,274 @@
+---
+categories:
+  - grpc
+  - management
+stability-level: default
+issue: https://github.com/wildfly/wildfly-proposals/issues/796
+# Provide the GitHub ids of the feature team members, organized by role.
+# Provide a single id for the 'assignee' role. Use a YAML list for the 'sme' and
+# 'outside-perspective' roles, even if there is only one person in a role.
+feature-team:
+ developer: jamezp
+ sme:
+  -
+ outside-perspective:
+  -
+---
+
+= Add gRPC Support to WildFly
+:author:            James R. Perkins
+:email:             jperkins@ibm.com
+:toc:               left
+:icons:             font
+:idprefix:
+:idseparator:       -
+
+== Overview
+
+Currently, gRPC support in WildFly is provided via an external feature pack that runs a standalone Netty server on a
+separate port. This proposal aims to integrate gRPC as a first-class citizen within the WildFly, allowing gRPC services
+to be deployed as CDI beans and served over the existing Undertow HTTP/2 listeners.
+
+=== User Stories
+
+1. As a Developer, I want to define a gRPC service using standard `.proto` files and implement it as a CDI bean so I
+can inject JPA/REST resources.
+2. As a System Administrator, I want gRPC services to be exposed on the same port as my REST endpoints (8080) to
+simplify network and firewall configurations.
+3. As a Security Officer, I want to use WildFly Elytron to authenticate gRPC calls using the same security domains used
+for the rest of the application.
+
+== Issue Metadata
+
+=== Related Issues
+
+* https://issues.redhat.com/browse/WFLY-15495[WFLY-15495]
+
+=== Affected Projects or Components
+
+* https://github.com/wildfly/wildfly[wildfly]
+* https://github.com/wildfly-extras/wildfly-grpc-feature-pack[wildfly-grpc-feature-pack] (superseded by this
+proposal)
+
+=== Other Interested Projects
+
+=== Relevant Installation Types
+
+* Traditional standalone server (unzipped or provisioned by Galleon)
+* Managed domain
+* OpenShift Source-to-Image (S2I)
+* Bootable jar
+
+== Requirements
+
+* Request Handling: The gRPC subsystem will add an Undertow `HttpHandler` to handle the `application/grpc` requests.
+* Automatic Service Discovery: The subsystem must use Jandex during the DEPLOY phase to identify all classes that
+extend `io.grpc.BindableService` without requiring manual registration.
+* CDI Integration: All discovered gRPC services must be managed as CDI beans, allowing for standard injection behavior.
+    ** Note that these beans will be scoped to `@ApplicationScoped`.
+* Port Unification: The system must multiplex gRPC traffic over the existing Undertow HTTP/2 listeners (defaulting to
+port 8080), avoiding the need for a secondary port.
+* Security Integration: The subsystem must authenticate gRPC requests using Elytron and support `@RolesAllowed`,
+`@PermitAll`, and `@DenyAll` annotations on service methods via a gRPC `ServerInterceptor`.
+* Context Propagation: The subsystem must ensure that the authenticated `SecurityIdentity` and Transaction state are
+available to the gRPC service thread.
+* Non-blocking Execution: The `HttpHandler` must avoid blocking the Undertow IO thread. Requests should be dispatched
+to the WildFly worker thread pool.
+* Minimal Overhead: The bridge between Undertow's `HttpServerExchange` and the gRPC-Java engine must minimize data
+copying.
+* Standard Compliance: The implementation must support the full gRPC wire protocol, including Unary, Server-streaming,
+Client-streaming, and Bidirectional streaming.
+* Lifecycle Awareness: gRPC services must be gracefully shut down when a deployment is stopped or the server is
+reloaded.
+* Deployment: Only one gRPC deployment will be allowed. The gRPC path is very specific in the way it is defined:
+`/serverName/methodName`. See the https://grpc.github.io/grpc/core/md_doc__p_r_o_t_o_c_o_l-_h_t_t_p2.html[documentation]
+for more details.
+    ** This means if more than one deployment is allowed, there is no way to use the context path to determine which
+       service to call.
+
+=== Changed requirements
+
+N/A
+
+
+=== Non-Requirements
+
+* gRPC Client Injection: No support for injecting gRPC Stubs.
+* Protobuf Compilation: No protoc execution within the server.
+* Non-CDI Services: Only CDI-managed beans will be eligible for registration.
+* Custom Transport: No support for gRPC-over-Unix-Domain-Sockets or other non-Undertow transports.
+* Transaction Support for Streaming RPCs: Only unary (request-response) gRPC methods will support `@Transactional`.
+Streaming methods (server-streaming, client-streaming, bidirectional) are inherently long-lived and unsuitable for
+transactional semantics.
+
+=== Future Work
+
+N/A
+
+== Backwards Compatibility
+
+This is a new feature and subsystem. However, it is **not compatible** with the existing wildfly-extras gRPC feature
+pack (https://github.com/wildfly-extras/wildfly-grpc-feature-pack). The new subsystem uses a fundamentally different
+architecture:
+
+* **Old feature pack:** Runs a standalone Netty server on a separate port
+* **New subsystem:** Integrates with Undertow's HTTP/2 listeners on the default port (8080)
+
+Users currently deploying the wildfly-extras feature pack will need to:
+1. Remove the old feature pack from their WildFly installation
+2. Reconfigure their gRPC services to work as CDI beans (no configuration changes to `.proto` files or service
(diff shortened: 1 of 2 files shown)


grpc/WFLY-15495-add-grpc-support.adoc
```diff
@@ -0,0 +1,271 @@
...
+N/A
+
+== Backwards Compatibility
+
+This is a new feature and subsystem. However, it is **not compatible** with the existing wildfly-extras gRPC feature
+pack (https://github.com/wildfly-extras/wildfly-grpc-feature-pack). The new subsystem uses a fundamentally different
+architecture:
+
+* **Old feature pack:** Runs a standalone Netty server on a separate port
+* **New subsystem:** Integrates with Undertow's HTTP/2 listeners on the default port (8080)
```
rhusar: I think there was a discussion somewhere in Zulip, the existing grpc FP is still in a low stability, so I don't think there is anything that precludes us from rewriting it into implementing this proposal in there. Certainly, this approach has certain benefits (no need to manage another module in WF) and disadvantages (alignment, etc).

Perhaps we can document that decision here too.

grpc/WFLY-15495-add-grpc-support.adoc
```diff
@@ -0,0 +1,271 @@
...
+issue: https://github.com/wildfly/wildfly-proposals/issues/796
+# Provide the GitHub ids of the feature team members, organized by role.
+# Provide a single id for the 'assignee' role. Use a YAML list for the 'sme' and
+# 'outside-perspective' roles, even if there is only one person in a role.
+feature-team:
+ developer: jamezp
+ sme:
+  -
+ outside-perspective:
+  -
```
rhusar: Just for book keeping this section needs people here :-)

grpc/WFLY-15495-add-grpc-support.adoc
```diff
@@ -0,0 +1,271 @@
...
+:toc:               left
+:icons:             font
+:idprefix:
+:idseparator:       -
+
+== Overview
+
+Currently, gRPC support in WildFly is provided via an external feature pack that runs a standalone Netty server on a
+separate port. This proposal aims to integrate gRPC as a first-class citizen within the WildFly, allowing gRPC services
+to be deployed as CDI beans and served over the existing Undertow HTTP/2 listeners.
```
rhusar: In the current grpc feature pack we have `org.wildfly.extension.grpc.WildFlyGrpcDeploymentRegistry` that allows registration of services programatically. Is the plan to no longer support this givne there is only mention of the CDI beans support?

rhusar (changes requested):

grpc/WFLY-15495-add-grpc-support.adoc
jamezp: Yes, I'm awaiting volunteers :)

grpc/WFLY-15495-add-grpc-support.adoc
```diff
@@ -0,0 +1,274 @@
...
+== Issue Metadata
+
+=== Related Issues
+
+* https://issues.redhat.com/browse/WFLY-15495[WFLY-15495]
+
+=== Affected Projects or Components
+
+* https://github.com/wildfly/wildfly[wildfly]
+* https://github.com/wildfly-extras/wildfly-grpc-feature-pack[wildfly-grpc-feature-pack] (superseded by this
```
marekkopecky, line 50: What is the plan with this repo? Will it be archived?

grpc/WFLY-15495-add-grpc-support.adoc
```diff
@@ -0,0 +1,274 @@
...
+* Traditional standalone server (unzipped or provisioned by Galleon)
+* Managed domain
+* OpenShift Source-to-Image (S2I)
+* Bootable jar
+
+== Requirements
+
+* Request Handling: The gRPC subsystem will add an Undertow `HttpHandler` to handle the `application/grpc` requests.
+* Automatic Service Discovery: The subsystem must use Jandex during the DEPLOY phase to identify all classes that
+extend `io.grpc.BindableService` without requiring manual registration.
```
marekkopecky, line 66: extend -> implement?

grpc/WFLY-15495-add-grpc-support.adoc
```diff
@@ -0,0 +1,274 @@
...
+* CDI Integration: All discovered gRPC services must be managed as CDI beans, allowing for standard injection behavior.
```
marekkopecky, line 67: > must be

What happens if not? Server doesn't start?

grpc/WFLY-15495-add-grpc-support.adoc
```diff
@@ -0,0 +1,274 @@
...
+    ** Note that these beans will be scoped to `@ApplicationScoped`.
+* Port Unification: The system must multiplex gRPC traffic over the existing Undertow HTTP/2 listeners (defaulting to
+port 8080), avoiding the need for a secondary port.
+* Security Integration: The subsystem must authenticate gRPC requests using Elytron and support `@RolesAllowed`,
+`@PermitAll`, and `@DenyAll` annotations on service methods via a gRPC `ServerInterceptor`.
+* Context Propagation: The subsystem must ensure that the authenticated `SecurityIdentity` and Transaction state are
```
marekkopecky, line 73: Do you mean `org.wildfly.security.auth.server.SecurityIdentity`? How this identity and transaction state can be used in the thread? Can you please elaborate it more?

grpc/WFLY-15495-add-grpc-support.adoc
```diff
@@ -0,0 +1,274 @@
...
+available to the gRPC service thread.
+* Non-blocking Execution: The `HttpHandler` must avoid blocking the Undertow IO thread. Requests should be dispatched
+to the WildFly worker thread pool.
+* Minimal Overhead: The bridge between Undertow's `HttpServerExchange` and the gRPC-Java engine must minimize data
+copying.
+* Standard Compliance: The implementation must support the full gRPC wire protocol, including Unary, Server-streaming,
+Client-streaming, and Bidirectional streaming.
+* Lifecycle Awareness: gRPC services must be gracefully shut down when a deployment is stopped or the server is
+reloaded.
+* Deployment: Only one gRPC deployment will be allowed. The gRPC path is very specific in the way it is defined:
```
marekkopecky, line 83: What server does if it detects more deployments with gRPC?

marekkopecky:
Can we have a call that could be CDI bean and Rest end-point, that implements io.grpc.BindableService ?

Eg. take this class
https://github.com/resteasy/resteasy/blob/main/testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/cdi/basic/resource/CDILocatorResource.java

and adds something like ` extends CustomGrpc.CustomImplBase`?

marekkopecky:
Will be grpc subsystem added to some configuration files? standalone.xml? standalone-microprofile.xml? etc?

jmesnil:
There is very good reasons to integrate gRPC directly from the Undertow stack as highlighted by James.
However, this would require to use API provided by grpc-java that are not meant for user consumption according to https://github.com/grpc/grpc-java#api-stability.

Any such code in WildFly (that would be backed by XNIO) could break in any grpc-java release without notice... It seems to be a heavy maintenance risk...
However, if WildFly would ever support Netty (either directly or through the Undertow/Netty stack), we could use the official io.grpc.netty package that is a stable API from the gRPC organization.
I would suggest to postpone this integration after Undertow/Netty is integrated in WildFly. In the mean time, we can continue to provide the existing gRPC feature pack and add missing capabilities (eg CDI) and make sure that we keep the migration to a unified gRPC stack in mind. 
What do you think?

grpc/WFLY-15495-add-grpc-support.adoc
jmesnil: I don't think we would need to restrict to only 1 gRPC deployment.
There could be multiple WAR deployments that provide gRPC services (as long as the name of their services are distinct).
At deployment time, a WAR deployment that provides gRPC services will try to register its services. If the names are not taken, deployment is fine. If there are existing gRPC services with these names, the deployment will be rejected.
The key change compared to WARs is that we can not compare their names (or context) to verify that there is no overlap, it must wait until the gRPC services tries to be registered in the internal registry.

jmesnil: (btw, this would also apply to the existing gRPC feature pack)

rhusar:
@jmesnil I agree with the first half of your concern, but I think the premise of the conclusion doesn't hold, so I'd like to push back before we postpone this.

What is agreed: the Undertow bridge as proposed does use unsupported API. The doc talks about bridging HttpServerExchange to InternalServer and "strictly following the gRPC-Java internal transport logic", which means io.grpc.internal.*... and that package leaves no room for mis-interpretation from Javadoc: "Interfaces and implementations that are internal to gRPC. All the content under this package and its subpackages are considered annotated with {@link io.grpc.Internal}". And thus no compatibility guarantee at all. That's a fair.

But the problem with the argument is here: `io.grpc.netty` is not a stable API! So the proposed alternative doesn't solve the problem it's meant to solve. The annotations are at class level, covering the whole public API we'd need; either @Experimental or @Internal: 

- https://github.com/grpc/grpc-java/blob/master/netty/src/main/java/io/grpc/netty/NettyServerBuilder.java#L75
- https://github.com/grpc/grpc-java/blob/master/netty/src/main/java/io/grpc/netty/NettyServerProvider.java#L25
- https://github.com/grpc/grpc-java/blob/master/netty/src/main/java/io/grpc/netty/NettyChannelBuilder.java#L82
- https://github.com/grpc/grpc-java/blob/master/netty/src/main/java/io/grpc/netty/GrpcSslContexts.java#L47
- https://github.com/grpc/grpc-java/blob/master/netty/src/main/java/io/grpc/netty/GrpcHttp2ConnectionHandler.java#L35

And the README rule you're citing applies to us specifically:

▎ APIs annotated with @ExperimentalApi are subject to change in future releases, and library code that other projects may depend on should not use these APIs.

On the tracking issue itself (https://github.com/grpc/grpc-java/issues/1784#issuecomment-1681129066, "Tracking Issue for Netty being Experimental", open since 2016), mainainer states:

▎ There are no plans to ever stabilize this API.

So moving to io.grpc.netty isn't internal to stable, it's @Internal to permanently @ExperimentalApi. Which is somewhat of an improvement  but it doesn't coimpletely justify "let's postpone until Undertow/Netty is integrated" for a stability argument, because the endpoint of that wait is still API that gRPC says library code should not depend on.


There's a third option I think we should evaluate before deciding: gRPC's own Servlet transport. io.grpc:grpc-servlet-jakarta which is also experimental... but would give us "gRPC multiplexed over the existing Undertow HTTP/2 listeners on the default port" with no io.grpc.internal usage in WildFly and no dependency on Netty. It is also @ExperimentalApi (https://github.com/grpc/grpc-java/issues/5066), but sseems to be used in the wild (jetty?).

So every option, has @ExperimentalApi / @Internal API  somewhere, I don't think "wait for Undertow/Netty" gives us the guarantee we'd want.

kabir:
What does Quarkus use for their integration with Netty (or Vert.X)?

rhusar:
> What does Quarkus use for their integration with Netty (or Vert.X)?

@kabir According to https://quarkus.io/guides/grpc/ -- Quarkus gRPC is based on [Vert.x gRPC](https://vertx.io/docs/vertx-grpc/java/) implementation -- which is Vert.x's implementation of gRPC (Vert.x gRPC is a gRPC implementation based on Vert.x, it aims to implement an application level API as well as a request/response oriented API to write gRPC middlewares like reverse proxies.)

IIUC  mounts it on the main Vert.x Web Router so they get the  port-unification model that the proposal wants to achieve for  WildFly.

Note that https://github.com/eclipse-vertx/vertx-grpc does not use io.grpc.netty under the hood - it only uses it for testing https://github.com/eclipse-vertx/vertx-grpc/blob/main/pom.xml#L146

PR #552, opened 2024-01-17
Add proposal for WFLY-15836 Add TLS support to JGroups TCP-based tran…

rhusar:
…sports.

PR #807, opened 2026-02-20
WFLY-21473 Promote "TLS support to JGroups TCP-based transports" to DEFAULT stability

rhusar:
Jira
https://issues.redhat.com/browse/WFLY-21473
https://redhat.atlassian.net/browse/EAP7-1865

Original proposal (COMMUNITY stability)
https://github.com/wildfly/wildfly-proposals/pull/552

WildFly Feature Development Process Issue
https://github.com/wildfly/wildfly-proposals/issues/808

### DIFF
--- a/clustering/jgroups/WFLY-15836_Add_TLS_support_to_JGroups_TCP-based_transports.adoc
+++ b/clustering/jgroups/WFLY-15836_Add_TLS_support_to_JGroups_TCP-based_transports.adoc
@@ -12,9 +12,9 @@ feature-team:
  outside-perspective:
   - darranl
 promotes:
-promoted-by:
+promoted-by: https://github.com/wildfly/wildfly-proposals/issues/808
 ---
-= WFLY-15836 Add TLS support to JGroups TCP-based transports
+= WFLY-15836 Add TLS support to JGroups TCP-based transport
 :author:            Radoslav Husar
 :email:             rhusar@redhat.com
 :toc:               left
@@ -24,7 +24,7 @@ promoted-by:
 
 == Overview
 
-The JGroups project has added support for TLS at the transport level.
+The JGroups project has added support for TLS at the transport level for TCP transport.
 This was implemented as part of https://issues.redhat.com/browse/JGRP-2487[JGRP-2487] and subsequent enhancements.
 Supporting this configuration in WildFly would effectively replace usage of `ASYM_ENCRYPT` in conjunction with `AUTH` protocol.
 This would simplify the process of securing a JGroups channel using standard TLS mechanisms.
@@ -35,8 +35,8 @@ User is wanting to secure internal cluster TCP communication using TLS leveragin
 
 == Issue Metadata
 
-* https://issues.redhat.com/browse/WFLY-15836[WFLY-15836 Add TLS support to JGroups TCP-based transports]
-// * https://issues.redhat.com/browse/EAP7-1865[EAP7-1865 Add TLS support to JGroups TCP-based transports]
+* https://issues.redhat.com/browse/WFLY-15836[WFLY-15836 Add TLS support to JGroups TCP-based transport]
+// * https://issues.redhat.com/browse/EAP7-1865[EAP7-1865 Add TLS support to JGroups TCP-based transport]
 
 === Related Issues
 
@@ -59,7 +59,7 @@ User is wanting to secure internal cluster TCP communication using TLS leveragin
 
 == Requirements
 
-The TCP-based transports (namely `TCP` and `TCP_NIO2`) defined in the JGroups subsystem can be configured with references to the Elytron SSL contexts.
+The TCP-based transport, specifically `TCP`, defined in the JGroups subsystem can be configured with references to the Elytron SSL contexts.
 A client and server SSLContexts are referenced (both are required if defined)
 
 === Changed requirements
--- a/clustering/jgroups/WFLY-21473_Promote_TLS_support_to_JGroups_TCP-based_transport_to_default_stability.adoc
+++ b/clustering/jgroups/WFLY-21473_Promote_TLS_support_to_JGroups_TCP-based_transport_to_default_stability.adoc
@@ -1,0 +1,142 @@
+---
+categories:
+  - clustering
+  - cloud
+  - openshift
+stability-level: default
+issue: https://github.com/wildfly/wildfly-proposals/issues/808
+feature-team:
+ developer: rhusar
+ sme:
+  - mkopecky
+ outside-perspective:
+  - rchettri
+promotes: https://github.com/wildfly/wildfly-proposals/pull/552
+promoted-by:
+---
+= WFLY-21473 Promote TLS support to JGroups TCP-based transport to default stability
+:author:            Radoslav Husar
+:email:             rhusar@redhat.com
+:toc:               left
+:icons:             font
+:idprefix:
+:idseparator:       -
+
+== Overview
+
+This proposal promotes the "TLS support to JGroups TCP-based transport" feature from `community` stability to `default` stability.
+This feature was implemented in https://issues.redhat.com/browse/WFLY-15836[WFLY-15836] with the corresponding community proposal at https://github.com/wildfly/wildfly-proposals/pull/552[wildfly-proposals#552].
+
+For complete details on the feature's design and implementation, refer to the link:WFLY-15836_Add_TLS_support_to_JGroups_TCP-based_transports.adoc[original proposal].
+
+=== User Stories
+
+User wants to secure internal cluster TCP communication using TLS leveraging Elytron subsystem capabilities at `default` stability level.
+
+== Issue Metadata
+
+* https://issues.redhat.com/browse/WFLY-21473[WFLY-21473 Promote 'TLS support to JGroups TCP-based transport' to 'default' stability]
+
+=== Related Issues
+
+* https://issues.redhat.com/browse/WFLY-15836[WFLY-15836 Add TLS support to JGroups TCP-based transport] (Original implementation)
+* https://redhat.atlassian.net/browse/WFCORE-7538[WFCORE-7538 Elytron SSLContext requirements cannot discern between client vs server SSLContext] (related usability issue)
+
+=== Affected Projects or Components
+
+* https://github.com/wildfly/wildfly
+
+=== Other Interested Projects
+
+* N/A
+
+=== Relevant Installation Types
+
+* Traditional standalone server (unzipped or provisioned by Galleon)
+* Managed domain
+* OpenShift Source-to-Image (S2I)
+* Bootable jar
+
+== Requirements
+
+All requirements from the link:WFLY-15836_Add_TLS_support_to_JGroups_TCP-based_transports.adoc[original proposal] remain in effect.
+
+=== Changed requirements
+
+All requirements from the link:WFLY-15836_Add_TLS_support_to_JGroups_TCP-based_transports.adoc[original proposal] remain in effect.
+
+=== Non-Requirements
+
+All requirements from the link:WFLY-15836_Add_TLS_support_to_JGroups_TCP-based_transports.adoc[original proposal] remain in effect.
+
+=== Future Work
+
+* Cloud feature pack can add support for this mechanism out-of-box as it uses TCP stack by default.
+* https://issues.redhat.com/browse/WFLY-19459[WFLY-19459 Add DTLS support to JGroups UDP-based transports w/multicast disabled]
+* https://issues.redhat.com/browse/WFLY-21665[WFLY-21665 Add TLS support to JGroups TCP_NIO2 transport]
+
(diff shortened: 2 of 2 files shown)


clustering/jgroups/WFLY-15836_Add_TLS_support_to_JGroups_TCP-based_transports.adoc
```diff
@@ -59,7 +59,7 @@ User is wanting to secure internal cluster TCP communication using TLS leveragin
 
 == Requirements
 
-The TCP-based transports (namely `TCP` and `TCP_NIO2`) defined in the JGroups subsystem can be configured with references to the Elytron SSL contexts.
+The TCP-based transport, specifically `TCP`, defined in the JGroups subsystem can be configured with references to the Elytron SSL contexts.
```
marekkopecky, line 62: It seems that TCP_NIO2 is back, can you please update the analysis and jira name / description?

clustering/jgroups/WFLY-21473_Promote_TLS_support_to_JGroups_TCP-based_transport_to_default_stability.adoc
```diff
@@ -0,0 +1,142 @@
...
+
+=== Non-Requirements
+
+All requirements from the link:WFLY-15836_Add_TLS_support_to_JGroups_TCP-based_transports.adoc[original proposal] remain in effect.
+
+=== Future Work
+
+* Cloud feature pack can add support for this mechanism out-of-box as it uses TCP stack by default.
+* https://issues.redhat.com/browse/WFLY-19459[WFLY-19459 Add DTLS support to JGroups UDP-based transports w/multicast disabled]
+* https://issues.redhat.com/browse/WFLY-21665[WFLY-21665 Add TLS support to JGroups TCP_NIO2 transport]
```
marekkopecky, line 92: Out of date?

clustering/jgroups/WFLY-21473_Promote_TLS_support_to_JGroups_TCP-based_transport_to_default_stability.adoc
```diff
@@ -0,0 +1,142 @@
...
+* Verification that the feature documentation is complete and accurate.
+
+== Community Documentation
+
+Documentation was included with the original implementation's pull request and resides in the High Availability Guide.
+No additional documentation changes are required for this promotion.
+
+== Release Note Content
+
+The "TLS support to JGroups TCP-based transport" feature has been promoted to `default` stability level.
```
marekkopecky: "TLS support to JGroups TCP-based transport" -> "TLS support to JGroups TCP-based transports"?

marekkopecky:
Please add info whether/how NIO2 would be supported, generally architecture questions from https://github.com/wildfly/wildfly/pull/20024 needs to be answered here in the analysis.

rhusar:
> Please add info whether/how NIO2 would be supported, generally architecture questions from [wildfly/wildfly#20024](https://github.com/wildfly/wildfly/pull/20024) needs to be answered here in the analysis.

It will not as explained in https://github.com/wildfly/wildfly/pull/19698#issuecomment-4482627969

In case anything in this document will be out-of-date after I push the update that addresses the above ^ comment I will update it along the code.

clustering/jgroups/WFLY-15836_Add_TLS_support_to_JGroups_TCP-based_transports.adoc
rhusar: Update since then, unfortunately TCP_NIO2 has been deemed unsupportable, thus this line is still relevant as is.

clustering/jgroups/WFLY-21473_Promote_TLS_support_to_JGroups_TCP-based_transport_to_default_stability.adoc
rhusar: Update since then, unfortunately TCP_NIO2 has been deemed unsupportable, thus this line is still relevant as is.

clustering/jgroups/WFLY-21473_Promote_TLS_support_to_JGroups_TCP-based_transport_to_default_stability.adoc
rhusar: Updated.

rhusar:
@pferraro Please review.

marekkopecky:
What is the reason for keeping FD_SOCK2 as default? Wouldn't be better to switch to FD_SOCK? (also with respect to virtual thread topic in https://github.com/wildfly/wildfly/pull/19698#issuecomment-4575304238 )

clustering/jgroups/WFLY-21473_Promote_TLS_support_to_JGroups_TCP-based_transport_to_default_stability.adoc
```diff
@@ -0,0 +1,167 @@
...
+The `TCP_NIO2` transport does not support TLS (see <<non-requirements>>).
+
+The resource definition registrar is split into two:
+
+* `SocketTransportResourceDefinitionRegistrar` — base registrar for socket-based transports without TLS support (used by `TCP_NIO2`).
+* `SecurableSocketTransportResourceDefinitionRegistrar` — extends the base with `client-ssl-context` and `server-ssl-context` attributes (used by `TCP`).
+
+When TLS is configured on the `TCP` transport and a protocol such as `FD_SOCK2` requests a non-blocking NIO channel, the channel is created without TLS and a warning is logged.
+This is because TLS support for NIO channels is not available.
+Users who need encrypted failure detection should use TCP's built-in failure detection (suspect events) or the socket-based `FD_SOCK` protocol.
```
marekkopecky, line 79: > TCP’s built-in failure detection (suspect events)

Can you please elaborate on this? How to configure this after FD_SOCK2 removal?

clustering/jgroups/WFLY-21473_Promote_TLS_support_to_JGroups_TCP-based_transport_to_default_stability.adoc
```diff
@@ -0,0 +1,167 @@
...
+
+The `client-ssl-context` and `server-ssl-context` attributes are now available only on the `TCP` transport resource.
+The `TCP_NIO2` transport does not support TLS (see <<non-requirements>>).
+
+The resource definition registrar is split into two:
+
+* `SocketTransportResourceDefinitionRegistrar` — base registrar for socket-based transports without TLS support (used by `TCP_NIO2`).
+* `SecurableSocketTransportResourceDefinitionRegistrar` — extends the base with `client-ssl-context` and `server-ssl-context` attributes (used by `TCP`).
+
+When TLS is configured on the `TCP` transport and a protocol such as `FD_SOCK2` requests a non-blocking NIO channel, the channel is created without TLS and a warning is logged.
```
marekkopecky, line 77: This seems to be risky. If user oversees the warning, user can think that security is configured but it isn't. Manual FD_SOCK2 removal would be better that this solution

clustering/jgroups/WFLY-21473_Promote_TLS_support_to_JGroups_TCP-based_transport_to_default_stability.adoc
marekkopecky, line 79: Main motivation of this RFE is to have more simple solution than we have originally - but extra manual switching doesn't seem to be simple.

rhusar: Sure, from user perspective just head over the the WF PR that documents this for the user (https://github.com/wildfly/wildfly/pull/19698/changes#diff-5d9c200f017b6e806ad674b13839304dc22f98ff9d718cb3a9eb6be09b4c6d74 but that link will break on next push given how GH works).

But to reiterate here, they don't have to do anything, it is enough to remove FD_SOCK2 for suspect events.

The logic that drives this is line 127 of SocketTransportResourceDefinitionRegistrar.java (that's already in main).

```
  if (!hasSocketBasedFailureDetectionProtocol && (transport.getReaperInterval() <= 0) && (transport.getConnExpireTime() <= 0)) {
      transport.enableSuspectEvents(true);
  }
```

clustering/jgroups/WFLY-21473_Promote_TLS_support_to_JGroups_TCP-based_transport_to_default_stability.adoc
rhusar: > This seems to be risky. 

This is immaterial.

There is no user data sent over the connections that the FD_SOCK/FD_SOCK2 create. The connection itself is the heartbeat: the pinger thread just calls read() which blocks until the socket closes.

>  If user oversees the warning, user can think that security is configured but it isn't. 

There is simply nothing we can do if people refuse to read documentation or the server log.

> Manual FD_SOCK2 removal would be better that this solution

Users are free to do so.

clustering/jgroups/WFLY-21473_Promote_TLS_support_to_JGroups_TCP-based_transport_to_default_stability.adoc
rhusar: A blanket statement like that is just simply not true.

> "simple solution than we have originally"

Not sure what you are referring to here, but jgroups "native" TLS support does not encrypt these FD connections, so our approach is superior. Moreover, this tightly integrates with elytron and thus formalizes the configuration and establishes proper model, which again is far superior. Last but not least, we also encrypt entire communication including headers using JDK verified concepts rather than custom encryption logic, not just payload, again, far superior.

> but extra manual switching doesn't seem to be simple.

We don't require manual switching, that's optional if people want to have those connections encrypted too.

rhusar:
> What is the reason for keeping FD_SOCK2 as default? Wouldn't be better to switch to FD_SOCK? (also with respect to virtual thread topic in [wildfly/wildfly#19698 (comment)](https://github.com/wildfly/wildfly/pull/19698#issuecomment-4575304238) )

If this is in reference to this RFE, the TLS support is opt-in and users need to configure it explicitly. Thus, we are NOT making any changes to the default stack in this RFE, that would be wrong.

If you are asking in general, then yes, [we want to eventually remove FD_SOCK2 and rely on TCP built-in detection](https://redhat.atlassian.net/browse/WFLY-20710) which we/I have been working on stabilizing and testing over the past months.

For the future virtual threads support refer to the epic https://redhat.atlassian.net/browse/WFLY-19564

clustering/jgroups/WFLY-21473_Promote_TLS_support_to_JGroups_TCP-based_transport_to_default_stability.adoc
marekkopecky: Thank you, I don't have permission to resolve this thread, but please consider it as resolved.

clustering/jgroups/WFLY-21473_Promote_TLS_support_to_JGroups_TCP-based_transport_to_default_stability.adoc
marekkopecky: Maybe I'm missing something here. So what happens if TLS is configured in transport and FD_SOCK2 protocol is in stack? Is FD_SOCK2 switched to FD_SOCK? Or is FD_SOCK2 still used, but its channel is not secured?

clustering/jgroups/WFLY-21473_Promote_TLS_support_to_JGroups_TCP-based_transport_to_default_stability.adoc
rhusar: Oh ok and I ll ask around if we can get you more permissions. This might be part of the 'triage' perms.

clustering/jgroups/WFLY-21473_Promote_TLS_support_to_JGroups_TCP-based_transport_to_default_stability.adoc
rhusar: As the design doc explains, 'When TLS is configured on the `TCP` transport and a protocol such as `FD_SOCK2` requests a non-blocking NIO channel, the channel is created without TLS and a warning is logged.' things continue to operate normally and a log message is issued. There is no auto magic. FD_SOCK2 is used creating non-TLS connection as it's an infrastructure connection, not a connection for data.

pferraro (approved):

marekkopecky (approved):
Thank you for the explanation! I'm approving current proposal here.

rhusar:
Thanks @pferraro and @marekkopecky for the reviews – proceeding to resolve the conversations then.

Issue #818, opened 2026-04-01
Provide an EE10 variant of standard WildFly once the main variant moves to EE11


bstansberry:
### Description

Standard WildFly will soon move to supporting EE 11. To help our users to more comfortably adapt to this change, for a period we will provide an additional variant of standard WildFly that still provides EE 10. Like the usual standard WildFly this will by default allow community stability and above features. It will provide all the standard WildFly appserver features that the EE 11 variant does, but the provided EE APIs and impls will be EE 10.

The intent is to provide this variant for a quarter or two after the primary variant moves to EE 10. Once the standard WildFly code base makes changes that are incompatible with EE 10 (e.g. removing SecurityManager support) or there are components with critical CVEs with no fixes in the EE 10 variant, we will stop providing the EE 10 variant.

### Issue Contact

bstansberry

PR #819, opened 2026-04-01
[Issue_818] Provide an EE10 variant of standard WildFly once the main…

bstansberry:
… variant moves to EE11

Analysis for https://github.com/wildfly/wildfly-proposals/issues/818

Also https://redhat.atlassian.net/browse/WFLY-21698


build/WFLY-21698_EE_10_variant_of_EE_11_WildFly.adoc
```diff
@@ -0,0 +1,170 @@
...
+* The MyFaces feature pack 2.x series will support the `jakarta-ee-10+` family, thus allowing `wildfly-ee-10` to be used as the base feature pack, instead of the default `wildfly-ee` provided by early WildFly releases the provided EE 10.
+** The expectation is a new 3.x series of this feature pack will provide EE 11. Unlike the other feature packs discussed in this analysis, the same release of this feature pack cannot be used with both an EE 10 and and EE 11 base WildFly, because the feature pack depends on EE APIs that differ between EE 10 and 11.
+* User BOMs specifying EE 10 dependencies will be provided that are equivalent to the usual boms (that will now provide EE 11 dependencies).
+* A WildFly Channels channel and manifest will be provided for the `wildfly-ee-10` feature pack.
+* A Galleon channel will be provided for the `wildfly-ee-10` feature pack.
+* The WildFly Container project will produce images based on a server provisioned using the `wildfly` feature pack in combination with the `wildfly-ee-10` feature pack.
```
jmesnil: @bstansberry the container images uses the tgz archive to install WildFly so:

The WildFly Container project will produce images based on the tgz archive of the WildFly fat build from the  `wildfly-ee-10` feature pack.
These images will be:
* `quay.io/wildfly/wildfly:${wildfly-version}-ee-10-jdk{java-version}`
* `quay.io/wildfly/wildfly:${wildfly-version}-ee-10-latest` (pointing to the jdk21 tag)

with `java-version` being `17` & `21`

build/WFLY-21698_EE_10_variant_of_EE_11_WildFly.adoc
```diff
@@ -0,0 +1,170 @@
...
+
+Feature packs provided by other repos that can alternatively use `wildfly-ee-10` as a dependency will add test profiles that exercise that dependency.
+
+== Community Documentation
+
+The https://docs.wildfly.org/39/WildFly_and_WildFly_Preview.html[WildFly and WildFly Preview, window=_blank] will be retitled and updated to cover this new variant.
+
+The https://docs.wildfly.org/39/WildFly_Maven_Plugin_Guide.html[WildFly Maven Plugin Guide, window=_blank] will be updated to illustrate using the `wildfly-ee-10` feature pack.
+
+A new guide will be created illustrating the use of the `wildfly-ee-10` feature pack.
```
jmesnil: Just to be clear, this guide would also illustrate the uses of the deliverables derived from/impacted by this feature pack such as:
* WildFly downloads archives
* container images
* WildFly Glow CLI
* others?

build/WFLY-21698_EE_10_variant_of_EE_11_WildFly.adoc
```diff
@@ -0,0 +1,170 @@
...
+
+
+=== Future Work
+
+* Impact on WildFly Glow and WildFly catalog that can be done once the new `wildfly-ee-10` feature-pack has been released:
+** WildFly Glow. We need to evolve tooling: WildFly Maven Plugin, WildFly Glow CLI, and possibly jbang's integration with the notion of a 'variant'. A new option to convey this information will be added.
+** WildFly Catalog. We today have a catalog per WildFly version for the 2 existing variants: WildFly and WildFly Preview. We will add a catalog for the new "EE-10" variant.
+** Some of this _might_ be done after the feature pack is release in WildFly beta but before the Final release; otherwise it will be done later.
+
+== Backwards Compatibility
```
jmesnil: There will be no backwards incompatibility (that's the whole reason of this work :) 

But what about forward compatibility?
Should we call out that this feature pack is *not* compatible with Jakarta EE 11 deployments?

build/WFLY-21698_EE_10_variant_of_EE_11_WildFly.adoc
bstansberry: Thanks; I will add that.

build/WFLY-21698_EE_10_variant_of_EE_11_WildFly.adoc
```diff
@@ -0,0 +1,170 @@
...
+
+The https://docs.wildfly.org/39/WildFly_Maven_Plugin_Guide.html[WildFly Maven Plugin Guide, window=_blank] will be updated to illustrate using the `wildfly-ee-10` feature pack.
+
+A new guide will be created illustrating the use of the `wildfly-ee-10` feature pack.
+
+Use of the feature pack will be a significant section of the release announcement for the first WildFly release where it appears.
+
+== Release Note Content
+
+This will be a major part of the release announcement, not just the typical few sentences that would be outlined in this part of the analysisc
```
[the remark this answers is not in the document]
bstansberry: @jmesnil Thanks.

Code review suggestions are good for this kind of thing as I'm just adding commits to this branch as I go and will squash at the end. So I'd just accept the change.

build/WFLY-21698_EE_10_variant_of_EE_11_WildFly.adoc
```diff
@@ -0,0 +1,170 @@
...
+MyFaces Feature pack
+WildFly container images
+
+=== Other Interested Projects
+
+=== Relevant Installation Types
+
+
+* Traditional standalone server (unzipped or provisioned by Galleon)
+* Managed domain
```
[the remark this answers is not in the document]
bstansberry: I added stuff related to this in the Requirements, Non-Requirements and Future Work sections.

build/WFLY-21698_EE_10_variant_of_EE_11_WildFly.adoc
```diff
@@ -0,0 +1,170 @@
...
+* The AI feature pack will support the `jakarta-ee-10+` family, thus allowing `wildfly-ee-10` to be used as the base feature pack, instead of the default `wildfly-ee`.
+* The deployment transformer feature pack will support the `jakarta-ee-deployment` family, thus allowing `wildfly-ee-10` to be used as the base feature pack, instead of the default `wildfly-ee`.
+* The MyFaces feature pack 2.x series will support the `jakarta-ee-10+` family, thus allowing `wildfly-ee-10` to be used as the base feature pack, instead of the default `wildfly-ee` provided by early WildFly releases the provided EE 10.
+** The expectation is a new 3.x series of this feature pack will provide EE 11. Unlike the other feature packs discussed in this analysis, the same release of this feature pack cannot be used with both an EE 10 and and EE 11 base WildFly, because the feature pack depends on EE APIs that differ between EE 10 and 11.
+* User BOMs specifying EE 10 dependencies will be provided that are equivalent to the usual boms (that will now provide EE 11 dependencies).
+* A WildFly Channels channel and manifest will be provided for the `wildfly-ee-10` feature pack.
```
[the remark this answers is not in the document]
bstansberry: Done.

build/WFLY-21698_EE_10_variant_of_EE_11_WildFly.adoc
bstansberry: Done.

build/WFLY-21698_EE_10_variant_of_EE_11_WildFly.adoc
```diff
@@ -0,0 +1,198 @@
...
+* Using a Domain Controller provisioned using the `wildfly-ee` feature pack to manage a secondary Host Controller provisioned using the `wildfly-ee-10` feature pack is not supported.
+
+
+
+=== Future Work
+
+* Impact on WildFly Glow and WildFly catalog that can be done once the new `wildfly-ee-10` feature-pack has been released:
+** WildFly Glow. We need to evolve tooling -- WildFly Maven Plugin and WildFly Glow CLI -- with the notion of a 'variant'. A new option to convey this information will be added.
+** WildFly Catalog. We today have a catalog per WildFly version for the 2 existing variants: WildFly and WildFly Preview. We will add a catalog for the new "EE-10" variant.
+** Some of this _might_ be done after the feature pack is release in WildFly beta but before the Final release; otherwise it will be done later.
```
yersan, line 123: s/is release/is released

build/WFLY-21698_EE_10_variant_of_EE_11_WildFly.adoc
```diff
@@ -0,0 +1,198 @@
...
+* Support for domains where a Domain Controller provisioned using the `wildfly-ee` feature pack can manage secondary Host Controllers provisioned using the `wildfly-ee-10` feature pack.
+** Support for this primarily a matter of adding testing. Conceptually it should just work.
+** Conceptually the opposite could work (EE 10 managing EE 11), but it's unlikely we would choose to test and support this.
```
yersan, line 126: +1 It is also against the requirements elaborated on this proposal, and the idea is to roughly keep ~2 WildFly releases with `wildfly-ee-10`. Mixed domains are there to support the transition period migrating your whole Domain to new Kernel version, to me this scenario with a Domain Controller using `wildfly-ee-10` is very unlikely.

build/WFLY-21698_EE_10_variant_of_EE_11_WildFly.adoc
```diff
@@ -0,0 +1,198 @@
...
+No incompatibility.
+
+=== Provisioning
+
+Users wishing to use this variant will need to update their provisioning configuration to specify use of the `wildfly-ee-10` feature pack.
+
+== Implementation Plan
+
+__<This section is optional. I've left it blank for now as I don't know
+   if using this analysis to describe this will be productive.>__
```
yersan, line 154: Can be removed, the description added below is useful

build/WFLY-21698_EE_10_variant_of_EE_11_WildFly.adoc
```diff
@@ -0,0 +1,198 @@
...
+* The WildFly Container project will produce images based on the tgz archive of the WildFly fat build from the wildfly-ee-10 feature pack. These images will be:
+** `quay.io/wildfly/wildfly:${wildfly-version}-ee-10-jdk{java-version}` (with `java-version` being 17 and 21)
+** `quay.io/wildfly/wildfly:${wildfly-version}-ee-10-latest` (pointing to the jdk21 tag)
+* A Domain Controller provisioned using the `wildfly-ee-10` feature pack will be able to manage secondary Host Controllers that are running previous versions of WildFly. Such a domain is referred to as a 'mixed domain'. Mixed domain support is limited to secondary installations that meet the following requirements:
+** They are running versions that are supported in a mixed domain by the EE 11 variant of the Domain Controller's version. (In other words, the EE 10 variant doesn't support additional versions.)
+** They are running a version that predates the introduction of the `wildfly-ee-10` feature pack, or they were provisioned using the `wildfly-ee-10` feature pack.
+** See also the Non-Requirements related to domain mode.
+* TODO -- what about `org.wildfly:wildfly-client-all` and other client jars?
```
yersan, line 99: `org.wildfly:wildfly-client-all` is the only one that ships Jakarta EE apis, right?

If so, as you mentioned, if none of the specs bundled in this client uber jar changed between EE 10 and 11, shipping the client jar with Jakarta EE 11 variants is compatible, so this TODO section can be resolved just by adding a note saying that client uber jars are compatible with both EE 10 and EE 11.

I assuming jboss-cli-client.jar is not affected by any Jakarta spec, correct?

build/WFLY-21698_EE_10_variant_of_EE_11_WildFly.adoc
```diff
@@ -0,0 +1,198 @@
...
+* The deployment transformer feature pack will support the `jakarta-ee-deployment` family, thus allowing `wildfly-ee-10` to be used as the base feature pack, instead of the default `wildfly-ee`.
+* The MyFaces feature pack 2.x series will support the `jakarta-ee-10+` family, thus allowing `wildfly-ee-10` to be used as the base feature pack, instead of the default `wildfly-ee` provided by early WildFly releases the provided EE 10.
+** The expectation is a new 3.x series of this feature pack will provide EE 11. Unlike the other feature packs discussed in this analysis, the same release of this feature pack cannot be used with both an EE 10 and and EE 11 base WildFly, because the feature pack depends on EE APIs that differ between EE 10 and 11.
+* User BOMs specifying EE 10 dependencies will be provided that are equivalent to the usual boms (that will now provide EE 11 dependencies).
+* A WildFly Channels channel and manifest will be provided for the `wildfly-ee-10` feature pack with the Maven GA `org.wildfly.channels:wildfly-ee-10`.
```
jbliznak, line 90: do we want to capture the wildfly-compat-ee-10 manifest here as well?

build/WFLY-21698_EE_10_variant_of_EE_11_WildFly.adoc
```diff
@@ -0,0 +1,198 @@
...
+* The AI feature pack will support the `jakarta-ee-10+` family, thus allowing `wildfly-ee-10` to be used as the base feature pack, instead of the default `wildfly-ee`.
+* The deployment transformer feature pack will support the `jakarta-ee-deployment` family, thus allowing `wildfly-ee-10` to be used as the base feature pack, instead of the default `wildfly-ee`.
+* The MyFaces feature pack 2.x series will support the `jakarta-ee-10+` family, thus allowing `wildfly-ee-10` to be used as the base feature pack, instead of the default `wildfly-ee` provided by early WildFly releases the provided EE 10.
+** The expectation is a new 3.x series of this feature pack will provide EE 11. Unlike the other feature packs discussed in this analysis, the same release of this feature pack cannot be used with both an EE 10 and and EE 11 base WildFly, because the feature pack depends on EE APIs that differ between EE 10 and 11.
+* User BOMs specifying EE 10 dependencies will be provided that are equivalent to the usual boms (that will now provide EE 11 dependencies).
```
jbliznak, line 89: does this include user client BOMs? they are based on `wildfly-standard-ee-bom`, do we need EE10 variant of them? 

wildfly-ejb-client-bom
wildfly-jaxws-client-bom
wildfly-jms-client-bom

jbliznak: tangent to that, with JAXWS being removed from EE11, what is the future of wildfly-jaxws-client-bom?

build/WFLY-21698_EE_10_variant_of_EE_11_WildFly.adoc
jbliznak: > There will be no backwards incompatibility (that's the whole reason of this work :)

+1,  but we should still persist it here, now this section is empty

build/WFLY-21698_EE_10_variant_of_EE_11_WildFly.adoc
```diff
@@ -0,0 +1,198 @@
...
+=== Provisioning
+
+Users wishing to use this variant will need to update their provisioning configuration to specify use of the `wildfly-ee-10` feature pack.
+
+== Implementation Plan
+
+__<This section is optional. I've left it blank for now as I don't know
+   if using this analysis to describe this will be productive.>__
+
+* Within the main WildFly code base, the term 'legacy' will be in non-end-user facing strings (e.g. directory names, comment text, artifactIds for things that are internal details). The intent there is to make as much code as possible generic in nature so that it can be repurposed easily in the future to provide, say, a `wildfly-ee-11` when standard WildFly moves on to EE 12. The term `wildfly-ee-10` or similar will be reserved for things the end user would type.
```
jbliznak, line 156: just curious, with regard to
> The hope is to provide it for two WildFly releases

could it happen the "legacy" code is cut out before moving to latest+legacy feature packs again with eg. EE1 and EE12?

build/WFLY-21698_EE_10_variant_of_EE_11_WildFly.adoc
```diff
@@ -0,0 +1,198 @@
...
+** The artifactIds of the thin and fat server builds.
+*** This last one is somewhat debatable. It impacts the file name of the downloadable tars/zips.
+
+== Admin Clients
+
+Admins clients will just work. A client connecting to a server or Host Controller provisioned using this feature will be manageable in the same way the EE 11 variant of the same release is. There may be differences in particular items in the server configuration, but every WildFly installation is expected to have differences versus what's in our OOTB configs.
+
+== Security Considerations
+
+None; this will be a continuation of the existing security functionality.
```
jbliznak, line 169: is it worth to cal out security manager here? it is stated on Requirements so probably not necessary

build/WFLY-21698_EE_10_variant_of_EE_11_WildFly.adoc
```diff
@@ -0,0 +1,198 @@
...
+
+In the main WildFly server code base, a new test execution profile or profiles that provide coverage analogous to the `base-server-tests` and `expansion-tests` profiles will be added that use the `wildfly-ee-10` feature pack instead of `wildfly-ee`. (Perhaps the profiles will simply tailor the execution of `base-server-tests` and `expansion-tests`.)
+
+A CI job will be added that execute these test profiles as part of PR testing.
+
+A full set of nightly CI jobs will be added that execute these test profiles.
+
+The EE 10 TCK will be run against a server using the `wildfly-ee-10` feature pack.
+
+Feature packs provided by other repos that can alternatively use `wildfly-ee-10` as a dependency will add test profiles that exercise that dependency.
```
jbliznak, line 182: this in worst case mean the required testing resources will double, is there any plan to consolidate the test combinations that is worth to note here too?

build/WFLY-21698_EE_10_variant_of_EE_11_WildFly.adoc
```diff
@@ -0,0 +1,198 @@
...
+** `quay.io/wildfly/wildfly:${wildfly-version}-ee-10-jdk{java-version}` (with `java-version` being 17 and 21)
+** `quay.io/wildfly/wildfly:${wildfly-version}-ee-10-latest` (pointing to the jdk21 tag)
+* A Domain Controller provisioned using the `wildfly-ee-10` feature pack will be able to manage secondary Host Controllers that are running previous versions of WildFly. Such a domain is referred to as a 'mixed domain'. Mixed domain support is limited to secondary installations that meet the following requirements:
+** They are running versions that are supported in a mixed domain by the EE 11 variant of the Domain Controller's version. (In other words, the EE 10 variant doesn't support additional versions.)
+** They are running a version that predates the introduction of the `wildfly-ee-10` feature pack, or they were provisioned using the `wildfly-ee-10` feature pack.
+** See also the Non-Requirements related to domain mode.
+* TODO -- what about `org.wildfly:wildfly-client-all` and other client jars?
+** Not used in a WF installation, but still deployed to JBoss Nexus / Maven Central
+** How does the 'recipe' for it end up driving what goes in the provisioning-time-generated jar?
```
jfdenise, line 101: If a shaded jar has the same dependencies in both variants (same GroupId/ArtifactId) then it can be shared between the variants. The versions will be resolved at provisioning time using the current provisioned variant fp or channel.
If dependencies are not the same (have different GA or the set of dependencies is different), then a new artifact must be defined (eg: wildfly-ee-10-client-all) with a different shaded model produced (eg: org.wildfly.wildfly-ee-10-client-all.shaded). At provisioning time it would have to be copied onto the same location (as seen in [this](https://github.com/wildfly/wildfly/blob/main/ee-feature-pack/galleon-shared/src/main/resources/packages/tools/pm/wildfly/tasks.xml#L9) provisioning task). So transparent once provisioned.

build/WFLY-21698_EE_10_variant_of_EE_11_WildFly.adoc
```diff
@@ -0,0 +1,170 @@
...
+
+* Traditional standalone server (unzipped or provisioned by Galleon)
+* Managed domain
+* OpenShift Source-to-Image (S2I)
+* Bootable jar
+
+== Requirements
+
+* A new `wildfly-ee-10` feature pack will be created.
+** It will have a Galleon Feature Pack Locator of `wildfly-ee-10@maven(org.jboss.universe:community-universe)`
```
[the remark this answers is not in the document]
bstansberry: We kept it.

build/WFLY-21698_EE_10_variant_of_EE_11_WildFly.adoc
bstansberry: @jbliznak The existing client boms for EE 10 are fine for EE 11 as the relevant specs did not update for EE 11.

Re Jakarta Webservices, the spec was removed from the EE 11 platform, but it still exists. EE just doesn't require vendors to implement it. WildFly still provides it and we have no plans to remove it. We'll run the relevant parts of the EE 10 TCK against our EE 11 server, not for the purposes of certifying compatibility, but just to use it as a test suite.

build/WFLY-21698_EE_10_variant_of_EE_11_WildFly.adoc
bstansberry: Yes; will do.

build/WFLY-21698_EE_10_variant_of_EE_11_WildFly.adoc
bstansberry: You are correct and what you suggest is the way to resolve this.

bstansberry: The jboss-cli-client.jar is not affected by this.

build/WFLY-21698_EE_10_variant_of_EE_11_WildFly.adoc
bstansberry: Thanks. It should be fine regardless as the relevant dependencies are all shared.

build/WFLY-21698_EE_10_variant_of_EE_11_WildFly.adoc
bstansberry: @jbliznak This section is meant to be empty -- it's just a parent header for the more specific subsections. If you have generic text you'd like to see, please make a suggestion.

build/WFLY-21698_EE_10_variant_of_EE_11_WildFly.adoc
bstansberry: @jbliznak I don't think we should remove it because there is a lot of work that goes into creating it, and it seems likely we'll want the same thing again.

Perhaps we'll disable building things by default, and disable the 'deploy' phase processing for any of the modules. So it can still occasionally be built and tested, as a way to mitigate against code rot, but would not be something we ship.

Note that if we kept it around any cross-cutting activities, e.g. removal of calls to SecurityManager APIs, would be applied to it. IOW if we decided to keep shipping this beyond WF 41 we'd drop SM support in it, and it therefore would not be possible to pass the EE 10 TCKs.

build/WFLY-21698_EE_10_variant_of_EE_11_WildFly.adoc
bstansberry: +1 I should add that.

build/WFLY-21698_EE_10_variant_of_EE_11_WildFly.adoc
bstansberry: Perhaps the nightly jobs running 'allTests' can be limited to the SecurityManager-enabled jobs, except for on SE 25 where the SM can't be enabled.

And then limit the bootable jar and Galleon layer jobs to a single SE version. I can't recall ever seeing any SE-specific issues in those jobs that didn't appear in the -DallTests jobs.

build/WFLY-21698_EE_10_variant_of_EE_11_WildFly.adoc
jbliznak: I see, no problem then :)

PR #822, opened 2026-04-09
[WFLY-21717] Promote HashiCorp Vault feature pack to the DEFAULT stab…

skyllarr:
…ility

resolves #821

PR #840, opened 2026-06-17
[WFLY-21717] Add the hashicorp-vault subsystem to the WildFly distribution at the Default stability level.

### LINKED ISSUE #821: [WFLY-21717] Promote HC Vault FP to DEFAULT stability
### Description

Promote WildFly Vault feature pack and hashicorp-vault layer/subsystem from community  to default stability

### Issue Contact

skyllarr

darranl:
Follow up from https://github.com/wildfly/wildfly-proposals/pull/822

Resolves #821


elytron/WFLY-21717-Add-the-hashicorp-vault-extension-to-wildfly.adoc
```diff
@@ -0,0 +1,305 @@
...
+[engine=TYPE][@mount-path][#]secret-path?key-path
+----
+
+Where:
+
+* `engine=TYPE` - Optional engine type specification (e.g., `KVv1`, `KVv2`, `database`, `pki`)
+** If omitted, uses the credential store's `default-engine-type` configuration (default: `KVv2`)
+* `@mount-path` - Optional custom mount path
+** If omitted, uses the credential store's `default-mount-path` configuration (default: `secret`)
+* `#` - Secret path marker (optional when no prefix is present, required when `engine=` or `@` is used)
```
honza-kasik, line 127: Maybe it won't be so nice looking, but doesn't it make sense to make `#` always required to rpevent any confusion?

elytron/WFLY-21717-Add-the-hashicorp-vault-extension-to-wildfly.adoc
```diff
@@ -0,0 +1,305 @@
...
+@mount%2Fname#secret%20path?db.host     # Multiple encoded chars
+----
+
+==== Legacy Format Support
+
+The legacy format `secret-path.key` (using dot as delimiter) can be supported via the `support-legacy-alias-format` configuration parameter on the credential store resource.
+
+**Format Detection:**
+
+* If the alias contains `?`, `#`, `@`, or starts with `engine=`, it uses the new format
```
honza-kasik, line 204: What if the legacy alias already contains `@` - e.g. `user@host`?

elytron/WFLY-21717-Add-the-hashicorp-vault-extension-to-wildfly.adoc
darranl: I think for users it will be nicer to enable them to omit so in the simplest form if they are using a secret engine mounted at "secret" they can just use secret_path?key_name so it keeps the alias very simple and they only need the complexity if they need to convey more information so I suspect the majority of users can use the simplest form.

elytron/WFLY-21717-Add-the-hashicorp-vault-extension-to-wildfly.adoc
darranl: I think this will just have to be a limitation, we would like users to migrate anyway and we are not going to support the legacy format at the Default stability level . In not too many releases time we should remove this support entirely.

PR #845, opened 2026-06-30
[WFLY-22046] Support fine-grained/individual metric filtering

### LINKED ISSUE #844: Support fine-grained/individual metric filtering
### Description

Add support for defining filters to control which metrics are gathered/reported.

### Issue Contact

jasondlee

jasondlee:
https://redhat.atlassian.net/browse/WFLY-22046

Create proposal document

Resolves: #844

### DIFF
--- a/observability/WFLY-22046-support-fine-grained-metric-filtering.adoc
+++ b/observability/WFLY-22046-support-fine-grained-metric-filtering.adoc
@@ -1,0 +1,231 @@
+---
+categories:
+- observability
+stability-level: community
+issue: https://github.com/wildfly/wildfly-proposals/issues/844
+feature-team:
+ developer: jasondlee
+ sme:
+  - TBD
+ outside-perspective:
+  - TBD
+promotes:
+promoted-by:
+---
+
+= Support fine-grained/individual meter filtering
+:author:            Jason Lee
+:email:             jasondlee@ibm.com
+:toc:               left
+:icons:             font
+:idprefix:
+:idseparator:       -
+
+== Overview
+
+Currently, a large number of system meters are automatically registered with the metrics system(s).
+This includes a variety of JMX beans and JVM statistics, as well as various subsystem properties.
+All of these are then published to an aggregation system of some type. In large deployments, this
+creates challenges for Operations, as the data for all the meters from a potentially large number
+of servers can cause memory capacity issues for the aggregation system. Often, many of these meters
+are of no interest to the system administrator(s), but they have to be managed, as there is no way to
+_not_ publish them.
+
+This RFE will provide a way for system administrators to filter which meters
+are actually published, enabling finer-grained control over the data produced -- and the memory
+required downstream to monitor them. The primary focus will be Micrometer, but a solution for
+OpenTelemetry will also considered.
+
+=== User Stories
+
+* As a system administrator, I need to be able to configure Micrometer to filter out meters that are of no interest to operations, etc.
+
+== Issue Metadata
+
+=== Related Issues
+
+* https://redhat.atlassian.net/browse/WFLY-22046[WFLY-22046]
+
+=== Affected Projects or Components
+
+* https://github.com/wildfly/wildfly[WildFly]
+
+=== Other Interested Projects
+
+* OpenShift
+
+=== Relevant Installation Types
+
+* Traditional standalone server (unzipped or provisioned by Galleon)
+* Managed domain footnote:[For push-based exports only. i.e., Prometheus is not supported.]
+* OpenShift Source-to-Image (S2I)
+* Bootable jar
+
+== Requirements
+
+=== Hard Requirements
+* The administrator must be able to list 0 or more meters to exclude from publishing
+** If no filter applies to a given meter, then it is allowed to publish.
+** If a meter _is_ matched to a filter, then it is not allowed to publish.
+* The user should be able to filter by
+** Meter id
+** Meter type
+** Meter tags
+* The filter definition should support regular expressions to give the user flexibility in defining
+  the match.
+* The filter should apply to all meters, including
+** JMX beans
+** System meters
+** WildFly subsystem meters
+** User-defined/application meters
+* The solution should consider both Micrometer and OpenTelemetry Metrics to ensure that the configuration
+of the two subsystems are as alike as possible, reducing complexity for users
+** Micrometer has an explicit `MeterFilter`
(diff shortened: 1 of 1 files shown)


observability/WFLY-22046-support-fine-grained-metric-filtering.adoc
```diff
@@ -0,0 +1,228 @@
...
+The current design is for meter filtering only, so there is no need to specify the signal type to filter. If and when support for other signals is defined, the model/schema can be updated to add those, with existing configurations migrated to specify that pre-existing filters target meters.
+
+[NOTE]
+====
+As we think through this design, we should strive to strike a balance between ease of use and functionality for the user, while trying to avoid an unbound proliferation of child resources to cover every conceivable filter type. Ideally -- perhaps -- we should be able to provide a simple management model (and XML schema) that provides the flexibility we're after without inundating the user with scores of types to remember and manage.
+
+It may turn out, however, that this flexibility makes for an awkward model/schema that hinders more than helps. We should be sensitive to this as the design matures and be prepared to pivot.
+====
+
+If an "accept" filter and a "reject" filter overlap, the accept filter will win, since the inclusion of the meter was explicitly configured.
```
bstansberry, line 143: Is this worth it? It adds quite a bit of complexity and can be unintuitive.

In a different context I've done quite a bit of work around filters using the JEP 290[1]  mechanism, which for the class filtering aspect uses accept/reject semantics, applying a list of filter expressions in the order they are provided. As soon as a given class matches, the accept or reject applies and processing ends. The user is responsible for providing a list that results in the desired behavior. Reading the list the user can readily intuit what will happen using step by step reasoning.

See java.io.ObjectInputFilter.Config.Global.checkInput

[1] https://openjdk.org/jeps/290

observability/WFLY-22046-support-fine-grained-metric-filtering.adoc
```diff
@@ -0,0 +1,228 @@
...
+* `false`|`false`|Invert the check defined (e.g., "starts with"->"does not start with")
+
+|`value`|`*`||The value to filter for
+|===
+
+The current design is for meter filtering only, so there is no need to specify the signal type to filter. If and when support for other signals is defined, the model/schema can be updated to add those, with existing configurations migrated to specify that pre-existing filters target meters.
+
+[NOTE]
+====
+As we think through this design, we should strive to strike a balance between ease of use and functionality for the user, while trying to avoid an unbound proliferation of child resources to cover every conceivable filter type. Ideally -- perhaps -- we should be able to provide a simple management model (and XML schema) that provides the flexibility we're after without inundating the user with scores of types to remember and manage.
```
bstansberry, line 138: This is a good comment and I'm not suggesting changing it, but it does seem like you have something quite simple.

observability/WFLY-22046-support-fine-grained-metric-filtering.adoc
```diff
@@ -0,0 +1,228 @@
...
+a|
+* `accept`
+* `reject`|`reject`|The most common action will be to reject, so the system will default to that
+
+
+|`field`
+a|
+* `meter.name`
+* `tag.name`
+* `tag.value`||
```
bstansberry: Please use '-' instead of '.'

In the CLI context, which is what really matters for the use of this, I'd rather avoid the use of '.' in literals. In other contexts it used to specify individual fields inside complex attributes and I'd rather not use it for some other purpose.

The dot also vaguely implies some kind of useful-to-the-user hierarchical relationship between the string before the dot and after, but that's not really the case.

If there actually *is* going to be some kind of useful-to-the-user hierarchical relationship, then two different fields may be probably better.

observability/WFLY-22046-support-fine-grained-metric-filtering.adoc
```diff
@@ -0,0 +1,228 @@
...
+
+|`condition`
+a|
+* `startsWith`
```
bstansberry: starts-with

observability/WFLY-22046-support-fine-grained-metric-filtering.adoc
```diff
@@ -0,0 +1,228 @@
...
+* `endsWith`
```
bstansberry: ends-with

observability/WFLY-22046-support-fine-grained-metric-filtering.adoc
```diff
@@ -0,0 +1,228 @@
...
+
+* Legacy metrics systems (i.e., WildFly/"base" Metrics) are out of scope.
+* Filtering of telemetry signals other than meters (i.e., spans, span events, logs, etc).
+* Transforming a meter (for example, see https://docs.micrometer.io/micrometer/reference/concepts/meter-filters.html#_transforming_metrics[here]) is not in scope, so the following actions are not supported:
+** Renaming a meter (id) or a tag
+** Adding tags to every meter
+** Dropping certain tags
+** Replacing tag values
+** Limiting the number of tags on a meter
+* A unified configuration for both Micrometer and OpenTelemetry, while interesting, given their similarities, is not in scope.
```
bstansberry: Please clarify what "A unified configuration" means.

Reading the above it sounds like the applicability to OpenTelemetry is not nailed down, which is ok at this stage. But it also sounds like there will be thinking about consistency (good!) and conceptually it seems like things can be consistent. So as this progresses if we end up being consistent, this non-requirement will be confusing unless it's explained.

observability/WFLY-22046-support-fine-grained-metric-filtering.adoc
jasondlee: Based on my preliminary experiments, if the accept filters are registered before the reject, then Micrometer just does the right thing. I think it will be a bit more manual on the OpenTelemetry side (via their view (?) API), but I think it should be pretty straightforward all around.

observability/WFLY-22046-support-fine-grained-metric-filtering.adoc
jasondlee: I don't expect anything much more complex than what's listed. It will be restricted pretty tightly by what the Micrometer (and otel) APIs give us access to, and I think this is pretty much it. It could expand, of course, as we dive deeper into design and discovery, but I don't think it will be to any great extent.

observability/WFLY-22046-support-fine-grained-metric-filtering.adoc
jasondlee: I've expanded the wording here. I hope that adequately addresses your concerns.

PR #848, opened 2026-07-16
WFLY-20810 was verified at default

bstansberry:
@jfdenise I see the downstream tracker for this is in Product Docs so to me this one is at default, not community.

I'm going to just merge this because I'm about the reference this analysis in the WF 41 release announcement. If I'm wrong we can revert this.


bstansberry:
I also updated https://github.com/wildfly/wildfly-proposals/issues/747

PR #855, opened 2026-09-18
Stop checking out fork code in the pull_request_target workflow

rhusar:
The Check Proposals workflow runs from pull_request_target, so it has the base repository's GITHUB_TOKEN, pull-requests: write and the default-branch cache scope, yet it checked out refs/pull/N/merge. That made two pieces of fork-controlled code run in the trusted context: bundler-cache installs from the pull request's Gemfile, and .scripts/check_proposals was the fork's own copy. actions/checkout now refuses this as a pwn request.

Check out the base branch instead and treat the pull request purely as data: list the changed files through the API, fetch the pull request head without checking it out, and extract only the changed AsciiDoc files for the checker to parse. The Gemfile, the checker and _data/wildfly-categories.yaml now always come from the base branch.

This also drops the tj-actions/changed-files dependency, which was only needed because the fork worktree was present.


rhusar:
CI failure is expected until merged.

See [#wildfly-developers > wildfly-proposals GH actions are failing](https://wildfly.zulipchat.com/#narrow/channel/174184-wildfly-developers/topic/wildfly-proposals.20GH.20actions.20are.20failing/with/625192695)

.github/workflows/check_process.yml
```diff
@@ -12,10 +12,12 @@ jobs:
...
 
     steps:
+      # NOTE: this job runs from 'pull_request_target', so it has the base
+      # repository's token and secrets. Never check out or execute the fork's
+      # code here: check out the base branch and treat the pull request's
+      # content as data only.
       - name: Checkout
         uses: actions/checkout@v4
-        with:
-          ref: "refs/pull/${{ github.event.number }}/merge"
```
rhusar, line 18: More context why this must be removed ASAP:

> Only allow pull_request_target when it is necessary. Workflows triggered by this event should not check out, build, or run code from an untrusted pull request with access to repository secrets or a privileged GITHUB_TOKEN.

https://docs.github.com/en/actions/reference/security/securely-using-pull_request_target#default-policy-for-pull_request_target

bstansberry:
Thanks @rhusar

PR #856, opened 2026-09-18
Only add or remove PR labels that actually need to change

rhusar:
The checker removed all stability-level/* labels and re-added the correct one on every run, which produced a 'removed and added' event on the pull request timeline even when nothing changed.

e.g. @github-actions github-actions Bot added stability-level/default and removed stability-level/default labels on May 19

Read the current labels once and only call gh for labels that are missing or stale.


darranl:
If multiple files are editing in a PR i.e. adding promotes and promoted-by this labelling should be restricted to the new file only.

rhusar:
> If multiple files are editing in a PR i.e. adding promotes and promoted-by this labelling should be restricted to the new file only.

Yes exactly, as discussed elsewhere. Let me see if I can include it here.

rhusar:
@darranl OK, included that fix here.

